PR #369, opened 2013-05-11
Recaptcha2, request #1382

pbojan:
Contains config variable description, option to make the config from `./setup` script and while enabled the user can't login without entering valid captcha on the login screen for cookie auth type. 

Removed "Test Commit", fixed CSP bug on the login screen and tested on Chrome and Firefox.

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -62,13 +62,13 @@ Basic settings
     :default: false
 
     Starting with version 2.3.0 phpMyAdmin offers a lot of features to
-    work with master / foreign – tables (see :config:option:`$cfg['Servers'][$i]['pmadb']`).  
-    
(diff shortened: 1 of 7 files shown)


crysman:
Please, what is the minimum PMA version containing this patch? It's to say, if I am running PMA 4.2.6 - if it's already there? :) Thanks

lem9:
@crysman Version is 4.4-dev which is the future 4.4 version.

crysman:
Hups, I see... so it will be available in the next release, ok, perfect. Will be the option to use CAPTCHA available also in PMA GUI? Or just some hidden config line? I would prefer the first one ;)

lem9:
@crysman You were talking about "this patch", which is about a newer version of captcha. Version 4.2 contains captcha support via two configuration lines, as explained in the documentation.

PR #985, opened 2014-03-02
fixed #4300 Grid edit textfield: can't place cursor or select text with ...

vidurangaw:
...mouse

Signed-off-by: Viduranga Wijesooriya vpowerrc@gmail.com

### DIFF
--- a/js/makegrid.js
+++ b/js/makegrid.js
@@ -1003,6 +1003,11 @@ function PMA_makegrid(t, enableResize, enableReorder, enableVisib, enableGridEdi
 
                     $input_field.datepicker("show");
 
+                    //move ui-datepicker-div inside cEdit div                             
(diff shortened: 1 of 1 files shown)


ashutoshdhundhara:
@zixtor Why are we moving datepicker manually as it automatically adjusts its location?

zixtor:
Not sure if I remember correctly but it may have been due to the use of horizontal scrolling in #sqlqueryresults div which necessitated manual position of datepicker. But it was later decided to put the scrolling in some outer div and someone forgot to remove this manual positioning.

ashutoshdhundhara:
Thanks Atul! This snippet was causing bug [#4712](https://sourceforge.net/p/phpmyadmin/bugs/4712). Just wanted to confirm about any side-effects.

PR #1440, opened 2014-12-30
SQL formatting with CodeMirror

madhuracj:
Not for immediate merge.

I started working on https://sourceforge.net/p/phpmyadmin/feature-requests/1582/ and Codemirror was the first choice since we use it for highlighting and auto completing. 
However, following drawbacks are there
1. Formatting functionality in CodeMirror is not actively developed. I'm using formatting.js from an old version here.
2. Even though the framework for formatting is there we have to define specific rules for SQL. I didn't find anything already developed and the current rules simple format by breaking at ; and brackets.

### DIFF
--- a/js/codemirror2/LICENSE
+++ b/js/codemirror2/LICENSE
@@ -1,0 +1,23 @@
+Copyright (C) 2012 by Marijn Haverbeke <marijnh@gmail.com>
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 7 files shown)


nisargjhaveri:
There is a nice SQL formatter, https://github.com/jdorn/sql-formatter. The only problem is that it is in PHP.

nisargjhaveri:
I've implemented basic formatter using [sql-formatter](https://github.com/jdorn/sql-formatter). See https://github.com/phpmyadmin/phpmyadmin/pull/1450

lem9:
I guess we no longer need this one.

PR #1444, opened 2015-01-01
some defaults for config.sample.inc.php

computersalat:
Thinking this will give the user more impression what could/should be done here.

Signed-off-by: Christian Wittmer &lt;chris (at) computersalat (dot) de&gt;

### DIFF
--- a/config.sample.inc.php
+++ b/config.sample.inc.php
@@ -10,66 +10,211 @@
  * @package PhpMyAdmin
  */
 
-/*
+/**
+ * Your phpMyAdmin url
+ *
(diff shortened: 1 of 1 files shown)


lem9:
Christian,
thanks for contributing; however, I don't see the value of this patch. We intentionally kept the sample config file short. Of course we can always argue about what is important and what is not important, but some of the lines you added, for example PmaAbsoluteUri, are not really needed for most users. Also, default config value were carefully chosen to inform users of potential problems (like SuhosinDisableWarning) and should not be reversed in the sample config file without a discussion.

PR #1445, opened 2015-01-02
Fix auto configuration issues

madhuracj:


lem9:
Madhura,
I started testing. It's a good improvement but I have found the following issues so far:
* The favorite table feature works but data is not stored in the pmadb so is not persistent across sessions
* User prefs are not applied when I come back to the db where pmadb tables were created
* Let's say I have access to two databases. If in the first one I have created
the pmadb tables, in the second one in Operations I am not offered to create the pmadb tables

madhuracj:
Hi Marc,
Thanks for reviewing. First two issues should be fixed with the new commits. Can you confirm?
Regarding the second issue, if I go into the second database right after login I am offered to create pma tables. However, if I go to the first database, it will detect that pma tables are there in the first table and subsequent visits to second database will not offer to create pma tables since the system now knows that they are already there in the first database. Do you seen this behavior? Do you think it should be changed?

lem9:
Madhura,
as soon as I start phpMyAdmin:
````
Fatal error: Can't use method return value in write context in libraries/RecentFavoriteTable.class.php on line 63
````

lem9:
Madhura,
about detecting that pma tables are in the first database, did you try closing all your browser windows, opening phpMyAdmin and going to the second database? I bet that the presence of pma tables in the first database won't be remembered.

madhuracj:
Strange I am not getting that error. Does
`$tables = (! empty($this->_getPmaTable()))
    ? $this->getFromDb()
    : array();
$_SESSION['tmpval'][$this->_tableType . '_tables'][$server_id]
    = $tables;`

fix the error for you

madhuracj:
No, the presence of pma tables are not expected to be remembered across sessions. 
When you login sometime later a message saying that configuration storage is not configured is displayed. 
Now, 3 scenarios,
- If you go to database where there are no pma tables you will be offered to create pma tables at its operations page. 
- If you go to a database where the pma tables reside, the system will detect it and configure itself to use those.
- If you first go to the database where pma table reside and then go to some other database, you will not be offered to create pma table in the latter.

lem9:
The proposed fix in two lines does not fix the problem, which moves to line 64. I'm using PHP 5.4.35.

lem9:
These scenarios are fine with me, but would need to be explained in the doc, especially now that we don't create pma tables in every database.

madhuracj:
The last commit should fix the issue

lem9:
I confirm that the fatal error is gone. The issue with favorite tables is fixed. The issue with user prefs is not fixed: I get "Your preferences will be saved for current session only. Storing them permanently requires phpMyAdmin configuration storage." even if I started by opening the db which hosts the pma tables.

madhuracj:
@lem9 
The issue with prefs should be fixed now. I updated the docs as well.

lem9:
All is fine now, thanks!

PR #1446, opened 2015-01-03
Fix #4645 and #4682

Tithugues:


lem9:
Thanks, all tests run fine!

PR #1450, opened 2015-01-06
SQL formatting with sql-formatter

nisargjhaveri:
Feature request, https://sourceforge.net/p/phpmyadmin/feature-requests/1582/

### DIFF
--- a/db_sql_format.php
+++ b/db_sql_format.php
@@ -1,0 +1,18 @@
+<?php
+/* vim: set expandtab sw=4 ts=4 sts=4: */
+/**
+ * Format SQL for SQL editors
+ *
+ * @package PhpMyAdmin
+ */
+
+require_once 'libraries/common.inc.php';
+require_once 'libraries/sql-formatter/lib/SqlFormatter.php';
+
+$query = isset($_POST['sql']) ? $_POST['sql'] : '';
+
+SqlFormatter::$tab = "\t";
+$query = SqlFormatter::format($query, false);
+
+$response = PMA_Response::getInstance();
+$response->addJSON("sql", $query);
--- a/js/functions.js
+++ b/js/functions.js
@@ -976,6 +976,30 @@ function insertQuery(queryType)
     if (queryType == "clear") {
         setQuery('');
         return;
+    } else if (queryType == "format") {
+        if (codemirror_editor) {
+            $('#querymessage').html(PMA_messages.strFormatting
+                + '&nbsp;<img class="ajaxIcon" src="'
+                + pmaThemeImage + 'ajax_clock_small.gif" alt="">');
+            var href = 'db_sql_format.php';
+            var params = {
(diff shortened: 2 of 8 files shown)


nijel:
This brings in yet another SQL parser to our codebase. I think having one in PHP (libraries/sqlparser.lib.php) and one in Javascript (CodeMirror) is enough. So either the formatter should use our parser or our parser should be replaced by this one.

nisargjhaveri:
Fair point. I missed that we already have a pretty-printing functionality in library.

nisargjhaveri:
I looked a bit more into it. The parser we have (sqlparser.lib.php) is not doing well enough to be used as a feature. But it does a lot more thing then just pretty-printing... The library (sql-formatter) is mainly focused on just formatting. It does not provide analyze or parse functionality.

What I suggest is replace formatting part of `sqlparser.lib.php` with the new library throughout pma, and use the old parser for other purposes.

I also looked into the possibility of using the new library as parser, but it doesn't seem possible unless we massively change/develop the library for this purpose.

nijel:
I know both libraries are quite different and does not match in features. However having both in codebase does not look like a good idea to me.

lem9:
Michal: not a good idea, even if we don't maintain the new formatter library? (in the same sense that we don't maintain CodeMirror).

nijel:
We could IMHO at least use their tokenizer instead of PMA_SQP_parse, that way we don't have two different set of reserved words and so on. But I agree it's quite a lot of additional work and I'm not sure if it's worth of it. On the other side our SQL parser is really hard to understand code and getting rid of it would be nice bonus...

Also there is still other option of using CodeMirror for formatting as implemented in https://github.com/phpmyadmin/phpmyadmin/pull/1440

nisargjhaveri:
The formatter implemented in #1440 is really very basic. And the major problem with it is no one is maintaining the library (formatting module of CodeMirror) we are planning to use.

Getting rid of the parser we have is a nice idea (Agreed, a little bit of additional work at first). But then we'll have to maintain the sql-formatter, too.

lem9:
By the way, in the GSoC 2015 ideas list, we have this:
https://wiki.phpmyadmin.net/pma/GSoC_2015_Ideas_List#Parser.2Fanalyzer_rewrite

lem9:
This pull request works fine and I am in favor of merging it.

lem9:
@nisargjhaveri can you prepare some unit tests for this code?

nisargjhaveri:
For the library, right? I'll look into it.

nijel:
Basically just for SqlFormatter::format and test test that with some usual queries...

nisargjhaveri:
I've added some tests. Please check.

And, the trailing spaces and tabs that the bot is complaining about are part of testdata. We can't get rid of it easily.

lem9:
Nice work!

PR #1456, opened 2015-01-09
Bug fix for Server-Process refresh rate

nobelium:
Server Status refresh not behaving as expected.

To reproduce the bug follow these steps:
1. Go to Server->Status->Processes
2. Start the auto refresh feature
3. Set the refresh rate to 20 min
4. Change the refresh rate to 2 sec

We expect the refresh to happen after 2 sec of change in refresh rate but it wait for a refresh to happen in 20 min and only then it refreshes for every 2 sec.

### DIFF
--- a/js/server_status_processes.js
+++ b/js/server_status_processes.js
@@ -152,6 +152,7 @@ AJAX.registerOnload('server_status_processes.js', function() {
     // Bind event handler for change in refresh rate
     $('#id_refreshRate').on('change', function(event) {
         processList.refreshInterval = $(this).val();
+        processList.refresh();
     });
     // Bind event handler for table header links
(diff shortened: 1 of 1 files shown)


nobelium:
Signed-off-by: Vignesh Panneerselvam me@vignesh.info

lem9:
Thanks. I redid your commit in your name, to include the Signed-off-by comment. However it was done on the QA_4_3 branch where bug fixes go.

See https://github.com/phpmyadmin/phpmyadmin/commit/0d87abfd7826de7e7b30db44aecdbd6301ebd20c

nobelium:
Thanks for looking into it!!!

PR #1458, opened 2015-01-12
Remove display direction

madhuracj:
Display direction has been deprecated for some time. Though it was planned to be removed much earlier, it never happened.

See 
http://phpmyadmin.996332.n3.nabble.com/Is-vertical-mode-useful-td1335.html
https://sourceforge.net/p/phpmyadmin/mailman/message/27694362/
https://sourceforge.net/p/phpmyadmin/mailman/message/27749740/


lem9:
Looks fine. This will be a good code cleanup for a feature rarely used. We took a step a few years ago to disable it by default so it should not be missed by many users.

madhuracj:
I think the pull request is ready now.

lem9:
Thanks!

PR #1459, opened 2015-01-13
FRE #1568: fold add column position radio into one 'add column after' 

smita786:
I have used 'After' and Begin as mysql uses 'AFTER' and 'FIRST' in queries not 'BEFORE'.

In the after dropdown, I have used 'At Begin' not sure if it looks fine. Please suggest if we can put it in better way.

### DIFF
--- a/js/tbl_structure.js
+++ b/js/tbl_structure.js
@@ -66,6 +66,13 @@ function reloadFieldForm() {
     $('#page_content').show();
 }
 
+function checkFirst() {
+    if ($("select[name=after_field] option:selected").data('pos') === 'first') {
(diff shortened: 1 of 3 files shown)


lem9:
Looks good. I suggest to replace "At Begin" with "Beginning of table" because the full phrase "after at begin" looks weird. With your choice of not putting "After" in each choice (which I support for clarity), we have to do something about the "At Begin" choice.

smita786:
Hi Marc,
Thanks. Yeah I changed it to "Beginning of table", it sounds better .

lem9:
Great; please merge.

smita786:
merged. Thanks :-)

PR #1461, opened 2015-01-19
bug#4712: Wrongly positioned date-picker while Grid-Editing.

ashutoshdhundhara:
Signed-off-by: Ashutosh Dhundhara ashutoshdhundhara@yahoo.com


lem9:
Thanks Ashutosh. My tests work fine with your patch and test case. I am a little unsure about any side-effect that will happen by removing this chunk of code, but let's do it.

ashutoshdhundhara:
I was also a little afraid of any side effects of this commit. I performed some tests of my own and I have confirmed from Atul for the same as well. This was introduced in #985

PR #1462, opened 2015-01-19
fix bug #4710

smita786:
FOR REVIEW
I have removed the searchClause parameter from ajax request to load child nodes. Which fixes the issue. For loading the child nodes why do we need searchClause ? is there any case where it could be a problem.

### DIFF
--- a/js/navigation.js
+++ b/js/navigation.js
@@ -517,17 +517,12 @@ function loadChildNodes($expandElem, callback) {
     }
     var $destination = $expandElem.closest('li');
 
-    var searchClause = PMA_fastFilter.getSearchClause();
-    var searchClause2 = PMA_fastFilter.getSearchClause2($expandElem);
-
     var params = {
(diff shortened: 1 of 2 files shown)


zixtor:
@smita786 searchClause parameters become useful in loadChildNodes, in case the searched database is on a separate page than the current pagination and you get results after clicking "x other results found". Removing searchClause fixes it for same page search but breaks it for a separate page.

zixtor:
It seems that for search SQL query, sometimes the resultset was smaller in size than the offset being used in LIMIT clause. I think https://github.com/zixtor/phpmyadmin/commit/bc840b07f9d32aa3de6e33caf30e3ad01605e4b9 should fix it.

smita786:
Hi 
yeah, that should fix it. pos offset is being used original one(without search string), instead we need to use the pos offset in the search result. 
I am closing this PR. You can open another one to merge your commit :-)

PR #1463, opened 2015-01-20
BugFix#4714

rakeshkumar4294:
Set a session variable to remember that a direct sql query is coming from console or query box .


lem9:
Nice work.

PR #1467, opened 2015-01-21
#4716 Collapse all in navigation panel is sometimes broken

madhuracj:
Would appreciate a second opinion on possible side effects of these changes.


lem9:
@madhuracj  Thanks, I could not find a side effect.

PR #1469, opened 2015-01-23
rfe #1574 Provide for a database select dropdown in navigation panel as in pre-4.0 versions

atul516:
For review, tests and subsequent merge


lem9:
Thanks Atul; I have started testing. 

The message "Whether to display navigation panel in tree style or in pre-4.0 style" needs improvement because when one unticks the option, she still gets some kind of tree, not at the db-level but below that level.

zixtor:
Thanks Marc for review!
For description message, I suggest:- "Navigate with the default database tree or have a database select-dropdown as in pre-4.0 versions".
And a clearer configuration option may be 'ReplaceDbTreeWIthSelect' with it's default value as false.

lem9:
@zixtor Well, this is a checkbox so the explanation cannot contains "or". It must describe what happens when the checkbox is ticked. I suggest "NavigationDatabasesSelector"  with the explanation "In the navigation panel, replace the databases tree with a selector". Of course, the logic would have to be reversed.

zixtor:
Makes sense, modified it..

lem9:
Thanks. There are two failing unit tests...

ibennetch:
Can I suggest either NavigationDatabasesDropdown or NavigationDatabasesAsSelector

The reason is because to me, NavigationDatabasesSelector makes Selector sound like a verb, which makes me wonder "make the databases a selector for what?" Clearly the documentation quickly sorts this out, but slightly changing the directive text might help make this more oblivious. 

I liked best ShowDatabasesNavigationAsTree (or if that's too long, ShowDatabasesAsTree) but it seems you didn't.

lem9:
I like your suggestion of ShowDatabasesNavigationAsTree. I don't remember seeing this suggestion before (I saw ShowNavigationAsTree).  The previous discussion was about removing completely the tree, which this pull request does not do.

zixtor:
Well, about completely removing the tree, I believe the main thing people really miss about pre-4.0 is switching databases easily with the selector, so I thought it's quite doable by modifying existing tree while keeping it's other features intact, hence no complete removal of the tree and it does falsify ShowNavigationAsTree.
I also like ShowDatabasesNavigationAsTree suggestoin. I will make the changes and look into failing tests tomorrow.
Talking about other features, I think we ought to implement the 'db fast filter' as well for this alternative way of navigation.

lem9:
@zixtor After the change of config parameter and the tests fix, I suggest you merge before adding the db fast filter.

PR #1471, opened 2015-01-31
bugs#3935 Table Header issue in Safari 5.0.5

whitepiegon:


whitepiegon:
This bug was occurring due to $col.outerWidth(true).
In safari 5.0.5 $col.outerWidth(true) apart from returning the width, it was also extending the width. On the other hand $col.outerWidth() is working fine for safari.
However, I did not change it for other browser as jquery 1.8 require .outerWidth with argument for includeMargin.
Please check if my fix works well for other versions of safari.

lem9:
Thanks. Could you resubmit a pull request on QA_4_3 (because 4.2 is outdated). And avoid using tab characters.

whitepiegon:
thanks.. here it is.. https://github.com/phpmyadmin/phpmyadmin/pull/1473
but i don't know why the The Travis CI build failed

PR #1473, opened 2015-02-02
bugs#3935 Table Header issue in Safari 5.0.5

whitepiegon:

### DIFF
--- a/js/makegrid.js
+++ b/js/makegrid.js
@@ -254,7 +254,9 @@ function PMA_makegrid(t, enableResize, enableReorder, enableVisib, enableGridEdi
             $('table.pma_table').find('thead th:first').removeClass('before-condition');
             for (var n = 0, l = $firstRowCols.length; n < l; n++) {
                 var $col = $($firstRowCols[n]);
-                $($resizeHandles[n]).css('left', $col.position().left + $col.outerWidth(true))
+                if ($.browser.safari) var colWidth = $col.outerWidth();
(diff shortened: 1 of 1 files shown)


lem9:
@whitepiegon To test your patch I need to be able to reproduce the problem. I am trying with Safari 5.0.5 on Windows 7 and am not able to reproduce, with either phpMyAdmin 4.2.9 or the current QA_4_3. Please describe your testing environment. In particular I am mostly interested to reproduce under QA_4_3, as 4.2.x is no longer supported.

whitepiegon:
Hi..
I can easily reproduce this bug in Windows 8 Safari 5.0.5 both in pma 4.2.9 and QA_4_3 in my localhost. Browse any table to see it.

qa43

.

lem9:
Thanks.

lem9:
Aayush,
your patch is now part of QA_4_3; unfortunately, $.browser is no longer available in newer jQuery so I am looking for another solution for the master branch.

whitepiegon:
Ok.. i ll make required change and will commit it asap.

nisargjhaveri:
Also, doing browser specific fixes like this doesn't seem like a good idea to me. Can't we try for some generic solution that works in all browser. Maybe check why it goes wrong in Safari?

lem9:
@whitepiegon No need, already done.

@nisargjhaveri You are welcome to propose a better solution.

PR #1475, opened 2015-02-02
Send server id to import.php in SQL console

muetob:
Include server id in ajax post request, when sending the SQL from the ajax
request from SQL console.

Signed-off-by: Tobias Mueller tobias.mueller@hostpoint.ch

### DIFF
--- a/js/console.js
+++ b/js/console.js
@@ -82,6 +82,7 @@ var PMA_console = {
             '<input name="is_js_confirmed" value="0">' +
             '<textarea name="sql_query"></textarea>' +
             '<input name="console_message_id" value="0">' +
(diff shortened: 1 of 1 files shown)


lem9:
Thanks; please describe what you are trying to fix with this patch.

muetob:
For our phpMyAdmin installation the sql console simple does not work. Without this patch every command I'm executing in the console leads to the following error:

```
 PHP Fatal error:  Call to a member function getTables() on a non-object in [path]/phpMyAdmin-4.3.8-all-languages/libraries/relation.lib.php on line 1890
```

lem9:
Thanks.

PR #1476, opened 2015-02-02
Remove unnecessary calculation, fix coding style

nisargjhaveri:
- Remove unnecessary calculation
- Remove trailing spaces

Someone please check that replacing `$(window).height() + $(document).scrollTop() - $('#pma_console').offset().top` doesn't have any side effects. Because I'm not sure why it was used in first place. 

`$('#pma_console').height()` should work just fine according to me.

### DIFF
--- a/js/navigation.js
+++ b/js/navigation.js
@@ -165,7 +165,7 @@ $(function () {
             $icon.attr('src', icon_reload_src);
         }, 1000);
     });
-    
+
     $(document).on("change", '#navi_db_select',  function (event) {
         $(this).closest('form').trigger('submit');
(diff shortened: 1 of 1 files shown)


D-storm:
Seems fine to me. Merged. Thanks.

D-storm:
@nisargjhaveri Its better to work on branches rather than working in master directly.

nisargjhaveri:
@D-storm, agreed. I normally do work on branches, but this time I mistakenly committed on master and didn't want to revert and commit again in new branch. Anyway, thanks. :)

PR #1479, opened 2015-02-04
Fix isCanvasSupported for new window

gzzhanghao:
Signed-off-by: Zhang Hao jason.daurus@gmail.com

### DIFF
--- a/js/pmd/iecanvas.js
+++ b/js/pmd/iecanvas.js
@@ -4,7 +4,7 @@
  */
 
 function isCanvasSupported() {
-    var el = document.getElementById("canvas");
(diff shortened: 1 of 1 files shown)


gzzhanghao:
This bug is not on the issue tracker, wondering whether it is welcomed :-)

It occurs when opening a new window in designer page, the code tried to get the canvas element before the document is ready. At that time the token field in `PMA_commonParams` is not set yet so it results in an `invalid token` error.

lem9:
The patch looks fine, thanks. Next time, please prepare your bugfix patch from the QA_4_3 branch. As this patch fixes something minor, we'll only apply it to the master branch.

PR #1480, opened 2015-02-04
Fix bug  #4706 Schema export doesn't handle dots in db/table name

kushalxp:
Signed-off-by: Kushal Bhatia bhatia.kushal@gmail.com


madhuracj:
This PR does not seem to fix the bug.

kushalxp:
I tested it before generating PR ...
Works fine in mine . . .

PR #1483, opened 2015-02-06
Feature #1558 Designer menu

gzzhanghao:
Redesigned the designer menu and add support for $cfg['TabsMode']
Works fine with chrome and firefox, needs help for test in IE

designer

PR #1485, opened 2015-02-07
New designer menu

gzzhanghao:
Moved from #1483


lem9:
This is better, but did you test all of the Designer features? I tried deleting a relation by clicking on the ending part (round) of the arrow and nothing happens. The arrow is supposed to turn red and I should be seeing a "Delete relation" confirmation dialog. I am testing with the employees sample database.

gzzhanghao:
@lem9 Yeah you are right, I didn't even know there are such a feature. I'll fix it and check if I'd missed anything else.

lem9:
@gzzhanghao Your ignorance of such feature made me realize that we should probably advertise it via the menu.

lem9:
@gzzhanghao I confirm your fix for the delete relation feature. Now could you take care of the trailing whitespace then rebase to produce just one commit? It would help us to review your patch.

gzzhanghao:
@lem9 Alright, committed.

gzzhanghao:
@lem9 I think it's better to check click on lines but not on the small circle. But it's likely to be much more difficult.

lem9:
@gzzhanghao After getting a second opinion from a developer, here is more feedback. Unless the user has a huge screen, the new side-menu eats a lot of the available space. We find ourselves always in the need of toggling the text in to see the descriptions, then out to see the working space. A solution would be to auto-expand the menu on hover and shrink it otherwise. What are your thoughts about this?

gzzhanghao:
@lem9 That's a good idea, but we can't follow the $cfg['TabsMode'] then.

lem9:
Excellent! Could you produce just one commit that includes the whole patch, not forgetting to sign it, and I'll merge for the upcoming 4.4 release.

gzzhanghao:
@lem9 Sure, nice to hear that :-)

lem9:
Nice job.

PR #1488, opened 2015-02-08
Sanity Check for table name

rakeshkumar4294:
A new variable '$reserved_keywords_fields' comprising of either column or table name is declared and previous function for checking column name is used for the table name and accordingly warning message is shown


lem9:
Please clean up your pull request, it should not contain merge commits and so on. Also your commit was not signed. The best is to create a new branch, see https://wiki.phpmyadmin.net/pma/Git#Working_on_new_features.

PR #1491, opened 2015-02-08
Sanity Check for table name : Feature request #723

rakeshkumar4294:
A new variable '$reserved_keywords_fields' comprising of either column or table name is declared and a previous function which was used for checking column name for being a reserved word is used for the table name and accordingly warning message is shown.
Also table name is trimmed before creation.


lem9:
Thanks, this is fine but see my message suggestion.

rakeshkaundilya:
Thanks.

PR #1492, opened 2015-02-09
rfe #1235 Relation view: move to main "Structure" page

madhuracj:


zixtor:
Thanks Madhura! the new relation tab sounds nice. There is conflict in ChangeLog.

lem9:
Thanks Madhura. I have merged because I think that this is an improvement but I'm not sure that this satisfies the original request of "Let users see them at once".

madhuracj:
Well, appending the relational view to the structure page make the page too long if you have more than a handful of column (For example, even for film table of sakila). This seemed to be the next best option.

PR #1495, opened 2015-02-09
Feature request 1312: Added configuration option for setting the timezone of the current database connection

dbroeks:
https://sourceforge.net/p/phpmyadmin/feature-requests/1312/

### DIFF
--- a/libraries/common.inc.php
+++ b/libraries/common.inc.php
@@ -981,6 +981,31 @@ if (! defined('PMA_MINIMUM_COMMON')) {
             $cfg['Server']['user'], $cfg['Server']['password'], false
         );
 
+        // Set timestamp for the session, if required.
+        if ($cfg['Server']['SessionTimeZone'] != '') {
+            $sql_query_tz = 'SET ' . PMA_Util::backquote('time_zone') . ' = '
+                . '\'' . PMA_Util::sqlAddSlashes($cfg['Server']['SessionTimeZone']) . '\'';
+
+            if(! $userlink->query($sql_query_tz)) {
+                $error_message_tz = sprintf(__('Unable to use timezone %s for server %d. '
+                    . 'Please check your configuration setting for '
+                    . '[em]$cfg[\'Servers\'][%d][\'SessionTimeZone\'][/em]. '
+                    . 'phpMyAdmin is currently using the default time zone of the database server.'),
+                    $cfg['Servers'][$GLOBALS['server']]['SessionTimeZone'],
+                    $GLOBALS['server'],
+                    $GLOBALS['server']
+                );
+
+                $GLOBALS['error_handler']->addError(
+                    $error_message_tz,
+                    E_USER_WARNING,
+                    '',
+                    '',
+                    false
+                );
+            }
+        }
+
         if (! $controllink) {
             $controllink = $userlink;
         }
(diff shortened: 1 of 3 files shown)


ibennetch:
This seems reasonable to me but I hope some others developers comment as well.

Could you also add this directive to doc/config.rst ?

d-bro:
Sure, I will do this tomorrow.

libraries/config.default.php
```diff
@@ -494,6 +494,20 @@
 $cfg['Servers'][$i]['MaxTableUiprefs'] = 100;
 
 /**
+ * Sets the time zone for the current database session. Leave blank to use the
+ * time zone of your database server.
+ *
+ * Useful when your database server uses a time zone which is different from the
+ * time zone you want to use in PhpMyAdmin.
```
ibennetch: Lines 500 and 501 probably can be removed from `libraries/config.default.php` and moved to `doc/config.rst`. We don't need full documentation here; just enough to give hints. Users can always refer to the full documentation if they don't understand a directive.

My suggestion is that here we say "Sets the time zone used by phpMyAdmin. Possible values are explained at http://dev.mysql.com/doc/refman/5.7/en/time-zone-support.html" -- and move the rest of what you wrote to config.rst.

libraries/common.inc.php
```diff
@@ -981,6 +981,31 @@
             $cfg['Server']['user'], $cfg['Server']['password'], false
         );
 
+        // Set timestamp for the session, if required.
+        if ($cfg['Server']['SessionTimeZone'] != '') {
+            $sql_query_tz = 'SET ' . PMA_Util::backquote('time_zone') . ' = '
+                . '\'' . PMA_Util::sqlAddSlashes($cfg['Server']['SessionTimeZone']) . '\'';
+
+            if(! $userlink->query($sql_query_tz) ) {
+                $error_message_tz = sprintf( 'Unable to use timezone %s for server %d. '
```
ibennetch: This string should be translatable; please see https://wiki.phpmyadmin.net/pma/Gettext_for_developers

lem9:
The patch seems reasonable to me too. I appreciate the fact that if left empty, the directive will do nothing (we don't want unnecessary queries sent to the MySQL server).

libraries/common.inc.php
```diff
@@ -981,6 +981,31 @@
...
+                    . 'Please check your configuration setting for '
+                    . '[em]$cfg[\'Servers\'][%d][\'SessionTimeZone\'][/em]. '
+                    . 'PHPMyAdmin is currently using the default time zone of the database server.',
+                    $cfg['Servers'][$GLOBALS['server']]['SessionTimeZone'],
+                    $GLOBALS['server'],
+                    $GLOBALS['server']
+                );
+
+                $GLOBALS['error_handler']->addError(
+                    __($error_message_tz),
```
ibennetch: I'm not 100% sure about this, but don't think this will work properly for translating. Normally you'd do something like `sprintf(__('Unable to use...'))`. That way, gettext can properly format the plural forms and digits.

lem9:
Thanks. Could you clean your commits, producing just one?
See https://wiki.phpmyadmin.net/pma/Git#Changing_history

lem9:
Thanks.

lem9:
See also 26f7584c50634d3eda1e3eb8cdde0e9e3264349c

PR #1497, opened 2015-02-12
Fixed Console screen error

djprawns:
Fixed Screen Console error by scrolling the bar back to its original position.

### DIFF
--- a/js/console.js
+++ b/js/console.js
@@ -259,6 +259,7 @@ var PMA_console = {
             'fast', 'easeOutQuart', function() {
                 PMA_console.$consoleContent.css({display:'none'});
                 $(window).trigger('resize');
+                $('body').scrollTop(window.s);
             });
         PMA_console.hideCard();
     },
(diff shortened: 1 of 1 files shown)


nisargjhaveri:
Which bug are you trying to fix? Is this for https://sourceforge.net/p/phpmyadmin/bugs/4744/? That bug is already fixed by 9e18a571f4d88bfa0a82053c0bf02cd66b6ce8e6.

If this is not about that bug, please mention.

vanarpsharma:
I believe this is the same bug, however, I had fixed this bug too, and so wanted to make a pull request. Maybe this is a better solution?

nisargjhaveri:
There are several points I think should be taken care of, 
1. I don't think there was anything wrong with `collapse`. Collapsing console was working fine AFAIK.
2. You are using `window.s`. Firstly, this pollutes the window namespace unnecessarily. And secondly, the name `s` doesn't mean anything. It would be better if we use some better name.

This could have been a better solution if we fix the original bug by changing just the `PMA_console.show` function. In the current fix we have to take care of this issue every time we use the function `PMA_console.show`

zixtor:
Pranav, we welcome your contributions, but the fix you propose is similar to the existing fix in the sense that there has to be manual scroll correction.

zixtor:
Nice observations Nisarg! I have refixed the bug addressing the original issue, see https://github.com/phpmyadmin/phpmyadmin/commit/b78ada45e6ec2f6555b67095c7fce2517c12f799

nisargjhaveri:
@zixtor, b78ada4 is a really neat fix. I was looking for such a fix from the day I reported that bug, rather then manual scroll correction. :)

PR #1498, opened 2015-02-13
rfe #1602 Save unfinished work

gzzhanghao:
A workaround for #1602 disallow redirect to login page on session end.
It synchronizes current query with console and save the query in console when session end.

### DIFF
--- a/js/console.js
+++ b/js/console.js
@@ -539,6 +539,15 @@ var PMA_consoleInput = {
         }
         $('#pma_console .console_query_input').keydown(PMA_consoleInput._keydown);
     },
+    getUnfinished: function () {
+        return PMA_consoleInput._unfinished;
+    },
+    on: function (event, handler) {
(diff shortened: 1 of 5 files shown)


lem9:
Nice work!

lem9:
Next time I suggest you create a branch specific to the feature request. It helps working on more than one thing at a time, in case you have more than one pull request pending.

gzzhanghao:
@lem9 Got it, I'll try next time : )

madhuracj:
I see a couple of problems with this implementation. The most annoying one is, now unrelated SQL queries are displayed in the SQL tab. For example if I visit SQL tab for a particular table, type some query and then move to SQL tab of some other database the old query which is not valid for the new database would still be displayed. Moreover, since this implementation depends on session to store the query it works only when the session file is not removed when the session expires.

PR #1500, opened 2015-02-14
For feature request: - #1605 After database creation, go to database structure page

djprawns:


lem9:
This patch is still not signed and still contains trailing whitespace. See above "This commit is missing Signed-Off-By".  Do you agree with our Developer's Certificate of Origin?

lem9:
This pull request should contain just one commit, like your second one which was properly signed.

PR #1504, opened 2015-02-16
bugfix 4770: Multiple delete on table browse ignoring foreign key checkbox

chirayuchiripal:
**_== Changelog ==**_
- Bugfix 4770: Multiple delete on table browse ignoring foreign key checkbox.
- Use single form instead of separate forms for each submit button.
- Logic to pass submit button name,value pair when submitting forms via ajax.

Signed-off-by: Chirayu Chiripal chirayu.chiripal@gmail.com


lem9:
Thanks.

PR #1511, opened 2015-02-19
Feature Request:#1608 Central columns allowing setting SIGNED UNSIGNED attribute to integer

rakeshkumar4294:
An extra column col_attribute is added to pma_central_column table and accordingly all the pages  are modified.

### DIFF
--- a/db_central_columns.php
+++ b/db_central_columns.php
@@ -22,17 +22,18 @@ if (isset($_POST['edit_save']) || isset($_POST['add_new_column'])) {
     $col_extra = $_POST['col_extra'];
     $col_isNull = isset($_POST['col_isNull'])?1:0;
     $col_length = $_POST['col_length'];
+    $col_attribute = $_POST['col_attribute'];
     $col_type = $_POST['col_type'];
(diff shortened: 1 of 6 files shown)


smita786:
Hey,
I think you need to update the sql/create_tables.sql as well, as you are adding a new column to pma_central_columns table. Please do update that accordingly.

madhuracj:
Changing the pma table structures should be the last resort IMO. Can't we use `col_extra` to store extra information as comma separated values and store SIGNED, UNSIGNED attribute in there?

rakeshkaundilya:
Hi Madhura,
I think it would be more simple to add an extra column as it would be little confusing to put attribute in extra.But I will do as you suggest ,if to put it in 'col_extra'.
Thanks.

madhuracj:
Adding an extra column is easy for the developer, but upgrading will not be easy for the user. We have to advertise that there is change to the pma table structures and even the code will have to handle situations where the user has installed the new version but has not upgraded the table structures. I suggest you start a discussion on this on the developer mailing list.

PR #1513, opened 2015-02-19
fix bug #4772 and other similar issues

smita786:
Git bisect tells the problem started from the commit where we updated jquery to version 1.11.1. 
And it seems in jQuery v 1.11.1 have some issues with checkboxes when used with ".attr". see the link , https://github.com/travist/jquery.treeselect.js/issues/9. 
So when we use ".prop" in place of ".attr", it works. I have replced all such ".attr" which i could find.


zixtor:
Thanks!

smita786:
@zixtor thanks :-)

PR #1522, opened 2015-02-25
Update PMAStandard as per PHP_CodeSniffer 2.x

chirayuchiripal:
Targeting QA_4_3  so that scrutinizer does not fail for patches on the same.

Signed-off-by: Chirayu Chiripal chirayu.chiripal@gmail.com


lem9:
Looks fine to me, thanks.

PR #1523, opened 2015-02-26
Alternative implementation for rfe #1602

madhuracj:
An alternative implementation is due to the problems described at https://github.com/phpmyadmin/phpmyadmin/pull/1498#issuecomment-76158097

### DIFF
--- a/js/console.js
+++ b/js/console.js
@@ -539,15 +539,6 @@ var PMA_consoleInput = {
         }
         $('#pma_console .console_query_input').keydown(PMA_consoleInput._keydown);
     },
-    getUnfinished: function () {
-        return PMA_consoleInput._unfinished;
-    },
-    on: function (event, handler) {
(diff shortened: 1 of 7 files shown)


lem9:
@madhuracj I agree that the other implementation has problems. However, as a user I'm trying to figure out what the "Auto-saved" button means. I mean, if something is automatic, why would I have to click a button for it?  I think that this button should be named "Get auto-saved query"; merging now and I'll change the button. Feel free to react to my change later.

madhuracj:
@lem9 "Get auto-saved query" sounds better. My only worry was whether it was too long for a button.

lem9:
@madhuracj Yes is a bit long but IMO it needed a clarification.

ibennetch:
I agree that "Get auto-saved query" is acceptable here.

PR #1527, opened 2015-02-26
Fix bug#4746:- Right-aligned columns have left-aligned header

tapanhalani:
Signed-off-by: tapanhalani tapanhalani231@gmail.com

### DIFF
--- a/libraries/DisplayResults.class.php
+++ b/libraries/DisplayResults.class.php
@@ -2257,6 +2257,20 @@ class PMA_DisplayResults
 
     } // end of the '_getSortOrderLink()' function
 
+    /**
+     * Check if the column contains numeric data . If yes, then set the column header's alignment right
+     *
+     * @param         $fields_meta                set of field properties
+     *
+     * @return string $value                      class that has to be added to th_class
+     * @see  _getDraggableClassForSortableColumns()
+     */
+    private function getClassForNumericColumnType($fields_meta,&$th_class){
+
+        if(preg_match('@int|decimal|float|double|real|bit|boolean|serial@i',$fields_meta->type)){
+            $th_class[] = 'right';
+        }
+    }
 
     /**
      * Prepare columns to draggable effect for sortable columns
@@ -2283,7 +2297,7 @@ class PMA_DisplayResults
         $draggable_html = '<th';
         $th_class = array();
         $th_class[] = 'draggable';
-
+        $this->getClassForNumericColumnType($fields_meta,$th_class);
         if ($col_visib && !$col_visib_j) {
             $th_class[] = 'hide';
         }
@@ -2336,7 +2350,7 @@ class PMA_DisplayResults
         $draggable_html = '<th';
         $th_class = array();
         $th_class[] = 'draggable';
-
+        $this->getClassForNumericColumnType($fields_meta,$th_class);
         if ($col_visib && !$col_visib_j) {
             $th_class[] = 'hide';
         }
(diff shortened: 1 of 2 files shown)


lem9:
Thanks. You have code duplication which has to be fixed before merging this. Also, we try to avoid inline style, it should be moved to the CSS of themes. By the way, why the margin-right?

tapanhalani:
The inline margin-right is for right margin of table header content , i.e. the span in which the table header name is written. While the external css addition is for content within the span. Hence ,I have to align both separately.

Secondly, I could not understand what code is duplicated here? Are you referring to the duplicate Regular expression checking function for sortable and non-sortable columns? If so, I can fix it. 

Lastly, I will try and make my inline css into external css.

Thanks.

lem9:
You added the exact same four lines twice: this is what I call code duplication.

tapanhalani:
I think you were right about margin-right. I mistakenly left it at it's place even after doing the same with external css. It's of no use and removed. 

Regarding regular expression match, I think it is only 3 lines of code now and would it be feasible to make a new function , implementing that 3 lines there and including their function call here?  As per my understanding , this 3 lines are of use only in this function.

Still , if you feel it is to be extracted from there , please reply me. Thank you.

lem9:
Please extract the duplicated code into a function. This will avoid this kind of problem: a developer changes the list of column types at just one place.
After your next commit, please squash the commits.

tapanhalani:
hello marc,

There seems to be a little problem here. I have modularised the duplicated code as you said. But now, the execution is giving me "internal Server error 500" , reason saying that I am calling an undefined function. I couldn't quite understand as I have defined the function just before the getDraggableClassForSortableColumns() function, in which I have placed the call for it.

How come its undefined??

lem9:
Maybe a typo.

tapanhalani:
solved it......it was indeed a typo....

tapanhalani:
Hey Marc,

I have squashed all my commits in the existing pull request..
Please inform me if anything needs to be done now.

Thank you.

lem9:
See my comments in the code. Also, it was missed for a previous commit from you, but we need your full name in the Signed-off-by line.

tapanhalani:
I understood your point. The updated code is pushed in the commit. Please check it and reply me.
Thank you

lem9:
Ok, we are almost there. I added a comment in the code review, and I still don't see your full name in the signature. As an example, my full name is "Marc Delisle".

tapanhalani:
I have changed the signed-off-by name to full name (Tapan Halani). But I can't change the author name to that.

lem9:
Thanks. You can have a look at 
https://wiki.phpmyadmin.net/pma/Git#Before_starting
maybe it can help about the author name. For the signature, next time please try to tidy the lines because there are many duplicated comments and signatures.

tapanhalani:
Will do. Thanks for the continous support and guidance.

lem9:
Also I forgot to tell you to verify Scrutinizer's reports. I fixed some coding style in 1a9015080ee0ba9694150e1aac7da5e47e77a120.

tapanhalani:
I saw it. Thanks. By the way, I saw the format of code and coding style
from the original code and replicated it. But still , I got bad coding
style when I pushed my changes. How to take care of that in future?

lem9:
For coding style, you can verify locally by installing CodeSniffer + the PMAStandard rules:
````
https://wiki.phpmyadmin.net/pma/CodeSniffer
````

PR #1528, opened 2015-02-27
Feature Request:#1608 Allowing signed/unsigned attribute central column

rakeshkumar4294:
The 'attribute' and 'extra' is stored in col_extra column in json_encoded form.During fetching of columns if col_extra is a json string then it is decoded and resulting array is stored in same variable as col_extra and col_attribute.Similarly it is done in javascript.
But If col_extra is not a json string (to make compatible with old column values) then simply col_attribute is assigned null string and remaining are done in similar way.
Signed-off-by: Rakesh Kumar rakeshkumar4294@gmail.com

### DIFF
--- a/db_central_columns.php
+++ b/db_central_columns.php
@@ -19,20 +19,24 @@ if (isset($_POST['edit_save']) || isset($_POST['add_new_column'])) {
         $orig_col_name = $_POST['orig_col_name'];
     }
     $col_default = $_POST['col_default'];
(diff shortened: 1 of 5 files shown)


smita786:
looks good. nice job!
@lem9 it would be great if you also can have a look! :-)

lem9:
Thanks.

rakeshkaundilya:
Thanks.

PR #1530, opened 2015-03-04
Fix for HHVM compatibility

bygiro:
this will fix the error with HHVM:

Fatal error: Uncaught exception 'Exception' with message 'Port is not numeric'

Signed-off-by


lem9:
Please read https://github.com/phpmyadmin/phpmyadmin/blob/master/DCO to see what a Signed-Off-By line should look like.

bygiro:
Hello,
sorry, but I do not have time to waste, and I already wasted too much time for a PR.....do whatever you want.

nijel:
Anyway forcing the conversion in all cases does not sound like a good idea, the code should rather fail if it gets something unexpected. In 2b23430 I've implemented handling of single value which can cause problems.

PR #1531, opened 2015-03-05
#1613 Central columns: action on multiple columns

rakeshkumar4294:
New check box,check-all,change,and remove from central column buttons are introduced.Selected checkbox containig column name is send using jQuery.
Page for multiple change option is cretated after requested which is inspired from the Table structure page.
Request for multiple delete is handled according to previously written code but for multiple change request , update column is looped.
Signed-off-by: Rakesh Kumar rakeshkumar4294@gmail.com

### DIFF
--- a/db_central_columns.php
+++ b/db_central_columns.php
@@ -66,10 +66,25 @@ $cfgCentralColumns = PMA_centralColumnsGetParams();
 $pmadb = $cfgCentralColumns['db'];
 $pmatable = $cfgCentralColumns['table'];
 $max_rows = $GLOBALS['cfg']['MaxRows'];
+
+if (isset($_REQUEST['edit_central_columns_page'])) {
+    $selected_fld = $_REQUEST['selected_fld'];
+    $selected_db = $_REQUEST['db'];
+    $edit_central_column_page = PMA_getHTMLforEditingPage($selected_fld, $selected_db);
+    $response->addHTML($edit_central_column_page);
+    exit;
+}
+if (isset($_POST['multi_edit_central_column_save'])) {
+    $message = PMA_updateMultipleColumn();
+    if (!is_bool($message)) {
+        $response->isSuccess(false);
+        $response->addJSON('message', $message);
+    }
+}
 if (isset($_POST['delete_save'])) {
     $col_name = array();
-    $col_name[] = $_REQUEST['col_name'];
-    $tmp_msg = PMA_deleteColumnsFromList($col_name, false);
+    parse_str($_POST['col_name'], $col_name);
+    $tmp_msg = PMA_deleteColumnsFromList($col_name['selected_fld'], false);
 }
 if (isset($_REQUEST['total_rows']) && $_REQUEST['total_rows']) {
     $total_rows = $_REQUEST['total_rows'];
@@ -106,6 +121,7 @@ $deleteRowForm = '<form method="post" id="del_form" action="db_central_columns.p
         . '<input type="hidden" name="delete_save" value="delete"></form>';
 $response->addHTML($deleteRowForm);
 $table_struct = '<div id="tableslistcontainer">'
+        . '<form name="tableslistcontainer">'
         . '<table id="table_columns" class="tablesorter" '
         . 'style="min-width:100%" class="data">';
 $response->addHTML($table_struct);
@@ -124,7 +140,10 @@ foreach ($result as $row) {
     $odd_row = !$odd_row;
     $row_num++;
 }
-$response->addHTML('</table></div>');
+$response->addHTML('</table>');
+$tablefooter = PMA_getCentralColumnsTableFooter($pmaThemeImage, $text_dir);
+$response->addHTML($tablefooter);
+$response->addHTML('</form></div>');
 $message = PMA_Message::success(
     sprintf(__('Showing rows %1$s - %2$s.'), ($pos + 1), ($pos + count($result)))
 );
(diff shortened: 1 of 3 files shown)


lem9:
Typo "mulit", your probably mean "multi".

rakeshkaundilya:
Yes,I missed it somehow.Thanks.

smita786:
Hey
Nicely done :-). I have few issues if you can fix those:
1. To be consistent, It should be better if you rename the With selected options "change" and "Remove from Central columns" to "Edit" and "Delete" recpectively. 
2. On the change page for Multiple selected columns, You have a table header "structure" followed by create_table documentation link, that doesn't seem to make sense there so if you can remove that. 
3. Again on change page for Multiple selected columns, I think it would be nice to have a cancel option (not mandatory for now, can be put later).
4. on change page for Multiple selected columns, if we select the type to be "Enum" or "Set", it is not showing the link "Edit ENUM/SET values".  Also in the case if a column type is already "Enum" or "Set", the change page doesn't show the link to edit values.
5. If a column has a default value set to "As defined" with some defined value then on change page, Default attribute dropdown shows "As defined" as selected but not show the actual defined value below it.
do revert back if something is not clear.
Thanks!

rakeshkaundilya:
Hi Smita,
I have fixed the issues.See if it can be further improved.
Thanks.

js/db_central_columns.js
```diff
@@ -41,6 +47,44 @@ AJAX.registerOnload('db_central_columns.js', function () {
...
+            return false;
+        }
+        PMA_ajaxShowMessage();
+        $("#del_col_name").val(multi_delete_columns);
+        $("#del_form").submit();
+    });
+    $('#tableslistcontainer button[name="edit_central_columns"]').click(function(event){
+        event.preventDefault();
+        var editColumnList = $('.checkall:checkbox:checked').serialize();
+        //console.log(editColumnList);
```
smita786: please remove the logging comments.

smita786:
Rakesh,
looks good. Just have a look at the Scrutinizer issues.

rakeshkaundilya:
Hi Smita,
I have tried to fix all the issues.The null issue gives notice of undefined offset.Thanks.

smita786:
Hi Rakesh,
Please squash the commits. 
Also, as you can see this PR is having conflicts, so if possible try to resolve the conflicts and then commit.

rakeshkaundilya:
Hi Marc,
Can we can optimize the multiple update by using Mysql 'case' to execute multiple update query? By this only one query will be finally executed thus saving time.
Link: http://stackoverflow.com/questions/3432/multiple-updates-in-mysql?rq=1#answer-84111

lem9:
Rakesh,
I prefer that you solve the issue of duplicate code by somehow calling PMA_updateOneColumn() from PMA_updateMultipleColumn(). It would produce cleaner logic.

smita786:
Hi Marc and Rakesh,
What about just having one function which handles both the cases, may be let say PMA_updateCentralColumns() or something? 
I think we don't need to have two functions as updating one column is just a basic case of updating multiple column.

lem9:
@smita786 Yes but it depends on how different the logic would be between the case of one column and the case of many. If the resulting function is too long, it's not an improvement.

rakeshkaundilya:
Hi Smita,
For a single update function we have to change assigning of variables in 'edit_save' if condition and in assigning of  $def variable(currently in PMA_updateOneColumn used for getting the insert query for adding one column),Secondly assigning of variables which is currently done in PMA_updateMultiple function will be done alongside IF in db_central_columns.php.
So I will do as suggested however IMO different function is good as it is no confusing at all.
Thanks.

smita786:
@lem9 yeah I agree.
@rakeshkaundilya I guess then you should go with whatever makes the code cleaner and solve the duplicate code issue.

smita786:
Rakesh,
Looks good to me. please squash the commits.

db_central_columns.php
```diff
@@ -66,10 +66,26 @@
...
+        $response->isSuccess(false);
+        $response->addJSON('message', $message);
+    }
+}
 if (isset($_POST['delete_save'])) {
     $col_name = array();
-    $col_name[] = $_REQUEST['col_name'];
-    $tmp_msg = PMA_deleteColumnsFromList($col_name, false);
+    parse_str($_POST['col_name'], $col_name);
+
```
smita786: Please remove this extra blank line.

libraries/central_columns.lib.php
```diff
@@ -680,6 +719,27 @@ function PMA_getCentralColumnsTableHeader($class='', $title='', $actionCount=0)
...
+ *
+ * @param array $header_cells headers list
+ *
+ * @return string html for table header in central columns multi edit page
+ */
+function PMA_getCentralColumnsEditTableHeader($header_cells)
+{
+    $html = '<table id="table_columns" class="noclick">';
+    $html .= '<caption class="tblHeaders">' . __('Structure');
+
```
smita786: another blank line.

smita786:
Hi Marc,
You might want to have a look. 
Looks good to me now, I guess it'll be fine to merge after removing those blank lines and squashing the commits.
Thanks.

lem9:
@smita786 Looks fine for me. If you want to merge (after removal of blank lines) please assign yourself to this one; otherwise assign me.

smita786:
Thanks Rakesh. Good job!

rakeshkaundilya:
Thanks.

PR #1532, opened 2015-03-05
Fix for #4788 : NULL check box hidden under date picker.

chathurawidanage:


chathurawidanage:
Why do my commits show lots of changes while I did only 10 line changes.. And File changed tab shows correctly as 10 line edited?

lem9:
@chathurawidanage Your commits removed and then added whitespace (not trailing whitespace). Maybe your editor did this, but we cannot accept such commits.

PR #1534, opened 2015-03-05
Fix for #4788

daniyals:


daniyalme:
Signed-off-by: daniyal daniyal36@gmail.com

PR #1536, opened 2015-03-05
Fix for #4788

daniyals:


lem9:
Your commit contains no change. By the way, I only need a pull request done from a commit on the QA_4_3 branch and targeting the QA_4_3 branch.

PR #1541, opened 2015-03-05
#4788 :  NULL check box hidden under date picker.

chathurawidanage:
PMA_addDatepicker is called with $editArea as the element to add the date picker and onSelect fundtion was over ridden to set the selected date to the $input_field.
Signed-off-by: Chathura Widanage chathurawidanage@gmail.com

### DIFF
--- a/js/makegrid.js
+++ b/js/makegrid.js
@@ -1025,13 +1025,19 @@ function PMA_makegrid(t, enableResize, enableReorder, enableVisib, enableGridEdi
                     }
 
                     // add datetime picker
-                    PMA_addDatepicker($input_field, $td.attr('data-type'), {
+                    PMA_addDatepicker($editArea, $td.attr('data-type'), {
                         showMillisec: showMillisec,
                         showMicrosec: showMicrosec,
-                        timeFormat: timeFormat
+                        timeFormat: timeFormat,
+                        onSelect: function (dateText, inst) {
+                            $input_field.val(dateText);
+                            $input_field.data('comes_from', 'datepicker');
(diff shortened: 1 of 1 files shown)


lem9:
Thanks but we got a more simple fix for this.

zixtor:
I think we should prefer to have this fix for bug#4788

js/makegrid.js
```diff
@@ -1025,14 +1025,20 @@ function PMA_makegrid(t, enableResize, enableReorder, enableVisib, enableGridEdi
...
-                        timeFormat: timeFormat
+                        timeFormat: timeFormat,
+                        onSelect: function (dateText, inst) {
+                            $input_field.val(dateText);
+                            $input_field.data('comes_from', 'datepicker');
+                        },
+                        onClose: function (dateText, dp_inst) {
+                            // The value is no more from the date picker
+                            $input_field.data('comes_from', '');
+                        }
```
zixtor: Why not add altField: $input_field, instead of onSelect: and onClose:?

daniyalme:
When I double click on a default NULL value, the datepicker doesn't appear using this comit. Datepicker apears fine if theres an exisiting value.

chathurawidanage:
"When I double click on a default NULL value, the datepicker doesn't appear using this comit. Datepicker apears fine if theres an exisiting value."

Found and fixed this issue. Working on the time resetting issue...

chathurawidanage:
Fixed and updated

https://github.com/phpmyadmin/phpmyadmin/pull/1545

PR #1542, opened 2015-03-05
#4788 Inline edit of DATE fields with NULL

daniyals:

### DIFF
--- a/js/makegrid.js
+++ b/js/makegrid.js
@@ -1026,6 +1026,12 @@ function PMA_makegrid(t, enableResize, enableReorder, enableVisib, enableGridEdi
 
                     $input_field.datepicker("show");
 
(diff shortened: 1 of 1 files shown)


lem9:
Merged, thanks.

zixtor:
This fix unconditionally adds a margin-top to the datepicker div, which is not quite desirable and produces following problems:-
1. For those columns that do not have DEFAULT NULL, we get blank spacing between input field and datepicker, due to margin-top.
2. Based on the state of page scroll, if the datepicker div is displayed on top of the input field instead of bottom, adding margin-top covers the input field.

zixtor:
I missed it on first look, but above problems seem to be happening after you have once used the datepicker on a column with DEFAULT NULL.

daniyalme:
Updated: https://github.com/phpmyadmin/phpmyadmin/pull/1543

lem9:
@zixtor Can you look at the updated patch?

PR #1543, opened 2015-03-06
Update on #4788

daniyals:

Update on bug #4788, removed the margin appearing on NULL datepicker column


zixtor:
The margin-top thing is fixed but for me, the null-checkbox still remains hidden on the first attempt of double clicking a null date field. After first change of the value using datepicker, the next time null-checkbox is visible.

daniyalme:
When the value is set to null and the null check box is set to "display: none;" thats why is not appearing when the field already has null and I think it makes sense since an already null field will need a value not a null again to be input if it requires change

zixtor:
Indeed, but I meant a properly valued date field and null checkbox wasn't appearing on the first attempt. Anyway,  thanks for the follow up, the bug ticket stands resolved.

PR #1544, opened 2015-03-06
RFE #1001. While processing large queries, release PHP session lock

udan11:
RFE [#1001](https://sourceforge.net/p/phpmyadmin/feature-requests/1001/). While processing large queries, release PHP session lock
https://sourceforge.net/p/phpmyadmin/feature-requests/1001/

### DIFF
--- a/libraries/Util.class.php
+++ b/libraries/Util.class.php
@@ -4627,6 +4627,26 @@ class PMA_Util
 
         return $html;
     }
+
+    /*
+     * Restarts the session.
+     *
+     * @return boolean Whether the session restarted.
+     */
+    public static function sessionRestart()
+    {
+        // Prevents a second call of `php_session_send_cookie()`.
+        static $setup = false;
+        if (!$setup) {
+            ini_set('session.use_only_cookies', false);
+            ini_set('session.use_cookies', false);
+            ini_set('session.use_trans_sid', false);
+            ini_set('session.cache_limiter', null);
+            $setup = true;
+        }
+        return session_start();
+    }
+
 }
 
 ?>
--- a/libraries/sql.lib.php
+++ b/libraries/sql.lib.php
@@ -1103,6 +1103,8 @@ function PMA_storeTheQueryAsBookmark($db, $bkm_user, $sql_query_for_bookmark,
  */
 function PMA_executeQueryAndStoreResults($full_sql_query)
 {
+    session_write_close();
+
     // Measure query time.
     $querytime_before = array_sum(explode(' ', microtime()));
 
@@ -1110,6 +1112,8 @@ function PMA_executeQueryAndStoreResults($full_sql_query)
         $full_sql_query, null, PMA_DatabaseInterface::QUERY_STORE
     );
     $querytime_after = array_sum(explode(' ', microtime()));
+
+    PMA_Util::sessionRestart();
 
     $GLOBALS['querytime'] = $querytime_after - $querytime_before;
 


udan11:
It looks like Scrutinizer timed out because of "External Code Coverage" took [too long](https://scrutinizer-ci.com/g/phpmyadmin/phpmyadmin/inspections/2c1d98ab-ce6f-45ee-8bfc-fa9a20c571aa).

I'll push this again.

lem9:
Correct reference for this:
https://sourceforge.net/p/phpmyadmin/feature-requests/1001/

udan11:
Sorry for the confussion, I have updated the tile and commit message.

D-storm:
To me it seems to have some duplicated code which can be removed. Also, close PHP session sounds misleading.

udan11:
Thank you for your feedback! I amended my commit.

lem9:
Maybe this is a good solution but I have a doubt about performance. Just by clicking on Browse for a simple table, the session is closed and restarted about 25 times.

lem9:
By the way, next time please create a feature-specific branch instead of using master.

D-storm:
Do we really need to call `ini_set()` everytime before `session_start()`?

D-storm:
Maybe, instead of applying this to every query. We should do it before some certain queries which are expected to take some time and for the queries which are executed in SQL console/tab.

lem9:
@D-storm How can we predict the execution time?

D-storm:
@lem9 We cannot but we can release session locks before export and for custom queries executed via SQL console/tab as we don't know what kind of query user is trying to execute.

udan11:
I have amended my commit based on your suggestions. I reduced the number of `ini_set` calls and the session will only be closed on SQL queries incoming from the console or the SQL tab. The session lock is still being released for fast queries, but I cannot think of any way to predict the execution time.

Maybe, a possible solution would be to keep track of queries that took a long time and what tables and operations were involved.

udan11:
I have amended my commit based on your suggestions. I reduced the number of `ini_set` calls (they are being called one time only) and the session lock will only be released on SQL queries incoming from the console or the SQL tab. Fast queries are still affected by this modification, but I cannot think of any good ways to predict the execution time.

A possible solution would be to keep track of queries that took a long time and what tables and operations were involved. For example: `SELECT * FROM test` took 20 seconds -> any operations involving `test` should release the session lock. However, `SELECT * FROM test LIMIT 0, 10` will probably be fast -> any operations that affect a small set of rows (e.g. less than 1000 rows) should not release the lock.

udan11:
I have amended my commit based on your suggestions. I reduced the number of `ini_set` calls (they are being called one time only) and the session lock will only be released on SQL queries incoming from the console or the SQL tab. Fast queries are still affected by this modification, but I cannot think of any good ways to predict the execution time.

A possible solution would be to keep track of queries that took a long time and what tables and operations were involved. For example: `SELECT * FROM test` took 20 seconds -> any operations involving `test` should release the session lock. However, `SELECT * FROM test LIMIT 0, 10` will probably be fast -> any operations that affect a small set of rows (e.g. less than 1000 rows) should not release the lock.

L.E.: A better solution I came up with: ensure that sessions are no longer needed after a query was executed (no more reading / writing to $_SESSION).

Any suggestions?

udan11:
I have amended my commit based on your suggestions. I reduced the number of `ini_set` calls (they are being called one time only) and the session lock will only be released on SQL queries incoming from the console or the SQL tab. Fast queries are still affected by this modification, but I cannot think of any good ways to predict the execution time.

A possible solution would be to keep track of queries that took a long time and what tables and operations were involved. For example: `SELECT * FROM test` took 20 seconds -> any operations involving `test` should release the session lock. However, `SELECT * FROM test LIMIT 0, 10` will probably be fast -> any operations that affect a small set of rows (e.g. less than 1000 rows) should not release the lock.

L.E.: A better solution I came up with: ensure that sessions are no longer needed after a query was executed (no more reading / writing to $_SESSION).

Any suggestions?

lem9:
@udan11 Some usages of session, for example those with 'tmpval', are there to hold user choices and we cannot rely on the phpMyAdmin Configuration storage for these, as it's an optional component.

lem9:
Closing for now as it does not seem to be ready for merge.

udan11:
Do you have any ideas on this topic? In my opinion, this is the best solution.

lem9:
Dan, what do you mean? that your patch is the best approach, or that a "better solution" that you explained in your comment is the best approach?

udan11:
Ideally, the best approach is the one I put down in that comment. However, I do not think that the complexity of implementing such a system to predict execution time is practical, because of the complexity it requires.

Taking this in the account, I think that commit f97a62328797af8ddfa24ea22e62687fd7c639ec is the best solution available right now.

I am to happy to hear others ideas.

nijel:
Why not just do session_write_close / session_start? I don't really see need for ini_set, all the session parameters should be already configured by libraries/session.inc.php.

lem9:
@udan11 I retested f97a62328797af8ddfa24ea22e62687fd7c639ec and it's an improvement. Indeed the restart no longer occurs 25 times. In my tests, sessionRestart() is called when clicking on Go for the SQL query box, but also when clicking on Browse, any idea why?

udan11:
@nijel Those `ini_set` are required to ensure that the `PHPSESSID` cookie is not sent again, which causes troubles. Without those `ini_set` calls, PHP emits a warning `PHP Warning: session_start(): Cannot send session cache limiter - headers already sent in ...`.

@lem9 I am at the university right now. I am going to investigate as soon as I get home.

nijel:
@udan11 That's true, but mostly harmless as usually no output happens in these cases. Anyway I've fixed that in 1b1b3cd.

PR #1545, opened 2015-03-06
#4788 update

chathurawidanage:
Datepicker works now when textfield value is null :fixed
Null check automatically un checks when date is picked : added
Null check field and date picker includes in the same div : improved
Date/Time reset when double click : fixed

Signed-off-by: Chathura Widanage chathurawidanage@gmail.com

### DIFF
--- a/js/makegrid.js
+++ b/js/makegrid.js
@@ -1029,15 +1029,13 @@ function PMA_makegrid(t, enableResize, enableReorder, enableVisib, enableGridEdi
                         showMillisec: showMillisec,
                         showMicrosec: showMicrosec,
                         timeFormat: timeFormat,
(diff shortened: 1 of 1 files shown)


zixtor:
Hi Chathura,
I get js error "datetime_value is not defined", maybe you meant "current_datetime_value".
Also with this fix, when using the datepicker to change the value of a column of type 'date', I get the error "Note: #1265 Data truncated for column 'date' at row 1".

Anyway, this cosmetic bug was actually being caused as a side effect of an earlier fix, as I commented on the ticket; please consider this bug as fixed.
And thanks, we welcome you to contribute to phpMyAdmin with other bugs/features/improvements you may like to deal with.

PR #1547, opened 2015-03-06
Remove unneeded closing tags

znck:
Closing tags at the end of file are not mandatory. It is often suggested to omit to closing tags at the end of file to prevent unwanted effects as PHP will start output buffering if there is any character after closing tag.

Also all the changed files are pure PHP.

Reference: http://php.net/manual/en/language.basic-syntax.phptags.php

### DIFF
--- a/browse_foreigners.php
+++ b/browse_foreigners.php
@@ -58,4 +58,3 @@ $html = PMA_getHtmlForRelationalFieldSelection(
 );
 
 $response->addHtml($html);
-?>
--- a/chk_rel.php
+++ b/chk_rel.php
@@ -23,5 +23,3 @@ $response = PMA_Response::getInstance();
 $response->addHTML(
     PMA_getRelationsParamDiagnostic($cfgRelation)
 );
-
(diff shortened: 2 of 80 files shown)


drainx:
I don't see what actual use this commit has. Rather than just remove all the closing tags (standard code convention is to have closing tags,) why not just check for whitespace on existing files and remove if present?

If committed, future php files will have to follow this new convention of not using the closing php tag.

znck:
PSR-2 suggests not to close tags in PHP only files.

lem9:
Added to the team meeting agenda (meeting in two days):
https://wiki.phpmyadmin.net/pma/2015-03_Meeting

udan11:
I consider this a great idea. An even better idea would be to fully support the PSR-1 and PSR-2 standards.

abdotayb:
Great notice that will help more with php execiting

nijel:
I think it makes sense to follow general recommendations for code, so I think we can merge this.

PR #1551, opened 2015-03-08
RFE #1570, part 2 (2). Added the possibility of permanently using Enter instead of Ctrl+Enter in console.

udan11:
RFE [#1570](https://sourceforge.net/p/phpmyadmin/feature-requests/1570/), part 2 (2). Added the possibility of permanently using Enter instead of Ctrl+Enter in console.
https://sourceforge.net/p/phpmyadmin/feature-requests/1570/

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -144,6 +144,17 @@ Basic settings
     set, and the :config:option:`$cfg['UserprefsDisallow']` directive should
     contain ``'SendErrorReports'`` in one of its array values.
 
+.. config:option:: $cfg['ConsoleEnterExecutes']
+
+    :type: boolean
+    :default: false
+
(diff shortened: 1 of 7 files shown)


lem9:
Nice work. This was missing:
https://github.com/phpmyadmin/phpmyadmin/commit/e0f6853e4b8ea087938c7ee60093aaf285162b75

udan11:
Regarding e0f6853, I did not consider this option a vital part of the setup process and left it out intentionally. I guess I should not have done that. Thank you for correcting.

lem9:
It's true that this is not vital but we try to put everything there. It might be a way to "advertise" this feature.

udan11:
Oh, that is great idea! Thanks for the explanation.

PR #1553, opened 2015-03-09
Feature Request: Add voice commands for PMA!

djprawns:
You can now switch tabs using voice commands like Show Databases, Show Users.
Developer Suggestions needed to guide the development and future of
this feature.
This seems to me a crucial feature for a software like phpmyadmin
which is meant to make the user experience convenient when managing databases.

We could possibly add more features like Drop Database which would directly
lead us to a page where a list of databases would be listed
and we can choose which database to drop simply by naming it.

The possibilities are endless. And I am excited! :)

As Requested on http://sourceforge.net/p/phpmyadmin/feature-requests/1585/

Also please use microphone for best results.

### DIFF
--- a/index.php
+++ b/index.php
@@ -264,6 +264,16 @@ if ($server > 0) {
 
 echo '</div>';
 
+echo '<div class="group" id="voice_div">';
+    echo '<h2>' . __('Voice Command') . '</h2>';
+    echo '<input type="checkbox" onclick="voice_init()">Enable Voice Command';
+    echo '<div>';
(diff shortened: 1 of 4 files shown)


lem9:
@vanarpsharma See my comments in
https://sourceforge.net/p/phpmyadmin/feature-requests/1585/

lem9:
Under discussion on the phpmyadmin-devel mailing list.

lem9:
The team decided against this feature.

PR #1555, opened 2015-03-10
Tries to implement RFE #1614.

devenbansod:
Tries to implement RFE #1614

Feedback appreciated.


lem9:
Nice work.

PR #1561, opened 2015-03-10
Option for filling random data in table

sneh1234:
Also default value of Auto incremented value is shown in form.


lem9:
The signature is missing (see reference above) and the test suite is failing (click on Details then on a job number).

PR #1563, opened 2015-03-11
Per server iv

madhuracj:
Related feature request https://sourceforge.net/p/phpmyadmin/feature-requests/1561/

This allows working with multiple servers. However the exact steps listed in 'How to reproduce' does not work. In the last step you get a token mismatch due to the new session and you need to reload to continue working.

Anyway if you log into multiple servers in the same tab and then open new tabs to work on other servers everything works fine.

### DIFF
--- a/libraries/plugins/auth/AuthenticationCookie.class.php
+++ b/libraries/plugins/auth/AuthenticationCookie.class.php
@@ -442,7 +442,7 @@ class AuthenticationCookie extends AuthenticationPlugin
 
         // check cookies
         if (empty($_COOKIE['pmaUser-' . $GLOBALS['server']])
-            || empty($_COOKIE['pma_iv'])
+            || empty($_COOKIE['pma_iv-' . $GLOBALS['server']])
         ) {
             return false;
         }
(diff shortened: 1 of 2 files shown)


lem9:
@madhuracj I am following the scenario in the feature request (and that looks like a bug to me) but with this PR, when going back to the first browser tab and trying some actions, my session is immediately expired and I'm back to the login page.

madhuracj:
@lem9 
That is strange. I get a token mismatch at that step.

lem9:
@madhuracj Can it depend on the browser? I tested on Windows 7 with FF 36. Also, when I do the step "Log in to a second server" I use the server selector in the navi panel.

madhuracj:
I doubt, i just tried on FF 36 on Windows 8.1. I also used the server selector to log in to the second server.

lem9:
@madhuracj Do you get a token mismatch before your patch or after?

madhuracj:
I am sent to the login page before the patch and I get a token mismatch after the patch.

lem9:
@madhuracj So we could say that the patch is not ready?

madhuracj:
@lem9 
As I mentioned in the initial comment, this does not fully solve https://sourceforge.net/p/phpmyadmin/feature-requests/1561/. However having iv per server is needed to support login in to multiple servers at the same time. With this patch, you can log in to multiple servers from the same tab and then switch between them without a problem (which was not possible earlier). Also once you are logged in to multiple servers from the same tab, you can then open multiple tabs to work on each server.

lem9:
@madhuracj Good point.

PR #1564, opened 2015-03-11
BUG #4725. Added integrity check for navigation tree.

udan11:
BUG [#4725](https://sourceforge.net/p/phpmyadmin/bugs/4725/). Added integrity check for navigation tree.
https://sourceforge.net/p/phpmyadmin/bugs/4725/

Performs an integrity check on the navigation tree. The hash is computed server side and send along with the navigation tree. When the navigation tree is stored, so is the hash. For each page reload, the two hashes (the one sent by the server and the one stored) are compared and if they don't match, the new tree, sent by the server, is used, else the old tree, from the previous page.

This actually fixes another bug which caused the navigation tree not to update. It can be easily reproduced by following these steps: log in phpMyAdmin; log in another database managing software; create a new database in that other software and reload the phpMyAdmin page; the new database will not show unless you log out and log in again or if you press the little reload icon above the navigation tree.

This patch adds integrity check only for databases.

### DIFF
--- a/js/navigation.js
+++ b/js/navigation.js
@@ -474,7 +474,8 @@ $(function () {
         ) {
             navTreeStateUpdate();
         } else if (PMA_commonParams.get('server') === storage.server &&
-            PMA_commonParams.get('token') === storage.token
+            PMA_commonParams.get('token') === storage.token &&
+            $('#pma_navigation_tree_hash').html() === storage.navHash
         ) {
             // Restore the tree from storage
             $('#pma_navigation_tree_content').html(storage.navTree);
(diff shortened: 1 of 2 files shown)


js/navigation.js
```diff
@@ -474,7 +474,8 @@ $(function () {
         ) {
             navTreeStateUpdate();
         } else if (PMA_commonParams.get('server') === storage.server &&
-            PMA_commonParams.get('token') === storage.token
+            PMA_commonParams.get('token') === storage.token &&
+            $('#pma_navigation_tree_hash').html() === storage.navHash
```
zixtor: This condition will be true even if there is lack of integrity in storage.navTree which is the actual data we are going to put into the page, so I am not sure if it is enough, without making a js hash of storage.navTree and comparing it with hash from server-side...

zixtor: Also, storage.navTree gets updated when another database node is opened(other than the selected database), but the hash from server side will not include node structure of any other databases except the current one.. though I am not sure about it..

udan11: I mentioned that this method adds integrity checks only for databases list. Tables are retrieved on demand and that complicates the things a little bit. I did not find any "beautiful" solutions to that (yet).

zixtor: In view of the complications, I have removed the need for storing html in https://github.com/phpmyadmin/phpmyadmin/commit/dc291426105f448b5039ae4392afaf2a2dc03ee1

PR #1568, opened 2015-03-11
Related to bug#4725 and rfe#1426: maintain navigation state with just needed info in sessionstorage

atul516:
- actually reload the navigation panel on page refresh while maintaining it's state as well.
- avoid storing bulky html into sessionstorage, instead store just the required paths in the tree.
- remaining is to do the same for ShowDatabasesNavigationAsTree = false case.

Signed-off-by: Atul Pratap Singh atulpratapsingh05@gmail.com

### DIFF
--- a/js/navigation.js
+++ b/js/navigation.js
@@ -114,6 +114,53 @@ function collapseTreeNode($expandElem) {
 }
 
 /**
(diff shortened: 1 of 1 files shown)


zixtor:
Merged to QA_4_4!

PR #1573, opened 2015-03-12
Dark theme for console & Minor bug fixes in designer

hkandala:
Generally people like to have their terminal dark. So, to give a more terminal kind of look, added an option of switching to dark theme for the console. This is a minor UI improvement. Some more improvements are needed for console. I am currently working on some of them.

And also minor bugs in designer are fixed:
- Tables List having z-index of 100 which is greater than floating menu bar(99), So, I changed it to 98
- When Tables List is in right side cursor to resize is showing NW instead of NE. Fixed it too
- Supporting http://sourceforge.net/p/phpmyadmin/bugs/4792/ his statement, changed the menu Item to Save Page instead of Save Position.

Feedback appreciated!

Signed-Off-By Harish Kandala kandalaharish95@gmail.com

### DIFF
--- a/js/console.js
+++ b/js/console.js
@@ -110,6 +110,10 @@ var PMA_console = {
                 if(ConsoleEnterExecutes === true) {
                     $('#pma_console_options input[name=enter_executes]').prop('checked', true);
                 }
+                if(tempConfig.darkTheme === true) {
+                    $('#pma_console_options input[name=dark_theme]').prop('checked', true);
+                    $('#pma_console>.content').addClass('console_dark_theme');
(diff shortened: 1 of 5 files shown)


harish095:
Codacy check is saying PMA_console_config need to be in camel case but that object is already defined in that way. please someone have a look into it

lem9:
Please resubmit as different pull requests. We prefer to deal with one issue per pull request. Anyway, bug fixes should target the QA_4_3 branch. Instead of working on your master branch, create branches specific to your fix and to your new feature.

PR #1575, opened 2015-03-13
Bug #4798 Missing column when exporting in sql 

andreaval:
Export failure when the comments contains reserved keyword (KEY, CONSTRAINT)
To resolve I removed the comment from the current line before checking.

Signed-off-by: Andrea Vallorani andrea.vallorani@gmail.com


lem9:
Please resubmit without tab characters.

PR #1579, opened 2015-03-16
Add option for exporting pretty-printed JSON

addaleax:
This adds a single option to the JSON export plugin,
which, if available (PHP >= 5.4.0) corresponds to passing
`JSON_PRETTY_PRINT` to `json_encode`.


lem9:
Thanks.

PR #1580, opened 2015-03-16
Fix bug#4807

nisargjhaveri:
Don't add `.` if table name is not set when trying to get column in
sort expression.


lem9:
Thanks.

PR #1584, opened 2015-03-17
Minor UI bug fix in designer

hkandala:
Resubmitting #1573 as different pull requests

Minor bugs in designer are fixed:
- Tables List having z-index of 100 which is greater than floating menu bar(99), So, I changed it to 98
- When Tables List is in right side, cursor to resize is showing NW instead of NE. Fixed it too

Signed-off-by: harish095 kandalaharish95@gmail.com


lem9:
Thanks.

PR #1586, opened 2015-03-17
MySQL 5.7 Syntax for password change

hkandala:
http://sourceforge.net/p/phpmyadmin/feature-requests/1620/
Will use the preferred syntax for changing password i.e ALTER USER if MySQL version is >= 5.7.6

Signed-off-by: harish095 kandalaharish95@gmail.com

### DIFF
--- a/libraries/server_privileges.lib.php
+++ b/libraries/server_privileges.lib.php
@@ -1786,7 +1786,7 @@ function PMA_updatePassword($err_url, $username, $hostname)
             . 'PASSWORD';
 
         // in $sql_query which will be displayed, hide the password
-        if (PMA_MYSQL_INT_VERSION >= 50706) {
+        if (PMA_Util::getServerType() === 'MySQL' && PMA_MYSQL_INT_VERSION >= 50706) {
             $sql_query = 'ALTER USER \''
                 . PMA_Util::sqlAddSlashes($username)
                 . '\'@\'' . PMA_Util::sqlAddSlashes($hostname) . '\' IDENTIFIED BY \''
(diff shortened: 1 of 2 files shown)


nijel:
Have you tested this with MariaDB as well? I think it does not support ALTER USER, while it reports higher version than you're testing.

harish095:
No, I haven't tested with MariaDB. It would be a great help if you do that for me. I don't think it would be a problem as I am checking it with PMA_MYSQL_INT_VERSION constant

lem9:
This constant applies to both MySQL and MariaDB.

harish095:
So, is there any way to check if the database is MySQL?

lem9:
See libraries/Util.class.php in the getServerType() function. This is used in index.php.

harish095:
Used the getServerType() to check that server is using MySQL. Hope it resolves the issue. Thanks a lot Marc!

lem9:
Ok, it works when tested on MySQL 5.7.6, however there is a unit test failure, can you fix it? See the travis report.

harish095:
Unit test is saying that PMA_MYSQL_VERSION_COMMENT constant is undefined. I checked the DatabaseInterface.class.php where that constant is defined to see if there is a possibility such that constant is undefined. I can't find anything there. Please suggest me. I will try to fix it.

lem9:
I fixed the unit test : d9fdf6b9aa6a0dcd494ffebfa2e26e1f1eb0050f. To explain the context: when running unit tests, the code is not really connected to a MySQL server so we have to compensate. I also squashed your commits and put a proper signature: dbba36e0b84a62573be7422d3d863cfb0a1d6af0

harish095:
Thanks Marc!

PR #1589, opened 2015-03-18
AutocompleteTableSql

pratik:
Signed-off-by: Pratik Sethia pratiksethia1@gmail.com

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,7 +1,7 @@
 phpMyAdmin - ChangeLog
 ======================
 
-4.3.12.0 (not yet released)
+4.3.12.0 (2015-03-14)
 - bug #4746 Right-aligned columns have left-aligned header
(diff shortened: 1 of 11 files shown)


lem9:
Thanks for submitting. A few issues:
* the pull request should not contain other commits (I see 3c2a851)
* the new directive should not be added to config.sample.inc.php because we try to keep this file short
* the new directive needs to be added to libraries/config.default.php
* when I go to Settings > SQL queries, I get "Field EnableAutocompleteForTables has no type"
* when I go to /setup I don't see the new choice
* the autocomplete is not only for table names but also for column names, so the new directive's name and explanation should reflect this
* ideally, doc/config.rst should explain the new directive

lem9:
Thanks, we are almost there. I still cannot merge because of the first commit (3c2a851) so please redo your pull request, squashing your commits into just one and avoiding this unrelated commit 3c2a851.

PR #1590, opened 2015-03-18
Update functions.js

herndlm:
favorite tables localStorage sync: avoid sending undefined objects or empty strings via ajax

we had problems with a php 5.6.6 CGI/FastCGI suphp setup (internal server errors because of those ajax posts. I have not been able to verify why those errors occured, but a simple jquery ajax call like

$.ajax({cache: false, type: 'POST', data: { '': undefined }})

triggered them and the favorites sync is doing exactly the same when the localStorate is supported but empty

Signed-off-by: Martin Herndl martin.herndl@gmail.com

### DIFF
--- a/js/functions.js
+++ b/js/functions.js
@@ -3789,7 +3789,7 @@ AJAX.registerOnload('functions.js', function () {
             cache: false,
             type: 'POST',
             data: {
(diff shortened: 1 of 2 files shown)


lem9:
Thanks.

PR #1591, opened 2015-03-18
Bug #4718 simulate query fails

andreaval:
It is not an elegant solution, but it seems functional.
It avoids all the errors caused by the presence of functions in the query.
The $extra_where_clause variable was valorized but never appended to the WHERE clause. (I think that was a mistake)

Signed-off-by: Andrea Vallorani andrea.vallorani@gmail.com


andreaval:
sorry for the tabs, you back the correct version

lem9:
Please fix the failing unit test.

PR #1595, opened 2015-03-19
mysql documentation title/tooltip in navigation header

kppf:

### DIFF
--- a/libraries/Util.class.php
+++ b/libraries/Util.class.php
@@ -483,12 +483,12 @@ class PMA_Util
         $link, $big_icon = false, $anchor = '', $just_open = false
     ) {
         $url = self::getMySQLDocuURL($link, $anchor);
-        $open_link = '<a href="' . $url . '" target="mysql_doc">';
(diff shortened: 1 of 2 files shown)


madhuracj:
Thanks for your pull request. 
PMA_Util:showMySQLDocu() seems to be called from other places as well. So rather than changing it, I prefer to fix the problem like https://github.com/phpmyadmin/phpmyadmin/commit/a9220d89cfd308fe9e29c7c9b7e88895c908877b
Moreover, the previous approach in _links() which relied on preg_replace is not very elegant IMO.

kppf:
Yea, I thought to implement it like that, but then I wasn't pretty sure whether I can do that in a way that it didn't affect anything. Secondly, you would have left it for me to do :( Would have been a contribution in UI improvements for GSOC..

kppf:
Yea, I thought to implement it like that, but then I wasn't pretty sure whether I can do that in a way that it didn't affect anything. Secondly, you would have left it for me to do :( Would have been a contribution in UI improvements for GSOC..

PR #1597, opened 2015-03-19
Mark all option for sub categories while changing privileges

hkandala:
While setting priviliges there is a check all option but it would be better if there is a check all option for sub categories too. Screenshot below will explain it clearly. Feedback appreciated.

screenshot 42

Signed-off-by: Harish Kandala kandalaharish95@gmail.com

### DIFF
--- a/js/functions.js
+++ b/js/functions.js
@@ -4301,6 +4301,35 @@ $(document).on("change", "input.checkall_box", function () {
 });
 
 /**
+ * Watches checkboxes in a sub form to set the sub checkall box accordingly
+ */
+var sub_checkboxes_changed = function () {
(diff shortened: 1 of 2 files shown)


lem9:
Good idea. Just one thing to fix: the newly added message has to be translatable. Then, please squash your commits.

lem9:
Thanks. I need a further fix because adding a checkbox to the "Require SSL" section does not make sense.

harish095:
Hello Marc,
    I didn't add any checkbox for 'Require SSL' , that checkbox is already defined to enable or disable 'Require SSL' section. Suggest me If I understood wrong.

lem9:
Sorry, I was mistaken.

PR #1599, opened 2015-03-20
handling failed ajax request

kppf:

Two things:
1. If an ajax request fails, the [Loading...] notification is replaced by [Request Failed] and is dismissed after default time (5000ms)
1. Figured out that there was a bug which prevented any AJAX request once any AJAX request has failed. So corrected that too by making AJAX.active = false; and AJAX.xhr = null;

### DIFF
--- a/js/console.js
+++ b/js/console.js
@@ -184,6 +184,9 @@ var PMA_console = {
                     PMA_console.ajaxCallback(data);
                 } catch (e) {
                     console.log("Invalid JSON!" + e.message);
+                    PMA_ajaxShowMessage(PMA_messages.strRequestFailed);
+                    AJAX.active = false;
+                    AJAX.xhr = null;
                 }
             });
 
--- a/js/messages.php
+++ b/js/messages.php
@@ -257,6 +257,7 @@ $js_messages['strCancel'] = __('Cancel');
 /* For Ajax Notifications */
 $js_messages['strLoading'] = __('Loading…');
 $js_messages['strAbortedRequest'] = __('Request Aborted!!');
+$js_messages['strRequestFailed'] = __('Request Failed!!');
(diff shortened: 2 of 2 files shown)


js/console.js
```diff
@@ -184,6 +184,9 @@ var PMA_console = {
                     PMA_console.ajaxCallback(data);
                 } catch (e) {
                     console.log("Invalid JSON!" + e.message);
+                    PMA_ajaxShowMessage(PMA_messages.strRequestFailed);
+                    AJAX.active = false;
+                    AJAX.xhr = null;
```
zixtor: This catch handles the case when the server response is not JSON and the console tries to catch all requests to look for the SQL query, it sometimes actually happens as reported on the error reporting server..
So this may be give you false positives when in fact the request didn't fail but fetched html maybe..

kppf: Yes. I figured that. This catch also handles when request fails, and the response is empty (Parsing empty response throws an exception). So, either the AJAX request terminated due to server/internet issues or the server sent a wrong JSON - in both the cases, the request actually failed for the end user. So I guess this does handle both the scenarios well.

zixtor: Ok, makes sense to treat the two cases on par from user's perspective. But it is blocking the existing mechanism for aborting an ongoing request and firing a new one on new click. Please make necessary changes to accommodate that.

js/console.js
```diff
@@ -184,6 +184,9 @@ var PMA_console = {
                     PMA_console.ajaxCallback(data);
                 } catch (e) {
                     console.log("Invalid JSON!" + e.message);
+                    PMA_ajaxShowMessage(PMA_messages.strRequestFailed);
```
zixtor: Enclose PMA_messages.strRequestFailed in a div with class "error"

PR #1601, opened 2015-03-20
Feature Request #1625 Autocomplete Sql

pratik:
Gives an option to user to toggle autocomplete of table and column
names in config settings

Signed-off-by: Pratik Sethia pratiksethia1@gmail.com

### DIFF
--- a/db_sql_autocomplete.php
+++ b/db_sql_autocomplete.php
@@ -19,6 +19,8 @@ if ($GLOBALS['cfg']['EnableAutocompleteForTablesAndColumns']) {
             );
         }
     }
+} else {
+    $sql_autocomplete = true;
 }
 $response = PMA_Response::getInstance();
 $response->addJSON("tables", json_encode($sql_autocomplete));
--- a/db_sql_autocomplete.php
+++ b/db_sql_autocomplete.php
@@ -8,17 +8,17 @@
 
 require_once 'libraries/common.inc.php';
 
-$db = isset($_POST['db']) ? $_POST['db'] : $GLOBALS['db'];
-$sql_autocomplete = array();
-
-if ($db) {
-    $tableNames = $GLOBALS['dbi']->getTables($db);
-    foreach ($tableNames as $tableName) {
-        $sql_autocomplete[$tableName] = $GLOBALS['dbi']->getColumns(
-            $db, $tableName
-        );
+if ($GLOBALS['cfg']['EnableAutocompleteForTablesAndColumns']) {
+    $db = isset($_POST['db']) ? $_POST['db'] : $GLOBALS['db'];
+    $sql_autocomplete = array();
+    if ($db) {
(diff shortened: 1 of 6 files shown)


lem9:
Pratik,
tested with IE 11 and Firefox 36. In both cases, when the feature is deactivated via Settings, whenever I start to type in a SQL box, I get a Javascript error.

pratiksethia1:
Thanks Marc, I'll try to fix it

pratiksethia1:
After multiple tries, in addition to IE and Firefox, the error is even generating in chrome(39/MacOS).
In past also, some people reported same problems but the issue seemed to be resolved by clearing cache/cookies or by reinstalling. But these workarounds are not working in these case.

lem9:
Pratik,
with f7c0896 the bug is gone for me, but it does not seem a good idea to send one POST per character typed.

pratiksethia1:
May be i'm not right but It's still sending the whole table with the response.

lem9:
Pratik,
is your comment related to my remark about the one POST per character?

pratiksethia1:
Yes Marc.

lem9:
Well I don't understand it. In the current master, without your patch, a POST request is not sent for each character typed in the SQL box. With your patch, there is one POST request per character typed, do you find this normal?

pratiksethia1:
I agree it isn't normal. I'll try to figure something out.

nisargjhaveri:
@lem9, this is probably because the autocomplete function in JavaScript [(here)](https://github.com/phpmyadmin/phpmyadmin/blob/master/js/functions.js#L1730) keeps trying until it gets the table list on each character typed. The PR just prevents tables/columns list from server. So, on next character entry, it tries again.

@pratiksethia1, The correct way would be to not send a request to server at all if this is disabled.

pratiksethia1:
Yea i'll add a die function and submit patch. Thanks

lem9:
@nisargjhaveri Good point.

pratiksethia1:
Marc,
would it be okay if I store the value of $GLOBALS['cfg']['EnableAutocompleteForTablesAndColumns'] in a variable by including a script in the php file and then accessing the value in function.js , so that it becomes one time thing as suggested be @nisargjhaveri

lem9:
If you need to access at the js level something that comes from the PHP level, see https://wiki.phpmyadmin.net/pma/Page_loader#Accessing_common_parameters_via_JS

pratiksethia1:
Thanks, I went through it already.
But I implemented the issue with https://github.com/pratiksethia1/phpmyadmin/commit/c3e0a58ec16655513dd43bf525b7bb296997280f

pratiksethia1:
Its tested in all the browsers and environments.

lem9:
Thanks.

pratiksethia1:
Took a little time to find the best solution!
Thanks for guiding.

PR #1602, opened 2015-03-21
Possible fix for bug 4805. 

ganeshaditya1:
Disables inline editing those result tables that were generated by queries containing functions.

Signed-off-by: Aditya Sastry ganeshaditya1@gmail.com


lem9:
Sorry, this does not fix the root cause. See my remarks in the bug tracker.

PR #1605, opened 2015-03-22
Possible fix for bug 4805

ganeshaditya1:
Fixes the problem with unique column identification.

Signed-off-by: Aditya Sastry ganeshaditya1@gmail.com


lem9:
This does not fix the initial cause of the bug.

PR #1608, opened 2015-03-23
Handling failed AJAX requests

kppf:
- If an ajax request fails, the [Loading...] notification is replaced by [Request Failed] and is dismissed after default time (5000ms)
- Figured out that there was a bug which prevented any AJAX request once any AJAX request has failed. So corrected that too by making AJAX.active = false; and AJAX.xhr = null;
- Also handles the cases when the returned data is not JSON, which is a fail for the client. Shows [Request Failed] for this too.

Update from previous pull request: No more blocking the mechanism of aborting a new call.

### DIFF
--- a/js/console.js
+++ b/js/console.js
@@ -184,6 +184,12 @@ var PMA_console = {
                     PMA_console.ajaxCallback(data);
                 } catch (e) {
                     console.log("Invalid JSON!" + e.message);
+                    if(AJAX.cancellingRequest++ > 0) {
+                        PMA_ajaxShowMessage($('<div />',{class:'error',html:PMA_messages.strRequestFailed}));
+                        AJAX.active = false;
+                        AJAX.xhr = null;
+                        AJAX.cancellingRequest = 0;
+                    }
                 }
             });
 
--- a/js/ajax.js
+++ b/js/ajax.js
@@ -21,6 +21,11 @@ var AJAX = {
      */
     lockedTargets: {},
     /**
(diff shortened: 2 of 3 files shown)


zixtor:
Now, on two subsequent link clicks, it allows the second request to get fired without actually aborting the first one, so now both requests show their effect, which is not the behavior as before this patch.

js/console.js
```diff
@@ -184,6 +184,12 @@ var PMA_console = {
                     PMA_console.ajaxCallback(data);
                 } catch (e) {
                     console.log("Invalid JSON!" + e.message);
+                    if(AJAX.cancellingRequest++ > 0) {
```
zixtor: Here we can avail AJAX.xhr.statusText..?
Modifying above condition to (AJAX.xhr.status === 0 && AJAX.xhr.statusText !== 'abort') seems to work..

js/console.js
```diff
@@ -184,6 +184,12 @@ var PMA_console = {
...
+                        PMA_ajaxShowMessage($('<div />',{class:'error',html:PMA_messages.strRequestFailed}));
```
zixtor: Additionally, we can use statusText here also, to give more information to the user, like 'timeout' or 'error'.

js/console.js
kppf: Thank you. I was confused with the call flow. That's why did the wrong thing. Thank you. Sent another pull request.

PR #1609, opened 2015-03-23
Handling failed AJAX requests

kppf:
- If an ajax request fails, the [Loading...] notification is replaced by [Request Failed (statusText)] and is dismissed after default time (5000ms)
- Figured out that there was a bug which prevented any AJAX request once any AJAX request has failed. So corrected that too by making AJAX.active = false; and AJAX.xhr = null;
- Also handles the cases when the returned data is not JSON, which is a fail for the client. Shows [Request Failed (statusText)] for this too.

Update from previous pull request: No more blocking the mechanism of aborting a new call.

### DIFF
--- a/js/console.js
+++ b/js/console.js
@@ -184,6 +184,11 @@ var PMA_console = {
                     PMA_console.ajaxCallback(data);
                 } catch (e) {
                     console.log("Invalid JSON!" + e.message);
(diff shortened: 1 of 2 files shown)


zixtor:
Thanks! Merged to QA_4_4! https://github.com/phpmyadmin/phpmyadmin/commit/8847df101a13c9dfe2ff29b5758aedde4b630dd8

PR #1610, opened 2015-03-23
Fixed localization issue in <database> --> Operations window

Anshu-xLR:
As pointed out in [Bug #4816](https://sourceforge.net/p/phpmyadmin/bugs/4816/), fixed a tiny localization issue in <database> --> Operations window

Signed-off-by: Anshuman Verma anshuman@shuffleconnect.com

### DIFF
--- a/libraries/operations.lib.php
+++ b/libraries/operations.lib.php
@@ -237,7 +237,7 @@ function PMA_getHtmlForChangeDatabaseCharset($db, $table)
         $html_output .= PMA_Util::getImage('s_asci.png') . '&nbsp';
     }
     $html_output .= '<label for="select_db_collation">' . __('Collation')
-        . ':</label>' . "\n"
+        . '</label>' . "\n"
         . '</legend>' . "\n"
         . PMA_generateCharsetDropdownBox(
             PMA_CSDROPDOWN_COLLATION,


ibennetch:
Thanks for resubmitting as a pull request! Someone will review and comment or merge when they have an opportunity.

libraries/operations.lib.php
```diff
@@ -237,7 +237,7 @@ function PMA_getHtmlForChangeDatabaseCharset($db, $table)
         $html_output .= PMA_Util::getImage('s_asci.png') . '&nbsp';
     }
     $html_output .= '<label for="select_db_collation">' . __('Collation')
-        . ':</label>' . "\n"
```
ibennetch: Shouldn't this also move the colon to the translatable string?

lem9: We don't really need the colon; in fact, in the RFE, the user believed it would be more logical to remove the colon.

ibennetch: Then should we also remove the colons in some of the other labels? :)

lem9: Indeed; do you want me to work on this?

ibennetch: I won't have an opportunity until this evening so sure, go right ahead.

lem9: Done.

PR #1611, opened 2015-03-24
#4447 PMA_SQP_parse() unexpected behaviour

chunqi:
Changed parser behaviour for unrecognised punctuation strings of length > 1 to output default type ‘punct’ without suffix.

Signed-off-by: Zhu Chunqi


lem9:
I'm not sure we should touch the parser at this point, as there should be a GSoC 2015 project to rewrite the parser.

PR #1612, opened 2015-03-26
Feature Request: Export Progress

djprawns:

Modified progress_sess.php to return proper json formatted code.
Added modal activation if being run on chrome.

Note: Please test on Safari and IE, and add the browsers to line #389 of export.js if the progress bar works in them.

### DIFF
--- a/export.php
+++ b/export.php
@@ -33,6 +33,12 @@ if (!defined('TESTSUITE')) {
         }
         exit;
     }
+    usleep(500000);
+    $temp = 'Initializing Parameters';
+    $_SESSION['export_progress'] = $temp;
+    session_write_close();
+    session_id($_COOKIE['phpMyAdmin']);
+    session_start();
     /**
      * Sets globals from $_POST
      *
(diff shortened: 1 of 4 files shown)


lem9:
Pranav,
testing with Firefox 36 I see no export progress bar and no call to progress_sess.php.

vanarpsharma:
Yes @lem9 , It has been excluded, see line #389 of js/export.js. I believe you had told me a solution that excludes firefox in the mailing list to my question regarding the issues with firefox. I havent tested them on safari and ie either since i am using ubuntu and safari and ie dont install in them(properly).

vanarpsharma:
to test on safari and ie, you simply have to add BrowserDetect.browser=="your_browser_name" to line 389. So it will only work on specific browsers that support this functionality. For the unsupported browsers, we can open a ticket and fix the problem for them. If that is ok.

lem9:
Pranav,
I tested with Chrome 41. I see the progress bar but I don't find the result interesting. The bar does not reflect the real progression of the transfer and gives me no indication about what to expect for the total time. This is because during almost the whole transfer, I am stuck in "Building file".

vanarpsharma:
I have some queries: - 
1. Does the progress bar eventually finish with Done! message?
2. As far as the real progression of the transfer is concerned, I think that can be handled by introducing more messages. The total time however will need some thought. So should I introduce more messsages?

lem9:
@vanarpsharma Yes I got the "Done" message. About the bar progressing or not, well, a user would expect the bar to be progressing in a steady manner, otherwise the user thinks that the export is stuck.

vanarpsharma:
@lem9 , I can introduce more messages and changes to the progress bar so that the progression is steadier. May I make another pull request when I have something concrete?

lem9:
Sure. Ideally, the progress bar should represent what's really going on and give a good estimate of the completion.

PR #1617, opened 2015-03-28
Tries to fix bug-#4820

devenbansod:
Submitting the pull request to get (and may be fix) the Unit testing results. Currently don't have the unit testing environment setup locally.

Removed the str_replace on $anchor as it seemed to break the anchor links.

Please suggest changes.


lem9:
Nice work. I was wrongly commenting out two lines and the other line was producing unit test failures.

devenbansod:
Thanks. :)

Yes. I thought the same first and commented it on the bug ticket as that. But later realized that removing only the 2nd line is enough.

PR #1622, opened 2015-03-30
#763 jquery validation feature

whitepiegon:

### DIFF
--- a/js/jquery/additional-methods.js
+++ b/js/jquery/additional-methods.js
@@ -1,0 +1,940 @@
+/*!
+ * jQuery Validation Plugin v1.13.1
+ *
+ * http://jqueryvalidation.org/
+ *
+ * Copyright (c) 2014 Jörn Zaefferer
+ * Released under the MIT license
+ */
+(function( factory ) {
+    if ( typeof define === "function" && define.amd ) {
+        define( ["jquery", "./jquery.validate"], factory );
+    } else {
+        factory( jQuery );
+    }
+}(function( $ ) {
+
+(function() {
+
+    function stripHtml(value) {
+        // remove html tags and space chars
+        return value.replace(/<.[^<>]*?>/g, " ").replace(/&nbsp;|&#160;/gi, " ")
+        // remove punctuation
+        .replace(/[.(),;:!?%#$'\"_+=\/\-“”’]*/g, "");
+    }
+
+    $.validator.addMethod("maxWords", function(value, element, params) {
+        return this.optional(element) || stripHtml(value).match(/\b\w+\b/g).length <= params;
+    }, $.validator.format("Please enter {0} words or less."));
+
+    $.validator.addMethod("minWords", function(value, element, params) {
+        return this.optional(element) || stripHtml(value).match(/\b\w+\b/g).length >= params;
+    }, $.validator.format("Please enter at least {0} words."));
+
+    $.validator.addMethod("rangeWords", function(value, element, params) {
+        var valueStripped = stripHtml(value),
+            regex = /\b\w+\b/g;
+        return this.optional(element) || valueStripped.match(regex).length >= params[0] && valueStripped.match(regex).length <= params[1];
+    }, $.validator.format("Please enter between {0} and {1} words."));
+
+}());
+
+// Accept a value from a file input based on a required mimetype
+$.validator.addMethod("accept", function(value, element, param) {
+    // Split mime on commas in case we have multiple types we can accept
+    var typeParam = typeof param === "string" ? param.replace(/\s/g, "").replace(/,/g, "|") : "image/*",
+    optionalValue = this.optional(element),
+    i, file;
+
+    // Element is optional
+    if (optionalValue) {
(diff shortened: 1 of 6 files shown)


lem9:
Please squash your commits.

lem9:
I like this behavior, even if the message for entering a letter in an INT is a little weird. One issue is the translatability of messages. I know that jQuery validate has translations because I had to implement these in another project, can you have a look at this?

whitepiegon:
I didn't thought of condition based messaging. If you want I can look for it. Then, it might show different messages if letter is entered in INT. Presently, it simply returns message on basis of min and max for int.
Yes, jQuery form validate api supports translation. I will look for this. :)
Meanwhile, I squashed all.. Please see if anything else is required.
Thank you..

whitepiegon:
Hi Marc, I was looking for translations. It is quite easy in applying in jQuery validation api. It simply requires a javascript file with translated messages for different check events. Should I try to incorporate that and commit here only?

lem9:
Did you plan to use src/localization which is part of jquery-validation-1.13.1 ?

whitepiegon:
Yes I was taking about the same. But I have not planed that. I am asking if that should be tried?
Didn't you like that?

lem9:
I am not sure that including those regional files will fit with our translation system. Have a look at the bottom of js/messages.php to see what we do for datepicker.

whitepiegon:
Hi Marc.. I was going through that.. I must tell this is some real cool idea. I will love to implement it.
Thanks for telling me about that.
Also I would like to ask, if you are happy with the current error messages or you need them to be changed..?

lem9:
In general, let's trust this validation library and keep the original messages.

whitepiegon:
Hi Marc, for default functions messages of validation library will work, but what about costumed functions?
namely:-
validationFunctionForDateTime returns error message "Please enter a valid date or time"
validationFunctionForHex returns error message "Please enter a valid HEX input"
validationFunctionForFuns(function) returns error message "Error"
are you fine with these? If so then I will commit changes required for translatability of messages. It is done.

lem9:
Yes these messages look fine.

whitepiegon:
I don't know why I added that, when it was not even required. I removed that. Please see if all is fine now.

whitepiegon:
Please see if all is well with this commit. So, that I can squash my commits for complete testing.
Thank you

lem9:
Aayush,
in a DATE I enter "dsasda" and get "warning: no message defined..."

whitepiegon:
Please see if this commit is working fine.. 
I will remove trailing space before final squashing.

whitepiegon:
Sorry for my previous commit. Due to cache it might have shown me wrong behaviour.

lem9:
Looks fine, please squash so I can have a final look.

whitepiegon:
Hi Marc.. I was going through my this patch and I found one typical behavior which I would like to share.
In insert page we have two separate input form. Now, if user inputs some data in lower form, ignore checkbox is unchecked and inputs in lower form is also validated and suppose the input of lower form fails against the validation rule and error message is displayed. Now, if user checks the ignore checkbox and tries to submit, still the lower form is validated and it stops from submiting the form, since both form comes under same <form> element.
Now you tell me, if this behavior is fine or lower form should be validated only when ignore checkbox is unchecked. If later behavior is required, then I will made another commit for it.
Thank you

lem9:
Merged, thanks.
Indeed, please prepare another pull request to handle the case you described: if an insert section is ignored, the validation should be deactivated.

Also, some formatting of the generated js would help. Try visiting directly js/messages.php, you'll see that we generated end of lines for each line, but the lines you added do not have end of lines.

whitepiegon:
Hi Marc, I was trying to modify the use of default functions provided by this api and add the feature of validation only when insert section is not ignored. But I am not able to find some good way. However my work will be very easy if I use costumed functions for all field (this way a condition can be added to check if section is ignored, then return true).
So, what do you say, can I use costumed functions for all the fields (after disabling default functions of this api) namely for field type int and char/varchar?
Also I saw js/messages.php. Changes required are 1. line change at required places 2. end comment.. Anything else?

lem9:
Hi Aayush,
you mean using http://jqueryvalidation.org/jQuery.validator.addMethod/ ? So basically you would replicate in the custom validation the code of the normal validation, plus a check to see if the section is ignored? 

I would prefer if the "ignore" checkbox just disables jquery validation, but you say this is not possible?

whitepiegon:
No No.. This is possible.. Finally I found the way. :) I will send the pull request for desired behavior asap.
If I should make separate pull request or add commit in this merged branch only?

lem9:
You can continue in the same branch then open a new pull request. By the way, next time please identify clearly your branch, for example rfe763.

lem9:
@whitepiegon I just found a bug: for a TEXT column, I enter a few characters and get "Please enter no more than 1 characters"

whitepiegon:
ok.. I got that. I missed one condition. Thanks for reporting.

PR #1623, opened 2015-03-31
Ui imporve 1.0.0

getmebasilpaul:
To replace the default Loading...(or please wait...) ajax message with an animated circular loader.

image

### DIFF
--- a/js/functions.js
+++ b/js/functions.js
@@ -2027,7 +2027,7 @@ function PMA_ajaxShowMessage(message, timeout)
     )
     .hide()
     .appendTo("#loading_parent")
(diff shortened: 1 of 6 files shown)


lem9:
Thanks. We have seen what you can do in this pull request, but I'm not sure that we should include this library, just to produce this animation.

getmebasilpaul:
We can use this library for future improvements of User Interface. Using this library we can achieve material design in our Interface.

lem9:
I know, but a decision has not been made to go this way.

getmebasilpaul:
Thank you for the update.

PR #1624, opened 2015-03-31
Bug #4754 pMA DB not detected properly

madhuracj:
As per the discussion on February team meeting it was decided to check for PMA tables inside phpmyadmin database if such exists under ZeroConf functionality [1]. This is implemented here.

However, It was decide to look inside the current db prior to checking the phpmyadmin database. But the problem with this is that the user needs to go into some database to trigger this detection. So here, when the user is at the server level tables are looked  for in the phpmyadmin database.

[1] https://wiki.phpmyadmin.net/pma/2015-02_Meeting#Bug_discussion:_Zero_Conf

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -11,6 +11,7 @@ phpMyAdmin - ChangeLog
 - bug       Missing sort icons in monitor
 - bug #4805 Inline edit broken when using functions in query
 - bug #4821 Timed-out import fails to restart when file represented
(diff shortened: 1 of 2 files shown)


lem9:
Hi Madhura,
please squash.

madhuracj:
Squashed.

lem9:
Ok, by the way did you get my two emails?

lem9:
Looks fine, thanks.

PR #1626, opened 2015-04-01
BUG #4827. TEXT formatting does not work after inline editing.

udan11:
BUG [#4827](https://sourceforge.net/p/phpmyadmin/bugs/4827/). TEXT formatting does not work after inline editing.
https://sourceforge.net/p/phpmyadmin/bugs/4827/

### DIFF
--- a/js/makegrid.js
+++ b/js/makegrid.js
@@ -641,32 +641,41 @@ function PMA_makegrid(t, enableResize, enableReorder, enableVisib, enableGridEdi
                         $this_field.addClass('null');
                     } else {
                         $this_field.removeClass('null');
-                        var new_html = data.isNeedToRecheck
+                        var value = data.isNeedToRecheck
                             ? data.truncatableFieldValue
                             : $this_field.data('value');
 
+                        // Truncates the text.
+                        $this_field.removeClass('truncated');
+                        if (PMA_commonParams.get('pftext') === 'P' && value.length > g.maxTruncatedLen) {
+                            $this_field.addClass('truncated');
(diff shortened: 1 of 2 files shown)


lem9:
Thanks Dan. There is a small problem: after editing, $cfg['LimitChars'] is no longer applied. Its default value of 50 means that the content should be truncated to 50, followed by "...", as in the ticket screenshot.

udan11:
I missed that, sorry. It should work as expected now.

js/makegrid.js
```diff
@@ -641,32 +641,55 @@ function PMA_makegrid(t, enableResize, enableReorder, enableVisib, enableGridEdi
...
+
                         var selector = 'span';
                         if ($this_field.hasClass('hex') && $this_field.find('a').length) {
                             selector = 'a';
                         }
-                        $this_field.find(selector).html(new_html);
+
+                        // Reformats the content.
+                        var $target = $this_field.find(selector);
+                        var $code = $target.find('code.json');
```
madhuracj: We have at least two more CodeMirror base transformations that produce code.sql and code.xml. It would be better if a generic solution can be implemented for these.

udan11:
Alrighy. I was not aware that these are used too. I will implement them as soon as I get home.

udan11:
`code.xml` and `code.sql` are handled now.

lem9:
Looks fine to me, thanks. Next time, please work on bug fixes on the quality assurance branch (now QA_4_4).  I cherry-picked your commit to that branch.

PR #1629, opened 2015-04-04
Tries to fix bug#4837

devenbansod:
Added an ID to the notice div.
While printing a new div, it will always removes the old div (if it exists) and then adds it. Thus prevents multiple notices.


lem9:
Nice work

pratiksethia1:
Also solves : https://sourceforge.net/p/phpmyadmin/bugs/4757/#1be0

PR #1631, opened 2015-04-04
Prevents console window from moving out of the screen height

Sukhtaaj:
Signed-off-by: sukhtaaj sukhtaaj@gmail.com

Console window moves out of the screen when expanded upwards and there is no way to bring it back. Which leaves the user puzzled for how to get back the normal view. And remains the same even after logging in again.

Merging this pull request will prevent the console window from going above the serverinfo bar. ( HTML div with id = "serverinfo" )

console_window_bug

solved_console_window_bug

### DIFF
--- a/js/console.js
+++ b/js/console.js
@@ -456,6 +456,9 @@ var PMA_consoleResizer = {
      * @return void
      */
     _mousemove: function(event) {
+        if (event.pageY < 35) {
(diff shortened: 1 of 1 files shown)


lem9:
Thanks. As per our DCO, the signature should contain your full name.

PR #1633, opened 2015-04-05
additional feature for form validation

whitepiegon:
adds the feature of validation only when insert section is not ignored.
It also tries to change error message displayed for int type field when input is non-integer.


lem9:
Thanks. I'll accept this PR but it would have been better to clear any validation messages when the "Ignore" checkbox is ticked.

PR #1634, opened 2015-04-06
clear any validation messages

whitepiegon:
Hi Marc. This commit tries to produce feature you asked for.
Clear any validation message when the "Ignore" checkbox is ticked and reappear when checkbox is unchecked again.

### DIFF
--- a/js/tbl_change.js
+++ b/js/tbl_change.js
@@ -346,12 +346,12 @@ function applyFunctionToAllRows(currId, functionName, copySalt, salt, targetRows
  */
 AJAX.registerTeardown('tbl_change.js', function () {
     $(document).off('click', 'span.open_gis_editor');
+    $(document).off('click', "input[name^='insert_ignore_']");
     $(document).off('click', "input[name='gis_data[save]']");
     $(document).off('click', 'input.checkbox_null');
     $('select[name="submit_type"]').unbind('change');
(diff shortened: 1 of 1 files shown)


lem9:
Hi Aayush,
a good improvement but incomplete for two reasons. First, as tested in Iceweasel (Firefox), the field's outline remains red. Second, you refer to "insert_ignore_1" but there may be more than two insert sections.

whitepiegon:
I personally felt that red outline is good as it keeps reflecting that input is invalid however act of removal of error message reflect that it doesn't affect the form validation. This was my view.
Also Marc all these functions, I am using, are default functions. They behave like this. But I can surely write customed functions if u want. So tell me if u want that red outline to be removed? I will modify/develop functions for that.
And yes i will correct this patch as per your 2nd point also. Thank you for telling about this.

whitepiegon:
hi Marc, I was looking for fixing 2nd issue you mentioned. I feel it can be easily taken care of, if "#insert_ignore_1" is replaced by "input[name^='insert_ignore_']". Please confirm if 1st issue also need change. So, I will make final changes and complete this feature.

lem9:
I accept your red outline explanation. Please implement your solution for the 2nd issue.

lem9:
Thanks.

PR #1637, opened 2015-04-07
Add option to show/hide git revision on main panel

Kdecherf:
By default phpMyAdmin shows current Git revision if it finds a .git folder. It can be annoying when PMA is maintained in a private Git repository and we don't want to show this kind of data on the main panel.

This option lets the administrator hide the Git revision even if .git folder exists.


lem9:
Thanks.

PR #1639, opened 2015-04-09
reported bug removal for text field

whitepiegon:
I missed a condition and hence maxlength validation condition got added to text field. Added the required condition, which solved the issue.


lem9:
Thanks.

whitepiegon:
Marc I was thinking, this validation system can be expanded for other field type also, like decimal, double etc. Your views??

lem9:
Well, everywhere a silent truncation can occur, yes. But please work on your other priority ;)

whitepiegon:
:) I got your point.. I am trying to conc. on that..
May be after my exams end and before other affairs begin, I will expand and finish this feature completely.

PR #1642, opened 2015-04-12
Refactoring tbl_view.lib and add PMA\Template

gzzhanghao:
- PMA\Template is a Template class we created for front-end templating. It enable us to split logical php codes from the HTMLs. We first apply it on tbl_chart.lib.php.
- I re-factored tbl_view.lib by moving  PMA_getNewTransformationDataSql() and PMA_getExistingTransformationData() into SystemDatabase.class. Meanwhile, PMA_getColumnMap() was moved into DatabaseInterface.class.
- SystemDatabase.class and DatabaseInterface.class were changed to Singleton class.


test/libraries/PMA_tbl_chart_test.php
```diff
@@ -1,296 +0,0 @@
...
-        );
-    }
-
-    /**
-     * Tests for PMA_getHtmlForTableChartDisplay() method.
-     *
-     * @return void
-     * @test
-     */
-    public function testPMAGetHtmlForTableChartDisplay()
```
zixtor: I think this test could be kept in a modified form to check the overall chart html output, though I am not sure how easy and relevant it is to make a test for such long html outputs when using template system like this.

zixtor:
Good work, looks good to me for merging!

Tithugues:
Hi,
Shouldn't we use phtml extension instead of php in templates folder? So it would be clearer that those files are for rendering and not processing?

zixtor:
Hi, I agree that it will become clearer, but will it not create any initial issues for some users with servers at first not processing .phtml as php? Should we be bothered about that?

lem9:
I don't think we should add a requirement for web servers to support .phtml extensions.  See also http://wordpress.stackexchange.com/questions/116454/can-i-use-phtml-instead-of-php

ibennetch:
Please correct me if I'm wrong, but it doesn't seem the files in the `templates/`folder are meant to be called directly but will be included by the Template class. In this case, it doesn't matter what the webserver allows because to the webserver, it's just using libraries/Template.class.php which then loads the files in `templates/`.

I tend to agree with @Tithugues that it makes sense to use a different extension for content that is parsed directly by the template engine and not designed to be called directly.

Tithugues:
Thanks @ibennetch .

There is something tricky here. I never remember having troubles with phtml files. We are using these files in my company and I believe that we didn't do anything special to work with it.

As @ibennetch said, the purpose of the templates/ folder isn't to call directly files inside, but only to be a store for other scripts.
Maybe we could do some tests on different environment, especially with unit tests on Travis to be sure there won't be any issues with this.
Whatever the result is, another solution might be to name the files <<myfile>>.tpl.php.

zixtor:
I agree @ibennetch that we will be using templates via php include only and not directly, so all php code in it will get executed irrespective of extension, so we can surely use .phtml.
Some say it may be a security risk for some applications if there is sensitive content in such files when accessed directly in case one just knows the path and has permissions and without being processed as php by the server, but I think that doesn't apply in our case.

ibennetch:
It is a good point that directly calling these files will most likely lead to the full source being displayed. I can't think of a specific problem with that — anything directly interacting with my SQL would have to be interpreted PHP anyway – but it may be worth finding an alternative anyway.

Perhaps using `.inc.php` or `.tmpl.php` instead would be more satisfactory to the group?

Tithugues:
But can't we use a .htaccess to block the direct access to the `templates/` folder ?

ibennetch:
We shouldn't depend on `.htaccess` -- phpMyAdmin works on other webservers than Apache and some hosts don't allow their users to use their own `.htaccess` files.

gzzhanghao:
Hello everyone,
I am tend to use `phtml` as extension of template files.
Because it tells the web server not to execute the file when the file was accessed directly. Otherwise the web server will execute the file and in most case, it will derive errors, which may cause some unexpected results.
Further, there are no security risk in this case because everyone can see the source code of pma here. Therefore, we don't need to protect the source code from being leaked.
I will add this feature in the PR #1671, and I'm welcome for any suggestions :-)

PR #1643, opened 2015-04-13
Ignore ./vendor directory of composer

alexprut:
The _./vendor_ directory of composer should be ignored by Git, the build script or CI script should install composer dependencies when needed.


lem9:
Thanks.

PR #1645, opened 2015-04-13
Bug #4845 White screen (Cloudflare)

alexprut:
Bug fix: https://sourceforge.net/p/phpmyadmin/bugs/4845/

_The data-cfasync='true' is a custom attribute used by CloudFlare, I don't thinks that this solution is the best one, so in case the PR can be closed._


lem9:
Thanks but I intend to apply the patch with the original author info.

PR #1646, opened 2015-04-14
Bug #4853 Relation view between 1600 and 1780 px

madhuracj:
I suggest to remove media queries since the available space also depends on the width of the navigation panel.


lem9:
@madhuracj This code came from https://github.com/phpmyadmin/phpmyadmin/commit/4007c092 as asked in https://sourceforge.net/p/phpmyadmin/feature-requests/1479/. Will the removal of the code bring back the older problem?

madhuracj:
@lem9  Thanks for pointing that out. With https://github.com/madhuracj/phpmyadmin/commit/9a04e80561a11f6c04e7765e857915905bd696fe this should be fine.

lem9:
Looks fine, thanks.

PR #1652, opened 2015-04-19
Feature Request: - Export Progress Bar

djprawns:
Made changes to progress_sess.php to have string keys instead of constant values.

Export progress is now a 10 part process if rows exceed by 20. Also the export progress process goes from the server level down to the row level.
Progress Bar is also animated.
Currently with these changes, my php error log shows no errors.

### DIFF
--- a/export.php
+++ b/export.php
@@ -33,6 +33,10 @@ if (!defined('TESTSUITE')) {
         }
         exit;
     }
+    /*
+    * Session based message to indicate initialization of parameters
+    */
+    sess_message('Initializing Parameters','2');
     /**
      * Sets globals from $_POST
      *
@@ -266,6 +270,8 @@ if (!defined('TESTSUITE')) {
         PMA_fatalError(__('Bad parameters!'));
     }
 
+    sess_message('Building File','3');
+
     // Merge SQL Query aliases with Export aliases from
     // export page, Export page aliases are given more
     // preference over SQL Query aliases.
@@ -408,31 +414,44 @@ if (!defined('TESTSUITE')) {
             if (! isset($db_select)) {
                 $db_select = '';
             }
+            sess_message('Exporting Server','1');
+            $tbl_size = PMA_getTableSizeForServer($db_select);
             PMA_exportServer(
                 $db_select, $whatStrucOrData, $export_plugin, $crlf, $err_url,
                 $export_type, $do_relation, $do_comments, $do_mime, $do_dates,
-                $aliases
+                $aliases, $tbl_size
             );
+            sess_message('Server Export Done','1');
+            sess_message('Done!','');
+            sess_message_close();
         } elseif ($export_type == 'database') {
             if (isset($lock_tables)) {
+                $tbl_size = PMA_getTableSizeForDb($tables, $db);
                 PMA_lockTables($db, $tables, "READ");
                 try {
                     PMA_exportDatabase(
                         $db, $tables, $whatStrucOrData, $export_plugin, $crlf,
                         $err_url, $export_type, $do_relation, $do_comments,
-                        $do_mime, $do_dates, $aliases
+                        $do_mime, $do_dates, $aliases, $tbl_size
                     );
                     PMA_unlockTables();
+                    sess_message('Database Export Done', '1');
+                    sess_message('Done!','');
+                    sess_message_close();
                 } catch (Exception $e) { // TODO use finally when PHP version is 5.5
                     PMA_unlockTables();
                     throw $e;
                 }
             } else {
+                $tbl_size = PMA_getTableSizeForDb($tables, $db);
                 PMA_exportDatabase(
                     $db, $tables, $whatStrucOrData, $export_plugin, $crlf, $err_url,
                     $export_type, $do_relation, $do_comments, $do_mime, $do_dates,
-                    $aliases
+                    $aliases, $tbl_size
                 );
+                sess_message('Database Export Done', '1');
+                sess_message('Done!','');
+                sess_message_close();
             }
         } else {
             // We export just one table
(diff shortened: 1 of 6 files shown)


lem9:
Tested in Chrome 42 (Linux)
````
[Sun Apr 19 08:11:59 2015] [error] [client ::1] PHP Notice:  Undefined index: export_progress in progress_sess.php on line 5
[Sun Apr 19 08:11:59 2015] [error] [client ::1] PHP Notice:  Undefined index: percentage in progress_sess.php on line 6
````

vanarpsharma:
Marc,
Could you replace lines 5 - 10 with this: - 
if(isset($_SESSION['export_progress']))
{
$result = $_SESSION['export_progress'];
}
else
{
$result = '';
}
if(isset($_SESSION['percentage']))
{
$percentage = $_SESSION['percentage'];
}
else
{
$percentage = '';
}
and comment if any error. 
Pardon any inconvenience, I dont see any other way to debug since I am not getting any errors on my system. Also my progress bar is working properly.

vanarpsharma:
Also, maybe you could try adding session_write_close(); as the last statement in progress_sess.php to see any errors vanish.

lem9:
I tested bce99300db5b34991e3e300bef5ac843a1fbb0a1 with Chrome 42 (Win 7). Things are better: the bar is progressing (but I'll have to retest on Linux). The only problem now can be seen in the error log, which shows notices (maybe your PHP is not set to show notices). When I export I get 16 lines of:
````
 PHP Notice:  A session had already been started - ignoring session_start() in progress_sess.php on line 4
````

lem9:
Closed until further feedback.

vanarpsharma:
I have set my error_reporting to On in php.ini.
Also turned on notices.
Still not getting anything. Very odd. Let me dig deeper.

vanarpsharma:
Marc, how is it behaving on Linux?

vanarpsharma:
Hey Marc, could you reopen this pull, I needed to make some changes.
Or you can make yourself by replacing lines 3-4 with: -
if(!isset($_SESSION))
{
session_id($_COOKIE['phpMyAdmin']);
session_start();
}
Maybe the errors go away.

lem9:
Done.

lem9:
I'll be able to retest on Linux in a few hours.

vanarpsharma:
ok Marc, please test a65a9a6

lem9:
The PHP notices about the session are fixed, thanks. Could you squash your commits?

vanarpsharma:
Marc, could you guide me how I could do that. These commits I have made through this interface and not command line.

lem9:
OK, I'll do the squash.

lem9:
I am working on the unit tests failures but haven't found a solution yet.

lem9:
By the way, I'm not sure that progress_sess.php is safe. Maybe it can be manipulated to generate XSS.

vanarpsharma:
Okay, Marc In that case, I will have to research a little about XSS and add appropriate changes. Do we need the safeguards immediately? Or is this acceptable for merging?

lem9:
We won't merge code that might be prone to an attack.

vanarpsharma:
So I have done some initial search, will using echo htmlspecialchars(json_encode($arr), ENT_QUOTES, 'UTF-8');
stop the XSS attack?

vanarpsharma:
I have just tested, I dont think that will work. I will have to research and then make appropriate changes. If you could give me a case how XSS could be done, I would get some direction.

lem9:
Instead of looking how XSS could be done, can you use our usual token protection? This is done by including libraries/common.inc.php, which verifies that the valid token is part of the request. Then all calls to progress_sess.php would need the token as a parameter. With Firefox + Firebug, you can see the token that is passed, for example when entering export.php.

vanarpsharma:
Hey Marc, I am able to pass the token as parameter to progress_sess.php. and have included libraries/common.inc.php in progress_sess.php. However it seems including it renders the response as entire html page response. I have looked at the documentation, and figured out that I can get the response using 
$response = PMA_Response::getInstance();
$response->addJSON('message', $message);
I use it but still get the same html page as response. What am i doing wrong?

lem9:
Maybe this would help:
http://wiki.phpmyadmin.net/pma/Generating_pages_and_ajax_responses_in_pma4#Bypassing_the_framework

vanarpsharma:
Hey @lem9 ,
Made appropriate changes. Token protection is being used as common.inc.php is included. Also standard pma response for ajax request is being used.
Wouldn't have been possible without your valuable guidance.
Thankyou :)

lem9:
With 2352621c2826c048937315bbc62044c60efae40d the progress bar stops at about half the distance, while my export has completed normally. I got the same results twice when exporting a table that produces a 14 MB file.

vanarpsharma:
@lem9 ,
On my system, the only time this happens is when cancel the file location system popup. Otherwise it works fine. I have also tested this multiple times using the employees db and its tables.
Maybe I could use some db or table you are using? and test on my system. I can put it in my system through the mysql terminal.

lem9:
I am testing the employees MySQL sample database.

vanarpsharma:
As in you are going to test, or the problem has occurred using the MySQL employees database?

lem9:
The problem happens with employees.dept_emp.

vanarpsharma:
Hey @lem9 ,
Hopefully all should go well now.

lem9:
@vanarpsharma I tested c1f45705b2fe61290744b4ebe382349b71c72460. Now the progress bar reaches approximately half the distance, but I know that the export is almost complete. When the export has completed, the bar reaches the right end and the "Done" message is displayed, but the bar does not reflect the true progression. It's more evident when exporting employees.salaries.

vanarpsharma:
Hey @lem9 ,
Give me some time, working on it.

vanarpsharma:
@lem9 ,
The reason why we are not getting true progression is not because of exactly a bug but rather the fact that sometimes, the client misses getting one of the session values since the php script has already executed one of the previous sess_messages. So instead of 10 requests we get nine. and some of the percentage is lost. So one request fetches the 8th table request and the second fetches the 10th, missing the 10th.
So the strategy that I am now thinking for this problem is to append data to the session variable instead of overwriting it.
In the client, I could determine what I have already shown in the progress bar and only show the new stuff from the session variables. This shall fix the issue of the "true progression", since some data is being lost.
So I guess my question is, is what I have just discussed an acceptable solution. One of the features of this solution is that it does not have that much net overhead. I am trying to make minimum requests while trying to achieve a satisfactory level of progression. What the solution does guarantee is that the last messages will be pretty near the 100% mark, which I think is still in happening most of the cases. It is not precisely accurate right now, but almost. The modified solution discussed will take some time to implement; If you ok it, I shall proceed. Please let me know what other requirements you would like?

lem9:
Pranav,
when you mention "it's not precisely accurate right now, but almost" does not correspond to what I see: a progression to about the center of the bar, then a last chunk that completes the bar.

lem9:
Closing for now.

PR #1628, opened 2015-04-02
Work-in-progress: Add support for xz compression

madhuracj:
This is for https://sourceforge.net/p/phpmyadmin/feature-requests/1621/

I was not successful in properly installing php-xz and getting it to work. Wasted couple of hours struggling to get it to work. So this is not tested at all. I am submitting as a PR hoping either that someone will be able to test this or this will be a starting point for someone who's trying to implement.

PR #1653, opened 2015-04-21
Add support for xz compression

madhuracj:
Finally, I managed to get the php xz extension installed on a linux machine and test the implementation.

See also https://github.com/phpmyadmin/phpmyadmin/pull/1628

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -1861,7 +1861,12 @@ Export and import settings
     :type: boolean
     :default: true
 
-    Defines whether to allow the use of zip/GZip/BZip2 compression when
+.. config:option:: $cfg['XZDump']
+
+    :type: boolean
+    :default: true
+
+    Defines whether to allow the use of zip/GZip/BZip2/xz compression when
     creating a dump file
 
 .. config:option:: $cfg['CompressOnFly']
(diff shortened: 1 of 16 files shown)


udan11:
Did you have any problems or found any limitations of the extension while using it? I would like to know in order to implement it and to help you ship this feature as soon as possible.

madhuracj:
Not anything particular.
Btw, support for xz was implemented only for importing. For exporting I couldn't find a xzencode or a similar function.

udan11:
Ok. I hope that I will implement this as soon as possible.

I believe that it would be better if you used `xzread` instead of that output buffering trick and `xzpassthru`.

For writing, you could use `xzwrite`, which writes directly to a file not to a buffer.

madhuracj:
Indeed, xzread would be better. Wonder how we could incorporate xzwrite to out export mechanism.

madhuracj:
Pull request update with xzread

udan11:
Incorporating `xzwrite` into the export mechanism would require quite a lot of work. I believe that it would be so much easier for you and a much better solution if I wrote a `xzencode`, a function that behaves like `gzencode`. It would be basically a drop-in replacement.

I am busy with some exams for now, but I hope that I'll find some time to implement it as soon as possible.

udan11:
I implemented `xzencode` and `xzdecode` in `php-xz` v0.1.1.

PR #1654, opened 2015-04-23
fix bug #4867

smita786:
If last column (selected value) is dropped, "add column" select dropdown was selecting by default to first value "at begining of table". So to fix this, when column is being dropped, I have selected the last value.


lem9:
Thanks.

PR #1656, opened 2015-04-24
#4871 Error on creating table - Work in progress

madhuracj:
I wonder whether there is a simpler solution


madhuracj:
Closing in favor of a simpler and better solution proposed by @lem9  at https://sourceforge.net/p/phpmyadmin/bugs/4871/

PR #1657, opened 2015-04-27
Support for virtual/persistent columns - Work in progress

madhuracj:
https://sourceforge.net/p/phpmyadmin/feature-requests/1517/


madhuracj:
Implementation halted due to http://sourceforge.net/p/phpmyadmin/feature-requests/1663

PR #1661, opened 2015-05-05
Bug 4884: make clean removes all documentation files

ibennetch:

This seems right to me, but I'd appreciate a second look before merging.

Should we make a ChangeLog entry for this?


lem9:
Looks fine. No need for a ChangeLog entry as this is internal stuff.

ibennetch:
Thanks Marc

PR #1512, opened 2015-02-19
enum editor and pick from central column dialogues were going out of screen

smita786:
If there are lots of columns in central list, the "pick from central column" dialogue on table creation page going out of screen. Same for enum editor. 
Its working fine on QA_4_3. Git bisect on master says "Update jquery-ui to version 1.11.2" is first bad commit. 
I have set the maxHeight attribute for both the dialogues. Please let me know if there is a better way to fix this issue.
Signed-off-by: Smita Kumari kumarismita62@gmail.com

PR #1662, opened 2015-05-06
RFE #1485

devenbansod:

### DIFF
--- a/js/messages.php
+++ b/js/messages.php
@@ -61,6 +61,7 @@ $js_messages['strConfirmNavigation'] = __('You have unsaved changes; are you sur
 $js_messages['strDropUserWarning'] = __('Do you really want to revoke the selected user(s) ?');
 $js_messages['strDeleteCentralColumnWarning'] = __('Do you really want to delete this central column?');
 $js_messages['strDropRTEitems'] = __('Do you really want to delete the selected items?');
+$js_messages['strDropPartitionWarning'] = __('Do you really want to DROP the selected partition(s)? This will also DELETE the data related to the selected partition(s)!');
 
 /* For modal dialog buttons */
 $js_messages['strSaveAndClose'] = __('Save & Close');
--- a/js/tbl_operations.js
+++ b/js/tbl_operations.js
@@ -5,6 +5,7 @@ AJAX.registerTeardown('tbl_operations.js', function () {
     $(document).off('submit', "#copyTable.ajax");
     $(document).off('submit', "#moveTableForm");
     $(document).off('submit', "#tableOptionsForm");
+    $(document).off('submit', "#partitionsForm");
     $(document).off('click', "#tbl_maintenance li a.maintain_action.ajax");
     $(document).off('click', "#drop_tbl_anchor.ajax");
     $(document).off('click', "#drop_view_anchor.ajax");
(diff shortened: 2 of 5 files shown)


devenbansod:
Hi, 
Scrutinizer failed because **Code Coverage was not sent to Scrutinizer**.

About Travis, sorry but I am unable to understand how my changes caused the test in **PMA_FormDisplay_Tpl_Test::testDisplayInput** to fail. 

Any help would be appreciated.

ibennetch:
> Scrutinizer failed because Code Coverage was not sent to Scrutinizer.

I'm not sure about this one

> About Travis, sorry but I am unable to understand how my changes caused the test in PMA_FormDisplay_Tpl_Test::testDisplayInput to fail.

Right, this is a problem from several commits prior. You can see by clicking on the `Details` for Travis, then go to the `Build history` tab, looking down it looks like test #13345 is the first failure.

ibennetch:
Just an update, I'm having trouble testing this due to a fatal PHP error, but I don't believe the error is related to your commit. I'll try to look again in a few hours.

lem9:
Maybe related, see my comment:
https://github.com/phpmyadmin/phpmyadmin/commit/34c28378

lem9:
I suggest rebasing your branch because of a recent fix in master.

devenbansod:
Thanks @lem9. I did the rebasing. It passes the tests now.

ibennetch:
Testing this, I get the error message `#1512 - DROP PARTITION can only be used on RANGE/LIST partitions ` but I don't know much about the use of partitions in this context so it may be that I haven't set something up properly. I'll do further research and see what I can find.

devenbansod:
Yes. You are right. Thanks for pointing that out.

This two links say :
> Dropping a partition from a table that is partitioned by either RANGE or by LIST can be accomplished using the ALTER TABLE statement with a DROP PARTITION clause
https://dev.mysql.com/doc/refman/5.1/en/partitioning-management-range-list.html

> You cannot drop partitions from tables that are partitioned by HASH or KEY in the same way that you can from tables that are partitioned by RANGE or LIST. However, you can merge HASH or KEY partitions using the ALTER TABLE ... COALESCE PARTITION statement. 
https://dev.mysql.com/doc/refman/5.1/en/partitioning-management-hash-key.html

Will look into it more.

devenbansod:
Now, a DROP or COALESCE radio button is added according to the type of partition method used.
Workflow when DROP is used is clear that the Partition(s) selected will be dropped while in case of COALESCE, the number of Partitions selected will be coalesced.

devenbansod:
PMA_Operations_Test::testGetHtmlForPartitionMaintenance

This test printed output: Not supported query: SELECT PARTITION_METHOD FROM information_schema.PARTITIONS WHERE TABLE_SCHEMA = "db" AND TABLE_NAME = "table"

I am unable to figure out what is its meaning exactly. Any heads-up regarding this ?

zixtor:
It appears that you are using a new query via PMA_getHtmlForPartitionMaintenance, so to pass testGetHtmlForPartitionMaintenance, you should add the query to libraries/dbi/DBIDummy.class.php with a valid dummy output for the testsuite to use.

devenbansod:
Thanks alot @zixtor :)

Have added it as you suggested. Now passing the previously failing test.

ibennetch:
Since DROP PARTITION is destructive to data stored in that partition, I think we should also prompt the user (similar to when calling DROP DATABASE or TRUNCATE table) for confirmation. What do you think?

ibennetch:
Since DROP doesn't make sense when used with hash or key partitions, do you think it's possible to hide "Drop" from the radio button list in that case? I believe you already do that with "Coalesce" so hopefully it's rather easy to do.

ibennetch:
Otherwise this looks good and is close to being ready to merge. The functionality works great in my testing.

devenbansod:
Yes. Such a check should be done. Because user might not be knowing and he might delete it accidentally. 

Also, would a simple check on 'submit' like this would be okay ? Or do you suggest using the PMA_confirm as used when clicked on the 'Drop Table' Link ?
I was thinking of this.
 
```javascript
        if ( $operation_selected == 'DROP'){
            if (confirm("Are you sure to drop the selected Partition(s)?
                   \nThis will delete the data related to the selected partition(s)")) {
            }
            else {
                event.preventDefault();
            }
        }
```

ibennetch:
I would prefer to use PMA_confirm as in the case of trying to Drop a database, but don't feel particularly strongly about the point; you could probably convince me to change my mind if you have a good reason.

devenbansod:
Added the Confirmation when DROP partition is selected.

ibennetch:
This looks great to me. Nice work. I'll do a bit more testing and merge later tonight, but it looks fine.

ibennetch:
Good job.

devenbansod:
Thanks. :)

devenbansod:
Please see #1681 .

PR #1664, opened 2015-05-08
RFE 657 : Realigning privileges after DB/Tbl/Col/Proc Rename/Copy/Move

devenbansod:
Adding an option to realign the database specific, table specific, column specific and procedures specific privileges after Copying / Renaming the database.

### DIFF
--- a/libraries/operations.lib.php
+++ b/libraries/operations.lib.php
@@ -76,7 +76,12 @@ function PMA_getHtmlForRenameDatabase($db)
 
     $html_output .= '<input id="new_db_name" type="text" name="newname" '
         . 'maxlength="64" size="30" class="textfield" value="" '
-        . 'required="required" />'
+        . 'required="required" />';
+    $html_output .= '<input type="checkbox" name="realign_privileges" value="1"'
+        . 'id="checkbox_realign_privileges" checked="checked" />';
+    $html_output .= '<label for="checkbox_realign_privileges">'
+        . __('Realign Privileges') . '</label><br />';
+    $html_output .= ''
         . '</fieldset>'
         . '<fieldset class="tblFooters">'
         . '<input id="rename_db_input" type="submit" value="' . __('Go') . '" />'
@@ -198,6 +203,10 @@ function PMA_getHtmlForCopyDatabase($db)
         . 'id="checkbox_constraints" />';
     $html_output .= '<label for="checkbox_constraints">'
         . __('Add constraints') . '</label><br />';
+    $html_output .= '<input type="checkbox" name="realign_privileges" value="1"'
+        . 'id="checkbox_privileges" checked="checked" />';
+    $html_output .= '<label for="checkbox_privileges">'
+        . __('Realign Privileges') . '</label><br />';
     $html_output .= '<input type="checkbox" name="switch_to_new" value="true"'
         . 'id="checkbox_switch"'
         . ((isset($pma_switch_to_new) && $pma_switch_to_new == 'true')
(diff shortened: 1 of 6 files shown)


devenbansod:
Work in Progress. Done only for renaming/ copying Databases. 


Also discussion is going on with respect to few doubts on the ticket.

lem9:
I suggest rebasing your branch because of a recent fix in master.

devenbansod:
Thanks @lem9 . Did the rebasing!

ibennetch:
Great work, but there's one small issue. I think you need to reload (flush) the privileges after making the change. This can be done manually from https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/server_privileges.lib.php#L4246-L4264 and https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/server_privileges.lib.php#L3717-3722 but in this case we should do it automatically.

devenbansod:
Yes. Flushing(reloading) should be automatic. I will run the 'FLUSH PRIVILEGES' sql query after updating all the privileges.

And I think whenever we have realigned the privileges, we should append to the original success message of 'Table Successfully Copied/ Moved' that 'The privileges have been realigned and the changes have been flushed'.

ibennetch:
Good idea, but I don't think we need that level of detail -- "The privileges have been realigned" should be enough. I think the user assumes we'll also perform the flush in this case.

Can you also make the checkbox selected by default? I think this should be the default behavior when moving a table.

devenbansod:
Yes. That seems a valid point too.

And yes, I will make it checked by default. It's right. The user should not have to specifically remember to realign the privileges. It should be done by default just that we provide him an option to choose not to do it.

devenbansod:
Added the privileges option for Table's Copy/ Rename/ Move as well Columns Renaming. 

Please suggest changes.

tbl_operations.php
```diff
@@ -109,7 +109,19 @@
     $warning_messages = array();
 
     if (isset($_REQUEST['new_name'])) {
+        // Get original names before rename operation
+        $oldTbl = $pma_table->getName();
```
ibennetch: Is it a problem to use `$oldTable` here to be a bit more readable?

ibennetch: I wonder if we should use `$oldTable` to make it a bit more readable. What do you think?

libraries/operations.lib.php
```diff
@@ -1594,6 +1751,91 @@ function PMA_getQueryAndResultForPartition()
...
+ * Realign the privileges after renaming/moving a table
+ *
+ * @param string $oldDb  Database name before table renaming/moving table
+ * @param string $oldTbl Table name before table renaming/moving table
+ * @param string $newDb  Database name after table renaming/ moving table
+ * @param string $newTbl Table name after table renaming/moving table
+ *
+ * @return void
+ */
+function PMA_RealignPrivileges_renameOrMoveTbl($oldDb, $oldTbl, $newDb, $newTbl)
```
ibennetch: I wonder if using `PMA_RealignPrivileges_renameOrMoveTable` instead would improve readability here, too.

tbl_operations.php
devenbansod: Now that you pointed out, I agree that for tables, $table is far more clear than $tbl.

Also, it appears $table is the standard used in our code, too.

I will make the appropriate changes and commit back.

ibennetch:
Looks good to me. Can you squash the commits after which I think it's ready for merge.

devenbansod:
Squashed the commits.

ibennetch:
Nice work Deven.

devenbansod:
Thanks. :)

devenbansod:
Please see #1683.

PR #1668, opened 2015-05-10
Fix bug on result info display

drainx:
Reference image from : http://puu.sh/hIn2q/237524ef12.png for a text BLOB field.

Bisect determined 34c2837894d712ae0148be6aad385d3b3f77995a to be the cause.

Signed-off-by: Daniel Rix Drainx1@live.com


lem9:
Thanks. Next time, for a bugfix, please target the current QA branch (now QA_4_4). I will backport to QA_4_4.

PR #1669, opened 2015-05-10
Added configuration directive to define logo filename within the theme

ge0ffray:
$cfg['NavigationLogo'] can be set to define a custom filename for the logo displayed at top of navigation panel.

Signed-Off-By: Geoffray Warnants gwarnants+github@gmail.com

### DIFF
--- a/libraries/navigation/NavigationHeader.class.php
+++ b/libraries/navigation/NavigationHeader.class.php
@@ -81,7 +81,10 @@ class PMA_NavigationHeader
         }
 
         $logo = 'phpMyAdmin';
-        if (@file_exists($GLOBALS['pmaThemeImage'] . 'logo_left.png')) {
+        if (@file_exists($GLOBALS['pmaThemeImage'] . $GLOBALS['cfg']['NavigationLogo'])) {
+            $logo = '<img src="' . $GLOBALS['pmaThemeImage'] . $GLOBALS['cfg']['NavigationLogo'] . '" '
+                . 'alt="' . $logo . '" id="imgpmalogo" />';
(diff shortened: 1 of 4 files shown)


ibennetch:
Hi and thanks for your contribution, but I don't really see the point -- if someone wishes to change the icon, they can just replace it in-place...I don't see really see how this is helpful to a user. Is there are reason replacing `logo_left.png` is less convenient than this?

gwarnants:
Hello, I have a custom theme having multiple logo versions. This feature allow users to choose a custom logo by the config without having to hack the theme.

libraries/navigation/NavigationHeader.class.php
```diff
@@ -81,7 +81,10 @@ private function _logo()
         }
 
         $logo = 'phpMyAdmin';
-        if (@file_exists($GLOBALS['pmaThemeImage'] . 'logo_left.png')) {
+        if (@file_exists($GLOBALS['pmaThemeImage'] . $GLOBALS['cfg']['NavigationLogo'])) {
+            $logo = '<img src="' . $GLOBALS['pmaThemeImage'] . $GLOBALS['cfg']['NavigationLogo'] . '" '
+                . 'alt="' . $logo . '" id="imgpmalogo" />';
+        } elseif ($GLOBALS['cfg']['NavigationLogo'] != 'logo_left.png' && @file_exists($GLOBALS['pmaThemeImage'] . 'logo_left.png')) {
```
Tithugues: Why add the first test (difference)? If parameter equals 'logo_left.png', if the file exists you will go into first "if", if it doesn't exist, you won't go into first, neither second if.
Isn't it possible to check/set the file path before the if, so you won't duplicate the  creation?

ibennetch:
Thank you for your effort, but the developer team doesn't feel this is useful to the majority of users.

PR #1671, opened 2015-05-11
Updates on PMA\Template class

gzzhanghao:
Updates on Template class.
- Add `trim` method to Template class. Now the usage is changed to: 
  `Template::get($template_script_name)->render($data, [$trim = true]);`
  The `$trim` parameter in the render function indicate whether it should eliminate white spaces between HTML tags and its values. This feature is set to ON by default; You can disable it by simply passing `false`.
- `trim` is also provided as a static method.
- Update tests for the Template class.
- Add templates for db_designer.lib.php.
- Change extensions of templates to ".pthml".


Tithugues:
Hi,
Did you create this for a specific need?

libraries/Template.class.php
```diff
@@ -0,0 +1,44 @@
...
+    {
+        $template = static::BASE_PATH . $this->name . '.php';
+        try {
+            extract($data);
+            ob_start();
+            if (file_exists($template)) {
+                include $template;
+            } else {
+                throw new \LogicException(
+                    'The template "' . $template . '" not found.'
```
Tithugues: I think that we should use translation here.

zixtor:
@Tithugues It must be related to https://wiki.phpmyadmin.net/pma/GSoC_2015_Projects#Codebase_Improvements:_OOP_and_refactoring_-_Zhang

libraries/Template.class.php
zixtor: I think it is intended more for the developers if they do happen to type a wrong template name, so we may choose to skip translating it. Its thrown again in catch, so if a user does see a Fatal error, can just copy it and report to us, as user not expected to really do something about it.

zixtor:
@gzzhanghao Please update it such as to have only changes on top of the current master. (so it will just include addition of trim function and all template class unit tests)

zixtor:
@gzzhanghao It would have been better had you done separate things in a new PR, but for now, please update the title/description as well for this PR, maybe as mentioned in your commit message.

templates/designer/database_tables.phtml
```diff
@@ -0,0 +1,140 @@
...
+                    <td width="10px" colspan="3" id="<?php echo $t_n_url; ?>.<?php echo urlencode($tab_column[$t_n]["COLUMN_NAME"][$j]); ?>">
+                        <div class="nowrap">
+                            <?php if (isset($tables_pk_or_unique_keys[$t_n . "." . $tab_column[$t_n]["COLUMN_NAME"][$j]])) : ?>
+                                <img src="<?php echo $_SESSION['PMA_Theme']->getImgPath('pmd/FieldKey_small.png'); ?>" alt="*" />
+                            <?php else :
+                                $type = 'pma/Field_small';
+                                if (strstr($tab_column[$t_n]["TYPE"][$j], 'char')
+                                    || strstr($tab_column[$t_n]["TYPE"][$j], 'text')
+                                ) {
+                                    $type = '_char';
```
zixtor: This must be .= here and at other similar places.

zixtor:
Overall looks very good to me, make the changes as mentioned in above comments, and its ready to merge.

templates/designer/database_tables.phtml
```diff
@@ -0,0 +1,140 @@
...
+                                   title="select_<?php echo urlencode($tab_column[$t_n]["COLUMN_NAME"][$j]); ?>"
+                                   store_column_param="<?php echo urlencode($GLOBALS['PMD_OUT']["TABLE_NAME_SMALL"][$i]); ?>,<?php echo htmlspecialchars($GLOBALS['PMD_OUT']["OWNER"][$i]); ?>,<?php echo urlencode($tab_column[$t_n]["COLUMN_NAME"][$j]); ?>">
+                        </td>
+                    <?php endif; ?>
+                    <td width="10px" colspan="3" id="<?php echo $t_n_url; ?>.<?php echo urlencode($tab_column[$t_n]["COLUMN_NAME"][$j]); ?>">
+                        <div class="nowrap">
+                            <?php if (isset($tables_pk_or_unique_keys[$t_n . "." . $tab_column[$t_n]["COLUMN_NAME"][$j]])) : ?>
+                                <img src="<?php echo $_SESSION['PMA_Theme']->getImgPath('pmd/FieldKey_small.png'); ?>" alt="*" />
+                            <?php else :
+                                $type = 'pma/Field_small';
```
zixtor: Typo here, pma -> pmd..

gzzhanghao:
@zixtor The typo problem was fixed in the newest commit. Please merge :-)

scripts/update-po
```diff
@@ -18,7 +18,7 @@ xgettext \
     --from-code=utf-8 \
     --keyword=__ --keyword=_pgettext:1c,2 --keyword=_ngettext:1,2 \
     --copyright-holder="phpMyAdmin devel team" \
-    `find . -name '*.php' -not -path './test/*' -not -path './po/*' -not -path './release/*' | sort`
+    `find . -name '*.php' -o -name '*.phtml' -not -path './test/*' -not -path './po/*' -not -path './release/*' | sort`
```
zixtor: @gzzhanghao One more thing, this find now seems to be resulting in ./test/* files as well, using regex seems ok to me:-
find -regex '.*\.\(php\|phtml\)' -not -path './test/*' -not -path './po/*' -not -path './release/*' | sort | wc -l

gzzhanghao:
@zixtor Thank you for providing a idea on solving this problem. Now the problem should have been fixed :-)

zixtor:
Thanks!

PR #1672, opened 2015-05-12
Don't throw errors in case reports.phpmyadmin.net is down

tomsommer:
Signed-off-by: Tom Sommer mail@tomsommer.dk


lem9:
Thanks. Backported to QA_4_4 and ported to master.

PR #1673, opened 2015-05-12
Fix undefined index 'collation_connection'

tomsommer:
Make sure $GLOBALS[$collation] is set

Signed-off-by: Tom Sommer mail@tomsommer.dk


lem9:
Thanks. As this is a bug fix I cherry-picked to the QA_4_4 branch, then ported to master. Next time, please use the QA_4_4 branch for fixed.

PR #1674, opened 2015-05-14
Return HTML instead of echo-ing directly in FormDisplay

nisargjhaveri:
The config/FormDisplay was directly echo-ing content.

I changed it to return HTML, so that the same functions can also be used with PMA_Response.

Someone please confirm what I am trying to do. After that, I'll fix the tests and the merge conflicts.

### DIFF
--- a/libraries/config/FormDisplay.class.php
+++ b/libraries/config/FormDisplay.class.php
@@ -213,23 +213,25 @@ class FormDisplay
      *
      * @return void
      */
-    public function display($tabbed_form = false, $show_restore_default = false)
+    public function getDisplay($tabbed_form = false, $show_restore_default = false)
     {
         static $js_lang_sent = false;
+
+        $htmlOutput = '';
 
         $js = array();
         $js_default = array();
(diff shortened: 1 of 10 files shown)


lem9:
Yes, this is the correct way.

nisargjhaveri:
Updated.

lem9:
The buttons in the "Configuration file" section (like Display, Download) do not work for me.

nisargjhaveri:
Check now?

lem9:
Yes, the Download button works now

nisargjhaveri:
Fixed remaining test too.

lem9:
Ha, we like these green "All is well" reports !

lem9:
Nisarg,
just found a problem with this commit. Try Settings > Import: some incorrect text is displayed between the lines of menu tabs.

nisargjhaveri:
o.O Looking into it.

nisargjhaveri:
See #1676.

Missed one `echo`. Sorry for that. -.-

PR #1675, opened 2015-05-15
[Work-in-progress] Prototype of page related settings

nisargjhaveri:
Known issues,
- Proper localization support
- Specific to one page
- Tests might fail (didn't check)

To check, go to DB SQL page, you'll see a small icon on top-right corner.

Thoughts?

### DIFF
--- a/db_sql.php
+++ b/db_sql.php
@@ -10,6 +10,7 @@
  *
  */
 require_once 'libraries/common.inc.php';
+require_once 'libraries/config/page_settings.lib.php';
 
 /**
  * Runs common work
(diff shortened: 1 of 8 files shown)


lem9:
Nisarg,
the result is correct, but please do not duplicate code. I saw duplication from prefs_forms.php.

nisargjhaveri:
Yes, I started with copying `prefs_forms.php` to `page_settings.lib.php`. But even now, some of the code is changed and I'm expecting that this file will see a lot of more changes.

And yes, if we think that it is too similar in future, I'll try unifying it again.

PR #1677, opened 2015-05-16
[Work in progress] Page related settings

nisargjhaveri:
Major changes
- Updated `page_settings.lib.php` and how page related settings work
- Fixed some JS
- Added page settings in some more pages to have clearer idea

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -34,6 +34,7 @@ phpMyAdmin - ChangeLog
 + rfe #1668 Procedures window shift-click should select multiple rows
 + rfe #1669 Designer: "Sticky" menu option
 + rfe #1671 Directly show table comments in structure view
++ rfe #1559 Page-related settings
 
 4.4.8.0 (not yet released)
 - bug       Allow accessing visual query builder when pmadb is not configured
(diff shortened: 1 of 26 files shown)


lem9:
Nisarg,
what's your plan with this part of the RFE: "There are also other ways to set options. In Browse, we have Options, and other remembered settings like column visibility, column sorting and order of display. I am not sure if these mechanisms should be integrated or not to the new way."

nisargjhaveri:
Marc,
I've been thinking about this part of the RFE. After some thought, I am really confused even about the usefulness of this. And also, implementation. I was just planning to mail `-devel` asking for opinions. Expect an email in few hours.

nisargjhaveri:
I've pushed a couple of commits. Please check.

lem9:
Looks fine, please have a look at the merge conflicts.

nisargjhaveri:
Please check this now. If you think it is fine, I'll squash the commits. Anything else can you remember that is still left?

lem9:
Merging works now. The hint on the top icon is missing the hyphen. Also what about the settings related to the navigation panel?

nisargjhaveri:
I'll fix the hyphen. 

For settings related to navigation panel, should we have separate settings icon along with the five buttons at the top of the panel?

lem9:
Yes it looks like the best place to put this icon.

nisargjhaveri:
So, should I do that with this or later in different commit? Because it'll be a different, special case of page-related settings.

lem9:
You can squash now, I'll do a final review, merge and you can work for the navi panel afterwards.

nisargjhaveri:
I'll squash it on a different branch and open a PR.

PR #1680, opened 2015-05-17
Fix HTML issue.

Tithugues:
A <a> tag wasn't closed.

Thanks to our wonderful browsers, there are no error today, but this is still an error.


lem9:
Thanks.

PR #1683, opened 2015-05-19
'Realign Privileges' should be shown only when EDITING

devenbansod:
'Realign Privileges' option should not be shown while adding a new Column.

Correcting my commit in Pull #1664 .


ibennetch:
Nice catch

PR #1684, opened 2015-05-19
Some fixes in page-related settings

nisargjhaveri:
- Update 'original' theme
- Fix lock icon location when settings icon is not available


lem9:
Thanks.

PR #1685, opened 2015-05-20
Navigation panel settings

nisargjhaveri:
Show an icon in navigation panel for settings related to navigation panel.

I was stuck with this for a really long time, so there is slightly high chances of mistakes or missing out something. Please check. :)

### DIFF
--- a/prefs_forms.php
+++ b/prefs_forms.php
@@ -90,3 +90,9 @@ if ($form_display->hasErrors()) {
     <?php
 }
 echo $form_display->getDisplay(true, true);
+
+if ($response->isAjax()) {
+    $response->addJSON('_disableNaviSettings', true);
+} else {
+    define('PMA_disableNaviSettings', true);
+}
--- a/js/ajax.js
+++ b/js/ajax.js
@@ -333,6 +333,12 @@ var AJAX = {
                 } else if (data._menuHash) {
                     AJAX.cache.menus.replace(AJAX.cache.menus.get(data._menuHash));
                 }
+                if (data._disableNaviSettings) {
+                    PMA_disableNaviSettings();
+                }
+                else {
+                    PMA_ensureNaviSettings(data._selflink);
+                }
 
                 // Remove all containers that may have
                 // been added outside of #page_content
--- a/js/config.js
+++ b/js/config.js
@@ -423,7 +423,7 @@ function validate_field(field, isKeyUp, errors)
             args = [];
         }
         args.unshift(isKeyUp);
-        result = functions[i][0].apply(field[0], args);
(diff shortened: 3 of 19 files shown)


lem9:
Nisarg,
found a buggy scenario. Visit first "Settings > Navigation panel" from the main panel, then visit the new icon in the navigation panel. In Chrome I see lots of "Not a positive number" messages.

nisargjhaveri:
That error should be fixed now.

But, it still has some problems on settings page because we have same settings twice which uses same scripts. So, I need to fix those.

lem9:
Yes, this error is fixed. Will wait for your other fixes. By the way, maybe enlarge the dialog's width.

nisargjhaveri:
The biggest problem is, there may be more than one elements with same id on the page. On is settings panel and one in new navi settings. What do we do with this? Any thoughts? Possibilities that I think of.
1. Ignore them (keep them)
2. Remove navi settings icon if on settings page.
3. Change current structure to accommodate this (maybe a lot of changes).

lem9:
I would choose suggestion 2.

nisargjhaveri:
Mark, did you have a look at this?

lem9:
Not yet, but there is a unit test failure.

nisargjhaveri:
Yes, I'm working on that. I was talking about other functionalities.

lem9:
The functionality itself is correct, I am reviewing the code.

lem9:
The code looks fine, but see my comments.

lem9:
After login, I click on a database in the navi panel and get:
````
Notice in ./navigation.php#26
 Undefined index: getNaviSettings
````

**Comment by nisargjhaveri** [comment]
Fixed.

**Comment by lem9** [comment]
Nice job. Please squash.

File: prefs_forms.php [hunk 1]
```diff
@@ -90,3 +90,9 @@
     <?php
 }
 echo $form_display->getDisplay(true, true);
+
+if ($response->isAjax()) {
+    $response->addJSON('_disableNaviSettings', true);
+} else {
+    define('PMA_disableNaviSettings', true);
```
nisargjhaveri: Scrutinizer says that constant name should be in all caps. But I am not sure here because it won't be readable after that. Any suggestion?

lem9: Yes, please use all caps. See https://wiki.phpmyadmin.net/pma/Developer_guidelines#Coding_style which points to http://pear.php.net/manual/en/standards.naming.php

ibennetch:
Nicely done

nisargjhaveri:
Thanks. :)

PR #1686, opened 2015-05-20
RFE#946 : Change Collation w/o changing data and Confirm before changing

devenbansod:
Asks for confirmation if the column collation is changed.

If confirmed, changes the column from any datatype to BLOB and then change the datatype to the original datatype and collation to whatever was originally requested .

### DIFF
--- a/js/messages.php
+++ b/js/messages.php
@@ -63,6 +63,14 @@ $js_messages['strDeleteCentralColumnWarning'] = __('Do you really want to delete
 $js_messages['strDropRTEitems'] = __('Do you really want to delete the selected items?');
 $js_messages['strDropPartitionWarning'] = __('Do you really want to DROP the selected partition(s)? This will also DELETE the data related to the selected partition(s)!');
 $js_messages['strTruncatePartitionWarning'] = __('Do you really want to TRUNCATE the selected partition(s)?');
+$js_messages['strChangeColumnCollation'] = __(
+        'This operation will attempt to convert your data to the new collation. In rare cases,
+        especially where a character doesn\'t exist in the new collation, this process could cause the data to appear incorrectly under the new collation;
+        in this case we suggest you revert to the original collation and refer to the tips at '
+    )
+    . '<a href="%s" target="garbled_data_wiki">' . __('Garbled Data') . '</a>.'
+    . '<br/><br/>'
+    . __('Are you sure you wish to change the collation and convert the data?');
 
 /* For modal dialog buttons */
 $js_messages['strSaveAndClose'] = __('Save & Close');
--- a/js/tbl_structure.js
+++ b/js/tbl_structure.js
@@ -102,6 +102,7 @@ AJAX.registerOnload('tbl_structure.js', function () {
          * @var    the_form    object referring to the export form
          */
         var $form = $(this);
+        var field_cnt = $form.find('input[name=orig_num_fields]').val();
 
         /*
          * First validate the form; if there is a problem, avoid submitting it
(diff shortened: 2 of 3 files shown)


devenbansod:
Hi @ibennetch ,  can you please suggest some steps to fix this "Errored" Scrutinizer test ?

ibennetch:
It appears that the test needs to be run again, but Scrutinizer says I don't have the permissions to do so. @nijel or @lem9 can you help with this failed test run?

js/messages.php
```diff
@@ -63,6 +63,7 @@ function () {
 $js_messages['strDropRTEitems'] = __('Do you really want to delete the selected items?');
 $js_messages['strDropPartitionWarning'] = __('Do you really want to DROP the selected partition(s)? This will also DELETE the data related to the selected partition(s)!');
 $js_messages['strTruncatePartitionWarning'] = __('Do you really want to TRUNCATE the selected partition(s)?');
+$js_messages['strChangeColumnCollation'] = __('Do you really want to change this column collation ? This might corrupt the data and data might get lost permanently.');
```
ibennetch: This warning sounds pretty severe; can we catch failure scenarios to ensure that we avoid corrupting user data? On one hand I understand informing them of the worst-case scenario, but on the other hand we don't want to scare them away, either.

ibennetch: Also, could you remove the space before the question mark at  "collation ?"

devenbansod: Yes. I thought so but I was more inclined towards the worst-case scenario information. I thought if an user is changing the collation, he/she should understand what he/she is doing. 

I think about catching the failure scenarios, I was unable to think of a way to guarantee that. But the way we are proceeding, to BLOB first and then moving to requested column definition, we make sure that even if data is getting corrupted, he/ she can always change back to original collation to recover the data. 

I think instead of such a warning, we could include a message like 'This might corrupt the data but you may be able to (*or 'can' ?*) recover it by changing back to original collation'. 

What would you suggest ?

devenbansod:
Please see my reply to your comment on the 'Outdated diff'. Github has hidden it on the main screen. ;)

ibennetch:
You're not wrong that we should warn the user of the worst-case scenario, but telling them their data might get permanently lost by doing some action is not ideal. I think you're on the right track with your suggestion. How about:

> This operation will attempt to convert your data to the new collation. In rare cases, this process could cause the data to appear incorrectly under the new collation; in this case we suggest you revert to the original collation and refer to the tips at http://wiki.phpmyadmin.net/pma/Garbled_data.
>
> Are you sure you wish to change the collation and convert the data?

ibennetch:
Also, it looks like the update caused the tests to start over, so we're all good there.

lem9:
I would not use the expression "permanently lost" because this presumes that the user does not have backups. Of course, all users have backups ;)

ibennetch:
@lem9 did you see my suggested text, which was posted around the same time as your comment?

lem9:
@ibennetch Your suggested text looks good.

devenbansod:
Thanks @ibennetch . Your suggested text looks great!

I will change the warning text to your suggested one. Recommit and squash the commits.

ibennetch:
In the example Marc provided in the tracker, the é character doesn't seem to exist in hebrew_bin, causing corrupted data, so I wonder if the warning message can be improved further.

Maybe "In rare cases, especially where a character doesn't exist in the new collation, this process..."

ibennetch:
I think it's ready, can you squash?

devenbansod:
Squashed.

PR #1688, opened 2015-05-22
#4918 Date field popup dialog position bug

madhuracj:
@zixtor @D-storm 
I would value your opinion on this since both of you have worked on related bugs.


zixtor:
@madhuracj I confirm the fix works well and in my tests all other functionalities of datepicker work unaffected. Lets hope it was the last remaining problem in the otherwise fragile datepicker which looks stable at the moment ;)

PR #1693, opened 2015-05-24
Fix undefined PMA_messages and valiadtion in /setup

nisargjhaveri:


lem9:
Thanks for the hint, but I implemented a better fix in QA_4_4 and master:
https://github.com/phpmyadmin/phpmyadmin/commit/a723e049eb51a5f9344e295c5271be0979a4ce80

PR #1694, opened 2015-05-26
RFE#657 Correction : Reselect Original DB after Renaming Table

devenbansod:
Re-select back the Original DB after Renaming Table.

While renaming, we have selected the 'mysql' db for adjusting the privileges.

Please comment if it still does not work appropriately.


lem9:
Deven,
I have tested this PR and it fixes the problem that I reported.

PR #1695, opened 2015-05-26
Fix bug#4925

nisargjhaveri:
- Don't show page-settings when not needed in db_structure

Signed-off-by: Nisarg Jhaveri nisargjhaveri@gmail.com


lem9:
Thanks.

PR #1697, opened 2015-05-26
RFE701 Print View - Rewrite

devenbansod:
Parse the HTML of db_structure.php and sql.php send JSONs back to php scripts(db_printview.php and sql_printview.php respectively) and print the view out on new frame(here tab)
Please suggest changes. Let me know if I am on completely wrong track.

Tasks remaining:
Rewrite : tbl_printview.php
Remove now-obselete parts of : sql.php and its library functions in sql.lib.php

### DIFF
--- a/db_printview.php
+++ b/db_printview.php
@@ -6,9 +6,6 @@
  * @package PhpMyAdmin
  */
 
-/**
- *
- */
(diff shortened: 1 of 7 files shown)


devenbansod:
We are choosing a different method to do this (using pure CSS). So closing this.

PR #1698, opened 2015-05-27
Disable foreign key checks in SQL query box

nisargjhaveri:
The problem with this is, you cannot enable foreign key checks if global `foreign_key_checks` is `OFF`. Which is also the current behavior of "Disable foreign key checks" in import. 

I think, even when global `foreign_key_checks` is `OFF`, one should be able to turn it on by unchecking the checkbox. If it looks good, I'll implement that. Another option is to completely hide these checkbox when global is disabled.

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -1275,6 +1275,16 @@ Generic settings
     middle-clicking for pasting the clipboard contents in some Linux
     distributions (such as Ubuntu) is not supported by all browsers.
 
+.. config:option:: $cfg['DefaultForeignKeyChecks']
+
+    :type: string
+    :default: ``'default'``
+
+    Default values for checkbox for foreign key checks, to disable/enable
+    foreign key checks for certain queries. Value can be either of ``'default'``,
+    ``'enable'`` or ``'disable'``. If set to ``'default'``, the value of
+    MySQL variable ``FOREIGN_KEY_CHECKS`` is used.
+
 .. config:option:: $cfg['AllowUserDropDatabase']
 
     :type: boolean
@@ -2019,8 +2029,8 @@ Languages
     :default: ``'utf8_general_ci'``
 
     Defines the default connection collation to use, if not user-defined.
-    See the `MySQL documentation for charsets 
-    <http://dev.mysql.com/doc/mysql/en/charset-charsets.html>`_ 
+    See the `MySQL documentation for charsets
+    <http://dev.mysql.com/doc/mysql/en/charset-charsets.html>`_
     for list of possible values. This setting is
     ignored when connected to Drizzle server.
 
@@ -2817,4 +2827,3 @@ Developer
 
     Enable to let server present itself as demo server.
     This is used for <http://demo.phpmyadmin.net/>.
-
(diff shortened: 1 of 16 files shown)


lem9:
Nisarg,
I agree with your suggestion (being able to turn it on by unchecking the checkbox) but also test as a non-privileged user.

By the way, in this PR, you are adding two instances of the SHOW VARIABLES... logic. There should be just one instance in the whole codebase. In the current master, if I count correctly there are already 3 instances, while there should be just one.

nisargjhaveri:
I've updated the PR, please have a look.

We have a config option for import to disable foreign key check by default, this slightly conflicts with global variable. What should be the default value of checkbox? according to config option or global mysql variable? Also, should we now move it from import and make it global? or remove it?

lem9:
The config option (and related Setting) should stay, but be moved outside of the Import because it will apply not just to Import. Also, this option should have priority over the global one and the session one. Note that an unprivileged user may issue a SET foreign_key_checks, but not a SET GLOBAL foreign_key_checks.

Let me know if all of this makes sense.

nisargjhaveri:
This'll mean that we just ignore GLOBAL and override it for every query, according to the config option.

Is there any way to know if the option was explicitly set by user or it comes from default? Because I think that we should override global only when user is completely aware of that, and not just because we set a default value for this option.

nisargjhaveri:
Or maybe we can just have a dropdown with "default", "enable" and "disable".

lem9:
About "Is there any way to know if the option was explicitly set by user or it comes from default? ", are you talking about libraries/config.default.php, config.inc.php and the user preferences? If so, it does not matter because, if I recall correctly, when you check something under $GLOBALS['cfg'], the appropriate priority has already been applied.

nisargjhaveri:
I've removed the additional library file.
Config option part is left. Not completely convinced about what to do. But I'll get back after some experiments.

lem9:
Ok. Please rebase on the current master, I hope that the Travis error will be gone.

nisargjhaveri:
Please look now, tests are not fixed though.

lem9:
I issued a pull request (https://github.com/phpmyadmin/phpmyadmin/pull/1704) to fix the test (the failure was not caused by your code) and the test passed, so please rebase.

About your pull request, it works well but I just got an idea to make the checkbox more consistent with the rest of the UI. At some point we made an effort to use positive wording, so I would prefer that the label for the checkbox in the SQL panel (and elsewhere) be "Foreign key checks" and that you reverse its meaning. Ticking it corresponds to "enable".

nisargjhaveri:
I already rebased the commits. I'll change the words.

nisargjhaveri:
Sorry, I didn't rebase after the fix. I'll do so with other changes.

nisargjhaveri:
I've updated it, please have a look.

lem9:
Nice. Please put the checkbox to the left (it's more standard) and remove the colon. Also, the Enabled/Disabled is too much information; a user is supposed to know what ticking a checkbox means.

nisargjhaveri:
Updated.

lem9:
Nice work.

PR #1699, opened 2015-05-27
RFE#726 - Batch changing column collations

devenbansod:
Adds a checkbox in Table Options form in Table Operations.

If checked, runs the query -> 

`ALTER TABLE tbl_name CONVERT TO CHARACTER SET charset_name [COLLATE collation_name];`

### DIFF
--- a/js/messages.php
+++ b/js/messages.php
@@ -73,6 +73,16 @@ $js_messages['strChangeColumnCollation'] = __(
     . '<a href="%s" target="garbled_data_wiki">' . __('Garbled Data') . '</a>.'
     . '<br/><br/>'
     . __('Are you sure you wish to change the collation and convert the data?');
+$js_messages['strChangeAllColumnCollationsWarning'] = __(
+    'Through this operation, MySQL attempts to map the data values between collations. '
+    . 'If the character sets are incompatible, '
+    . 'there may be data loss and this lost data may <b>NOT</b> be recoverable simply '
+    . 'by changing back the column collation(s).'
+    . '<b> To convert existing data, it is suggested to use the column(s) editing feature '
+    . '(the "Change" Link) on the table structure page. </b>'
+)
+. '<br/><br/>'
+. __('Are you sure you wish to change all the column collations and convert the data?');
 
 /* For modal dialog buttons */
 $js_messages['strSaveAndClose'] = __('Save & Close');
--- a/js/tbl_operations.js
+++ b/js/tbl_operations.js
@@ -78,10 +78,35 @@ AJAX.registerOnload('tbl_operations.js', function () {
         event.stopPropagation();
         var $form = $(this);
         var $tblNameField = $form.find('input[name=new_name]');
+        var $tblCollationField = $form.find('select[name=tbl_collation]');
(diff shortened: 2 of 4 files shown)


cybernet:
it would be awesome to batch change tables collations :)

cybernet:
is this something we should worry about ?
https://scrutinizer-ci.com/g/phpmyadmin/phpmyadmin/inspections/ef805df7-1db8-45fe-8965-9daed93489a5/issues/files/tbl_operations.php?status=new&orderField=path&order=asc

devenbansod:
Hi, 
Do you mean to provide a mult_submit such that one may select one or more tables on db_structure.php and select an option 'Change all column collations' from 'With Selected' 
OR
Provide a similar checkbox in db_operations.php that will run the query for all the tables in database. ?

Also, one thing missing I found is I have to add is warning message to let user know that if the collations are not compatible then there might be a loss of data.

ibennetch:
This is an interesting idea, but I'm not sure it's needed because one can pretty quickly change collation by selecting all columns then use the Change link on the Structure page -- it's not changing them in a batch, but I'm not convinced we need a batch change for this.

devenbansod:
@ibennetch : Were you talking about the whole feature ? Or the addition that @cybernet suggested.

Because, I feel that the feature itself may be useful, since if the table has a large no of columns, selecting the collation-change in each column would be cumbersome.

This feature would make that cumbersome work easy.

devenbansod:
Only problem here would be if data lost, we may not be able to simply recover his data even if the user changes a column collation back. 

So the warning message should make that clear to the user along with the documentation line : 
> This operation uses `ALTER TABLE` to convert columns from one character set to another and MySQL attempts to map the data values. If the character sets are incompatible, there may be data loss and this loss may NOT be recoverable simply by changing back the column collation.

ibennetch:
Okay, on further review my comment doesn't make much sense.
I want to give this some more thought. There may be a different way to present this to the user that's more consistent with the existing interface.

ibennetch:
My concern here is that the user is looking at the operations that are available to be performed on the table itself, not individual columns. I'm considering that a better place for this may be as a popup link from the table structure page (along with "Unique," "Index," and "Add to central columns." I also considered whether we might just reword it to make it more clear (I think just saying "also change all column collations" or something similar would be enough). I like the idea of this but want to make sure we find the best place for it.

devenbansod:
On 3 Jun 2015 08:19, "Isaac Bennetch" <notifications@github.com> wrote:
>
> My concern here is that the user is looking at the operations that are
available to be performed on the table itself, not individual columns. I'm
considering that a better place for this may be as a popup link from the
table structure page (along with "Unique," "Index," and "Add to central
columns."

I think that would make it more confusing, may be for the users and may be
also developers then.

Because the links that you mentioned are, I guess, submit_mult links where
u select one or more columns and click this links.

Now this operation, we are talking about, is not related to selected
columns and is related to the table itself and will be applied to all the
columns of the tables nonetheless.

>I also considered whether we might just reword it to make it more clear (I
think just saying "also change all column collations" or something similar
would be enough). I like the idea of this but want to make sure we find the
best place for it.

devenbansod:
Made the changes and squashed.

PR #1700, opened 2015-05-28
Complete templates for table related scripts

gzzhanghao:
- Some tests are still in failure status because we didn't come up with good solution on front-end related scirpts test yet
- It's not perfect — that means we should keep improving these scripts in our future work


zixtor:
Very nice work! from next time please try using separate pull requests with more history and lesser sized commits. See my comments for some things which get broken when merging it.

zixtor:
templates/tbl_relation/foreign_key_row is not working, as I don't see input fields to add foreign keys in Table=>Structure=>Relation View.

zixtor:
Can't use Print View under Table=>Structure=>Print View (tbl_printview.php), though it works well at other places.

templates/index_form.phtml
```diff
@@ -0,0 +1,237 @@
...
+                'multilinestring',
+                'multipolygon',
+                'geomtrycollection'
+            ); ?>
+            <tbody>
+
+                <?php /* @var $column PMA_Index_Column */
+                foreach ($index->getColumns() as $column): ?>
+
+                    <tr class="$html .= $odd_row ? 'odd' : 'evennoclick">
```
zixtor: Typo 'evennoclick', noclick should be separate from this line.

templates/index_form.phtml
```diff
@@ -0,0 +1,237 @@
...
+                        <label for="input_index_comment">
+                            <?php echo __('Comment:'); ?>
+                        </label>
+                    </strong>
+                </div>
+
+                <input type="text"
+                       name="index[Index_comment]"
+                       id="input_index_comment"
+                       size="30"
```
zixtor: maxlength="1024" is missing here!

templates/index_form.phtml
```diff
@@ -0,0 +1,237 @@
...
+                       value="<?php echo htmlspecialchars($index->getComment()) ?>" />
+            </div>
+        </div>
+        <!-- end of indexoptions div -->
+
+        <div class="clearfloat">
+        </div>
+
```
zixtor: A \</div\> seems to be missing here!

zixtor:
About tests, as you intend to fix them in subsequent weeks as some clarity emerges on keeping HTML parts(or not) and other ideas, then for now I think you can comment out all the failing tests maybe with TODO remarks, so that Jenkins doesn't complain in the meanwhile as we merge it.

PR #1701, opened 2015-05-29
RFE#701 - Print View (using CSS) - Rewrite

devenbansod:
Rewrite of RFE#701 - Print View using CSS
Please suggest changes.

Operations remaining : Print View of Multiple Tables from db_structure.php


devenbansod:
Rebased but still fails the test. : 

PMA_Tracker_Test::testChangeTrackingData : Failed asserting that true matches expected 'executed_2'

ibennetch:
I've seen that Scrutinizer error quite often in the past few days but don't know the cause. I'll try to look at it in the morning and see if I can find anything.

About the travis one, I'm also kind of stumped on that but get different messages when run on my local machine. Hmmmmm....

ibennetch:
Try merging master now, Marc just merged a fix for testChangeTrackingData.

devenbansod:
Yes. Already rebased to current master. 
Thanks.

ibennetch:
Ha, nice speed on that

ibennetch:
I'll do some further tests on this before merging, but at first look everything looks quite good.

js/tbl_structure.js
```diff
@@ -244,6 +245,18 @@ AJAX.registerOnload('tbl_structure.js', function () {
     }); //end of Drop Column Anchor action
 
     /**
+     * Attach Event Handler for 'Print View'
+     */
+    $(document).on('click', "#printView", function (event) {
+        event.preventDefault();
+
+        // Unhide the indexes
+        $("a.ajax[href^=#indexes]").click();
```
ibennetch: Do you think it's better to force showing the indexes or to leave it shown or hidden based on what the user is already seeing?

ibennetch: I tend to think we should just leave it shown or hidden based on what the user is doing, but I could be convinced to change my mine.

ibennetch:
Looks good, can you squash the commits?

js/tbl_structure.js
devenbansod: Ya. On more thought, I too think that we should leave it as hidden/ shown according what the user wants. Also, we already have a [ Config option related to the slider ] (http://docs.phpmyadmin.net/en/latest/config.html#cfg_InitialSlidersState) that the user can use.

js/db_structure.js
```diff
@@ -345,6 +346,19 @@ AJAX.registerOnload('db_structure.js', function () {
...
 
+    /**
+     * Attach Event Handler for 'Print View'
+     */
+    $(document).on('click', "#printView", function (event) {
+        event.preventDefault();
+
+        // Fetch the real rows count
+        $('a.real_row_count').click();
+
```
devenbansod: By the way, 
this .click is done for getting the true number of rows for VIEWS. Do you suggest to keep this or remove this?

If we are keeping this, I had a little issue.
I am unable to find a method so that printPage is **called only after all clicks have been done and the row_counts have been fetched**.

Otherwise if I click 'Print View', it shows the 'Currently Working' symbol first because the row_counts have not been fetched yet. Then I have to close the Print Preview and wait for all counts to be fetched and again click 'Print View'.

Any help/ suggestions would be greatly appreciated.

devenbansod:
Hi, I just realized a few things.
1. Remove the now rendered unnecessary/ unused code relating to the earlier implementation of this feature (All of the printview things may not be removed as features like 'data dictionary' may also use the same piece of code).
2. Also, print.css was used by a few more features like 'data dictionary' and with the current changes, I have also removed the existing print.css content. I should not remove it completely but just that should move it to @media screen{} in print.css.

Do let me know if you realize any more changes.

devenbansod:
Closing this.

Please see #1708

PR #1708, opened 2015-05-31
RFE#701 - Print View using CSS - Rewrite

devenbansod:
Rewrote the implementation of **Print View** Feature. Now using CSS to design the printing of the page.


devenbansod:
Most of the related discussion can be found at : #1701

ibennetch:
In the other pull request we were discussing whether it was possible to get the exact view count and how to do so.

I didn't have a good answer to that question and was spending some time thinking about it and looking in to it.

Can we use an approximate count immediately?

devenbansod:
I am slightly confused here.

Because at present, the print view of Views does not mention the row - count at all.
[PDF of Print View on Demo Server](https://devenbansod.files.wordpress.com/2015/05/print-view-phpmyadmin-4-5.pdf)

And if we print the whatever is shown before those clicks, user might fall prey to wrong notions.

ibennetch:
I mean in the normal, screen-based view you can see an approximate row count, then if you click it you see the exact row count. 

Let's move discussion to the -devel mailing list. I'll send a message now.

ibennetch:
On the mailing list Madhura points out something I had overlooked; if the user wishes to get an exact row count they can do it manually before pressing print. So we should leave it how it is and not force the user to get the exact view count; just print whatever is displayed in the field whether it's approximate or exact.

Were there any other concerns about this pull request?

libraries/Header.class.php
```diff
@@ -658,6 +658,9 @@ private function _getLinkTags()
             $retval .= '<link rel="stylesheet" type="text/css" href="'
                 . $basedir . 'phpmyadmin.css.php?'
                 . 'nocache=' . $theme_id . $GLOBALS['text_dir'] . '" />';
+            // load Print view's CSS latest, so that it overrides all other CSSs while 'printing'
```
ibennetch: s/latest/last/

ibennetch:
I think we should move `printview.css` to the theme directory instead of the root; that way theme developers have the option to change the look of a printout.

devenbansod:
Yes. It makes sense to leave it as it is and let the user click if he/she wants to print the real rows. I will make the required changes.

devenbansod:
About the printview.css, that seems a valid point. I did not think of it before. I will move it to both the themes.

devenbansod:
Also, should I implement the 'Get all Row Counts' link in this feature itself, or should I open a new feature-request ticket and then make a new PR ?

devenbansod:
Now, Scrutinizer errored because it could not find the dependency **guzzle**

ibennetch:
Definitely as a new pull request, but it's beyond the scope of the requirements for 701 so you can work on it if you feel you have time as long as it doesn't interfere with your GSoC schedule :)

devenbansod:
Thanks. 
Opened a PR #1709

ibennetch:
I don't know about this guzzle issue, I'll have to look in to it later today.

ibennetch:
Great work on this. I like that it shows your ability to work through completely different parts of the codebase in different capacities, because this touches several different areas.

I hope you are pleased with your work, it looks good.

devenbansod:
Thanks. :)

Also, had made a small error. Corrected in #1711

ibennetch:
@devenbansod can you take a look at https://sourceforge.net/p/phpmyadmin/bugs/4957/ which may be related to this?

PR #1713, opened 2015-06-02
RFE#1359 - Ctrl Navigation in Grid Editor

devenbansod:
Added 'Ctrl + Arrow Key' navigation through the cells in JS Grid Editor in sql.php (Table Browse)

Suggestions/ Feedback welcome.

### DIFF
--- a/js/makegrid.js
+++ b/js/makegrid.js
@@ -604,8 +604,17 @@ function PMA_makegrid(t, enableResize, enableReorder, enableVisib, enableGridEdi
 
                     g.currentEditCell = cell;
                     $(g.cEdit).find('.edit_box').focus();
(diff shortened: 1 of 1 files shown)


ibennetch:
I emailed Deven about this, but for compatibility with Apple computers, which use the control key for the window manager, we should catch both `ctrlKey` and `altKey`. `metaKey` would also work, but `altKey` is consistent with existing behavior and the existing documentation (http://docs.phpmyadmin.net/en/latest/faq.html?highlight=ctrl#how-can-i-simply-move-in-page-with-plenty-editing-fields).

ibennetch:
Thanks, can you squash the commits?

devenbansod:
Squashed.

PR #1714, opened 2015-06-04
#4942 Export to gzip saves plain text under Chrome

madhuracj:
See my comments at https://sourceforge.net/p/phpmyadmin/bugs/4942/


lem9:
See my comments on the bugs tracker.

lem9:
Fixed by another commit.

PR #1718, opened 2015-06-08
Refactoring tbl_indexes.lib.php & tbl_chart.lib.php

gzzhanghao:
Creating a basic MVC structure for table related scripts.
In current stage, I will migrate some logic processing functions from tbl_*.lib.php to tbl_*.php for convenience. More works will come out in next few days.


zixtor:
Great, thoughtful strategy to begin with, should be helpful in long term, eager to see upcoming work. See my comments for some issues.

templates/preview_sql.phtml
```diff
@@ -0,0 +1,11 @@
...
+    <?php if (empty($query_data)): ?>
+        <?php echo __('No change'); ?>
+    <?php elseif (is_array($query_data)): ?>
+        <?php foreach ($query_data as $query): ?>
+            <?php echo PMA_Util::formatSql($query); ?>
+        <?php endforeach; ?>
+    <?php else: ?>
+        <?php echo PMA_Util::formatSql($query_data); ?>
+    <?php endif; ?>
+</div>
```
zixtor: getting "500 Internal Server Error" when using "Preview SQL" feature when creating an index under Table=>Structure..

gzzhanghao: @zixtor The error should have fixed now.

PR #1720, opened 2015-06-09
RFE#1594 : More plain english config words

devenbansod:
Replaces the existing use of 'script_names' like 'db_structure.php', 'sql.php' etc. as config options with more simple english words like 'structure', 'browse' etc.

Feedback welcome. Please suggest changes. 

### DIFF
--- a/libraries/config.values.php
+++ b/libraries/config.values.php
@@ -36,19 +36,19 @@ $cfg_db['OBGzip'] = array('auto', true, false);
 $cfg_db['MemoryLimit'] = 'short_string';
 $cfg_db['NavigationLogoLinkWindow'] = array('main', 'new');
 $cfg_db['NavigationTreeDefaultTabTable'] = array(
-    'tbl_structure.php', // fields list
-    'tbl_sql.php',       // SQL form
-    'tbl_select.php',    // search page
-    'tbl_change.php',    // insert row page
-    'sql.php'            // browse page
+    'structure' => __('Structure'), // fields list
+    'sql' => __('SQL'),             // SQL form
+    'search' => __('Search'),       // search page
+    'insert' => __('Insert'),       // insert row page
+    'browse' => __('Browse')        // browse page
 );
 $cfg_db['NavigationTreeDefaultTabTable2'] = array(
     '', //don't display
-    'tbl_structure.php', // fields list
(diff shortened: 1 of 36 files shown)


ibennetch:
I've begun reviewing this and so far everything looks good.

One note; we should also update `doc/config.rst` at least in the entries for
* `.. config:option:: $cfg['NavigationTreeDefaultTabTable']`
* `.. config:option:: $cfg['NavigationTreeDefaultTabTable2']`
* `.. config:option:: $cfg['NavigationTreeServer']`
* `.. config:option:: $cfg['NavigationTreeDatabase']`

I'll do some further tests and let you know if there are any problems.

devenbansod:
Added the docs.

ibennetch:
I believe we can improve the look of the setup form, which I think you were asking about by email, by making this change to `libraries/config.values.php`

<pre>
-    'structure', // fields list
+    'structure' => __('Structure'), // fields list
</pre>

And so on.

libraries/config.values.php
```diff
@@ -36,19 +36,19 @@
...
-    'tbl_structure.php', // fields list
-    'tbl_sql.php',       // SQL form
-    'tbl_select.php',    // search page
-    'tbl_change.php',    // insert row page
-    'sql.php'            // browse page
+    'structure', // fields list
+    'sql',       // SQL form
+    'search',    // search page
+    'insert',    // insert row page
+    'sql'        // browse page
```
ibennetch: Do you mean `'browse'` here?

libraries/config.values.php
```diff
@@ -36,19 +36,19 @@
...
```
ibennetch: And `'browse'` here as well?

ibennetch:
Nice work. If you can fix the three items listed above -- two 'sql' lines where I think you meant 'browse' and adding the `=>` syntax in `config.values.php`, then squash it all and I'll do one more quick review before merging.

devenbansod:
Thanks for pointing that out! I don't know how could I miss it!

Anyways, updated and squashed.

ibennetch:
Nice job Deven.

devenbansod:
Thanks.

devenbansod:
Please see #1729

devenbansod:
Also, please see #1739

PR #1721, opened 2015-06-11
Fix bug#4948

nisargjhaveri:


lem9:
Nisarg,
this does not fix the bug for me. To reproduce, open a database, click on SQL and type "EXPLAIN SELECT * FROM tablename" (of course, using a real table name)

nisargjhaveri:
Umm, it looks like it is working for me! Can you please check again?

lem9:
You see the results of the EXPLAIN statement? For me, it only shows "# Rows: 1".

nisargjhaveri:
Yes, I can see the result of EXPLAIN statement.
On previous commit a069ab2, it shows "# Rows: 1" as you said.

lem9:
When testing 7c5fb73 I don't see the result of EXPLAIN statement. Are you sure that you are *typing* the statement in the SQL query box?

nisargjhaveri:
Yes, login -> select DB from navi panel -> Click on SQL from tabs -> type EXPLAIN statement -> click on go.

Git revision: 7c5fb73 from issue_4948 branch
PHP version: 5.5.9-1ubuntu4.9

nisargjhaveri:
Okay, seems like it fails when Foreign key checks is unchecked. Looking into it. Is that the case with you too?

nisargjhaveri:
Doesn't work :/
Closing.

PR #1722, opened 2015-06-11
Control the ability to check for the lastest version available.

sibprogrammer:

### DIFF
--- a/libraries/Util.class.php
+++ b/libraries/Util.class.php
@@ -4391,6 +4391,10 @@ class PMA_Util
      */
     public static function getLatestVersion()
     {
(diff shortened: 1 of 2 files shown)


madhuracj:
This is already available. See https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/config.default.php#L607..L611

So, by adding $cfg['VersionCheck'] = false; in your config.inc.php you can disable version check

sibprogrammer:
Direct access to /version_check.php is not restricted via $cfg['VersionCheck'] variable. I'll re-work the patch to use this variable.

PR #1726, opened 2015-06-13
Foreign key checkbox for empty/drop in table operation

nisargjhaveri:


lem9:
Yes, a function would do.

nisargjhaveri:
Updated, please check now.

js/functions.js
```diff
@@ -1692,6 +1692,14 @@ function getForeignKeyCheckbox() {
     return html;
 }
 
+function getJSConfirmCommonParam(elem) {
```
nisargjhaveri: A better name?

lem9: This is a fine function name.

PR #1727, opened 2015-06-13
Decompose tbl_*.lib.php scripts

gzzhanghao:
I am going to create a TableController to handle logic just I mentioned before. So I decomposed them and move them into corresponding tbl_*.php files, as it can be easily extracted from tbl_*.php when I started to create the TableController class with these logic represented as methods. 

The decomposed tbl_relation.lib.php & tbl_columns_definition.lib.php will available later tonight.


gis_data_editor.php
```diff
@@ -89,10 +88,17 @@ function escape($variable)
...
+$_visualization = PMA_GIS_Visualization::getByData($data, $visualizationSettings);
+$visualization = null;
+if ($format == 'svg') {
+    $visualization = $_visualization->asSvg();
+} elseif ($format == 'png') {
+    $visualization = $_visualization->asPng();
+} elseif ($format == 'ol') {
+    $visualization = $_visualization->asOl();
+}
+
```
zixtor: After refactoring this condition appears at many more places, maybe we can treat it internally in a common public method of PMA_GIS_Visualization, passing $format while instantiating and optionally as an argument to the common method to accommodate this next line which is using asOl..

zixtor:
I will be testing the functionality tomorrow, you may add the remaining changes in the meanwhile..

zixtor:
About removing PMA_getHtmlForColumnAttributes (as said in https://jasondaurus.wordpress.com/2015/06/13/the-oop-refactor-project-weekly-report-5/), can't we just make a common columns_definitions/column_attributes template (which would include all other columns_definitions/* templates used by PMA_getHtmlForColumnAttributes)?

And great work overall, seems fine to merge after making said changes..

PR #1728, opened 2015-06-13
[WIP] Implementing sql-parser

udan11:
Uses [sql-parser](https://github.com/udan11/sql-parser) for parsing queries.

Work in progress.


lem9:
I will have more time to test tomorrow. Please have a look at the failing test.

udan11:
The failing test is related to a keyword that has been marked as forbidden in phpMyAdmin, but it isn't.

lem9:
According to http://dev.mysql.com/doc/refman/5.6/en/keywords.html the keyword "GENERAL" was a reserved word until MySQL 5.6.1. I guess you do not handle a list of keywords by version?

lem9:
Hmm, according to http://dev.mysql.com/doc/refman/5.5/en/keywords.html "GENERAL" became reserved in 5.5.3 and unreserved in 5.5.8. We could remove it from the list of keywords but there must be other cases where a list by version would make sense.

lem9:
Removed "GENERAL": 9acf48878da6288427ab50d76b2568dc44d5ac44

lem9:
Please have a look at Scrutinizer's issues.

udan11:
Multiple contexts will be defined soon for each version.

lem9:
Dan, I like the way that the library is taking shape and how it's integrated to the rest of the codebase.

udan11:
Thank you, Marc. I know there is so much room for improvement, but I am as well pretty proud of the library.

lem9:
Please rebase your branch to avoid merge commits.

lem9:
Dan, please have a look at Scrutinizer's issues (especially the major ones). For example, it reports that some properties does not seem to exist.

udan11:
Some of those errors related to properties are false-positives, because of the object inheritance. I'll try to find a fix for them.

udan11:
FIxed major issues. I still have to fix the documentation and duplication issues.

udan11:
Fixed most minor issues as well.

lem9:
Dan,
see also http://sourceforge.net/p/phpmyadmin/bugs/4970

udan11:
At this moment, I am replacing calls to the old parser from the `parse_analyze.inc.php` and `sql.lib.php`. This should fix that among other issues.

udan11:
Functions and files that rely on the old parser (because the new parser doesn't have a query builder, yet).

- [ ] ExportSql.class.php
- [x] import.php
- [x] sql.php
- [ ] sql.lib.php
    - [x] PMA_executeQueryAndGetQueryResponse
        - [x] PMA_isRememberSortingOrder
        - [x] PMA_handleSortOrder
        - [x] PMA_isAppendLimitClause
        - [x] PMA_getSqlWithLimitClause
        - [x] PMA_executeTheQuery
            - [x] PMA_isJustBrowsing
            - [ ] PMA_countQueryResults
        - [ ] PMA_getQueryResponseForNoResultsReturned
            - [x] PMA_isDeleteTransformationInfo
            - [ ] PMA_deleteTransformationInfo
            - [x] PMA_getMessageForNoRowsReturned
            - [ ] PMA_DisplayResults::getCreateViewQueryResultOp
        - [ ] PMA_getQueryResponseForResultsReturned
            - [ ] PMA_getHtmlForSqlQueryResultsTable
            - [ ] PMA_DisplayResults::getTable

lem9:
Trying database > query, choosing a table and a column, then updating the query, reports this:
````
Runtime Notice in ./libraries/sql-parser/src/Fragments/OptionsFragment.php#26
 Declaration of SqlParser\Fragments\OptionsFragment::build() should be compatible with SqlParser\Fragment::build(SqlParser\Fragment $fragment)
````

**Comment by udan11** [comment]
Fixed.

**Comment by lem9** [comment]
Dan,
in database > query, after I submit a query I get:
````
./libraries/sql.lib.php#670
 Undefined property: SqlParser\Fragments\WhereKeyword::$condition
````

**Comment by udan11** [comment]
Fixed that too.

**Comment by lem9** [comment]
Fixes confirmed.

**Comment by lem9** [comment]
Looks like sorting order is not remembered in the session (see PMA_handleSortOrder() ).

**Comment by udan11** [comment]
I redid my last commit and overwrote it.

From now on, I'll specify the current version of the library so you can track the changes to the library as well.

**Comment by lem9** [comment]
About renaming, I'm not sure. For example, what would happen to this?
````
class AlterFragment extends Fragment
````

Also, my previous question was mostly about "field" which was not generic enough. Maybe "fragment" should not be renamed, but "field" should become "component"  because it can mean column, function or subquery?

**Comment by udan11** [comment]
That would become:
```
class AlterOperation extends Component
```

I had doubts since the beginning about the name "Fragment" and now that you mentioned it I gave another thought about renaming them so they would reflect what they actually are and mostly get rid of the word "fragment".

lem9:
Dan,
what about "field"? any renaming?

udan11:
Yes, to `Expression`.

lem9:
Yes, `Expression` looks good.

udan11:
The only traces of the old parser are in the `PMA_Util` class. I hope that I'll finish replacing it tomorrow. After that, I will rename the fragments to something more appropriate.

lem9:
Great; after this we can think of merging this part.

udan11:
I removed the old parser completely. I didn't have too much time to check if it's alright, but, in theory, it should be fine.

Marc, can you also check out this [code](https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/Util.class.php#L1106). It looks dead to me, that's what I removed, but I would like you to confirm it.

lem9:
This does not look like dead code to me. Have you moved elsewhere the logic that changes the limit and the order clause?

udan11:
It looks like the test was failing because the test was bad, it contained one query that didn't make sense at all, but the old parser had no problem with it. 

Also, `ExportSql::getTableDef()` is using some regexes for parsing, so I'll have to implement the parser here.

lem9:
Dan,
do you feel we are ready to merge?

udan11:
I'd like to do more testing to the code I changed today. From my first tests, it seems to be fine, but I didn't get to test too much.

lem9:
Nice work!

PR #1731, opened 2015-06-14
Prohibit the ability to check version if it was disabled.

sibprogrammer:

### DIFF
--- a/libraries/Util.class.php
+++ b/libraries/Util.class.php
@@ -4491,6 +4491,10 @@ class PMA_Util
      */
     public static function getLatestVersion()
     {
+        if (!$GLOBALS['cfg']['VersionCheck']) {
+            return array();
+        }
+
         // wait 3s at most for server response, it's enough to get information
         // from a working server
         $connection_timeout = 3;


lem9:
Thanks but I'm not sure about this. The goal of this directive is "Enables check for latest versions using javascript on main phpMyAdmin page." and with this patch, you are making this directive control version check on pages different than the main page (setup and direct access to version_check.php).

sibprogrammer:
The goal of the directive "Enables check for latest versions using javascript on main phpMyAdmin page." is a visible part. The direct access to version_check.php in an invisible part. From this point of view it's ok to keep the current description for this directive. From feature complete point of view I need to hide version checking functionality completely (even via direct call to version_check.php).

There was another pull request that was rejected. I've introduced one more directive to control this behaviour:
https://github.com/phpmyadmin/phpmyadmin/pull/1722
It seems reasonable to control the whole behaviour via single directive (VersionCheck) instead of creating several directives that would control the only parts of the whole picture.

lem9:
Thanks.

PR #1734, opened 2015-06-15
Ability to contol the minimal MySQL version via config.

sibprogrammer:

### DIFF
--- a/libraries/common.inc.php
+++ b/libraries/common.inc.php
@@ -1027,10 +1027,10 @@ if (! defined('PMA_MINIMUM_COMMON')) {
         /* Log success */
         PMA_logUser($cfg['Server']['user']);
 
-        if (PMA_MYSQL_INT_VERSION < 50500) {
+        if (PMA_MYSQL_INT_VERSION < $cfg['MysqlMinVersion']['internal']) {
             PMA_fatalError(
                 __('You should upgrade to %s %s or later.'),
-                array('MySQL', '5.5.0')
+                array('MySQL', $cfg['MysqlMinVersion']['human'])
             );
         }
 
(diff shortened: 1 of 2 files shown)


lem9:
Not sure about this. The supported version is hard-coded because it corresponds to what we test and indicate on http://www.phpmyadmin.net/home_page/downloads.php. Please explain your need.

sibprogrammer:
I'm the one of developers of Plesk control panel - http://www.odin.com/products/plesk/
As far as you may know Plesk has an integration with phpMyAdmin. We have a bunch of patches and want to decrease the amount of patches in favour of customisations. We also will be glad to help phpMyAdmin project by contributing the code in the future (not only related to Plesk customisations).

Some Linux distros has MySQL which is old enough. E.g. CentOS 6 has MySQL 5.1 out of the box. We need to support a bunch of OSes where MySQL is lower than 5.5. In the other hand we have millions of end-users that are doing very trivial tasks like check the list of tables, view the table content and that's all. From our point of view it's much better to give these users the ability to accomplish their trivial tasks instead of blocking the whole functionality or providing an old version of phpMyAdmin (it's also a bad idea from security point of view). So we decrease the minimal supported version of MySQL for our integration.

You may not like the idea. But if the minimal supported version is hardcoded in config.defaults.php instead of common.inc.php nothing has been changed for you. But this change will make our integration life a little better (minus one patch and an ability to customise this behaviour).

The most elegant solution could be if phpMyAdmin could work on MySQL starting from 5.0 and use the concept known as "graceful degradation" (is some feature is not supported by particular MySQL version, then there is no such feature in UI). But of cause it will lead to much more efforts on development and testing.

So instead of asking you to implement the true "graceful degradation" I want to ask you to give us the ability to control minimal supported version.

lem9:
Thanks for the nice justification.

PR #1736, opened 2015-06-15
[WIP] SQL debugging in console

nisargjhaveri:
Known issues and remaining tasks
- [x] Have to think something about self-handled requests
- [x] Have some problem with newly refreshed page
- [x] Update `original` theme
- [x] Cleanup code for old mechanism

### DIFF
--- a/js/ajax.js
+++ b/js/ajax.js
@@ -348,7 +348,6 @@ var AJAX = {
                     .not('#page_nav_icons')
                     .not('#page_content')
                     .not('#selflink')
-                    .not('#session_debug')
                     .not('#pma_header')
                     .not('#pma_footer')
                     .not('#pma_demo')
@@ -403,9 +402,6 @@ var AJAX = {
                 var msg = '';
                 if(data._errSubmitMsg){
                     msg = data._errSubmitMsg;
-                }
-                if (data._debug) {
-                    $('#session_debug').replaceWith(data._debug);
                 }
                 if (data._errors) {
                     $('<div/>', {id : 'pma_errors'})
--- a/js/console.js
+++ b/js/console.js
@@ -124,6 +124,7 @@ var PMA_console = {
             PMA_consoleInput.initialize();
             PMA_consoleMessages.initialize();
             PMA_consoleBookmarks.initialize();
+            PMA_consoleDebug.initialize();
 
             PMA_console.$consoleToolbar.children('.console_switch').click(PMA_console.toggle);
             $(document).keydown(function(event) {
@@ -150,6 +151,10 @@ var PMA_console = {
                 PMA_console.showCard('#pma_console_options');
             });
 
+            $('#pma_console .button.debug').click(function() {
+                PMA_console.showCard('#debug_console');
+            });
+
             PMA_console.$consoleContent.click(function(event) {
                 if (event.target == this) {
                     PMA_consoleInput.focus();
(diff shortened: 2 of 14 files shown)


nisargjhaveri:
Some of the requests are self-handled, as in it does not use AJAX library in `ajax.js`  (For example running a query in SQL box). What should we do with these request? Right now, it does not show any debug info.

One way is to try to get debug info for every AJAX request made. But it'll also include other unintentional request, for example autocomplete.

nisargjhaveri:
To test, run a query in Console. It works here for sure. :P 
If it does not work somewhere(for some actions), please let me know.

lem9:
Nisarg,
It makes sense to show in the console only what is accumulated by _dbgQuery() from the PMA_DatabaseInterface class.

nisargjhaveri:
But the debugging info is accumulated by `_dbgQuery()` even in case of submitting SQL query box and autocomplete, but right now it is never shown, as they are not managed by `AJAX.responseHandler`.

lem9:
About submitting SQL query box, this is a user action for which debugging info is useful (unless I don't understand what you mean).

For autocomplete, well this generates indirect queries that I don't believe we are interested in, so please have a look to see how we could avoid debuggin info for this.

lem9:
I have started to test and so far I like the interface. I will continue my tests, then review the code. 

One minor thing: after deactivating "Debug SQL" via Settings, the "Debug SQL" console tab is still there.

lem9:
The previous behavior was to regroup all instances of the same query, showing the number of instances. What's your opinion on this?

nisargjhaveri:
I am not sure of usefulness as well as uselessness of grouping queries,so as Madhura suggested on RFE, I was going to implement an option for this. Still thinking about how the interface should be, though.

nisargjhaveri:
Updated, with fix to problem with newly refreshed page and problem after disabling debugging.

nisargjhaveri:
See new commit. I'm not sure of how many more places are there where we should pass `no_debug`. Right now, I added it for navigation reload, and autocomplete.

lem9:
While I continue my tests, please have a look at the failing test and at Scrutinizer's issues.

nisargjhaveri:
Already on it.

lem9:
Can you reproduce this? With your latest commit, I click on Settings and get a js error.

lem9:
About "no_debug", this is a good solution and I am not aware of other places.

nisargjhaveri:
No. Can you please check and tell what the error is?

lem9:
The error is gone none, I'll pay more attention if I see it. 

Another point: the debug panel is updated when I click on various actions (for example browse, search) which is fine; but it's not update when I type in the SQL box and submit.

nisargjhaveri:
Some more info? It was the whole point of the last commit. :/

nisargjhaveri:
I've updated the `original` theme, too. Please check.

nisargjhaveri:
And, did that work after the SQL query box?

lem9:
Sorry, I don't understand your question.

nisargjhaveri:
Sorry.
Does it now show debug information when typing in SQL query box and submitting it?

lem9:
Yes it's fixed now.

lem9:
About the original theme, all seems to work correctly. Please squash the commits so I can better review the code.

nisargjhaveri:
Squashed.

lem9:
There are non-English developers too.

lem9:
Nice work!

nisargjhaveri:
Okay. I'll make it translatable.

And, Thanks :)

lem9:
When the messages are translatable, please squash and I can merge. Then you can work on the grouping query option.

lem9:
How about showing the percentage of total time that each query (or group of queries) took?

nisargjhaveri:
That is a nice idea. I'll do it.

lem9:
Nisarg,
I would prefer merging a branch containing translatable messages but not containing your work on query grouping or percentage.

nisargjhaveri:
Hope I haven't missed any messages.

lem9:
Looks fine.

PR #1732, opened 2015-06-14
Save 'Angular Links' and 'Snap to Grid' settings in Designer

devenbansod:
RFE#1060 : Settings in 'Designer' should be remembered.

Currently done for 'Angular Links' and 'Snap to Grid' buttons.

Required to be done for 'Small/ Big All' button.

Signed-off-by: Deven Bansod devenbansod.bits@gmail.com

PR #1737, opened 2015-06-15
Designer settings saved in a new table in configuration storage

devenbansod:
Added a new table in PMA configuration storage.

Stores/ updates the values of designer settings in this new table.

### DIFF
--- a/config.sample.inc.php
+++ b/config.sample.inc.php
@@ -62,6 +62,7 @@ $cfg['Servers'][$i]['AllowNoPassword'] = false;
 // $cfg['Servers'][$i]['navigationhiding'] = 'pma__navigationhiding';
 // $cfg['Servers'][$i]['savedsearches'] = 'pma__savedsearches';
 // $cfg['Servers'][$i]['central_columns'] = 'pma__central_columns';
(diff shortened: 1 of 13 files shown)


ibennetch:
Looks good!

ibennetch:
For future reference, see also the other (session-based) implementation we decided against at https://github.com/phpmyadmin/phpmyadmin/pull/1732

madhuracj:
@devenbansod 
Please note that I have changed the column name from 'user' to 'username' to be consistent with other phpMyAdmin tables in https://github.com/phpmyadmin/phpmyadmin/commit/3ff7a264269187298ddc93194c1dfcdd5b5da6a9.

I also think it would be better to avoid using MySQL keywords as column names (`index` and `value`), especially when we show a warning when users are trying to create tables with keywords as column names.

devenbansod:
Thanks @madhuracj . I will take care of it from now on.

About the column names, thanks, I had not realized it before. I will keep a check on keywords while introducing such things in future. 

Also, do you think, `setting` and `preference` for 'index' and 'value' respectively, would make suitable substitutes ?

ibennetch:
Madhura this is a good point, thanks.

Do you see any use to making this table name more generic to potentially store other various phpMyAdmin settings in the future?

PR #1738, opened 2015-06-16
Add support for GPG signing releases

nijel:
I think this all we need in case we decide to use personal GPG keys for signing releases.


lem9:
Looks good.

PR #1741, opened 2015-06-18
Group/Ungroup queries in console SQL debug

nisargjhaveri:


lem9:
Nicely done! Very clear code.

nisargjhaveri:
Thanks :)

Now I'll work on sorting queries according to time taken, order of execution, count, etc. Anything else?

lem9:
Showing the percentage of time taken.

nisargjhaveri:
Yes. About that, what should I show for grouped queries. For queries belonging to a group, the percentage should be shown with respect to total time or with respect to the group?

lem9:
I believe that what most helps is the percentage of this group with respect to total time.

PR #1746, opened 2015-06-19
Ability to sort queries in console SQL debugging

nisargjhaveri:


lem9:
This is a fine implementation.

PR #1748, opened 2015-06-19
Templating getStartAndNumberOfRowsPanel function, recursivity and failures

Tithugues:
Please, do not merge this PR!

I wanted to do some templating on \PMA_Util::getStartAndNumberOfRowsPanel. But after doing the template, sending the right parameters and loading the page to check the result… I saw that it didn't work…
I think that this is because of use of ob_\* functions in \PMA\Template::render.

To understand, you need to know that:
- tbl_chart.php use tbl_chart.phtml
- tbl_chart.phtml use \PMA_Util::getStartAndNumberOfRowsPanel
- \PMA_Util::getStartAndNumberOfRowsPanel use startAndNumberOfRowsPanel.phtml (in this PR only)

As you can see, 2 phtml are used. So the code is going twice into \PMA\Template::render. So ob_start() is run twice (it seems that's not a big deal) and mostly, ob_get_clean() is run twice… And after the first one, the include is printing the data.

I see 2 ways to fix this:
- Create an "ob" object that manage the fact that ob_start / ob_get_clean is run many times (and so won't run it if already called)
- Return a value into phtml (see http://us2.php.net/manual/en/function.include.php); but I don't think this is a good thing, because that means that phtml are written in full PHP

Do you have any comment or idea to fix this issue?

### DIFF
--- a/libraries/Util.class.php
+++ b/libraries/Util.class.php
@@ -9,6 +9,9 @@ if (! defined('PHPMYADMIN')) {
     exit;
 }
 
+require_once 'libraries/Template.class.php';
+use PMA\Template;
+
 /**
  * Misc functions used all over the scripts.
  *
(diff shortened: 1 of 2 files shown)


Tithugues:
I provided a solution to described issue in https://github.com/phpmyadmin/phpmyadmin/pull/1749.
After PR1749 merge into master, I didn't find anymore issue on this templating.

zixtor:
Hi Hugues,

by applying just this PR (i.e. without changes from 1749), under 'Table=>Browse=>Display Chart', I can see the chart page with 'start and number row' panel being displayed.. do I need something else to reproduce this?

zixtor:
On the two template calls used in this, I believe as ob_get_clean is positively being run subsequent to ob_start, before returning from the render function, it should not interfere with another underlying ob_start and ob_get_clean sequence as I think these buffers can be stacked also in php?.. Could you please explain more how this problem could be happening..

Tithugues:
Hi @zixtor ,
You're right… As you can see, I did a fix after my initial commit. Maybe it was the origin of my issue…
So, if it's OK for you, let's merge this one. I'll send a PR to use OutputBuffering class in Template.
Thanks for your feedback.

zixtor:
Great, thanks for templating!

PR #1749, opened 2015-06-20
Improve output buffering class.

Tithugues:
Improve output buffering class.
Use PMA_OutputBuffering class in PMA\Template class.
Add output buffering test.

Signed-off-by: Hugues Peccatte hugues.peccatte@gmail.com

### DIFF
--- a/libraries/OutputBuffering.class.php
+++ b/libraries/OutputBuffering.class.php
@@ -17,8 +17,10 @@ class PMA_OutputBuffering
 {
     private static $_instance;
     private $_mode;
-    private $_content;
-    private $_on;
+    private $_contents;
+    private $_current;
+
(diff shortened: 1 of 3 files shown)


gzzhanghao:
Hi Hugues,
I think that the problem you mentioned in PR #1748 is nothing to do with the PMA\Template class, because ob_start is implemented as stackable in PHP. It should be something wrong in other place.
Also, I agree that we should use \PMA_OutputBuffering in PMA\Template for a better control of output buffers. But it seems to be some test issues in this PR, can you try to fix them so we could merge it?

Tithugues:
Hi,
You're right. I'll send another PR to use OutputBuffering into Template.
Thanks.

PR #1750, opened 2015-06-21
Ctrl/Alt + Enter submits Table->Insert form

devenbansod:
RFE#1679 : Ctrl + Enter (for Windows/ Linux) and Alt + Enter (for Mac) submits the Table->Insert (tbl_change.php) form.

Please suggest changes if required.

### DIFF
--- a/js/functions.js
+++ b/js/functions.js
@@ -4671,3 +4671,26 @@ function isStorageSupported(type)
     }
     return false;
 }
+
+/**
+ * Unbind all event handlers before tearing down a page
+ */
+AJAX.registerTeardown('functions.js', function(){
+    $(document).off('keydown', 'form input, form textarea, form select');
+});
+
+AJAX.registerOnload('functions.js', function () {
+    /**
+     * Handle 'Ctrl/Alt + Enter' form submits
+     */
+    $('form input, form textarea, form select').on('keydown', function(e){
+        if((e.ctrlKey && e.which == 13) || (e.altKey && e.which == 13)) {
+            $form = $(this).closest('form');
+            if (! $form.find('input[type="submit"]') ||
+                ! $form.find('input[type="submit"]').click()
+            ) {
+                $form.submit();
+            }
+        }
+    });
+});


ibennetch:
This seems like a good improvement to me. Is there an existing feature request about this?

devenbansod:
Yes. The feature request has been made by Nisarg [here] (http://sourceforge.net/p/phpmyadmin/feature-requests/1679/)

But, Madhura suggested that it would go against the current standard HTML5 behaviour of submitting the form on pressing Enter on standard text fields (except Long Text areas).

ibennetch:
Good, thanks. I had remembered some previous discussion on it but for some reason wasn't having luck finding it on my own.

nisargjhaveri:
@devenbansod, I just tested it.

My intention was to add ctrl+enter method to submit the form, not replace standard method of submitting by enter in non-textarea fields. Madhura is right, we should not modify the standard method, as many people might be used to it. But the addition would probably be nice.

Also, this has a problem. I cannot insert newline by pressing enter in textarea, with this PR. So, if we decide to merge this, it should be fixed first.

devenbansod:
Hi, Nisarg. 
Thanks for testing and pointing out the error.

I have made the changes. Can you once confirm if it works fine ?

nisargjhaveri:
Works. :)

I'm not sure if it should submit when pressing ctrl+enter on function dropdowns or not.

lem9:
Looks fine but how will the user know to use ctrl+enter? I suggest to show a hint when hovering over a textarea on this page -- or maybe better, when hovering over the Go button on this page.

devenbansod:
Thanks Marc for pointing it. Indeed, the user has to be conveyed somehow that he/she can use Ctrl + enter to submit the form.
I would prefer to go with a tooltip over the 'Go' button saying :
'You can also use Ctrl + Enter (Mac: Alt + Enter) to submit this form'

Any suggestions regarding the tooltip text?

lem9:
I would use something very short, similar to what we show in table Browse for the column headers. So "Ctrl + Enter (Mac: Alt + Enter) to submit" on the Go buttons. No need to mention the "also" stuff because they know they can click the button.

devenbansod:
Thanks for the suggestions :)

Made the changes.

zixtor:
Guys, IMHO we should refrain from over facilitating things in the UI and maintain reasonable consistency, I feel this addition will be seldom used and of course seems like intruding into the normal way things work in an html text area. But since its an addition and not quite a change, I will take it moderately on the consistency front.

nisargjhaveri:
The original RFE was there to maintain consistency with other places in PMA. We use ctrl+enter at many places to submit whatever we are doing (For example, SQL query box, console, grid editing, etc).

I am not sure how many will use it, though.

lem9:
Atul, in what way is this inconsistent?

zixtor:
On those lines of consistency as mentioned by Nisarg, it does seem to make sense, though a raw form texarea is quite different from sql query box or console where one would naturally expect a function for the enter key in combination with ctrl maybe, but less people would expect the same in a raw textarea, even less so when they are used to clicking a button.
And my mistake I used the wrong word, by consistency, I was more concerned about continuity in the behavior of a mature UI expecting changes only when absolutely useful.

That said, I do take it as a moderately useful addition with stated reservations.

zixtor:
We seem undecided on this, so added to IRC meeting: https://wiki.phpmyadmin.net/pma/2015-07_Meeting#RFE_Ctrl.2BEnter_submit_for_textarea_in_Table_Insert_form

libraries/insert_edit.lib.php
```diff
@@ -1574,7 +1574,9 @@ function PMA_getHeadAndFootOfInsertRowTable($url_params)
         . ' <tfoot>'
         . '<tr>'
         . '<th colspan="5" class="tblFooters right">'
-        . '<input type="submit" value="' . __('Go') . '" />'
+        . '<input type="submit" value="' . __('Go') . '" title="'
+        . __('Ctrl + Enter (Mac: Alt + Enter) to submit')
+        . '" />'
```
zixtor: This informative title should appear only for tables having at least one textarea field and may be should also tell that just Enter also works on other fields.

zixtor:
At the meeting, we have decided to accept this proposal, but we would need to comprehensively cover all forms across PMA, so effectively, Ctrl/Alt+Enter would become a way to submit a form with focus on any of its constituent fields (additionally we will retain the implicit way of Enter submission on input fields).

devenbansod:
Made some changes for testing.
I tested out some of the forms (still testing on more) and it works fine for those tested till now.
Please report any errors/ suggestions. 
Also, please comment if this is not the approach expected towards this feature.

zixtor:
Testing on table insert page, when submitting the form by clicking Ctrl+Enter on a plain Input field, gives following js error on the resulting '1 row inserted' page. (no js error when doing the same thing with just Enter)
<pre>
{
    "pma_version": "4.5.0-dev",
    "browser_name": "CHROME",
    "browser_version": "37.0.2062.120",
    "user_os": "Linux",
    "server_software": "Apache/2.4.12 (Ubuntu)",
    "user_agent_string": "Mozilla/5.0 (X11; Linux i686) AppleWebKit/537.36 (KHTML, like Gecko) Ubuntu Chromium/37.0.2062.120 Chrome/37.0.2062.120 Safari/537.36",
    "locale": "en",
    "configuration_storage": "enabled",
    "php_version": "5.5.26-1+deb.sury.org~precise+1",
    "exception_type": "js",
    "exception": {
        "mode": "onerror",
        "message": "Uncaught TypeError: Cannot read property 'settings' of undefined",
        "stack": [
            {
                "line": 135,
                "func": "rules",
                "context": [
                    "    rules: function( command, argument ) {",
                    "        var element = this[ 0 ],",
                    "            settings, staticRules, existingRules, data, param, filtered;",
                    "",
                    "        if ( command ) {",
                    "            settings = $.data( element.form, \"validator\" ).settings;",
                    "            staticRules = settings.rules;",
                    "            existingRules = $.validator.staticRules( element );",
                    "            switch ( command ) {",
                    "            case \"add\":",
                    "                $.extend( existingRules, $.validator.normalizeRule( argumen//..."
                ],
                "filename": "jquery/jquery.validate.js"
            }
        ],
        "useragent": "Mozilla/5.0 (X11; Linux i686) AppleWebKit/537.36 (KHTML, like Gecko) Ubuntu Chromium/37.0.2062.120 Chrome/37.0.2062.120 Safari/537.36",
        "name": "TypeError",
        "uri": "tbl_change.php?"
    },
    "script_name": "tbl_change.php",
    "microhistory": null
}
</pre>

zixtor:
Testing on table search page (submitting the form with Ctrl+Enter on a plain Input field), when trying to navigate away from the results page after the search results are displayed, gives "You have unsaved changes" alert.

zixtor:
On a related note, I think it would make additional sense if we automatically focus a form's first field when a form page is loaded, sounds in line with the keyboard only handling of forms.

devenbansod:
Thanks @zixtor for the reports.

I will be frank, but I am not able to identify the exact problem behind the tbl_change.php error. Any pointers to the problem will be very helpful.

About the other issue, of tbl_select.php problem, I tested it out and found that the AJAX.lockPageHandler  function is being called *before* the response while I press just Enter. But it is called *after* the ajax_request has been made and the response has been received, when I press Ctrl+Enter. Any specific issue leading to this behavior, that one can think of ?

zixtor:
@devenbansod The first problem should get fixed with https://github.com/phpmyadmin/phpmyadmin/commit/80a9ed2e8ccecedc35b497e20b7b53f947507743, I will look into the second problem about locking, later. Merging now. Thanks!

devenbansod:
@zixtor Thanks for the fix!

PR #1752, opened 2015-06-22
rfe#1557, Hide groups from navi panel

nisargjhaveri:


libraries/navigation/Nodes/Node.class.php
```diff
@@ -741,7 +741,40 @@ private function _getWhereClause($columnName, $searchClause = '')
      */
     public function getHtmlForControlButtons()
     {
-        return '';
+        $ret = '';
+        $cfgRelation = PMA_getRelationsParam();
+        if (isset($cfgRelation['navwork']) && $cfgRelation['navwork']) {
+            if ($this instanceof Node_DatabaseChild_Container
+                || $this instanceof Node_DatabaseChild
```
nisargjhaveri: Any better method? Scrutinizer is a little sad about this.

lem9:
Nisarg,
it would have been better to discuss the implementation before doing it. In the RFE, one of the reasons is "it slows down the page" and I believe that your implementation does not help in this regard. I checked the SQL debugging and, for example, when hiding functions, there is still a "SHOW FUNCTION status..." executed.

In conclusion, maybe we can keep the group hiding feature you propose, but either you add a configuration option to avoid retrieving functions and procedures, or you change the logic to avoid this retrieval when the corresponding group is hidden.

nisargjhaveri:
Please check now, it should now avoid retrieval of hidden group.

I am also planning to add a config option for this to remove a group completely from the navi panel (not db specific).
About that, should we add one config option that contains an array of all groups to hide or different config option for each group.
I think a single config option would be better, but I am not sure how to show array input in GUI, and I think we should have this option in GUI settings too. Thoughts?

lem9:
Yes retrieval is now fine. About array input in GUI (user preferences and setup) I have found examples in setup server configuration (only_db, hide_db) but the GUI is not intuitive. It might be easier (and more consistent with the GUI) to offer separate boolean options).

libraries/navigation/Nodes/Node.class.php
lem9: This is what Scrutinizer recommends, correct? I don't know a better method but if there is still a warning, you can ask on phpmyadmin-devel.

nisargjhaveri: Yes, this is what it suggests. But still, it classifies issues as bug.

lem9: I'm not an expert on this matter, but it looks like a false alarm.

nisargjhaveri: It looks like so. Still I'll ask once on devel.

nisargjhaveri:
I've added some config options for this, please have a look.

lem9:
This is fine, please fix the missing brackets for "if" statements as reported by Scrutinizer.

nisargjhaveri:
Updated and squashed.

lem9:
Nice work.

PR #1754, opened 2015-06-22
Proper support of the table aliases in CodeMirror

cmyker:
Signed-off-by: Igor Prokopenkov cmyk777@gmail.com

Before:
snapshot1
After:
snapshot2
snapshot3

### DIFF
--- a/js/codemirror/addon/hint/sql-hint.js
+++ b/js/codemirror/addon/hint/sql-hint.js
@@ -127,13 +127,13 @@
 
     if (columns) {
       addMatches(result, string, columns, function(w) {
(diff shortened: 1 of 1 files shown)


madhuracj:
Since the issue is in a third party library, we prefer to fix it in upstream, so that we do not have to patch the library every time we update it. 
I have sent a pull request to the upstream project for this issue and it has already been merged. See https://github.com/codemirror/CodeMirror/commit/172a78163d3ad80636d99983001acde633f3433b. 
I am waiting till CodeMirror version 5.4.0 is released with this change to update the CodeMirror library in phpMyAdmin.

PR #1757, opened 2015-06-23
Make text and password fields in config forms smaller

nisargjhaveri:
The size of these field in config forms was a little too big and was causing some problems with navigation panel settings dialig box


nisargjhaveri:
Sorry, for the failing tests. My phpunit installation is broken for some reason. :'(

Anyway, it should be fixed now.

lem9:
Thanks.

PR #1758, opened 2015-06-24
WIP: Introduce Controllers

gzzhanghao:
This PR introduce Controllers to PMA.

The structure of controllers has multiple layers, and in each layer it can provide some utility functions or variables that could be accessed from lower layers. But always notice that we can only holds logic codes in these controllers.

So far I've created a Controller class as the base class for any other controllers and it provides `$dbi` and `$response` for general usages. In future I would like to create a `$request` object to help handling access of `$_REQUEST`.

Let me introduce the Dependency Injection Container:
- We can insert dependent elements into the container using `Container::set`.
- Services and Factories can be set up using `Container::service` and `Container::factory` respectively.
- Elements that in the container can be renamed, aliased or removed using `Container::rename`, `Container::alias` and `Container::remove` respectively.

Further, all of the controllers should be created in /libraries/controllers with PMA\Controller namespace, and follow the PSR-4 rules.

For more details, please visit my blog post:
http://jasondaurus.wordpress.com/2015/06/24/the-oop-refactor-project-weekly-report-6/


zixtor:
Very good start, this is brave work, I will be testing functionality, later..

libraries/controllers/table/TableChartController.class.php
```diff
@@ -0,0 +1,234 @@
...
+use PMA_Message;
+use PMA\Template;
+use PMA\Controllers\TableController;
+
+require_once 'libraries/common.inc.php';
+require_once 'libraries/Util.class.php';
+require_once 'libraries/Message.class.php';
+require_once 'libraries/Template.class.php';
+require_once 'libraries/controllers/TableController.class.php';
+
```
zixtor: For .class.php files, I think over course of development we can replace these multiple require statements in all controllers, with a properly defined __autoload function to load classes in libraries/ and libraries/controllers/..  this function could be defined in a common file.

libraries/controllers/table/TableChartController.class.php
```diff
@@ -0,0 +1,234 @@
...
+        $this->url_query = &$GLOBALS['url_query'];
+        $this->table = &$GLOBALS['table'];
+        $this->db = &$GLOBALS['db'];
+        $this->cfg = &$GLOBALS['cfg'];
+    }
+
+    /*
+     * Execute the query and return the result
+     */
+    public function indexAction()
```
zixtor: I think we ought to define indexAction() as an abstract method in the base controller, to make it mandatory..

zixtor:
About creating Dependency Injection Container, I agree it will become complex given current state of our work and codebase, but I believe that for current purpose you can create a basic container on factory pattern, much like the way you are creating objects right now in zero argument constructors of controllers, so these constructor codes can be moved to a factory container with some reusability introduced.. Other opinions are welcome..

zixtor:
Further, I agree with creating template helper class for tbl_definition_form.inc.php (as said in your blog post), and I also like the idea of creating $request object for handling access to $_REQUEST

libraries/controllers/table/TableChartController.class.php
gzzhanghao: I suggest not to define `indexAction()` as an abstract method, because I think it's not necessary for every controller. I just put it here for convenience now, and I suggest we should eliminate it in the future for a better structure, though I haven't figured out how to achieve that goal yet. Any suggestion are welcome.

libraries/controllers/table/TableChartController.class.php
gzzhanghao: I'd like to use an existing library for that feature. Any suggestion?

gzzhanghao:
@zixtor I agree with your opinion on the container. I will pin down a timeline for new tasks soon.

libraries/controllers/table/TableChartController.class.php
zixtor: Seems like it can be done neatly for new Controller classes which follow namespaces and PSR-4 as done on https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-4-autoloader-examples.md
For other classes in our codebase, I wonder if we should first refactor our codebase to use namespaces and PSR-4 to have full consistency throughout and then only worry about autoloading, or maybe for current purpose, we could just look for class files in folders based on substrings in the class names.
Anyway, you can leave it for now, and focus on the schedule ;).

libraries/controllers/table/TableChartController.class.php
zixtor: No problem if you plan it to be non-mandatory later on, though now it seemed to be mandatory.

zixtor:
@gzzhanghao Great, firstly I hope that few remaining table related scripts get covered by end of this week; maybe do improvements like request object, container, autoloading etc. in the next refining week if time permits.

gzzhanghao:
Hi there,
I will finish my works on DIContainer tomorrow (which is out of my plan, though), then I will submit my commits here.
After that I will get my hands on the Template Helper and the `tbl_columns_definition_form.inc.php` file, I'll apply the Template Helper on it and I believe it will be more simple after that. It should be done by July 1st.
Further, I will refine `PMA_Table` and `PMA_TableSearch` class by July.2nd. It should takes about 1 week (July 9th).
I will spend about 2.5 weeks on `structure.lib.php`, including create templates, optimize it and OOP it (around July 25st).
Then in the rest of July I will integrate my works with `import.lib.php` and `export.lib.php` (July 31st).
After finishing my works mentioned in my proposal, I will do some works on controllers.

zixtor:
Makes sense to set aside the controller stuff after integrating DIC and temlpate helper(for tbl_columns_definition_form) with this PR, as it seems to be taking too much time from the schedule.
But, I think the quality detours and challenges we have seen so far in this project will serve as motivation for future movement towards MVC.

gzzhanghao:
@zixtor Hi, the latest commit fixed problems that exists in the TableGisVisualizationController and other scripts. Please review them and start merging process :-)

zixtor:
Thanks. Things seem to work fine in my tests. But I noticed the introduced DIC is not actually being used in most scripts in this PR and is not being used properly, I will make some changes to that effect, soon. Merging now.

zixtor:
See https://github.com/phpmyadmin/phpmyadmin/commit/ca61c3ace83942195b40a49b25c8dfa0f71adaf6, made the changes to achieve constructor parameter injection using the new DIC, though there remains scope for lot of refining.

PR #1759, opened 2015-06-24
Changed the mysql reserved words conflicting column names in Designer settings

devenbansod:
Changed the conflicting column names of `index` and `value` to `setting` and `stored_value`.

Any suggestions over these renaming are welcome.

### DIFF
--- a/db_designer.php
+++ b/db_designer.php
@@ -97,7 +97,7 @@ if (isset($_REQUEST['operation'])) {
                 . '.' . PMA_Util::backquote($cfgDesigner['table'])
                 . ' VALUES("' . $cfgDesigner['user'] . '", "'
                 . $_REQUEST['index'] . '", "' . $_REQUEST['value'] . '") ON DUPLICATE KEY '
-                . 'UPDATE VALUE = "' . $_REQUEST['value'] . '";';
+                . 'UPDATE `stored_value` = "' . $_REQUEST['value'] . '";';
 
             $success = $GLOBALS['dbi']->query($query);
             $response->isSuccess($success);
--- a/libraries/db_designer.lib.php
+++ b/libraries/db_designer.lib.php
@@ -186,60 +186,60 @@ function PMA_getSideMenuParamsArray()
 
     if ($GLOBALS['cfgRelation']['designer_settingswork']) {
         // angular_direct
-        $query = 'SELECT `value` FROM ' . PMA_Util::backquote($cfgRelation['db']) . '.'
+        $query = 'SELECT `stored_value` FROM ' . PMA_Util::backquote($cfgRelation['db']) . '.'
             . PMA_Util::backquote($cfgRelation['designer_settings']) . ' WHERE ' . PMA_Util::backquote('username') . ' = "'
-            . $GLOBALS['cfg']['Server']['user'] . '" AND ' . PMA_Util::backquote('index') . ' = "angular_direct"';
+            . $GLOBALS['cfg']['Server']['user'] . '" AND ' . PMA_Util::backquote('setting') . ' = "angular_direct"';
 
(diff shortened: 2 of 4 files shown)


lem9:
Deven,
I just noticed that you are using the entity-attribute-value model. See item 3 in https://www.simple-talk.com/sql/database-administration/five-simple--database-design-errors-you-should-avoid/

devenbansod:
Thanks for pointing that out Marc.

Yes. Reading the article (and especially the problems related to EAV model), I think we should change away from that. 
I will make the changes in a separate PR.

lem9:
I know that there are pros and cons to the EAV model, but so far we have tried to avoid using it in the pmadb, for the reasons cited in the article and other reasons (in the past I found an even more "con" article that I cannot find at the moment).

madhuracj:
@lem9 

My knowledge on EAV model and its advantages/disadvantages is quite limited. However, I see that it's a good fit for our purpose. Here is my reasoning.

1. It facilitates new features to the designer - Whenever we add a new feature and want to save its state we do not have to send update scripts to add a new column. I think update scripts are confusing to the users (especially if they have to apply a number of them in a certain order) and should try to avoid them as much as possible. 

2. We already use a variant of EAV - Even though not exactly the same, storing values json encoded does pretty much the same. See pma__table_uiprefs or pma__userconfig

3. Most of the disadvantages described in the article you shared do not apply to configuration storage.
i. We do not expect millions of rows in these tables, so we will not have performance issues
ii. We do not want to define foreign keys for this table
iii. We do not expect complex queries for this table. The most complex query will have a where clause referring `username` and `setting` column.

The only real disadvantage I see is the inability to have type constraints, which we need to handle at the application level.

devenbansod:
Hi Madhura,
I think, in that case, instead of going for a pure EAV, can't we simply use the same structure(a variant of EAV) as `pma__userconfig` (may be even without the timestamp) ?
This will surely reduce the number of rows about 6 times and querying would be done in a single simple query.

Suggestions welcome.

madhuracj:
@devenbansod 
That make sense since you'd want to read all the settings for a particular user when loading the designer. Anyway, wait for more feedback.

lem9:
@madhuracj I see your point and I realize that we never discussed this issue in details. Regarding performance, your affirmation is difficult to evaluate, as the configuration storage can be used for a shared installation. I agree that EAV would be easier for upgrades. However, not being able to define foreign keys is anti-relational; for example in pma__userconfig we cannot validate that each attribute or each value belongs to a valid set. But maybe we don't have this need. I just wanted to make use think more about the issue, since others have been bitten by its limitations.

ibennetch:
Thanks to everyone for sharing their thoughts.

Does anyone have a strong feeling about resolving this? I favor a similar style to `pma__userconfig` as proposed above (a single row per user with JSON encoding).

ibennetch:
Thanks to everyone for sharing their thoughts.

Does anyone have a strong feeling about resolving this? I favor a similar style to pma__userconfig as proposed above (a single row per user with JSON encoding).

I've done a quick test and don't notice any speed difference between a table having 10 rows and 1,000,000 rows, so I don't think there's a performance problem with many users.

lem9:
Isaac,
Not a strong feeling but I agree with your recommendation. Anyway, in the future if we find a roadblock about this, we can modify the structure.

madhuracj:
I also agree with Isaac's recommendation.

devenbansod:
Thanks everyone for their suggestions and thanks Isaac for the efficiency test. That really helped.

We will , at present, go ahead with a structure similar to `pma__userconfig` and modify it accordingly if we find a difficulty later.

I will make a new PR to update the structure soon.

PR #1760, opened 2015-06-24
Syntax error and refactoring.

udan11:
The `break` statement was outside any loop or `switch` statement.


lem9:
Thanks.

PR #1762, opened 2015-06-25
RFE #1561 Being able to use multiple servers at the same time when using cookie auth

madhuracj:
Submitting as a pull requests to get your opinion on changes to common.inc.php. Wonder whether it poses any security threat.

### DIFF
--- a/libraries/common.inc.php
+++ b/libraries/common.inc.php
@@ -872,7 +872,9 @@ if (! defined('PMA_MINIMUM_COMMON')) {
 
         if (! $auth_plugin->authCheck()) {
             /* Force generating of new session on login */
-            PMA_secureSession();
+            if ($token_provided) {
+                PMA_secureSession();
+            }
             $auth_plugin->auth();
         } else {
             $auth_plugin->authSetUser();
--- a/libraries/select_server.lib.php
+++ b/libraries/select_server.lib.php
@@ -35,7 +35,6 @@ function PMA_selectServer($not_only_options, $omit_fieldset)
                 $GLOBALS['cfg']['DefaultTabServer'], 'server'
             )
             . '" class="disableAjax">';
-        $retval .= PMA_getHiddenFields(array('token' => $_SESSION[' PMA_token ']));
 
         if (! $omit_fieldset) {
             $retval .= '<fieldset>';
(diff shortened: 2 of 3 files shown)


lem9:
I don't see a security threat; I believe that a threat could happen only if you changed the $token_mismatch logic.

madhuracj:
@lem9 Thanks for the feedback

libraries/common.inc.php
```diff
@@ -872,7 +872,9 @@
 
         if (! $auth_plugin->authCheck()) {
             /* Force generating of new session on login */
-            PMA_secureSession();
+            if ($token_provided) {
+                PMA_secureSession();
+            }
```
jamesyooda: This code modification does nothing for me, I still cannot connect to multiple servers at the same time, and I still get error: token mismatch

libraries/select_server.lib.php
```diff
@@ -35,7 +35,6 @@ function PMA_selectServer($not_only_options, $omit_fieldset)
                 $GLOBALS['cfg']['DefaultTabServer'], 'server'
             )
             . '" class="disableAjax">';
-        $retval .= PMA_getHiddenFields(array('token' => $_SESSION[' PMA_token ']));
```
jamesyooda: In PMA version 4.6 this line does not exist

test/libraries/PMA_select_server_test.php
```diff
@@ -125,12 +125,6 @@ public function testPMASelectServer()
             $html
         );
 
-        //PMA_URL_getHiddenInputs
-        $this->assertContains(
-            '<input type="hidden" name="token" value="token" />',
-            $html
```
jamesyooda: In PMA 4.6 none of these lines exist

jamesyooda:
Your file modifications do not solve the problem, I'm using PMA 4.6. With auth_type in cookie mode PMA is generatiing a different token each time I connect to different sql server, so I will always get a token mismatch, when trying to be connected to multiple databases at the same time. 

Where I work we use lots of different databases, this is a huge problem for us.

juokelis:
As you can see on last comment issue is not solved, with 4.6, now with 4.8 changing server disconnects from previously connected one. Impossible to work with multiple servers in parallel.

PR #1763, opened 2015-06-25
rfe#1603 select structure or data for each table when exporting

nisargjhaveri:
Finally, I had to keep the the current options, too. As they also control whether to export database structure and all. So, now the current radio buttons enables or disables data/structure selectors for tables. Also, if there is no choice available by plugin, both the checkboxes(data,structures) are disabled.

I was stuck with this for a really long time, so the chances of errors in slightly high.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -69,6 +69,7 @@ phpMyAdmin - ChangeLog
 + rfe #1531 Cant use external config file
 + rfe #1552 CSV import: Allow "Columns escaped with" to be optional
 + rfe #1561 Being able to use multiple servers at the same time when using cookie auth
++ rfe #1603 select structure or data for each table when exporting
 
 4.4.11.0 (not yet released)
 - bug       Missing selected/entered values when editing active options in visual query builder
--- a/db_export.php
+++ b/db_export.php
@@ -37,21 +37,25 @@ if ($num_tables < 1) {
     exit;
 } // end if
 
-$multi_values  = '<div>';
-$multi_values .= '<a href="#"';
-$multi_values .= ' onclick="setSelectOptions(\'dump\', \'table_select[]\', true);'
-    . ' return false;">';
-$multi_values .= __('Select All');
-$multi_values .= '</a>';
-$multi_values .= ' / ';
-$multi_values .= '<a href="#"';
-$multi_values .= ' onclick="setSelectOptions(\'dump\', \'table_select[]\', false);'
-    . ' return false;">';
-$multi_values .= __('Unselect All');
-$multi_values .= '</a><br />';
-
-$multi_values .= '<select name="table_select[]" id="table_select" size="10"'
-    . ' multiple="multiple">';
+$multi_values  = '<div class="export_table_list_container">';
+if (isset($_GET['structure_or_data_forced'])) {
+    $force_val = htmlspecialchars($_GET['structure_or_data_forced']);
+} else {
+    $force_val = 0;
+}
+$multi_values .= '<input type="hidden" name="structure_or_data_forced" value="' . $force_val . '">';
+$multi_values .= '<table class="export_table_select">'
+    . '<thead><tr><th></th>'
+    . '<th>' . __('Tables') . '</th>'
+    . '<th class="export_structure">' . __('Structure') . '</th>'
+    . '<th class="export_data">' . __('Data') . '</th>'
+    . '</tr><tr>'
+    . '<td></td>'
+    . '<td class="export_table_name all">' . __('Select all') . '</td>'
+    . '<td class="export_structure all"><input type="checkbox" id="table_structure_all" /></td>'
+    . '<td class="export_data all"><input type="checkbox" id="table_data_all" /></td>'
+    . '</tr></thead>'
+    . '<tbody>';
 $multi_values .= "\n";
 
 // when called by libraries/mult_submits.inc.php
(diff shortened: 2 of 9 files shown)


lem9:
Nisarg,
I looked at just the UI for now. IMO, keeping these old options under "Format-specific options" is too confusing for the user. If you absolutely need to keep these in the form, they will have to be invisible.

nisargjhaveri:
The problem with just removing those options is, even if none table is selected for export structure, still the export may have database structure, functions, etc., by selecting the old options.

Currently, the old options override table specific options. For example, if I select data from Format-specific options, checkboxes for structure in table select are disabled. If 'structure and data' is selected from Format-specific options, then only user can select data or structure for individual tables.

Yes, I agree that is little confusing, but don't know the better alternative. 

One way might be to automatically deduce that user want only data and no structures in export if he/she has no tables selected for structure. But I don't think that is the best choice.

Ideas, suggestions?

lem9:
I'm not sure I understand your first sentence "The problem...". In current master's database export (before your pull request), if I unselect all tables, no export occurs and this is expected. Also, I tried with your pull request and if no table is selected, no export occurs (even if in the old options, structure and data is chosen) and this is fine.

About your second sentence, the old options are too far away from the new options for this to be useful. The standard way is to put a checkbox on the header line, that permits to check all / uncheck all everything in the corresponding column.

Also I don't find the old options a little confusing, but much confusing.

nisargjhaveri:
What I wanted to say with the first statement was,
Selecting 'data' from old options is not equivalent to selecting 'structure and data' from old option and deselecting all new checkboxes for structure (and keeping checkboxes for data checked).
In the first case, it just export "insert into.." statements. Whereas in the second case, it also exports definition for functions, view, etc., along with insert statements for table data.

I intentionally didn't add check all / uncheck all checkboxes because of this confusion.

**Possible solutions?**

1. We can work on moving the old options closer to the new table select, keeping the functionality as it is, and probably updating text to reflect what it does properly.

2. We can remove old options and assume following.
 - all structure checkboxes unchecked => equivalent to 'data' from old options
 - all data checkboxes unchecked => equivalent to 'structure'
 - else equivalent to 'structure and data' (obviously, respecting selection for individual tables)

I hope, I made it clear now. Sorry that I couldn't convey it properly the first time.

Also, do we really support browsers without JavaScript?

lem9:
I'll review your last answer. About JavaScript, we officially require JavaScript-enabled browsers, see http://docs.phpmyadmin.net/en/latest/require.html#web-browser

lem9:
* In the header for checkboxes, please remove "Export"; this is redundant as the whole goal of the page is to export
* We want to avoid confusion and have no conflicting choices, so the choices in "Object creation options > Add statements" should be the only ones that control whether or not functions, procedures, etc will be exported, regardless of the state of the table-specific choices.

Tell me if this makes sense.

nisargjhaveri:
Logically makes sense, but think about a user who used to export just data with one or two extra clicks. Now, if we do this, he would have to deselect structures from tables list, then deselect those options for functions, procedures, view etc. Wouldn't it become a complex task?

Or we should also add list of functions, procedures, view and events along with tables? But, this would be again specific to SQL and other few plugins which can export these. So, the similar problem will come up again.

lem9:
When we designed the Quick export (which does structure+data), it was assumes that most users would choose that. 

About your scenario, well, I assume that https://sourceforge.net/p/phpmyadmin/feature-requests/812/ would be the best general solution. Meanwhile, these choices can be set via either cfg or Settings. We cannot have the defaults that suit everyone.

nisargjhaveri:
Okay.

Still there is the "Add CREATE TABLE statement" which conflicts with new checkboxes, and overrides if not checked (i.e. does not export any table structures).

Also, what should be done with plugins which does not support structure or data?

lem9:
This "Add CREATE TABLE statement" option could be grayed out (for database export only); I don't see the point of this option with the new UI.

For plugins which do not support structure or data, how about not showing the corresponding checkboxes (structure or data) in the table selector?

nisargjhaveri:
How should default value of structure or data for each plugin affect table select checkboxes?

nisargjhaveri:
I've updated the behaviour, please have a look.

Also, with this update, the default value of plugin does update the checkboxes. If a plugins default is 'data', it'll uncheck all structures from table list. Similarly for 'structure'. If it is 'structure and data', it does not do anything. It never checks a checkbox. Please review this behaviour, too.

lem9:
I'll review in a few hours; meanwhile can you try a merge and see what is conflicting?

nisargjhaveri:
Conflict is in ChangeLog, so will fix it when it is ready for merge.

lem9:
Looks fine, but I just noticed one issue. Let's assume we are using the SQL format. In the previous version, depending on the choice structure / data / structure and data, only the relevant sections Object creation options / Data creation options were shown. Now these sections are always shown and I wonder if it makes sense. My suggestion is that Object creation options should be shown only if there is at least one table with Structure ticked; the same for the other section. Opinion?

nisargjhaveri:
Object creation options also contains options for exporting function, procedures etc. These settings apply regardless of tables selected for exporting structure data. So, I don't think we should hide this section.

The data creation options, on other hand, doesn't seem to have anything like that, and can be hidden/shown based on table selections. I personally wouldn't prefer that. But, we can probably safely do it for this section.

lem9:
I accept your judgement. Now, please make messages translatable.

nisargjhaveri:
Done.
Anything else? Or should I squash now?

nisargjhaveri:
Doesn't the checks run on PRs with merge conflict? Or is there something else?

lem9:
Please make Scrutinizer happy :)

lem9:
I think I found a bug; to reproduce:
- initially with SQL format, Structure and Data checkboxes are ticked (this is fine)
- switch format to PDF, Structure checkboxes are unticked (this is fine)
- switch to SQL, Structure checkboxes are still unticked (bug)

nisargjhaveri:
Yes, I am aware of this behaviour. And this is why I specifically write "It never checks a checkbox" in one of my comments!

The intention was to prevent unwanted selection of tables to export. As the selection checkboxes comes first in the page. Let's say I unchecked all structures manually, I change export type to PDF and then again to SQL and all the structures are now checked. I was trying to prevent this case.

Anyway, what we can do is, 

1. Remember previous states and revert back to that state.
2. Make the column identical to other column. For example, in this case, when selecting SQL back, check all structure checkboxes for those tables which has data checked.

Suggestions?

lem9:
When I am initially on the SQL format, the default from $cfg['Export']['sql_structure_or_data'] is applied. When I switch to PDF, the default from $cfg['Export']['pdf_structure_or_data'] is applied. When I go back to SQL, I would expect the default from $cfg['Export']['sql_structure_or_data'] to be applied again. Or am I missing something?

Note that if the solution to my test scenario involves too complex code, maybe it's better not to solve it.

nisargjhaveri:
Yes, but the checboxes are to override default values. And the checkboxes are before selecting options. So, we can't just override user filled values with some default values. The current behaviour(with this PR) also partially does that.

It would be best if we move the format selection dropdown above table selection, and we can do what you suggested(fill default checkboxes according to default values for format, and then let user override values). I think I'll try doing this.

lem9:
Yes, I had the same idea.

nisargjhaveri:
Please check now.

lem9:
Excellent, please squash.

nisargjhaveri:
Done.

lem9:
Nice work.

nisargjhaveri:
Thanks. :)

PR #1764, opened 2015-06-28
Few fixes related RFE#946 & RFE#657 - PR #1686 &  PR #1664

devenbansod:
Fixes:
1. Move JS function definitions before usage to make Scrutinizer happy. (PR #1686)
2. Run "FLUSH PRIVILEGES" only when at least one of the column names have been changed (PR #1664 ).
3. Also, changing collation along with some other changes to the column definition (esp. to column name) was not working as expected, as it was using the new column name instead of the old one in an intermediate query. This commit should fix it. 


ibennetch:
Thanks Deven!

PR #1769, opened 2015-06-30
'As Separate files' export works with ZIP compression only

devenbansod:
When exporting 'as_separate_files', compression method should be fixed as 'zip', as it currently supports only ZIP compression.

### DIFF
--- a/libraries/config.default.php
+++ b/libraries/config.default.php
@@ -1403,7 +1403,7 @@ $cfg['Export']['compression'] = 'none';
 $cfg['Export']['lock_tables'] = false;
 
 /**
- * Whether to export databases/tablse as seperate files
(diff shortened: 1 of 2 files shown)


libraries/config.default.php
```diff
@@ -1403,7 +1403,7 @@
 $cfg['Export']['lock_tables'] = false;
 
 /**
- * Whether to export databases/tablse as seperate files
+ * Whether to export databases/tables as seperate files
```
ibennetch: This should be "separate" instead of seperate.

devenbansod: Thanks for pointing it out. 
Fixed in the new commit.

ibennetch:
Thanks!

PR #1771, opened 2015-06-30
#812 store export definitions for reuse

madhuracj:

### DIFF
--- a/libraries/config.default.php
+++ b/libraries/config.default.php
@@ -491,6 +491,15 @@ $cfg['Servers'][$i]['central_columns'] = '';
 $cfg['Servers'][$i]['designer_settings'] = '';
 
 /**
+ * table to store export templates
+ *   - leave blank to disable saved searches feature
+ *     SUGGESTED: 'pma__exporttemplates'
+ *
+ * @global string $cfg['Servers'][$i]['exporttemplates']
+ */
+$cfg['Servers'][$i]['exporttemplates'] = '';
+
+/**
  * Maximum number of records saved in $cfg['Servers'][$i]['table_uiprefs'] table.
  *
  * In case where tables in databases is modified (e.g. dropped or renamed),
--- a/config.sample.inc.php
+++ b/config.sample.inc.php
@@ -63,6 +63,7 @@ $cfg['Servers'][$i]['AllowNoPassword'] = false;
 // $cfg['Servers'][$i]['savedsearches'] = 'pma__savedsearches';
 // $cfg['Servers'][$i]['central_columns'] = 'pma__central_columns';
 // $cfg['Servers'][$i]['designer_settings'] = 'pma__designer_settings';
+// $cfg['Servers'][$i]['exporttemplates'] = 'pma__exporttemplates';
 /* Contrib / Swekey authentication */
 // $cfg['Servers'][$i]['auth_swekey_config'] = '/etc/swekey-pma.conf';
 
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -812,6 +812,20 @@ Server connection settings
     * put the table name in :config:option:`$cfg['Servers'][$i]['savedsearches']` (e.g.
       ``pma__savedsearches``)
 
+.. _exporttemplates:
+.. config:option:: $cfg['Servers'][$i]['exporttemplates']
+
+    :type: string
+    :default: ``''``
+
+    Since release 4.5.0 you can save and load export templates.
+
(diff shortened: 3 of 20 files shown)


lem9:
Started testing. I know that I can use the "Find out why" link to create the missing table, but I should not have an error when clicking on Export if I forgot to create it.

lem9:
So, the template is per table? Why?

madhuracj:
@lem9 
Thanks for testing Marc.
On another server where neither the configuration table nor the entry in config.inc.php is setup, I cannot seem to see any errors. Is it on server level export page you get the error?

Templates can be for server (both `db_name` and `table_name` is `NULL`), db (`table_name` is `NULL`) or table.

lem9:
The error happens when exporting at the table level. The generated FROM clause contains only
````
`phpmyadmin`.
````

About templates, when I am at the table level, it would be helpful to load a template that I made for another table. That's the purpose of having a template, right?

**Comment by madhuracj** [comment]
What I had in mind was repeat exports of same database or tables. But I see that ability to use a template made for another db or table can be useful. I'll change accordingly.

**Comment by madhuracj** [comment]
Please have a look now

**Comment by lem9** [comment]
The error is now fixed and the UI behaves correctly, whether I have the missing pmadb or not. I tried table templates and now I can reuse a template for another table. I'll continue testing.

One confusing thing: when I click on Export, I see the Existing templates interface but I have no way of telling whether or not a template has been loaded. There is no difference in the interface, before or after loading (apart from checking if the export options have changed). Maybe the UI should tell me that there is a loaded template (and which one).

**Comment by madhuracj** [comment]
Please check https://github.com/madhuracj/phpmyadmin/commit/f2cf3e5a72d80d54e84e76acd22b5412bc779379 and https://github.com/madhuracj/phpmyadmin/commit/5dbc0c45056f0454773a8426f3d3b6cc5fcc8dcb to see whether the UI has improved

File: libraries/config.default.php [hunk 1]
```diff
@@ -491,6 +491,15 @@
 $cfg['Servers'][$i]['designer_settings'] = '';
 
 /**
+ * table to store export templates
+ *   - leave blank to disable saved searches feature
+ *     SUGGESTED: 'pma__exporttemplates'
+ *
+ * @global string $cfg['Servers'][$i]['exporttemplates']
+ */
+$cfg['Servers'][$i]['exporttemplates'] = '';
```
ibennetch: We're not very consistent with naming these variables regarding whether or not to use an underscore between words; I've added this to https://wiki.phpmyadmin.net/pma/2015-07_Meeting#Naming_style_for_Configuration_Storage_variables . There are more with an underscore than without, so perhaps `export_templates` would be more consistent, but that's just my suggestion.

ibennetch:
I like this feature and have a few suggestions.

* It would be nice to be able to reset to the defaults; say for instance I load a template and realize I don't like it and instead want the defaults; could such a button be added easily?
* If I select to "View output as text" and proceed to the next page by clicking Go, everything happens like I expect. If I then press Back, I'm returned to the Export page, again as expected. All the settings are the same, because phpMyAdmin stores and resets the export options based on what I was just viewing, but the Template dropdown returns to "-- Select a template --"; it would be nice if instead the current template was re-selected.

About Marc's suggestion about the UI showing when a template was loaded, I think your improvement was good but what about going further and also showing some text like "Current template: foo" perhaps under the dropdown? I suppose we'd then have to either hide it or add another text if the export settings are changed from what the template is...so I'm not sure about my idea.

ibennetch:
It's also a bit confusing to save a template in the table export page and not see it listed from the database export page. I understand the reason, and I don't have any suggestion on how to fix it, but as a user it would annoy me.

lem9:
What does "Template wss updated" means?

libraries/config.default.php
madhuracj: Thanks for the suggestion. I'll update the table name if required after the IRC meeting.

madhuracj:
@lem9 and @ibennetch 
Thanks for the suggestions, I am attending to them.

madhuracj:
@ibennetch 
Easiest way to load the defaults would be to click on the 'Export' menu link.
Chosen template is correctly selected on 'Back' button now.
About having to separate template to 'server', 'db' and 'table' levels and list only relevant templates, I do not see a good solution either.

madhuracj:
@lem9 
Typo fixed.
Template dropdown auto submits now to load the template.

lem9:
Looks fine, thanks.

PR #1772, opened 2015-07-01
Trial Approach to solve privileges bug for RFE#657

devenbansod:
**Bug description:**
Adjusting privileges implementation requires `SELECT`, `INSERT`, `UPDATE`, `DELETE` privileges on `mysql`.`db`, `mysql`.`cols_priv`, `mysql.tables_priv`, `mysql`.`procs_priv` 
And `FLUSH` privilege for reloading the privileges.
If user without these privileges, tries to adjust privileges while copying/ renaming/ moving db, table, columns, procedures then the operation might not give complete properly and gives an access denied error.

**Approach:**
_Try_ making similar queries that we are going to need to make while adjusting the privileges.
If any of these fail, means that the user does not have enough privileges to perform this action.

So, while we display the checkbox field for 'Adjust Privileges', we check if user has required privileges. If he/she does not possess the required privileges, then we display a _disabled_ checkbox with a tooltip explaining the situation.

### DIFF
--- a/db_operations.php
+++ b/db_operations.php
@@ -21,6 +21,7 @@ require_once 'libraries/mysql_charsets.inc.php';
 /**
  * functions implementation for this script
  */
+require_once 'libraries/check_user_privileges.lib.php';
 require_once 'libraries/operations.lib.php';
 
 // add a javascript file for jQuery functions to handle Ajax actions
--- a/db_routines.php
+++ b/db_routines.php
@@ -16,6 +16,7 @@ require_once 'libraries/mysql_charsets.inc.php';
 /**
  * Include all other files
  */
+require_once 'libraries/check_user_privileges.lib.php';
 require_once 'libraries/rte/rte_routines.lib.php';
 
 /**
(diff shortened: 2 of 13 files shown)


ibennetch:
I like this improvement. Thanks.

ibennetch:
Deven, good research on this issue. Nice work.

ibennetch:
I now get a notice when going to the Structure page and clicking Change for an existing column:

<pre>
Notice in ./libraries/tbl_columns_definition_form.inc.php#338
Undefined variable: privs_available

Backtrace

./libraries/structure.lib.php#2495: include(./libraries/tbl_columns_definition_form.inc.php)
./tbl_structure.php#83: PMA_displayHtmlForColumnChange(
string 'stack',
string '1',
NULL,
string 'tbl_structure.php',
)
</pre>

I believe it's related to these commits, could you take a look?

devenbansod:
Hi, 
Sorry I could not look into this earlier.

Anyways Madhura fixed this issue due to an extra unused parameter, in  this commit : https://github.com/phpmyadmin/phpmyadmin/commit/944166de03f6e830a1951408ec99215ac9890925

PR #1773, opened 2015-07-02
Fix tests, use proxy settings from env

nisargjhaveri:


nijel:
Sorry to break things for you, but I've rewritten response processing in 5a5ceec to avoid manual parsing of HTTP response headers.

nisargjhaveri:
That is much better than this. I'll update this PR. :)

nisargjhaveri:
Updated.

lem9:
Thanks.

PR #1774, opened 2015-07-03
Edit link should be active only for non read-only system variables

devenbansod:
RFE#1683 : Edit links should be active only for non Read-only variables

Edit link should be active only for non read-only system variables.
Add a tooltip for disabled links explaining the situation.


devenbansod:
Reference : https://dev.mysql.com/doc/refman/5.0/en/dynamic-system-variables.html

ibennetch:
I think it would be better to completely hide the Edit link here, though I also see value in helping the user see why they can't edit it.

Maybe it would work showing the [red circle with an X](https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/display_create_database.lib.php#L55-L58) (same as is used on the "Database" tab when a user doesn't have privileges to create a new database) instead of the edit icon and Edit text; with the same tool tip you used.

devenbansod:
Thanks for the suggestions.
Yes. Completely hiding the link would not be very useful for the user. But your suggestion looks great!

I will make the required changes.

devenbansod:
Hi Isaac,

Made the changes you suggested.

ibennetch:
Looks good to me now, thanks.

There's no bug/feature ticket about this, correct?

devenbansod:
Yes. Actually there is a feature request.

RFE#1683 : Edit links should be active only for non Read-only variables

PR #1779, opened 2015-07-09
Move to html5 history api

atul516:
Changes:
- clearer URL with easier manipulation of URL script name or query params to navigate.
- as we use no-cache, so browser now makes new requests on pressing back/forth buttons, under HTML5 API. Earlier with hash microhistory, we were manually caching the no-cache content of pages.

### DIFF
--- a/js/ajax.js
+++ b/js/ajax.js
@@ -354,7 +354,7 @@ var AJAX = {
                 if (data._menu) {
                     if (history && history.pushState) {
                         history.replaceState({
-                            menu : data._menu
+                                menu : data._menu
                             },
                             null
                         );
@@ -545,7 +545,7 @@ var AJAX = {
          *
          * @return void
          */
-        load: function (files) {
+        load: function (files, callback) {
             var self = this;
             self._scriptsToBeLoaded = [];
             self._scriptsToBeFired = [];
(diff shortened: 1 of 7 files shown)


zixtor:
For error reports, I have removed sending microhistory cache data with reports, it sounded useful, but personally to me it came of little use practically. Removal is necessitated because we are now loading microhistory code conditionally only when html5 is not supported. But if any objections to removal, we can consider keeping a cache just to send with reports.
Also, it might require some changes on error reporting server, I could not test if report submission still works after this removal, due to very very slow error reporting server.

ibennetch:
I'm okay with removing microhistory from the error reports. Like you said, it seemed useful at the time but probably isn't very practical at this point.

zixtor:
Thanks, with that, things seem fine to me for merging.

zixtor:
So, with the fast error reporting server I am able to verify that the report submission worked, and I see "Microhistory: null" as on Incidents page [0], so I think the 'Microhistory' field should be removed from the Incidents UI and database of the server. @smita786, please queue this request.

[0] https://reports.phpmyadmin.net/incidents/view/284716

smita786:
@zixtor thanks. I have opened an issue here https://github.com/phpmyadmin/error-reporting-server/issues/83 . 'll fix it soon.

zixtor:
Thank you! :)

PR #1781, opened 2015-07-10
RFE#1163

devenbansod:
RFE#1163 : This is more of a bug rather than a feature.

In the existing scenario, the variables `character_set_results` and `character_set_client`
are always stuck at `utf8mb4' or`utf8` (for earlier versions of MySQL).

So, we provide 2 more drop-downs on the welcome page (i.e. index.php) similar to the existing one for `collation_connection` to select the values for these 2 variables.

Please suggest changes if needed. Suggestions regarding the complete approach to the problem are welcome as well.

Also, suggestions regarding the icons for these 2 drop-downs are also welcome. Currently I have used the same icons as used for `collation_connection` drop-down.

### DIFF
--- a/index.php
+++ b/index.php
@@ -23,6 +23,8 @@ $drops = array(
     'lang',
     'server',
     'collation_connection',
+    'character_set_client',
+    'character_set_results',
     'db',
     'table'
 );
@@ -198,8 +200,8 @@ if ($server > 0 || count($cfg['Servers']) > 1
                 );
             }
         } // end if
-        echo '    <li id="li_select_mysql_collation" class="no_bullets" >';
-        echo '        <form method="post" action="index.php">' . "\n"
+        echo '    <form method="post" action="index.php">' . "\n";
+        echo '        <li id="li_select_mysql_collation" class="no_bullets" >'
            . PMA_URL_getHiddenInputs(null, null, 4, 'collation_connection')
            . '            <label for="select_collation_connection">' . "\n"
            . '                ' . PMA_Util::getImage('s_asci.png') . " "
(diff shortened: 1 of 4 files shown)


ibennetch:
Any comments on this from the rest of the development team? Deven and I don't see any drawbacks to this approach, but we're being a bit cautious about merging it.

lem9:
As I wrote in 2009, "I admit I did not test extensively what you're talking about, however don't you wonder why there are no general complaints about this? Maybe there is something non-standard about your data?". So adding more drop-downs with a possibility of users mangling their data, is risk. Can Deven reproduce the problem mentioned in the RFE? and if so, can't we just SET NAMES behind the scene?

devenbansod:
In fact, I could not reproduce the problem as it is. But just through experimenting with different values and methods found out that the `character_set_results` and `character_set_client` are always stuck at utf8mb4 or utf8 (for older versions).

Now that Marc pointed out 'Can't we just use "SET NAMES" instead ?', I think we can use it for sure. As, just selecting the value from `collation_connection` dropdown, all three (`character_set_results`, `character_set_clients` and `collation_connection`) would be set with that selected value through just one "SET NAMES" query. 

But just to note, if we implement this, we would not be able to have difference in values of these 3 variables.
Ex. One could never have `collation_connection` as 'latin1_swedish_ci', `character_set_results` as 'utf8' and `character_set_client` as 'utf8mb4'.

I, myself, have never dealt with a varied collation and charsets data, so, at this point, I am unable to decide if this noted point, of not being able to set different values, would, in fact, be a problem later.

Any suggestions?

lem9:
If we cannot reproduce the problem, I conclude that this was a case of garbled data (see https://wiki.phpmyadmin.net/pma/Garbled_data) and that we should close this RFE and do nothing.

devenbansod:
On more thought, Marc's suggestion to close this issue seems right way to move.

If we encounter some related bug reports in future, we can get a proper feedback from the ticket creator and then make appropriate corrections to fix that.

ibennetch:
Okay,
Marc thanks for the feedback, and Deven thanks for the thorough research and good pull request. I'll close the ticket for now.

PR #1783, opened 2015-07-12
Fixed a small error

devenbansod:
Removed the extra '?' from the URL while Filtering the process list.


zixtor:
Thanks!

PR #1785, opened 2015-07-12
Format backtrace when required, debug console

nisargjhaveri:
It was taking too much time with high number of SQL queries.


lem9:
Thanks, I can notice the speedup.

PR #1788, opened 2015-07-13
[WIP] Added lint.

udan11:
Added lint. 

Work in progress. 
- [X] Basic support.
- [X] Add configuration entries.
- [ ] Write more strict rules for parsing depending on context (see udan11/sql-parser#2).
- [x] Reorganize `lint.php`.

### DIFF
--- a/js/codemirror/addon/lint/lint.css
+++ b/js/codemirror/addon/lint/lint.css
@@ -1,0 +1,73 @@
+/* The lint marker gutter */
+.CodeMirror-lint-markers {
+  width: 16px;
+}
+
+.CodeMirror-lint-tooltip {
+  background-color: infobackground;
+  border: 1px solid black;
+  border-radius: 4px 4px 4px 4px;
+  color: infotext;
+  font-family: monospace;
+  font-size: 10pt;
+  overflow: hidden;
+  padding: 2px 5px;
+  position: fixed;
+  white-space: pre;
+  white-space: pre-wrap;
+  z-index: 100;
+  max-width: 600px;
+  opacity: 0;
+  transition: opacity .4s;
+  -moz-transition: opacity .4s;
+  -webkit-transition: opacity .4s;
+  -o-transition: opacity .4s;
+  -ms-transition: opacity .4s;
+}
+
+.CodeMirror-lint-mark-error, .CodeMirror-lint-mark-warning {
+  background-position: left bottom;
+  background-repeat: repeat-x;
+}
+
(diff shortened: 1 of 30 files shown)


lem9:
Interesting feature. I tested typing errors in the data types and fixing them. I fear about the performance (calling a script almost once per character typed).

Also, something curious: if I type CREATE TBLE then continue with my definition, I have an error below on my VARCHAR not being a correct statement. I guess that this will be fixed with more parsing rules.

udan11:
The accuracy will improve when more rules will be written. I already wrote a couple of new ones and I'll commit the new library to phpMyAdmin soon.

Regarding performance, CodeMirror will make a new request only when you pause typing. For example, if you were to type continuously and only stop when the query is finished, only one request would be made. Also, long queries aren't analyzed.

I guess that I could set a timer to send a request only a couple of seconds (maybe 10) and only if the contents changed. Would that be better?

lem9:
I was not typing fast enough in my test! So I withdraw my comment about performance.

lem9:
Please use the normal token protection inside lint.php; we don't want XSS via this script.

lem9:
Not sure if it's because you're using strlen() instead of mb_strlen(), but I can generate a js error by typing
````
select from école
````

udan11:
I believe there is absolutely no vulnerability in this script. All it does is to accept one user input, parse it and send back a JSON response. Before displaying the result to the user, the response is escaped and any tag that may cause issues is escaped.

Also, I fixed the issue with UTF-8 strings.

lem9:
About strlen(), see https://github.com/phpmyadmin/phpmyadmin/commit/66b4be9b and look at my comments. Can you use mb_strlen() with the '8bit' parameter?

lem9:
If you believe there is no vulnerability, please indicate your reasoning as comments at the beginning of lint.php to document why there is no token protection.

lem9:
Changes are fine; awaiting completion of the WIP.

udan11:
I believe this is ready to be merged. The rules have been improved significantly and further updates will come in the future versions of the library.

lem9:
Nice work. I'm merging but I hope you are aware that some mistakes like "WHRE" are not detected. Also, looks like the linter does not have effect when typing in the console?

udan11:
"WHRE" might not be a mistake. The following query: "SELECT * FROM actor WHRE" is valid.

lem9:
True about the alias case. Now, what about typing in the console?

udan11:
I'll take care of that. :)

PR #1790, opened 2015-07-14
Fixed a small error

devenbansod:
# prefs_autoload was being printed in the Print View output.

It was not needed in print output, hence removed.


ibennetch:
Thanks!

PR #1796, opened 2015-07-18
Fixed some errors visible in all travis builds

devenbansod:
All the recent Travis builds' logs showed the messages with
'Unsupported Query', which were introduced in (d014dbc)

This PR should fix that errors.


lem9:
Nice catch!

PR #1799, opened 2015-07-19
Enabled lint for console.

udan11:
Enabled lint for console.
Updated sql-parser library to udan11/sql-parser@5e554b3.

### DIFF
--- a/js/console.js
+++ b/js/console.js
@@ -553,7 +553,12 @@ var PMA_consoleInput = {
                 mode: 'text/x-sql',
                 lineWrapping: true,
                 extraKeys: {"Ctrl-Space": "autocomplete"},
-                hintOptions: {"completeSingle": false, "completeOnSingleClick": true}
+                hintOptions: {"completeSingle": false, "completeOnSingleClick": true},
+                gutters: ["CodeMirror-lint-markers"],
+                lint: {
+                    "getAnnotations": CodeMirror.sqlLint,
+                    "async": true,
+                }
             });
             PMA_consoleInput._inputs.console.on("inputRead", codemirrorAutocompleteOnInputRead);
             PMA_consoleInput._inputs.console.on("keydown", function(instance, event) {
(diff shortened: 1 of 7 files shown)


Tithugues:
Hi,
Would it be possible to get your sql-parser from comporser, instead of including it in pMA sources ?
Thanks.

udan11:
@Tithugues I proposed using Composer in the past. There's been a discussion on the mailing list. I can't link to you now because Sourceforge is in maintenance mode, but as soon as it's up you can look for "Outdated libraries", that was the subject of the discussion.

lem9:
@udan11 Apart from the "Outdated libraries" discussion, I believe that you and I had a discussion specifically about this library. I can't seem to find it right now, but I remember that the outcome was to not use Composer for this library. Any insight you could give about this other discussion?

lem9:
Nice work.

udan11:
From "[Phpmyadmin-devel] Parser and uppercase characters". The quoted text is my question.

```
> The other question I have is about autoloading classes. For this library
> I used the autoloader that is generated by Composer. However, phpMyAdmin
> does not use Composer for managing dependencies and showed no interest
> in starting using it (I proposed it in the past), so now I have to find
> a way to include the required files. Possible solutions: a) manually
> include each class in each file; b) include all files in the beginning;
> c) use the generated Composer autoloader; d) write another smaller
> autoloader. I think that c) is the best solution, but I would love to
> hear your opinions as well.

Looks like c) is the best, but I'm not familiar enough with Composer to
be sure, so I'll welcome other opinions.
```

lem9:
Given that the library's error messages will be localized using phpMyAdmin's gettext mechanism, is it still relevant to include it via Composer instead of being directly in the repo?

udan11:
I did my best to make the library independent of any other component, but making it independent of phpMyAdmin's gettext mechanism would require implementing a new system for handling translations and it's not worth it and would also complicate the translation process a lot.

I still think that the best method to manage the libraries is via Composer, even though they are localized by phpMyAdmin.

lem9:
But I suspect that the way you implemented it in the other pull request, won't work.

PR #1800, opened 2015-07-19
Fix undefined index when executing some queries.

udan11:
`$analyzed_sql_results['select_tables']` is set only when `$analyzed_sql_results['select_from']` is true.

Use `CREATE DATABASE test;` for testing.


lem9:
Thanks.

PR #1801, opened 2015-07-20
'Host name' is the terminology used by MySQL

madhuracj:
'Host name' is the terminology used in http://dev.mysql.com/doc/refman/5.7/en/account-names.html


lem9:
Thanks.

PR #1807, opened 2015-07-20
Improved localization.

udan11:
Improved localization.
Updated the sql-parser library to udan11/sql-parser@e5fc868.

I didn't get to test it. I will tomorrow.

I had to refactor the code a little bit to be able to use user's preferences (language selection).

### DIFF
--- a/js/codemirror/addon/lint/sql-lint.js
+++ b/js/codemirror/addon/lint/sql-lint.js
@@ -30,7 +30,8 @@ CodeMirror.sqlLint = function(text, updateLinting, options, cm) {
         method: "POST",
         url: "lint.php",
         data: {
-            'sql_query': text
+            sql_query: text,
+            token: PMA_commonParams.get('token'),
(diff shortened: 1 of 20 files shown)


udan11:
Tested. Works fine after `*.po` and `*.mo` files are updated.

udan11:
It didn't find the messages because they weren't constants, but stored in variables. Now, I modified the library to use the `__` function. I tried to make the library independent from the `__` function, but it resulted a mess. The new approach is better.

lem9:
Please squash so I can have a better look.

lem9:
Dan,
when I merge your branch to the current master branch and I try it, I have a problem. Let's say I type
````
SLECT
````
in the SQL query form. The hint on the red X says
````
Unexpected beginning of statement. (near %s)
````
So parameter subtitution does not happen.

lem9:
Nice job.

PR #1809, opened 2015-07-21
Fix unexpected behaviour of Processlist filter

devenbansod:
Removed the unnecessary and problem-causing URL parameter for `showExecuting` as there is already a checkbox input related to it in the form and the checked property for that is properly set, keeping in mind the submitted request parameters.

Moreover, this extra url param causes a problem in the behavior that, once checked, the checkbox always remains checked as $_REQUEST['showExecuting'] was always 1.

This PR should fix the above issue.


madhuracj:
Looks good. Thanks

PR #1810, opened 2015-07-22
Show form group header properly in settings forms

nisargjhaveri:
It was broken since 431847b39ea32cbd20a7e58117df5029540f0e3c


lem9:
Nice catch.

PR #1814, opened 2015-07-23
Fix #1813, Delete rows using "Check All" is broken

### LINKED ISSUE #1813: Delete rows using "Check All" is broken
Hello,

I am using the latest PMA 4.4.12.

Today I lost the whole content of a table, because the "Check All" button is broken if you want to deselect some of the entries.

Steps to reproduce:

Create a simple table and add some rows to it.
Browse the table and click "Check All".
Deselect some entries.
Click "Delete".

Result: All rows will be deleted, even the deselected ones.

This is a very serious bug. I consider it as critical.

Best regards
Michael
pma_step1
pma_step2
pma_step3

nisargjhaveri:


js/functions.js
```diff
@@ -787,25 +788,15 @@ AJAX.registerOnload('functions.js', function () {
             // usual click
 
             // XXX: FF fires two click events for <label> (label and checkbox), so we need to handle this differently
-            var $checkbox = $tr.find(':checkbox');
-            if ($checkbox.length) {
```
nisargjhaveri: I don't think this condition makes any sense as the event is fired only on checkboxes. Did I miss something?

lem9:
Looks fine, thanks!

PR #11323, opened 2015-07-27
Avoid caching JS and CSS files across phpMyAdmin versions

madhuracj:
We can see quite a lot of JS errors in our error reporting server mainly due to not clearing browser cache upon version upgrade. This patch avoid caching upon version upgrade by appending version as a parameter.


lem9:
Thanks.

PR #11333, opened 2015-07-29
Replication improvements

madhuracj:


lem9:
@madhuracj Something looks wrong: with this PR the panel for slave configuration looks very different than the one in current master.

PR #11335, opened 2015-07-30
Replace inefficient jQuery selectors.

Tithugues:
See tips 4th on http://www.sitepoint.com/efficient-jquery-selectors/

### DIFF
--- a/js/console.js
+++ b/js/console.js
@@ -72,10 +72,10 @@ var PMA_console = {
         }
 
         // Vars init
-        PMA_console.$consoleToolbar = $('#pma_console>.toolbar');
-        PMA_console.$consoleContent = $('#pma_console>.content');
-        PMA_console.$consoleAllContents = $('#pma_console .content');
-        PMA_console.$consoleTemplates = $('#pma_console>.templates');
+        PMA_console.$consoleToolbar = $("#pma_console").find(">.toolbar");
+        PMA_console.$consoleContent = $("#pma_console").find(">.content");
+        PMA_console.$consoleAllContents = $('#pma_console').find('.content');
+        PMA_console.$consoleTemplates = $('#pma_console').find('>.templates');
 
         // Generate a from for post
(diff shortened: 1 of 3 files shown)


zixtor:
Thanks, makes sense! Just to be sure I ran some profiling at http://jsperf.com/pma, using PMA's HTML covering most selectors touched by this PR, observations are:-
- Confirmation that using $("#id").find(".class") is much faster compared to $("#id .class") selector.
- Exception to this observation was selector of the form $('#pma_console_options input[name=always_expand]'), which was faster than using find, in all other cases find was faster.
- About timescales and practical difference: for thousands of iterations it's about few milliseconds, so it seems that it would make some noticeable difference only if we use some bad selector repeatedly in a very long loop, since users are used to way longer to/fro timings with server. Anyway, we should consciously aim at best selectors.

Tithugues:
Thanks Atul for your feedback. I didn't know this tool which seems to be quite useful.
So you're right, maybe the selector you mentioned can be reverted.

Last thing, about the last selector. Like said in the link I gave, #page_name should be unique, so no need to search for a parent. Maybe, that would be more efficient.
I updated your test case and added this last selector with only the second ID. Let's see the result.

Tithugues:
Oh, it seems that if you want to see "my" test case, it's a different URL: http://jsperf.com/pma/2

Tithugues:
Sorry, because of missing information, my added test wasn't saved… Here is the new version WITH the new test: http://jsperf.com/pma/3

Tithugues:
No possible comparison… $("#page_name") is faster than $("#name-panel").find("#page_name") which is faster than $("#name-panel #page_name").

To change:
- revert $('#pma_console_options input[name=always_expand]') modification
- replace $("#name-panel").find("#page_name") by $("#page_name")

Thanks Atul!

zixtor:
Hi Hugues, I agree and confirm better profiling result with the last improvement as well as the suggested revert. Thanks.

Tithugues:
Done!

PR #11337, opened 2015-07-31
Implemented new formatter in the sql-parser library

udan11:
Implemented a new formatter in the [sql-parser](https://github.com/udan11/sql-parser) library.

Some improvements can be seen by using this query:

``` sql
 -- This is a MySQL

          -- dump.

-- incomplete query
SELECT DISTINCT * FROM foo WHERE

-- working query with inline clauses
SELECT * FROM foo WHERE id > 10 LIMIT 0, 10

-- this statement caused issues with the old formatter
START TRANSACTION;


                -- some

    -- more

            -- comments


CREATE TABLE `tb` (`uid` INT UNSIGNED NOT NULL
`position` INT NOT NULL,
PRIMARY KEY ( `uid` ) ,
INDEX ( `position` )
) ENGINE = InnoDB;

COMMIT;

CREATE DEFINER /* another comment */ = 'admin'@'localhost' PROCEDURE account_count()
SQL SECURITY INVOKER
BEGIN
  SELECT 'Number of accounts:', COUNT(*) FROM mysql.user;
END;

-- this one also caused issues
-- it put the ; near DELIMITER and created an invalid query
DELIMITER     ;

SELECT * FROM db.table;
```

P.S. In the current master, after formatting the above query a server error occurs because the lexer crashes when it tries to tokenize the invalid "DELIMITER;" statement. That was fixed too in udan11/sql-parser@576b70df3dc93b7b723a9a16fafee5265ab95803 which is already included in this pull request.


lem9:
Nice work; it seems that you know your library very well.

PR #11341, opened 2015-08-01
Fix #11340.

### LINKED ISSUE #11340: Not all results are displayed
Hello,

I am not sure if this is intentional or not, but I see that the behavior of the following query changed:

    USE sakila;
    SELECT * FROM actor;
    SELECT * FROM address;

In [master](http://demo.phpmyadmin.net/master-config/), it looks like it displays results only for the third query and in [4.4](http://demo.phpmyadmin.net/QA_4_4/?pma_username=root) it looks like it displays results only for the first and the second query.

At first, I thought I broke something with the new parser, but I tested against 053f7fc, which is the last commit of the pull request that implemented the [new parser](https://github.com/phpmyadmin/phpmyadmin/pull/1728) and it worked fine.

udan11:

### DIFF
--- a/import.php
+++ b/import.php
@@ -638,7 +638,7 @@ if (! $error) {
         // Do the real import
         try {
             $default_fk_check = PMA_Util::handleDisableFKCheckInit();
-            $import_plugin->doImport();
+            $import_plugin->doImport($sql_data);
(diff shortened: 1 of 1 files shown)


lem9:
Thanks.

PR #11343, opened 2015-08-02
[WIP] Rewrote a part of the import mechanism.

### LINKED ISSUE #11340: Not all results are displayed

udan11:
Rewrote a part of the import mechanism.
Updated sql-parser library to udan11/sql-parser@9130cca0d17169b4647d39b644cc6483b600623c

This is still work in progress. I didn't get to test very much. This should fix ~~#11340~~ 11339 too.

### DIFF
--- a/libraries/plugins/import/ImportSql.class.php
+++ b/libraries/plugins/import/ImportSql.class.php
@@ -26,61 +26,6 @@ class ImportSql extends ImportPlugin
     const READ_MB_TRUE = 1;
 
     /**
-     * @var string SQL delimiter
-     */
-    private $_delimiter;
-
-    /**
-     * @var int SQL delimiter length
-     */
-    private $_delimiterLength;
-
-    /**
-     * @var bool|int SQL delimiter position or false if not found
-     */
-    private $_delimiterPosition = false;
-
-    /**
-     * @var int Query start position
-     */
-    private $_queryBeginPosition = 0;
-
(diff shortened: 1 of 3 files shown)


lem9:
Dan, could you add a comment containing checkboxes describing the remaining work?

lem9:
@udan11 I get an error when trying to import actor.sql (from sakila).

udan11:
In my opinion it is done, but I didn't get to test too much and that's why I said it is work in progress. I haev a feeling that the MySQL specific statements will be ignored (those between /*! */) and I'll have to test against that too.

udan11:
For large dumps, lexing the entire dump, initializing tokens and parsing them to delimit queries is a lot slower than the parser the import mechanism already uses.

PR #11351, opened 2015-08-05
Added WAMPs to be more impartial, updated links

archer87pl:
Signed-off-by: Dariusz Handzlik handzlik.d@gmail.com


lem9:
To really be impartial, we should not mention any package. Please submit a new pull request that follows this intention.

PR #11359, opened 2015-08-06
Fixes cURL timeout.

### LINKED ISSUE #10250: Displayed git revision info is not set
On the PHPMyAdmin home page after logging in, it shows the git revision branch, date and user, but they are not set after I've checked out the code from git. Screen shot attached.

---
* **Original URL**: https://sourceforge.net/p/phpmyadmin/bugs/3860/
* **Original author**: coolbru
* **Found in version**: Latest_Git
* **Attachment**: http://sourceforge.net/p/phpmyadmin/bugs/3860/attachment/Screen%20Shot%202013-03-26%20at%2011.48.24.png

udan11:
cURL tries to reverse DNS the server and because it cannot it waits for the timeout.
It depends on machine configuration.

Fixes #10250.

### DIFF
--- a/libraries/Config.class.php
+++ b/libraries/Config.class.php
@@ -778,6 +778,7 @@ class PMA_Config
         curl_setopt($handle, CURLOPT_SSL_VERIFYPEER, 0);
         curl_setopt($handle, CURLOPT_CONNECTTIMEOUT, 5);
         curl_setopt($handle, CURLOPT_TIMEOUT, 5);
+        curl_setopt($handle, CURLOPT_IPRESOLVE, CURL_IPRESOLVE_V4);
         if (! defined('TESTSUITE')) {
             session_write_close();
         }


udan11:
This fixed #10250 for me because the request to GitHub API failed. GitHub doesn't use IPv6 for their API and that's why the reverse DNS failed.

```
➜  ~  curl --ipv6 api.github.com
curl: (6) Could not resolve host: api.github.com
➜  ~  curl --ipv4 api.github.com
curl: (7) Failed to connect to api.github.com port 80: Connection refused
```

lem9:
Looks fine, thanks.

udan11:
Unfortunately, that bug report is quite old and I am not 100% sure if this was the issue in the first place.

I tried debugging the code responsible and it could not find the commit details in the `.git` folder which made it send a request to GitHub which failed.

I believe that there should be a way to get all details from the .git folder and get rid of GitHub API, but I am not sure how.

lem9:
Well, fixing one possible issue is a good way, and we'll see if this reappears.

libraries/Config.class.php
```diff
@@ -778,6 +778,7 @@ public function checkHTTP($link, $get_body = false)
         curl_setopt($handle, CURLOPT_SSL_VERIFYPEER, 0);
         curl_setopt($handle, CURLOPT_CONNECTTIMEOUT, 5);
         curl_setopt($handle, CURLOPT_TIMEOUT, 5);
+        curl_setopt($handle, CURLOPT_IPRESOLVE, CURL_IPRESOLVE_V4);
```
nijel: Sorry but this completely breaks on IPv6 only networks...

PR #11365, opened 2015-08-07
Remove leading underscores on protected fields.

Tithugues:
Rename fields with pMA standards.

### DIFF
--- a/libraries/controllers/StructureController.class.php
+++ b/libraries/controllers/StructureController.class.php
@@ -42,73 +42,73 @@ class StructureController extends Controller
     /**
      * @var string  Indicate the db_structure or tbl_structure
      */
-    protected $_type;
+    protected $type;
 
     /**
      * @var string  The database name
      */
-    protected $_db;
+    protected $db;
 
     /**
      * @var string  The table name
      */
-    protected $_table;
+    protected $table;
 
     /**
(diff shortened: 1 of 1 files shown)


D-storm:
I am not sure whether we should remove leading underscores from protected fields as they are good indicators that field is not public and a standard naming convention.

Tithugues:
Le ven. 7 août 2015 à 19:28, Chirayu Chiripal <notifications@github.com> a
écrit :

I am not sure whether we should remove leading underscores from protected
fields as they are good indicators that field is not public and a standard
naming convention.

—
Reply to this email directly or view it on
<https://github.com/phpmyadmin/phpmyadmin/pull/11365#issuecomment-128772813>
GitHub
<https://github.com/phpmyadmin/phpmyadmin/pull/11365#issuecomment-128772813>.


Hi,

I agree with this and would like to talk about this. I didn't do it because
I agree with this, but because this is one of pMA standards.

For me too, everything not-public should be prefixed with an underscore.

Can we review our position regarding this standard?

H.

Tithugues:
That's strange, I don't have this error running Travis on my repository: https://travis-ci.org/Tithugues/phpmyadmin/builds/74488023

Tithugues:
This PR is closed. I'll see this modifications later when trying to use PSR rules.

PR #11366, opened 2015-08-07
Improved schema svg export

nobodypb:


lem9:
Please explain what you are improving here.

nobodypb:
Sorry, got not much time ;)
In general it's readability. I changed the colors to be better visible. 
I also removed the "fill" style from Line elements and replaced it with "stroke" as it has no impact on such thin lines. Same with of tables wasn't working correctly, I've fixed that too.
And finally I've added the viewBox attribute to the svg tag, and used the right dimensions for the svg.

I also replaced all line elements with polyline paths in a local version for better rendering in a local version of mine, but thats not part of this commit.

lem9:
Thanks. I rebased your commits into just one: e29119cb64436e35b6b4faa4138e74e425540176 and merged that. See inside this commit for this line:
````
Signed-off-by: Sebastian Ott <sebastian@it-ott.de>
```` 
as we only need this line, not the whole DCO.

Issue #11367, opened 2015-08-07
Export after search, missing WHERE clause


profhun:
When you use the Search panel, and the phpmyadmin shows only one record, then push the export button bellow ( choose: 'view output as text' option, don's download the sql ).
Normally this is return the selected/filtered records.
In this (active 4.4.12) version, the phpmyadmin returned the all records, without any filter.

lem9:
It was expected that you would select which rows you wanted (in your case, the one row) before clicking on the "With selected: Export" link below. Maybe we should deactivate these links when nothing has been selected?

devenbansod:
The Export Button in the `Query results operations` section below works fine.

But, the Multi - submit options of Edit, Delete and Export, that @profhun has mentioned do have an unexpected behavior.

I think Marc's suggestion seems good. 
OR
May be we can replicate the behavior of the mult-submits on Structure page which is : 
If one submits/clicks these links without selecting any columns it shows a message saying 'No Column selected'. We can do a similar thing for Browse page as well.

Which of these seems more better ?

profhun:
Another aspect:
You have a table with a million record, then you run a query with where... you gain some hundred records. You see just the first 25/50/100/200 (), and the another few hundred doesn't...

I think the checkboxes are not a good solution. please redo the old method back, or create a new "export whole resultset" button

lem9:
@profhun This other case is correctly handled with Query results operations / Export.

lem9:
@devenbansod Your suggestion seems better, with "No row selected".

devenbansod:
Thanks for the feedback, Marc.

Please see my PR #11369 which tries to fix this issue.

lem9:
@Tithugues IMO this is a bug.

PR #11369, opened 2015-08-08
Fix Issue #11367

devenbansod:
Force the user to select at least one row from the query results if he/she wants to perform an action on some of the rows.

Shows a `No row selected` message if none selected.


lem9:
Nice fix, thanks.

PR #11370, opened 2015-08-08
Recover test coverage

gzzhanghao:
This PR is going to recover test coverage after I made changes to the codebase in the GSoC project. I will first recover some tests on business logics, then write tests for the templates.

Signed-off-by: Jason jason.daurus@gmail.com


zixtor:
Thanks!

PR #11371, opened 2015-08-08
Update sql-parser.

udan11:
54e80e9 just refactores field to something more relevant depending on the context a411931 introduces support for queries inside comments (e.g. `/*! SELECT * FROM test; */`).

Support for queries inside comments was limited (it could be executed, but sorting or any other operation did not work) in 4.4 and when I replaced the parser I removed the support completely. Now, I am reintroducing it.


lem9:
Thanks.

PR #11372, opened 2015-08-09
Manage templates unit tests.

Tithugues:
Attempt to structure unit tests on templates.

I'm not sure about where declaring this in phpunit.xml.\* files, so if something is wrong (for example, phpunit.xml.hhvm not updated, I don't know if I should because "libraries" is "missing"), please tell me.

I also used a new package in template files: PhpMyAdmin-template. Is this OK? Should we use another name? Should we use package?

Thanks.

### DIFF
--- a/test/templates/Logo_test.php
+++ b/test/templates/Logo_test.php
@@ -1,0 +1,93 @@
+<?php
+/* vim: set expandtab sw=4 ts=4 sts=4: */
+/**
+ * Test for templates/logo view
+ *
+ * @package PhpMyAdmin-test
+ */
+
+require_once 'libraries/Template.class.php';
+
+/**
+ * Test for templates/logo view
+ *
+ * @package PhpMyAdmin-test
+ */
+class Logo_Test extends PHPUnit_Framework_TestCase
+{
+    const VIEW = 'logo';
+
+    /**
+     * Sets up the fixture, for example, open a network connection.
+     * This method is called before a test is executed.
+     *
+     * @return void
+     */
+    protected function setUp()
+    {
+        $GLOBALS['pmaThemeImage'] = './themes/pmahomme/img/';
+    }
+
+    /**
+     * Test view
+     *
+     * @return void
+     */
+    public function testView()
+    {
+        $html = PMA\Template::get(self::VIEW)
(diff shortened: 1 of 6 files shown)


test/templates/Logo_test.php
```diff
@@ -0,0 +1,93 @@
...
+     */
+    public function testViewWOParameters()
+    {
+        $html = PMA\Template::get(self::VIEW)->render();
+
+        $this->assertContains(
+            '<img src="' . $GLOBALS['pmaThemeImage'] . 'logo_left.png" '
+            . 'alt="phpMyAdmin" id="imgpmalogo" />',
+            $html
+        );
```
zixtor: Hi Hugues,

As regards unit testing of templates (which essentially return just html); at IRC meeting [0], we felt not to carry forward the practice of asserting presence of few HTML tags in a way bigger HTML which may prove a rather strict and undesirable way of testing in long term and will frequently break with changes in HTML during development. So instead of that we thought it would be better to "assert the data [variables being passed into the template against model values] and just parse the whole HTML to see if there is DOM error". So far, we don't have an example where it's being so done, so I am not further sure about actual mechanics of doing it, but surely sounds feasible.

Also, I think, may be in short run at least, we should attach more priority to increasing unit test coverage of actual logic code over template code.

[0] https://wiki.phpmyadmin.net/pma/2015-06_Meeting#Unit_tests_on_HTML

Tithugues: Hi,

Thanks, I didn't see this IRC meeting feedback. I should admit that I don't really see how to assert data. Do you mean the result of the logic before calling the template?

My idea wasn't to add unit tests of all templates, but only to give a structure so we'll be able to do it later.
But I agree that the logic might be more sensitive and so should be the priority.

zixtor:
I think @gzzhanghao will give a try to this proposed way, after the preferred unit tests are recovered.

gzzhanghao:
@zixtor @Tithugues 
Hello, I am writing tests on templates and actions in controllers. I created a stub for the PMA_Response object which receive HTML strings and finally concatenate them together, on which we can check for the validity of the variables passed into the template easier. By this way we can also write tests for actions in controllers.
I will push some tests with the `ResponseStub` at today mid-noon.

PR #11390, opened 2015-08-12
Fix #11389 ReCaptcha produce deprecated messages under PHP 7

### LINKED ISSUE #11389: ReCaptcha produce deprecated messages under PHP 7
Possible solution: https://github.com/google/recaptcha. In their current master there are more than one file to implement the full logic and in the newer ReCaptcha.php they now use the proper __construct() which should get rid of the deprecated message. Watch for API changes but they have examples.

Target: QA_4.4

madhuracj:
Upgraded ReCaptcha library to version 1.1.1. 
However, 1.1.1 does not have a autoload.php file to support installations that do not use composer. autoload.php file is taken from the master branch.


lem9:
Thanks.

PR #11391, opened 2015-08-12
Fix #11387 phpseclib < 2.0 produces deprecated messages on PHP

### LINKED ISSUE #11387: phpseclib < 2.0 produces deprecated messages on PHP 7
We are using phpseclib < 2.0. I contacted one of their developers who told me that we could try https://github.com/phpseclib/phpseclib/tree/2.0 which does use __construct(). However, due to it being namespace'd, the API is slightly different.

Target for this should be QA_4_4 because we have to be ready for PHP 7.

madhuracj:
Update phpseclib to version 2.0.0


lem9:
Thanks.

PR #11395, opened 2015-08-13
Improve partition support

madhuracj:
Work in progress


madhuracj:
I think this is ready to be merged.

lem9:
Nice improvement! Maybe put the Partitions section inside a slider as we do for Indexes?

Issue #11399, opened 2015-08-15
Illegal offset type


lem9:
Using the git master branch: Open a db, click on Tracking. In the list of untracked table, click on "Track table" next on one table name:
````
in ./libraries/Index.class.php#244
 Illegal offset type
````
git bisect tells me that the culprit is 66361530713fcf99b9e090b82ccd016dc3c287b2

gzzhanghao:
@lem9 
Hello, sorry I cannot find the "Tracking" after opening a db, is there any requirements on the database engine or format etc?

lem9:
@gzzhanghao See
 http://docs.phpmyadmin.net/en/latest/config.html?highlight=tracking#cfg_Servers_tracking

gzzhanghao:
@lem9 
Okay, I had successfully reproduced the bug. Try fixing.

gzzhanghao:
@lem9 
Hello, now the bug should be fixed. Please review my [PR#11418](https://github.com/phpmyadmin/phpmyadmin/pull/11418).

PR #11418, opened 2015-08-20
Fix bug#11399.

gzzhanghao:
The bug fix for issue #11399.

The root cause of this bug is the variable `$table` which originally holds table name information was got override by a new `$table` that holding PMA_Table object (around line 195 of tbl_tracking.php).
Also, some changes were made to prevent same kind of bug from occurring again.

Signed-off-by: Jason jason.daurus@gmail.com

### DIFF
--- a/libraries/Menu.class.php
+++ b/libraries/Menu.class.php
@@ -165,8 +165,8 @@ class PMA_Menu
     private function _getBreadcrumbs()
     {
         $retval = '';
-        $table = new PMA_Table($this->_table, $this->_db);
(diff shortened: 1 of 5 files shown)


lem9:
Thanks @gzzhanghao. In my tests the bug is fixed, but see my comment and see also the failing tests.

gzzhanghao:
Thanks @lem9. I have fixed the typo now.
I noticed the failing test, it seems it's not caused by my commits because it starts failing from the upstream.

lem9:
Great. Rebased and merged.

Issue #11407, opened 2015-08-17
phpMyAdmin SQL Dump - Export/Import (CREATE TABLE IF NOT EXISTS)


llcool:
-- phpMyAdmin SQL Dump
-- version 4.2.12deb2
-- http://www.phpmyadmin.net
--
-- Host: localhost
-- Generation Time: Aug 12, 2015 at 10:09 AM
-- Server version: 5.6.25-0ubuntu0.15.04.1
-- PHP Version: 5.6.4-4ubuntu6.2

CREATE TABLE IF NOT EXISTS `alert` (
  `alert_id` int(11) NOT NULL,
  `user_id` int(11) DEFAULT NULL,
  `email` varchar(80) DEFAULT NULL
) ENGINE=MyISAM AUTO_INCREMENT=29 DEFAULT CHARSET=latin1;
.
.
.
ALTER TABLE `alert`
 ADD PRIMARY KEY (`alert_id`);

If table structure exists ALTER TABLE will fail with duplicate key and the import exits any other modification will not run.

Why not go back to CREATE TABLE with PRIMARY KEY.

CREATE TABLE IF NOT EXISTS `alert` (
  `alert_id` int(11) NOT NULL,
  `user_id` int(11) DEFAULT NULL,
  `email` varchar(80) DEFAULT NULL,
  PRIMARY KEY (`alert_id`)
) ENGINE=MyISAM AUTO_INCREMENT=29 DEFAULT CHARSET=latin1;

lem9:
There was a reason to split the statement like this.

llcool:
lem9, You statement seems incomplete, or am I missing something...

nisargjhaveri:
See, #6366 and #5703.

But, yes, it is really frustrating when importing a dump into an already existing databases and it fails with similar error. Can we do something about it?

llcool:
Thank nisargjhaveri, for the info.

Why can't we have ADD PRIMARY KEY IF NOT EXIST

lem9:
@llcool When I have time, I'll search to complete my above comment and clarify the reason why there are two statements.  About your suggestion, looking at http://dev.mysql.com/doc/refman/5.6/en/alter-table.html I don't see that it would be possible.

lem9:
Thanks to @nisargjhaveri comment I have read in #6366 that "creating index and foreign keys only after inserting data makes inserting much efficient. "; but the case @llcool was not taken into account.

[...]

PR #11419, opened 2015-08-20
Create indexes with table creation if "IF NOT EXISTS" is on

nisargjhaveri:


lem9:
Nisarg,
see my objection in #11407 about the intention we can conclude (or not) due to "IF NOT EXISTS" being checked by default.

nisargjhaveri:
I've updated the PR.

lem9:
Nice work.

PR #11422, opened 2015-08-23
fix for routine radio and checkbox width

bkdotcom:
fixes execute enum/set options and the "is deterministic" checkbox on edit


lem9:
Thanks. I tried with Chrome and Firefox the change for the checkbox; I see your intention but it's not working for me. Only the ".rte-table input" has an effect on this checkbox.

bkdotcom:
slop quick fix on my end...  i used `.rte-table` rather than `.rte_table`

lem9:
Rebased and merged, thanks.

PR #11423, opened 2015-08-23
MariaDB 10.0.21 does not support sha256_password auth plugin

devenbansod:
MariaDB 10.0.21 (latest stable version) also does not support sha256_password auth plugin.

I could not pin point a documentation reference but the latest version on the demo server as well as my local instance does not list `sha256_password` in the `SHOW PLUGINS;` query results.

Even if the newest beta version(10.1.6) adds the required support (could not find a docu. ref. again), we would to have a different condition to check which can be added later. 
So, this PR can surely be merged for now.


lem9:
Thanks.

PR #11424, opened 2015-08-23
Fixed an unsupported query in Travis log

devenbansod:


devenbansod:
I cannot reproduce the error but it might be there. My phpunit does not have runkit configured, so it skips the test. I will try to re-install it and then retry the tests.

By the way, I think the build history at Travis [(link)] (https://travis-ci.org/phpmyadmin/phpmyadmin/builds) shows the error started from da1d274

PR #11425, opened 2015-08-23
Improve jQuery selectors

Tithugues:


zixtor:
The failing tests here don't relate to changes in this PR, right?

lem9:
Indeed they don't relate.

PR #11430, opened 2015-08-26
Fix #6297 Use GET only for read only operations 

### LINKED ISSUE #6297: Use GET only for read only operations
Currently we perform some operations by GET and some by POST requests. We should change this to use GET only for read only operations, while anything what performs changes should be done by POST. This way we can remove CSRF token from URLs and pass it only in forms, making URLs nicer.

Per RFC 7231, the POST method should be used for any context in which a request is non-idempotent: that is, it causes a change in server state each time it is performed, such as submitting a comment to a blog post or voting in an online poll. GET is defined to be nullipotent, with no side-effects (in contrast to "no side effects on second or future requests" as with idempotent operations).

---
* **Original URL**: https://sourceforge.net/p/phpmyadmin/feature-requests/1684/
* **Original author**: nijel

madhuracj:
I think this is ready to be merged now.

### DIFF
--- a/js/db_operations.js
+++ b/js/db_operations.js
@@ -137,7 +137,7 @@ AJAX.registerOnload('db_operations.js', function () {
         );
         $(this).PMA_confirm(question, $(this).attr('href'), function (url) {
             PMA_ajaxShowMessage(PMA_messages.strProcessingRequest);
-            $.get(url, {'is_js_confirmed': '1', 'ajax_request': true}, function (data) {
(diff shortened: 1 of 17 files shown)


Tithugues:
@madhuracj you have the same Travis error I had. @nijel fixed it. You should try to rebase your branch to run again Travis.

madhuracj:
Rebased.

nijel:
This looks good, I think we can merge it.

PR #11437, opened 2015-08-27
Fix #11433

### LINKED ISSUE #11433: '%' does not work as it is with sprintf
`%` as hostname value while copying a new user and adding a new user does not work as it is with sprintf.
It may have to be escaped as `%%`.

This creates a problem of *empty query* (esp. for $create_user_stmt) while copying an user in version 4.5.0-beta1 on my local installation.
But I am not able to `copy an user` on demo.phpmyadmin.net and `creating new user` is working fine here.
Can anyone confirm if this problem exists ? or is this problem just with my local machine only ?

devenbansod:
Should fix issue #11433 

`%` character causes problem with sprintf().
And has to be escaped using `%%`

### DIFF
--- a/libraries/server_privileges.lib.php
+++ b/libraries/server_privileges.lib.php
@@ -4905,10 +4905,23 @@ function PMA_getSqlQueriesForDisplayAndAddUser($username, $hostname, $password)
     $slashedUsername = PMA_Util::sqlAddSlashes($username);
     $slashedHostname = PMA_Util::sqlAddSlashes($hostname);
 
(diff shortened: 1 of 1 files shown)


Tithugues:
Hi,
I think that we might have the same issue with the password...
I'll have a look.

lem9:
Thanks.

devenbansod:
@Tithugues : Thanks for bringing it up. I had thought so, but then I could not reproduce any problem with password containing `%` so skipped it in this PR. 
Please let me know if you are able to find an issue.

PR #11439, opened 2015-08-27
login form style fix, sprites css caching fix and extended security-related HTTP headers

herndlm:
Just a bunch of smaller changes we're using locally and thought would be useful to have upstream.

The CSS "box-sizing" feature is in our instance since a few months and should be compatible with all major browsers according to http://caniuse.com/#search=box-sizing

We're using the modified CSS sprites URL even longer and it made upgrades easier (no cached sprites file which results in different images). I just updated the fix to match the method used in cd76cb6acc1af19331dc543db45aa0f60eaaefe6

The HTTP headers are something we're using on different products since almost a year or so. The "X-XSS-Protection" re-enables user disabled XSS filters (they should be on by default), "X-Content-Type-Options" should prevent mime type sniffing attacks on IE and Chrom(e|ium) and "X-Permitted-Cross-Domain-Policies" seems to be an Adobe specific header that prevents clients like e.g. Flash from getting content from phpMyAdmin. It does NOT interfere with "X-Content-Security-Policy".

Signed-off-by: Martin Herndl martin.herndl@world4you.com


monojp:
sorry, messed up

monojp:
Sorry for that. I'm re-doing this in a more clean way

PR #11440, opened 2015-08-27
login form style fix, sprites css caching fix and extended security-related HTTP headers

herndlm:
Just a bunch of smaller changes we're using locally and thought would be useful to have upstream.

The CSS "box-sizing" feature fixes the width differences of the input and select elements on the login page. It is in our instance since a few months and should be compatible with all major browsers according to http://caniuse.com/#search=box-sizing
see before [1] and after [2]

We're using the modified CSS sprites URL even longer and it made phpMyAdmin upgrades easier (no cached sprites file which results in different images and needing to tell customers to press CTRL-F5 or smth. like that). I just updated the fix to match the method used in cd76cb6acc1af19331dc543db45aa0f60eaaefe6

The security-related HTTP headers are something we're using on different products since almost a year or so. The "X-XSS-Protection" re-enables user disabled XSS filters (they should be on by default), "X-Content-Type-Options" should prevent mime type sniffing attacks on IE and Chrom(e|ium) and "X-Permitted-Cross-Domain-Policies" seems to be an Adobe specific header that prevents clients like e.g. Flash from getting content from phpMyAdmin. It does NOT interfere with "X-Content-Security-Policy". More information to those headers are in [3] and [4].

Signed-off-by: Martin Herndl martin.herndl@world4you.com

[1] 
image

[2] 
image

[3] https://www.owasp.org/index.php/List_of_useful_HTTP_headers
[4] http://www.adobe.com/devnet/articles/crossdomain_policy_file_spec.html

### DIFF
--- a/libraries/Header.class.php
+++ b/libraries/Header.class.php
@@ -593,6 +593,22 @@ class PMA_Header
             . $captcha_url
             . ";"
         );
+        // Re-enable possible disabled XSS filters
+        // see https://www.owasp.org/index.php/List_of_useful_HTTP_headers
+        header(
+            'X-XSS-Protection: 1; mode=block'
+        );
+        // "nosniff", prevents Internet Explorer and Google Chrome from MIME-sniffing a
+        // response away from the declared content-type
+        /// see https://www.owasp.org/index.php/List_of_useful_HTTP_headers
+        header(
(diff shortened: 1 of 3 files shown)


lem9:
Thanks. I will also backport the caching theme sprites and the irregular length patches to QA_4_5.

PR #11444, opened 2015-08-31
Fix #11442 MySQL 5.7 and SHOW VARIABLES

### LINKED ISSUE #11442: MySQL 5.7 and SHOW VARIABLES
A new behavior in MySQL 5.7 might cause some problems. Indeed, it seems like many (all?) SHOW VARIABLES statements are disabled by default for unprivileged users, unless show_compatibility_56 is set to ON.  Tested on MySQL 5.7.8.

See http://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_show_compatibility_56_show

madhuracj:


lem9:
Thanks. I'll open a follow-up issue.

PR #11467, opened 2015-09-07
Fixes #11465.

### LINKED ISSUE #11465: Using shell_exec for very large sql files
I was trying to upload a large sql file using phpMyAdmin but there were many errors. When i switched to terminal to execute it then in my surprise there was no error. So can we add an option in phpMyAdmin to evoke a shell exec command for import.[See link on stackoverflow](http://stackoverflow.com/questions/4027769/running-mysql-sql-files-in-php)

udan11:


lem9:
Thanks. Merging as this fixes one of the two problematic scenarios of #11465.

PR #11471, opened 2015-09-10
Fix #11464 phpMyAdmin suggests upgrading to newer version not usable on that system

### LINKED ISSUE #11464: phpMyAdmin suggests upgrading to newer version not usable on that system
phpMyAdmin suggests by default upgrading to a newer version which is not usable on that system. Example: phpMyAdmin 4.0.10.10 is the latest stable version that can be run on RHEL/CentOS 6, because it ships PHP 5.3.3 and MySQL 5.1.73. However:

> Version information: 4.0.10.10, latest stable version: 4.4.14

and:

> A newer version of phpMyAdmin is available and you should consider upgrading. The newest version is 4.4.14, released on 2015-08-20."

I am aware that these checks can be disabled, but shouldn't these checks also pay attention to the versions they depend on? From my point of view the version checks should be maybe per branch/release series (4.0.x) if a newer version (4.4.x) would not run with the versions a server provides.

madhuracj:


version_check.php
```diff
@@ -14,15 +14,23 @@
 // Always send the correct headers
 header('Content-type: application/json; charset=UTF-8');
 
-$version = PMA_Util::getLatestVersion();
+$versionDetials = PMA_Util::getLatestVersion();
```
Tithugues: Not a big deal, but did you mean "details" here?

madhuracj:
With this new version check someone can get hints on what the PHP and MySQL versions are. I wonder whether this constitutes to a security vulnerability.

nijel:
Can you please include testcase for this? There is quite some logic involved...

madhuracj:
PHP Unit does not allow mocking static methods. So I moved all the methods related to version checking to a new class for better organization and testability.

Tithugues:
Thanks, you didn't name the file .class.php and you respected the case of file name regarding class name. The merge with namespacing branch should be quite easy.

lem9:
Maybe I missed something: we no longer use .class.php extension?

lem9:
To test this, I forced current to be '304140100' in PMA_current_version() and forced the PHP version to be '5.0.0' in evaluateVersionCondition(). I got a notice "A newer version..." even though my system with a simulated PHP version is too old to upgrade to phpMyAdmin 4.4.14.1.

madhuracj:
@lem9 
The behavior is due to https://github.com/phpmyadmin/phpmyadmin/pull/11471/files#diff-9b30aec46af47b2d8b2d29269bd89b94R30.
Since phpMyAdmin does not have a (security) supported version that is compatible, the version check falls back to suggest the latest stable version available. I think this makes more sense rather than not suggesting a version upgrade.

lem9:
@madhuracj But the whole point of this issue is to cease suggesting an upgrade that is not compatible with the system, and this fallback defeats this.

madhuracj:
@lem9 Should be fixed now.

madhuracj:
No, if there is a compatible version for upgrading I see the version value there.

Having forced current to be '304140100' in PMA_current_version(), forcing PHP version to following values, I see different versions of phpMyAdmin being suggested for upgrade
PHP = 5.3.0 --> 4.4.15
PHP = 5.2.0 --> 4.0.10.10
PHP = 5.0.0 --> No upgrade suggestion

madhuracj:
@lem9 
My above tests were in a multi server environment, which ignores the MySQL condition. May be, the bug fixed in f70add8 was the reason?

lem9:
Thanks @madhuracj I am happy with your tests.

lem9:
@madhuracj Will you do the cherry-pick for older versions?

madhuracj:
Will do that.

madhuracj:
@lem9 
Cherry-picked to QA_4_4 and QA_4_5 branches. Should I cherry-pick to 4.0 version as well?

lem9:
Yes, in MAINT_4_0_10.
QA_4_4 had been deleted. Please work on MAINT_4_4_15 then redelete QA_4_4.

madhuracj:
Changes pushed to MAINT_4_4_15. Porting to MAINT_4_0_10.

PR #11488, opened 2015-09-16
Browse table bookmark form: Clickable labels

mynetx:
Make text field and check box labels clickable in the Browse table page,
create SQL bookmark form.


lem9:
Hi,
we are trying to clarify whether or not we can accept contributions without a person name; meanwhile, is it possible for you to redo this commit, using a full name, like we show in https://github.com/phpmyadmin/phpmyadmin/blob/master/DCO ?

jublo:
@lem9 lmk about the outcome of your clarification.

lem9:
Thanks!

PR #11494, opened 2015-09-23
Fix #11492, AUTO_INCREMENT statements are partly missing from exports

### LINKED ISSUE #11492: AUTO_INCREMENT statements are partly missing from exports
Hi, I noticed that in 4.5.0 there seem to be bugs in the export. We have a Joomla! database which was missing ~50 % of the statements that add an AUTO_INCREMENT to a column.

I was able to reproduce this easily with a simplified table on the demo server:

```sql
CREATE TABLE `a` (
  `a` int(11) NOT NULL,
  `b` int(11) NOT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8;

ALTER TABLE `a`
  ADD PRIMARY KEY (`a`);

ALTER TABLE `a`
  MODIFY `a` int(11) NOT NULL AUTO_INCREMENT;
```

results in an export that is missing the last ALTER TABLE part regarding the AUTO_INCREMENT. I double checked that adding the AUTO_INCREMENT was checked on the export page and that it really showed up on the structure page.

nisargjhaveri:


lem9:
Nice fix, thanks.

madhuracj:
Thanks @nisargjhaveri

PR #11499, opened 2015-09-24
Fix #11490 UI for defining partitioning in create table window

### LINKED ISSUE #11490: UI for defining partitioning in create table window
Even though phpMyAdmin 4.5 has improved support for partitioning, it is still limited to viewing information on partitions and performing various operations on partitions. There is only a text field in create table window to specify partitioning criteria and it does not guide the user in anyway.

Instead, phpMyAdmin should have an interface where the user can specify the following with regards to partitioning.
1. partitioning algorithm and its expression or column list
2. number of partitions
3. sub partitioning algorithm and its expression or column list
4. number of sub partitions

For each partition and sub partition
1. values included in it
2. storage engine
3. comment etc.

madhuracj:
Ready to be merged.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -4,6 +4,7 @@ phpMyAdmin - ChangeLog
 4.6.0.0 (not yet released)
 + issue #11456 Disabled storage engines
 + issue #11479 Allow setting routine wise privileges
++ issue #11490 UI for defining partitioning in create table window
 
 4.5.1.0 (not yet released)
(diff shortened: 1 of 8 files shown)


lem9:
Nice improvement. Maybe it would be interesting to have access to this interface also on table Structure (to edit partition definitions, or to add partitions to an existing table).

madhuracj:
Sure, I'll give it a try.

PR #11513, opened 2015-09-29
Fix for minor space changes

Anirudhk94:
Minor space changes in README


lem9:
Thanks; please recommit and resubmit if you agree with our DCO (see above), using your full name (see http://wiki.phpmyadmin.net/pma/Git#Before_starting) and the proper Signed-Off-By line.

Issue #11516, opened 2015-09-30
Some partitions are missing in copied table


madhuracj:
Seen on version 4.5
Create the table with 
```
CREATE TABLE `ss` (
 `id` int(11) NOT NULL,
 `purchased` date NOT NULL
) ENGINE=InnoDB DEFAULT CHARSET=latin1
/*!50100 PARTITION BY RANGE (YEAR(purchased))
SUBPARTITION BY HASH (TO_DAYS(purchased))
(PARTITION p0 VALUES LESS THAN (1990)
(SUBPARTITION p0s0 COMMENT = 'a' ENGINE = InnoDB,
 SUBPARTITION p0s1 COMMENT = 'b' ENGINE = InnoDB),
PARTITION p1 VALUES LESS THAN (2000)
(SUBPARTITION p1s0 COMMENT = 'c' ENGINE = InnoDB,
 SUBPARTITION p1s1 COMMENT = 'd' ENGINE = InnoDB),
PARTITION p2 VALUES LESS THAN MAXVALUE
(SUBPARTITION p2s0 COMMENT = 'e' ENGINE = InnoDB,
 SUBPARTITION p2s1 COMMENT = 'f' ENGINE = InnoDB)) */
```
And try to make a copy of it in table operations page.
Some partitions are missing in the copied table

udan11:
This is because of a bug in the SQL parsing library.

madhuracj:
@udan11 
Thank for the fix. I incorporated it in my commit.
There are a couple of points I noted and they are
1. `STORAGE` is not used without `ENGINE`
2. Only `RANGE` and `LIST` type partitions support 
```
[VALUES 
            {LESS THAN {(expr) | MAXVALUE} 
            | 
            IN (value_list)}]
```
section.

udan11:
@madhuracj Can you give me an example that wasn't parsed properly due to point 2?

Thanks!

madhuracj:
@udan11 

```
CREATE TABLE `a` (
 `a` int(11) NOT NULL,
 `b` int(11) NOT NULL
) ENGINE=InnoDB DEFAULT CHARSET=latin1
/*!50100 PARTITION BY HASH (a)
(PARTITION p0 MAX_ROWS = 10 MIN_ROWS = 1 COMMENT = '1' ENGINE = InnoDB,
PARTITION p1 MAX_ROWS = 10 MIN_ROWS = 1 COMMENT = '2' ENGINE = InnoDB) */
```

PR #11517, opened 2015-09-30
Allow adding and editing partitions

madhuracj:
When editing a partitioned table, some partition information are missing due to https://github.com/phpmyadmin/phpmyadmin/issues/11516


lem9:
Very nice!

PR #11525, opened 2015-10-01
Fix #11520 and some other improvements

### LINKED ISSUE #11520: Incorrect syntax creating an user using mysql_native_password with MariaDB
According to MariaDB docs (https://mariadb.com/kb/en/mariadb/development-pluggable-authentication/), the correct syntax to create an user with mysql_native_password plugin is:
```
CREATE USER asd IDENTIFIED VIA mysql_native_password USING '***';
```
And today phpMyAdmin 4.5.0.2 is using this syntax:
```
CREATE USER 'asd'@'%' IDENTIFIED WITH mysql_native_password BY '***';
```

This is generating errors "1065 - Query was empty" error above the printed query and "The selected user was not found in the privilege table." right on top of the user creation wizard.

I was able to create the user directly in SQL Query with the syntax pointed at MariaDB docs. (got password hash using `select PASSWORD('test');`).

Versions:
phpMyAdmin 4.5.0.2
MariaDB 10.0.20-MariaDB

devenbansod:
Tries to fix #11520

Also, some improvements in the related code in a separate commit.
Previously, we were not allowing to create users with user-added auth plugins, now this is allowed.

As this involves many different versions of MySQL as well as MariaDb, please report if you get any failures in any operation, due to this commit.

### DIFF
--- a/libraries/display_change_password.lib.php
+++ b/libraries/display_change_password.lib.php
@@ -77,73 +77,33 @@ function PMA_getHtmlForChangePassword($username, $hostname)
         . $chg_evt_handler . '="nopass[1].checked = true" />'
         . '</td>'
         . '</tr>';
+
+    $active_auth_plugins = PMA_getActiveAuthPlugins();
+
     $default_auth_plugin = PMA_getCurrentAuthenticationPlugin(
         'change', $username, $hostname
     );
 
-    // See http://dev.mysql.com/doc/relnotes/mysql/5.7/en/news-5-7-5.html
-    if (PMA_Util::getServerType() == 'MySQL'
-        && PMA_MYSQL_INT_VERSION >= 50705
-    ) {
-        $html .= '<tr class="vmiddle">'
-            . '<td>' . __('Password Hashing:') . '</td>'
-            . '<td>'
-            . '<input type="radio" name="pw_hash" id="radio_pw_hash_mysql_native" '
-            . 'value="mysql_native_password"';
-        if ($default_auth_plugin == 'mysql_native_password') {
-            $html .= '" checked="checked"';
+    $html .= '<tr class="vmiddle">'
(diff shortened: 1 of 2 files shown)


lem9:
@devenbansod I was testing this PR and saw a problem. When listing user accounts, for one of them a red "No" is shown in the password column, but this account has a password and QA_4_5 shows "Yes".

devenbansod:
@lem9 : Thanks for the test. Can you please provide the MySQL/MariaDb version information ?

About squashing, I had thought that both were for different purposes(one fixed it, other made code improvements), so I had kept them as different commits. 
Anyways, will squash them once I fix the error you pointed out.

lem9:
MySQL 5.5.44. About fixes and code improvements: in QA_4_5 only fixes should go. I suggest sending the code improvements to master in a separate pull request. So please do not squash :)  Closing for now.

devenbansod:
Sorry, but I feel that the code improvements(I don't know what else to call it) also indeed fix a possible shortcoming that users with independent authentication plugins might face.

Explanation:
While creating in a new user, since we have hard-coded the options for choosing the authentication plugin (namely `mysql_native_password`, `mysql_old_password` and for 5.6+ `sha256_password`), the users wiil not be able to create users with any other auth plugin that they might have installed.

So should we consider it as a bug and target QA_4_5 or consider as a new feature and target master instead ?

lem9:
Looks like a bug. Please open a separate issue and include the description above, then a separate pull request to solve this other issue. It's easier to review code fixing just one issue at the time.

lem9:
@devenbansod Hi Deven, were you able to reproduce my "red No" password?

devenbansod:
Hi, Sorry @lem9 . I was not able to devote time as exams had come up. 
Can I can look over it tomorrow ?

Also, I will open a new issue and open 2 PRs separate to fix them separately as you suggested.

PR #11532, opened 2015-10-02
Fix #11531.

### LINKED ISSUE #11531: Invalid SQL in table definition when exporting table
Steps to reproduce:

1. Create a table with
    1. an `id` field of type INT as primary key,
    2. and another field of type ENUM, where the enumeration contains an equal sign as a possible value that is selected by default.
2. Export the table.

The expected result and the result the latest release gives can be found in the attached file. The expected result was produced by version 4.0.10; the invalid SQL was generated by the latest release (4.5.0.2). Server (Ubuntu 14.04) runs Apache 2.4.7 with PHP 5.5.9.

udan11:
Fix #11531.
Cherry-picked udan11/sql-parser@e5b1cd830d2156127ba090d0d173e344db5fcfd1.

### DIFF
--- a/libraries/sql-parser/src/Components/OptionsArray.php
+++ b/libraries/sql-parser/src/Components/OptionsArray.php
@@ -197,7 +197,7 @@ class OptionsArray extends Component
                 }
             } elseif ($state === 1) {
                 $state = 2;
-                if ($token->value === '=') {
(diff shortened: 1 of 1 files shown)


lem9:
Thanks Dan.

PR #11537, opened 2015-10-05
Fix #11522 Strange behavior on table rename

### LINKED ISSUE #11522: Strange behavior on table rename
So, by accident I renamed a table a from chemicals_copy_two to chemicalsCopy2 and an issue with phpmyadmin (4.4.14+) not being able to read the config of the table arose.  As far as I can tell this is a PMA issue as the name is valid based on MySQL guidelines.  I don't know what the issue is...

See https://youtu.be/_neeYqlh6TU

madhuracj:


lem9:
I haven't tested but it makes sense.

PR #11552, opened 2015-10-08
Fix #11551 Fatal error when switched to master from QA_4_5

### LINKED ISSUE #11551: Fatal error when switched to master from QA_4_5
```
Fatal error: main(): The script tried to execute a method or access a property of an incomplete object. Please ensure that the class definition &quot;PMA_Theme_Manager&quot; of the object you are trying to operate on was loaded _before_ unserialize() gets called or provide a __autoload() function to load the class definition in C:\wamp\apps\phpmyadmin\libraries\common.inc.php on line 662
```

Clearing cache does not help. Clearing cookies fixes this. 
However we should be able to clear/invalidte cookie automatically upon upgrading

madhuracj:


lem9:
Does not work for me, going back to QA_4_5 from your branch. By the way, looking at the logic for pmaCookieVer in 4.5, this cookie cannot be created for someone who starts to use phpMyAdmin in some browser, because the setCookie() is inside the "if".

madhuracj:
Not sure what I was thinking when I did the commit 54b0409. I've sent a pull request for 4.5 at https://github.com/phpmyadmin/phpmyadmin/pull/11554

lem9:
All is fine now, thanks.

PR #11559, opened 2015-10-10
Warning "PHP 5.5+ is required" changed in common.inc.php 

abhishekkanojia:
The warning " PHP 5.5+ is required " has been replaced with the new warning which indicates the current version installed in the user's system and also provided a link to download the latest version of PHP. This provides a basic idea to the user about what PHP version is running on their system and provides a link to update the same.
Signed-off-by: Abhishek kanojia abhishekka3193@gmail.com

### DIFF
--- a/libraries/common.inc.php
+++ b/libraries/common.inc.php
@@ -56,7 +56,7 @@ if (getcwd() == dirname(__FILE__)) {
  * PHP 5 function, so cannot easily localize this message.
  */
 if (version_compare(PHP_VERSION, '5.5.0', 'lt')) {
-    die('PHP 5.5+ is required');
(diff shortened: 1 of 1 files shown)


lem9:
Thanks. Not sure we want to add this option because it gives the false impression that changing PHP is simply a matter of downloading it. Anyway, please have a look at http://www.smashingmagazine.com/2012/06/links-should-never-say-click-here/ and also, squash your commits.

lem9:
As the majority of our users are running phpMyAdmin from a prepackaged suite, a distro or an ISP, they are powerless regarding a PHP change. You can resubmit a pull request that just shows the current PHP version. The Signed-Off-By line should be part of the git commit comments and not the github comments. Your author name should be configured in git, see http://wiki.phpmyadmin.net/pma/Git#Before_starting.

PR #11560, opened 2015-10-10
Fix #11520

### LINKED ISSUE #11520: Incorrect syntax creating an user using mysql_native_password with MariaDB

devenbansod:

### DIFF
--- a/libraries/server_privileges.lib.php
+++ b/libraries/server_privileges.lib.php
@@ -3422,26 +3422,21 @@ function PMA_getHtmlTableBodyForUserRights($db_rights)
             $html_output .= '<td>';
 
             $password_column = 'Password';
+            $serverType = PMA_Util::getServerType();
 
-            if (PMA_Util::getServerType() == 'MySQL'
-                && PMA_MYSQL_INT_VERSION >= 50606
-                && PMA_MYSQL_INT_VERSION < 50706
+            $check_plugin_query = "SELECT * FROM `mysql`.`user` WHERE "
+                . "`User` = '" . $host['User'] . "' AND `Host` = '"
+                . $host['Host'] . "'";
+            $res = $GLOBALS['dbi']->fetchSingleRow($check_plugin_query);
+
+            if ((isset($res['authentication_string'])
+                && ! empty($res['authentication_string']))
+                || (isset($res['Password'])
+                && ! empty($res['Password']))
             ) {
-                $check_plugin_query = "SELECT * FROM `mysql`.`user` WHERE "
-                    . "`User` = '" . $host['User'] . "' AND `Host` = '"
-                    . $host['Host'] . "'";
-                $res = $GLOBALS['dbi']->fetchSingleRow($check_plugin_query);
-                if (isset($res['plugin'])
-                    && $res['plugin'] == 'sha256_password'
-                    && isset($res['authentication_string'])
-                ) {
-                    $password_column = 'authentication_string';
-                    if (! empty($res['authentication_string'])) {
-                        $host[$password_column] = 'Y';
-                    } else {
-                        $host[$password_column] = 'N';
-                    }
-                }
+                $host[$password_column] = 'Y';
+            } else {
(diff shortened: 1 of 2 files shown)


lem9:
Deven,
with this PR under MySQL 5.7.8, I create an account deven on any host with password bingo and mysql native password. Afterwards I cannot log in with this account. The same works fine under QA_4_5.

lem9:
When I look at the authentication_string it's quite different than the one generated under QA_4_5.

devenbansod:
@lem9 : Thanks for the test. I could reproduce it. It was a problem with wrong coding of the syntax (`BY` instead of `AS`). It should be fixed now. Please let me know if it still gives a problem.

Also, from the link that you provided in the other PR, it is clear that we can use the syntax `IDENTIFIED WITH auth_plugin AS hash_string` for all 5.5.7+, while using a separate `SET PASSWORD` for lower versions. These changes have been made in the squashed commit.

lem9:
Looks fine but in my phpunit I get 1 error and 2 failures.

devenbansod:
Sorry. Had missed changing the tests accordingly.
Should be fixed now. Let me know if they don't.

lem9:
Nice work.

devenbansod:
I tested on demo server QA_4_5, but if I create a new user in MySQL(which is 5.5.44), I am not able to login using that user. But if I use the appropriate syntax applicable from the ones stated below, for 5.5.44, it works properly.

**But I think I have now figured out the correct solutions :** 

**For MySQL**, 
Also, some references here - 
1. https://dev.mysql.com/doc/refman/5.7/en/create-user.html
2. https://dev.mysql.com/doc/refman/5.6/en/create-user.html
3. https://dev.mysql.com/doc/refman/5.5/en/create-user.html

MySQL - 5.5.0 -> 5.5.6 - AND SET the hashing method properly -- NOT TESTED but should work.
```
CREATE USER 'run1'@'%'; SET PASSWORD FOR 'run1'@'%' = PASSWORD('test');
```

MySQL - 5.5.7 -> 5.7.5 - AND SET the hashing method properly -- TESTED
```
CREATE USER 'run1'@'%' IDENTIFIED WITH mysql_native_password; SET PASSWORD FOR 'run1'@'%' = PASSWORD('test');
```
*Note: With this PR merged, we use the syntax *`IDENTIFIED WITH auth_plugin AS hash_string` for 5.5.7+. But with this syntax in 5.5.7 -> 5.7.5, it leaves one of the `authentication_string` or `password` field in `mysql`.`user` table empty, and then the plugin tends to use that blank only and not the set password. So this change is required*

MySQL - 5.7.6+ -- TESTED
```
CREATE UESR 'run1'@'%' IDENTIFIED WITH mysql_native_password BY 'test';
```

**For MariaDB,**
MariaDB - pre-5.2.0 -- NOT TESTED but should work
```
CREATE USER 'run1'@'%'; SET PASSWORD FOR 'run1'@'%' = PASSWORD('test');
```

MariaDB - 5.2.0+ -- TESTED
```
SELECT PASSWORD('test'); // Get hashed string
CREATE UESR 'run1'@'%' IDENTIFIED VIA mysql_native_password USING 'hashed_string_here';
```

Issue #11558, opened 2015-10-10
Can't execute SQL with subquery


sonanchenko:
Hello Developers,

PMA: 4.5.0.2
OS: Centos 6.5
DB: MariaDB-server-10.0.21-1.el6.x86_64

I am trying to execute SQL:
```sql
SELECT pa.id, pa.product_id, pa.supplier_code, pa.sid, p.* FROM `products_all` pa LEFT JOIN products p ON p.sid=pa.sid and p.supplier_code=pa.supplier_code WHERE pa.`product_id` IS NULL AND pa.`sid` IS NOT NULL AND pa.`outdated` = 0 and EXISTS (SELECT id from products p where p.sid=pa.sid and p.supplier_code=pa.supplier_code) ORDER BY `product_id` ASC
```
but face the following error:

====

Static analysis:

2 errors were found during analysis.

    Unrecognized keyword. (near "EXISTS" at position 237)
    Unexpected token. (near "(" at position 244)

SQL query: Documentation

SELECT pa.id, pa.product_id, pa.supplier_code, pa.sid, p.* FROM `products_all` pa LEFT JOIN products p ON p.sid=pa.sid and p.supplier_code=pa.supplier_code WHERE pa.`product_id` IS NULL AND pa.`sid` IS NOT NULL AND pa.`outdated` = 0 and EXISTS ( LIMIT 0, 25

MySQL said: Documentation
#1064 - You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'LIMIT 0, 25' at line 1

===

it seems that query is cut.

Your hep would be very appreciated.
Do not hesitate to ask if you need additional information.

sonanchenko:
The same with next query:

select * from extras where tariff_id not in (select id from tariff)
==========
Error

Static analysis:

2 errors were found during analysis.

    Unrecognized keyword. (near "not in" at position 37)
    Unexpected token. (near "(" at position 44)

SQL query: Documentation

select * from extras where tariff_id not in ( LIMIT 0, 25

MySQL said: Documentation
#1064 - You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'LIMIT 0, 25' at line 1

=========

Please assist.

udan11:
I am going to submit a pull request for the first issue, but I couldn't reproduce the second one. 

I used the following query which seems to be the same (but with different tables to reproduce the issue) and it worked as expected.

[...]

PR #11564, opened 2015-10-10
Partial fix for #11558.

udan11:
Partial fix for #11558.
Updated sql-parser library to udan11/sql-parser@42c34ddd1fee261da43331e855c80bc432d66247 (v3.0.3).

The partial (just for the first issue) fix for #15588 is in those two files:
- libraries/sql-parser/src/Statement.php
- libraries/sql-parser/src/Components/Condition.php

I included the other files just to keep in-sync with the latest version of the library.


lem9:
Thanks.

PR #11566, opened 2015-10-11
Fix error in PR #11560

devenbansod:
Fixes an error in PR #11560 and implements the syntax for creating users as described [here](https://github.com/phpmyadmin/phpmyadmin/pull/11560#issuecomment-147158219)


lem9:
Looks fine, thanks for the research.

PR #11570, opened 2015-10-14
Fixed error in PR : 11569

devenbansod:
I need a slight advice/help here. The problem is mainly for MySQL versions 5.5.7 -> 5.7.5.

MySQL introduced ALTER USER syntax in 5.6.6, but only from 5.7.6, it can be used to change the authentication_plugin used to authenticate an user. But, as a GUI tool, we should be able to provide it. 

So in Edit Privileges > Change Password, we provide the logged-in user with different options to change the user's(i.e. user being edited) plugin to along with changing the user being edited's password. This is done by Updating the `plugin` field in `mysql`.`user`, flushing the privileges, setting the old_passwords value (0 for mysql_native and 2 for sha256) and then running the SET PASSWORD query.
This works as expected.

And, for the main page's 'Change Password', we do the same thing, but here the currently logged-in user and user being edited is same.

Now, the only problem here is that for changing passwords anywhere, we would require update_priv for `mysql.user`, the `FLUSH` priv as well. So the users which don't have this privileges would not be able to change their passwords at all (atleast not using PMA).

What I suggest we do ?
We keep the above explained method in Edit Privileges > Change Password, as we expect that if the user who is changing credentials for some other user must be privileged enough and his/her actions would go through. So, no problem.
For main page's Change Password, I think we remove the plugin changing options altogether and just provide for changing password. 

Currently this PR implements the privileged user version for both Change Password in Edit Privileges as well as on the main page. I will make the changes as we decide.

### DIFF
--- a/libraries/display_change_password.lib.php
+++ b/libraries/display_change_password.lib.php
@@ -12,12 +12,14 @@ if (! defined('PHPMYADMIN')) {
 /**
   * Get HTML for the Change password dialog
   *
+  * @param string $mode     where is the function being called?
+  *                         values : 'change_pw' or 'edit_other'
   * @param string $username username
   * @param string $hostname hostname
   *
   * @return string html snippet
   */
-function PMA_getHtmlForChangePassword($username, $hostname)
+function PMA_getHtmlForChangePassword($mode, $username, $hostname)
 {
     /**
      * autocomplete feature of IE kills the "onchange" event handler and it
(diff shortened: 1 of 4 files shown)


lem9:
I agree with "For main page's Change Password, I think we remove the plugin changing options altogether and just provide for changing password."  but the plugin changing option could be left there for privileged users (when $is_superuser is true):
````
$is_superuser = $GLOBALS['dbi']->isSuperuser();
````

devenbansod:
Yes. That would make sense.
I will make the changes and let you know.

.

devenbansod:
Tests are passing now.
Please let me know if you could spot some error in the expected behavior for some version.
Otherwise I believe this is good to go.

lem9:
Is it expected that in MySQL 5.5.44, a privileged user who clicks on Change password does not see the hashing dialog?

devenbansod:
@lem9 : I think yes. Let me explain.

Because, before 5.5.7, there were no authentication plugins, just 2 password hashing methods namely `mysql_native_password` and `mysql_old_password`. Of these, PHP with mysqlnd does not support mysql_old_password, and as we discussed [here] (https://github.com/devenbansod/phpmyadmin/commit/22c0cf752566725f1b38bf875e9d9b99126efb3c#commitcomment-13720649) , we decided not to provide option regarding it, to the users.
Thus, the only choice remains is `mysql_native_password`. 

So I had removed showing it altogether. 
Let me know if you believe we should show the drop-down with  `mysql_native_password` as the only option. I will then make the appropriate changes.

lem9:
Explanation accepted.

lem9:
How about the unused variable?

devenbansod:
Sorry again! 
But, in fact, I am contradicting myself here it seems. 

Because, according to this [lines] (https://github.com/phpmyadmin/phpmyadmin/pull/11570/files#diff-4683a14fdc99b0a7a11842d79302f994R135), I think it should be visible to all the users in older versions.
Was it not visible to you on an older MySQL version ? Then that is unexpected.

Also, I will fix the unused variable soon.

lem9:
I confirm that in MySQL 5.5.44, with a privileged user (root@localhost) or an unprivileged one, on the home page I click on "Change password" and "Password hashing" is not shown.

devenbansod:
Just one more question, is the above mentioned output same on the 'Edit Privileges' > 'Change Password'  page as well?

lem9:
No, in User accounts > Edit privileges > Change password, the Password hashing part is there (with the sole "Native MySQL authentication" choice).

devenbansod:
Sorry. It seems to be getting too confusing for me at this moment!

I misunderstood your MySQL version to be lower than 5.5.7 (which it is not in fact) and thus provided link to an unrelated code!

So yes. The above behavior is as expected, since the second condition [here](https://github.com/phpmyadmin/phpmyadmin/pull/11570/files#diff-4683a14fdc99b0a7a11842d79302f994R100) will be true only for Edit Privileges > Change Password and not for Main page's Change Password.

So, **for versions 5.5.7 -> 5.7.5, no user can change its *own* authentication plugin** which is as per our agreement in this [comment] (https://github.com/phpmyadmin/phpmyadmin/pull/11570#issuecomment-148056217).

I will remove the unused variable and let you know.

lem9:
Looks fine, thanks.

devenbansod:
Thanks for all the help and advice throughout the PR! :)

PR #11578, opened 2015-10-15
Fix #11565 No UI to relog when using auth_type = config

### LINKED ISSUE #11565: No UI to relog when using auth_type = config
When using auth_type 'config' the login credentials are pulled from the config file, and therefore the login page and the logout button is not shown. This conflicts with certain functionality that requires you to log out and back in. For example, to reload the config file after updating to a newer version and adding the new phpmyadmin tables.

As it stands now, to make this happen I have to get rid of the old php session data, either by deleting the session file serverside, or by deleting the browser cookie. Or use a hack like ```index.php?old_usr=```. Restarting the webserver does not help since the config file seems to be embedded in the session data.

madhuracj:


lem9:
To test I introduced a syntax error in config.inc.php and clicked on the new button. No error is reported. Maybe this is because AuthenticationConfig.php does not check old_usr?

PR #11591, opened 2015-10-20
Documentation about which features can be removed to reduce the insta…

ibennetch:
…ll footprint


lem9:
Very nice!

Issue #11605, opened 2015-10-26
Changing hostname kills password


holyfive:
Hello everyone,

we use phpMyAdmin 4.5.1 with MariaDB 10.0. After upgrading from 4.5.0 we notice that changing a user's credentials, especially the hostname or ip range, "destroys" the password. It has to be set manually again to re-enable login. Can you please have a look at this? Thanks in advance!

CU, Patrick

lem9:
This patch has been merged for the upcoming 4.5.2 version, please confirm:
https://github.com/phpmyadmin/phpmyadmin/commit/1e6f128202922a94a9153f021ea3c9efea462a8b.patch

holyfive:
Unfortunately it does not help. Still the problem occurs: 

* Create a new user with password
* Change the user's ip range

=> User has no password, it has to be set again

lem9:
Please clarify "Change the user's ip range". I assume you go to User accounts > Edit privileges > Login information, but give more details about what you do next.

holyfive:
Indeed. E.g. I create a user "abc"@"134.95.%" with password "xyz". Then I want to change the ip range and go to User accounts > Edit privileges > Login information. I edit "134.95.%" to "134.95.125.%". After saving that, the user doesn't have a password anymore.
Everything was fine in 4.5.0, but since 4.5.1 we have this issue.

lem9:
At the bottom of the "Login information" dialog, by default it's "keep the old one". Do you use this default?

holyfive:
No, I use "delete the old one from the user tables."

lem9:
Can you try this patch?
https://github.com/devenbansod/phpmyadmin/commit/d604a3613523d09c30ddcef6c700c67b0a317899.patch

holyfive:
Yes, it works. Thanks a lot.

lem9:
The patch will be included in 4.5.2.

holyfive:
I'm sorry, I have to re-open this issue. The patch leads to the following behavior, which I don't understand at all: Changing the login information of a user and deleting the old one from the user tables leads to an account with a password, but with an *empty* hash. BUT the user can still login to the server and has to use the correct password. How is that possible? And how can I get back the password hashes in "mysql.user" table?

holyfive:
I've created a short screencast so you can see what is actually happening:

http://www.uni-koeln.de/dbadmin.ogv

[...]

PR #11613, opened 2015-10-27
Fix issue #11605

devenbansod:
We were hashing the already hashed value of password. I have changed it here.

Please comment if still you find some anomaly in the expected behavior, in any of versions of MySQL/ MariaDB.


lem9:
Looks fine, thanks.

PR #11629, opened 2015-10-31
fix #11627, CREATE TABLE/INSERT INTO executed twice (ctrl+enter)

### LINKED ISSUE #11627: [4.5.1] CREATE TABLE/INSERT INTO executed twice (ctrl+enter)
When I execute an INSERT INTO/CREATE TABLE query it's executed twice.

Example:
The table was EMPTY (I tested it twice).
Screen: http://imgur.com/pQbv0pN

I executed that query with an empty table and it returned that the AUTO INCREMENT id is 2 (so the query was executed twice).

Table after the query was executed: http://imgur.com/eTt5kac

It happened to me with some tables so it may be possible to be a general bug.

Server info: http://imgur.com/5bEWQwh

nisargjhaveri:


lem9:
Thanks.

PR #11640, opened 2015-11-04
Add support for PHP builtin http server

ghost:
Run php builtin http server, just type

cd phpmyadmin; php -S 0.0.0.0:8587

Then open browser http://xxx.xxx.real.ip:8587

Make sure mysql/mysqli were installed and config.inc.php
configured right

Signed-off-by: ZhiFeng Hu hufeng1987@gmail.com

### DIFF
--- a/index.php
+++ b/index.php
@@ -1,4 +1,12 @@
 <?php
+if (php_sapi_name() === 'cli-server') {
+        $path = realpath(__DIR__ . parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH));
+            if (__FILE__ !== $path && is_file($path)) {
+                        return false;
+                            }
+            unset($path);
+}
+
 /* vim: set expandtab sw=4 ts=4 sts=4: */
 /**
  * Main loader script


lem9:
Thanks. Please explain the need for this patch, as I just tried  the build-in server with PHP 7.0.0RC6 (without the patch) and could not see any problem.

netroby:
if static file exists, just stop php praser.php buitin server will request static file direct


通过 Android 版 Outlook 发送





On Thu, Nov 5, 2015 at 5:04 AM -0800, "Marc Delisle" <notifications@github.com> wrote:












Thanks. Please explain the need for this patch, as I just tried  the build-in server with PHP 7.0.0RC6 (without the patch) and could not see any problem.

lem9:
Sorry, I don't understand your reply.

netroby:
http://php.net/manual/en/features.commandline.webserver.php

Here is the PHP document said:
```
If a URI request does not specify a file, then either index.php or index.html in the given directory are returned. If neither file exists, the lookup for index.php and index.html will be continued in the parent directory and so on until one is found or the document root has been reached. If an index.php or index.html is found, it is returned and $_SERVER['PATH_INFO'] is set to the trailing part of the URI. Otherwise a 404 response code is returned.
```
The code i contribute, check file exists. if file exists . Don't continue execute php scripts, just tell server read the real file

ibennetch:
Does it work to use your browser to go to `http://192.0.2.1/phpmyadmin/index.php` directly?

It seems to me this is unnecessary as its really not phpMyAdmin's job to select the desired path (and besides this server seems to be for development only)

netroby:
Yes. it been tested. and all other php project may using the same way to resolve the problem.
This file just like routes configure for PHP built-in server

For example:

https://github.com/zendframework/ZendSkeletonApplication/blob/master/public/index.php

lem9:
Is there a problem running phpMyAdmin under the build-in web server? Again, I tried it and got no problem, using the example you gave.

PR #11644, opened 2015-11-05
Fix import tab broken by ImportLdi.

drainx:
The return before the class would cause the class to not be found, breaking https://github.com/phpmyadmin/phpmyadmin/blob/QA_4_5/libraries/plugin_interface.lib.php#L79
which in turn caused a 500 error. Other importers might need a fix similar.
Can be seen on the demo server, click the import tab.

Signed-off-by: Daniel Rix Drainx1@live.com

### DIFF
--- a/libraries/plugins/import/ImportLdi.class.php
+++ b/libraries/plugins/import/ImportLdi.class.php
@@ -13,12 +13,6 @@ if (! defined('PHPMYADMIN')) {
 /* Get the import interface */
 require_once 'libraries/plugins/import/AbstractImportCsv.class.php';
 
-// We need relations enabled and we work only on database
-if ($GLOBALS['plugin_param'] !== 'table') {
-    $GLOBALS['skip_import'] = true;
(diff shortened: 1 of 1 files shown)


madhuracj:
Not sure how that can happen when `skip_import` is set to true at https://github.com/phpmyadmin/phpmyadmin/blob/QA_4_5/libraries/plugins/import/ImportLdi.class.php#L18 and it being checked at https://github.com/phpmyadmin/phpmyadmin/blob/QA_4_5/libraries/plugin_interface.lib.php#L77

drainx:
There seems to be a discrepancy on master and QA.
https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/plugin_interface.lib.php#L85

The include_once is not present on Master

drainx:
Found it: blame 0680c243d196a8695e68546d07195558eb9a6388

drainx:
New PR created #11645

Issue #11504, opened 2015-09-27
Reactivate cut&paste possibility in print view


ck-on:
The previous print view in phpmyadmin was useful to copy and paste tables into other documents, email, etc.

The new print view in 4.5 makes that impossible because it works completely differently and does not show on the screen and javascript is used to trigger automatic print dialog in browsers.

You need to put an option back into config to disable the automatic print dialog javascript and to just display it on screen (so it can be shown in another tab, etc.)

Achilles-96:
Hi,
I would like to work on this.
But one thing. If you want users to copy and paste tables. Isn't it better if we show it in full text format.

nisargjhaveri:
Hi @Achilles-96,

About full text, please have a look at #11634.

lem9:
The team had an internal discussion about this. 

I think that the new "Print view" is fine and that the real need is
copying to clipboard; we suggest a new feature for master in Browse > Query
results operations > Copy to clipboard.

Achilles-96:
Sure, I'll try this one.

ck-on:
@lem9 

What if you just want to copy one column of a table to the clipboard like you could do easily with the old print view and selective copy in most browsers?

Just disable the automatic print, that behavior is completely obnoxious.

Do whatever stylesheet manipulations you think are better than the old print view (they aren't better) and then give the users a big PRINT button.

That would be proper.

lem9:
@ck-on I don't see your point about copying one column; do you mean copying one column of one row?

Achilles-96:
@lem9 Should we copy only the contents of table to clipboard or the entire content which is going to be printed.
One more thing: In the earlier method, users will easily be able to copy a part of the table or content what they need out of what is going to be printed. Do you think we should reconsider?

lem9:
@Achilles-96 I think we should have the same as what appears on screen when we click on Print view (version 4.5). I don't think we should reconsider, as it's possible to select and cut (ctrl/c) text from the Browse output.

Achilles-96:
@lem9 What icon do you think must be placed for Copy to clipboard link

lem9:
@Achilles-96 Maybe the same as the one used for the Copy action on each line when you Browse.

ck-on:
Thrilled at the idea I will now need a third program in between copying columns of data from phpmyadmin to email.

Why is there so much resistance to simply disabling the auto print and making it manual?

Why not include the old print view as a generic plain view of the page?

[...]

PR #11656, opened 2015-11-11
Fix for issue #11504

Achilles-96:
This fixes issue #11504.

The following is my approach:
Currently we are writing rules for print using CSS media query.
For previewing the print, we can follow a similar approach, by defining a class 'previewprint' which has similar but not entirely same rules as media query and switch the class of body when 'Print view' button is pressed.
We trigger print when 'Print' button is pressed.
We disassociate body from this class when 'Back' button is pressed.

### DIFF
--- a/js/functions.js
+++ b/js/functions.js
@@ -4338,6 +4338,11 @@ function PMA_getCellValue(td) {
     }
 }
 
+$(window).on('popstate', function (event, data) {
+    $('#printcss').attr('media','print');
+    return true;
+});
+
 /**
  * Unbind all event handlers before tearing down a page
  */
@@ -4383,11 +4388,48 @@ AJAX.registerOnload('functions.js', function () {
 });
 
 /**
- * Print button
+ * Produce print preview
  */
-function printPage()
+function printPreview()
 {
-    // Do print the page
+    $('#printcss').attr('media','all');
+    createPrintAndBackButtons();
+}
+
+/**
+ * Create print and back buttons in preview page
+ */
+function createPrintAndBackButtons() {
+
+    var back_button = $("<input/>",{
+        type: 'button',
+        value: PMA_messages.back,
+        id: 'back_button_print_view'
+    });
+    back_button.click(removePrintAndBackButton);
+    back_button.appendTo('#page_content');
+    var print_button = $("<input/>",{
+        type: 'button',
+        value: PMA_messages.print,
+        id: 'print_button_print_view'
+    });
+    print_button.click(printPage);
+    print_button.appendTo('#page_content');
+}
+
+/**
+ * Remove print and back buttons and revert to normal view
+ */
+function removePrintAndBackButton(){
+    $('#printcss').attr('media','print');
+    $('#back_button_print_view').remove();
+    $('#print_button_print_view').remove();
+}
+
+/**
+ * Print page
+ */
+function printPage(){
     if (typeof(window.print) != 'undefined') {
         window.print();
     }
--- a/themes/pmahomme/css/printview.css
+++ b/themes/pmahomme/css/printview.css
@@ -1,168 +1,171 @@
 @media print {
-
-    /* For removing element from Print View */
-    .print_ignore {
+    #back_button_print_view, #print_button_print_view {
         display: none;
     }
+}
 
-    .nowrap {
-        white-space: nowrap;
-    }
+/* For removing element from Print View */
+.print_ignore {
+    display: none;
+}
 
-    .hide {
-        display: none;
-    }
+.nowrap {
+    white-space: nowrap;
+}
 
-    /* Standard CSS */
-    body, table, th, td {
-        color:             #000;
-        background-color:  #fff;
-        font-size: 12px;
-    }
+.hide {
+    display: none;
+}
 
-    /* To remove link text decoration */
-    a:link {
-        color:#000;
-        text-decoration:none
-    }
+/* Standard CSS */
+body, table, th, td {
+    color:             #000;
+    background-color:  #fff;
+    font-size: 12px;
+}
 
-    /* To remove any image borders */
-    img {
-        border: 0;
-    }
+/* To remove link text decoration */
+a:link {
+    color:#000;
+    text-decoration:none
+}
 
-    /* Table specific */
-    table, th, td {
-        border: .1em solid #000;
-        background-color: #fff;
-    }
+/* To remove any image borders */
+img {
+    border: 0;
+}
 
-    table {
-        border-collapse:   collapse;
-        border-spacing:    0.2em;
-    }
+/* Table specific */
+table, th, td {
+    border: .1em solid #000;
+    background-color: #fff;
+}
 
-    thead {
-        border-collapse:   collapse;
-        border-spacing:    0.2em;
-        border: .1em solid #000;
-        font-weight: 900;
-    }
+table {
+    border-collapse:   collapse;
+    border-spacing:    0.2em;
+}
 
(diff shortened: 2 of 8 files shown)


devenbansod:
Hi @Achilles-96 , thanks for the PR and an apology for my delayed response.

A few comments about the PR.
1. Would it better if the newly added buttons could have their text as 'Print' and 'Back', instead of all lower-case letters ?
2. Also, can this newly added buttons be hidden from the printed output? I don't think they are needed on the final output.
3. Also, I go to a db structure page, click on Print View and print the page and then click back. The top menu bar is expanded to the full, blocking the table headers. (which are generally folded into 'More'). See screenshot. Obviously, refreshing the page would bring it to as before, but can we do that folding automatic ?
topmenuexpanded


Some general notes:
It is always good to work on different branches for different issues, instead of working on master.
Also, though not necessary, its always better to write one-two sentences about what exactly is your approach to solving the mentioned issue in the PR description. 


I am not very sure if we should consider as feature request or a bug-fix. Because if it is bug-fix, we would have to target a different branch. @ibennetch Any help/comments ?

Achilles-96:
@devenbansod ,
About points 1 and 2, I'll fix them and update the PR.
But about point 3, I can't reproduce what you said on my machine. They remain folded when I press back on my machine. Please look into that once more.
About final point, I think this is a feature more than a bug as lacking this is not considered as  bug.
I'll proceed according to your reply.

Achilles-96:
@devenbansod , 
Any updates?

js/functions.js
```diff
@@ -4338,6 +4338,15 @@ function PMA_getCellValue(td) {
     }
 }
 
+$(window).on('popstate', function (event, data) {
+    var body_elements = document.getElementsByTagName('body');
+    console.log(body_elements[0].className.toString());
```
nisargjhaveri: I believe this is just for debugging. Can you please remove it?

js/functions.js
```diff
@@ -4383,16 +4392,51 @@ AJAX.registerOnload('functions.js', function () {
...
+    createPrintAndBackButtons();
+}
+
+/**
+ * Create print and back buttons in preview page
+ */
+function createPrintAndBackButtons() {
+    var back_button = document.createElement("input");
+    back_button.type = "button";
+    back_button.value = "Back";
```
nisargjhaveri: These buttons are not translatable, please make them so.

js/functions.js
```diff
@@ -4383,16 +4392,51 @@ AJAX.registerOnload('functions.js', function () {
...
+    var body_elements = document.getElementsByTagName('body');
+    body_elements[0].className = body_elements[0].className + "previewprint";
+    createPrintAndBackButtons();
+}
+
+/**
+ * Create print and back buttons in preview page
+ */
+function createPrintAndBackButtons() {
+    var back_button = document.createElement("input");
```
nisargjhaveri: We generally use jQuery for DOM interactions. It would be good idea to use jQuery here too.

themes/pmahomme/css/printview.css
```diff
@@ -165,4 +165,181 @@
...
     }
+
+    #back_button_print_view, #print_button_print_view {
+        display: none;
+    }
+
+}
+
+
+body.previewprint .print_ignore {
```
nisargjhaveri: Isn't this the copy of styles above? It might become difficult to maintain this in future. Isn't there any better way to do this?

Achilles-96:
@nisargjhaveri ,
Yes, we can do it without repeating rules. We can remove media query because by the time they press print button all the styles are properly set and we don't need a special media query for that.

nisargjhaveri:
But then, if someone simply presses ctrl+P, or uses right click -> print, it wouldn't work.

@devenbansod, Any ideas about this?

lem9:
After thinking about that and the initial issue #11504 it looks like it would be better to keep the print view as is, and offer a new copy to clipboard feature (probably for 4.6).

devenbansod:
Thanks @NisargJhaveri for the comments on the commit! :)

Hi @Achilles-96, we have reopened this since the team agrees that this would be indeed a good approach to address some issues in #11504. We would be treating this as a feature for 4.6.0 (so target would be master so no need to change).

Now, could you please redo the commit with first the changes suggested by Nisarg in his comments on the commit ? I will add mine soon if I find some. 
Also, it will be good to rebase your commit to current master to remove the merge conflicts.

Achilles-96:
Yeah, sure.

Achilles-96:
I'll update the PR, something went wrong.

Achilles-96:
Any update on this?

devenbansod:
Sorry for the delay. I was waiting for your comment that you have fixed the thing you said had gone wrong. 
Anyways, I will have a look at this towards the evening. 

Also, it would be good if you *rebase* your commit over current master instead of one more commit to fix the merge. See : https://git-scm.com/book/en/v2/Git-Branching-Rebasing.

themes/pmahomme/css/printview.css
devenbansod: Currently, I am not able to think of a concrete solution of not duplicating this. 
So, in my opinion, we can just add a comment explaining the situation, just above the `@media print` query pointing the developer to these as well.

Also, @Achilles-96, these new styles should be added for the original theme as well.

js/functions.js
```diff
@@ -4338,6 +4338,14 @@ function PMA_getCellValue(td) {
...
 }
 
+$(window).on('popstate', function (event, data) {
+    var body_elements = document.getElementsByTagName('body');
+    if(body_elements[0].className.toString() == 'previewprint'){
+        body_elements[0].className = '';
+    }
+    return true;
+});
+
```
devenbansod: Can we use the JQuery selectors here, too ?

themes/pmahomme/css/printview.css
nisargjhaveri: I was looking at it.. Looks like we can keep a `link` tag with `media="print"` without `@media` in CSS file, and then change `media` attribute with JS when we want those styles to apply on screen. I've not tried it properly, but I think it is worth giving a try.

Achilles-96: @nisargjhaveri , I looked at media attribute. Does this approach look good:
We add `<link rel="stylesheet" type="text/css" href="printview.css">` so that ctrl+p is taken care of and we also load printview.css when print button is clicked.

devenbansod: Thanks @nisargjhaveri for the workaround.
 
@Achilles-96 this seems partially correct to me. 
I don't  know *exactly* what Nisarg was thinking. But I think, by default, we add a printview.css (which does not have the `@media` query) with `media="print"` (So Ctrl+P and right click is taken care of). When user clicks 'Print' link on table's output, we change it to dynamically to `media="all"`, along with adding those buttons and adding the `previewprint` class to the body. 
And when the user presses the newly added `back` button, we revert it back to `media="print"`. 

I have not tried it but I think it should work.
@nisargjhaveri , any suggestions/changes ?

nisargjhaveri: Almost. But I don't think we even need to add `previewprint` class to body. The `printview.css` will contain the old css rules, just without @media query. When applied on screen, it shows the print view.

devenbansod: Yes. That seems correct. 

@Achilles-96, could you please make the suggested changes ?

devenbansod:
Looks good to me!
Thanks @Achilles-96! Nice work :)

Thanks for the help, @nisargjhaveri! :)

PR #11671, opened 2015-11-15
[WIP] Start respecting PSR1 standard

Tithugues:
Some weeks ago, many pMA developers agreed that respect PSR-1 and maybe PSR-2 might be a good thing (see https://lists.phpmyadmin.net/pipermail/developers/2015-August/019049.html).
So I started to update the pMA standard and update the code to respect PSR1, which is the first step if we want to respect PSR2.
Seeing the first modifications and the impact, should we continue in this direction?

This might be a huge work. I put "[WIP]" in the subject, but maybe we should merge it in a continuous mode and not one shot at the end. What do you think about it?

Thanks for your feedback.


Tithugues:
And here are the standards for PSR-1: http://www.php-fig.org/psr/psr-1/

lem9:
Yes it should be merged continuously.

marginchan:
good works!!!!!    must respect PSR-1 or up

lem9:
Nice cleanup!

Tithugues:
Thanks Marc.
Now, we see that there are a lot of new errors… We should now work too remove logic code in declaration files.

Issue #11684, opened 2015-11-22
mysql_native_password with MariaDB bug


AbbyNormal:
So I checked this thread https://github.com/phpmyadmin/phpmyadmin/issues/11520

but there is a bug in phpmyadmin.

If you create new user through phpmyadmin this is the syntax that is used:

CREATE USER 'test'@'localhost' IDENTIFIED VIA mysql_native_password USING '***';

And it works. But when you login with that username and change password it says changed but it really is not. If you logout and login with new password you will fail. You cannot login with new password but with old one you can. This is shown in the log:

21:37:08 [Warning] 'user' entry 'test@localhost' has both a password and an authentication plugin specified. The password will be ignored.

If you delete in mysql.user of the user test field plugin, then new password is working. Any idea?

lem9:
@AbbyNormal Which phpMyAdmin and MariaDB versions? P.S. tomorrow, version 4.5.2 will be released.

AbbyNormal:
Hello, phpmyadmin 4.5.1 and mariadb 5.6.x latest one in centos 7.

devenbansod:
Hi @AbbyNormal , are you sure it is 5.6.x ? I cannot seem to find it in https://downloads.mariadb.org/mariadb/+releases/ .
Or did you mean the latest in 5.5.x ?

By the way, bug confirmed in MariaDB 5.5.46. 
The problem is the different behaviors of different versions of MySQL and MariaDB.

MySQL 5.6.16
===========
Create a user with phpMyAdmin, it sets the `Password` field and the `authentication_plugin` field.
Changing the password changes the `Password` field. Works fine.

MariaDB 10.0.21
=============
Create a user with phpMyAdmin, it sets the `authentication_string` field and the `authentication_plugin` field.
Changing the password changes the `Password` field. But I dont know why this warning is not shown in the log and I am able to login properly.

MariaDB 5.5.46
=============
Create a user with phpMyAdmin, it sets the `authentication_string` field and the `authentication_plugin` field.
Changing the password changes the `Password` field. But warning is not shown in the log and I am not able to login properly.

AbbyNormal:
Sorry, I was writting away from computer. The version is mariadb 5.5.44-1. If you change the password you can't login with new password but old one works. If you delete the content of plugin in mysq.user for that user and logout you can login back with new password.

AbbyNormal:
[...]

PR #11686, opened 2015-11-24
Fix issue 11684

devenbansod:
Tries to fix issue #11684 

This involves many different conditions for different server versions. I am continuing my own tests but it would be great if as many as possible people can test it too. 


lem9:
Tested the #11684 scenario with MySQL 5.5.44 and 5.7.8: all is fine.

devenbansod:
Great! Thanks! :)

I will just test it with MySQL 5.6.16 and MariaDB 10.0.21 installations, before merging later today.

PR #11697, opened 2015-11-27
Fix for not being able to create users on Percona Server

bjdelange:
Inspired by the patch https://github.com/phpmyadmin/phpmyadmin/commit/9bf88cbd32f5f5b4a18bd7ec60e89c5b75bcf947.patch

### DIFF
--- a/libraries/server_privileges.lib.php
+++ b/libraries/server_privileges.lib.php
@@ -5156,7 +5156,7 @@ function PMA_getSqlQueriesForDisplayAndAddUser($username, $hostname, $password)
 
     // 'IDENTIFIED WITH auth_plugin'
     // is supported by MySQL 5.5.7+
-    if ($serverType == 'MySQL'
+    if (($serverType == 'MySQL' || $serverType == 'Percona Server')
         && PMA_MYSQL_INT_VERSION >= 50507
         && isset($_REQUEST['authentication_plugin'])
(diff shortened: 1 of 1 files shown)


ibennetch:
Hi and thanks for your contribution. 

As the bot mentioned, if you agree to our DCO the best way to indicate that is to sign your commit and push it again (you can amend the commit such as http://stackoverflow.com/a/15667644/2385479 then force push it back to your repo).

Do you have access to a Percona server to confirm whether this works?

Since we don't officially support Percona, I'm not sure how much conditional testing we should do to make it work. I think we should include this, but we'll see what the other developers think.

Thanks.

lem9:
I was wondering: if Percona behaves exactly like MySQL (not sure of this), maybe getServerType() should return 'MySQL' even for Percona?

bjdelange:
@ibennetch This works on my Percona Server 5.6.24-72.2-log, I can create new users now.

bjdelange:
Signoff added !

lem9:
Tracked on https://github.com/phpmyadmin/phpmyadmin/issues/11698

lem9:
Thanks. I will merge this to our QA_4_5 branch for the upcoming 4.5.3 version.

ibennetch:
Thanks @bjdelange

PR #11704, opened 2015-11-29
Show table header and create bookmark on empty results

### LINKED ISSUE #11668: show table header along with the empty resultset message when on browse tab of empty tables
currently, the browse tab (which is also the default tab) does not display any information when selected on empty tables, other than showing up a message saying - "MySQL returned an empty result set (i.e. zero rows). (Query took 0.0003 sec)".

at-least we can show a table header row with all the columns listed in their respective order so that anyone selecting different tables from left panel need not click again on "Structure" tab to see the columns.

currently there is no way one can see the table header row of an empty table and on the other hand, we are wasting default tab's valuable screen with no information about the table being displayed.

nisargjhaveri:
Fix #11668.

### DIFF
--- a/libraries/DisplayResults.php
+++ b/libraries/DisplayResults.php
@@ -4327,7 +4327,7 @@ class DisplayResults
                 $message, $this->__get('sql_query'), 'success'
             );
 
-        } elseif (! isset($printview) || ($printview != '1')) {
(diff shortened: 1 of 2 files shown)


lem9:
Looks fine, thanks.

PR #11716, opened 2015-12-04
Fixes #11679

### LINKED ISSUE #11679: Cannot export results of some queries
Seen in 4.5.0 to 4.5.2. Test case:
````
CREATE TABLE `product` (
  `product_id` int(11) NOT NULL,
  `price` int(11) NOT NULL

INSERT INTO `product` (`product_id`, `price`) VALUES
(1, 10),
(2, 20);

CREATE TABLE `product_special` (
  `product_id` int(11) NOT NULL,
  `price` int(11) NOT NULL

INSERT INTO `product_special` (`product_id`, `price`) VALUES
(1, 10);

ALTER TABLE `product`
  ADD PRIMARY KEY (`product_id`);

ALTER TABLE `product_special`
  ADD PRIMARY KEY (`product_id`);
````

Run this query:
````
SELECT a.product_id, b.price
 FROM product a, product_special b
 WHERE a.product_id = b.product_id
 LIMIT 1
````

then Query results operations > Export. We don't reach the export dialog.

### PR DESCRIPTION
**Pull request by Achilles-96**

As referred in #11679, I tried ignore checking existence of table if `tbl_common.inc.php` is included by `tbl_export.php`

### DIFF
--- a/libraries/tbl_common.inc.php
+++ b/libraries/tbl_common.inc.php
@@ -48,6 +48,12 @@ $err_url = PMA\libraries\Util::getScriptNameForOption(
 
 /**
  * Ensures the database and the table exist (else move to the "parent" script)
+ * Skip test if we are exporting as we can't tell whether a table name is an alias (which would fail the test).
  */
-require_once './libraries/db_table_exists.lib.php';
+$URL = explode('/',$_SERVER['REQUEST_URI']);
+if ($URL[2])
+    $requested_function = explode('?',$URL[2])[0];
+
+if (!$URL[2] || $requested_function != 'tbl_export.php')
+    require_once './libraries/db_table_exists.lib.php';
 

### CODE REVIEW

**Comment by ibennetch** [comment]
Hi @Achilles-96 unfortunately we do ask for a real name in the signed-off-by comment rather than your Achilles-96 username. If you don't mind doing so, you can amend the commit and force push it back up to reflect the new change.

File: libraries/tbl_common.inc.php [hunk 1]
```diff
@@ -48,6 +48,12 @@
 
 /**
  * Ensures the database and the table exist (else move to the "parent" script)
+ * Also check if it is export function as it we get only aliases and aliases may not exist.
```
ibennetch: Would it be accurate to re-phrase this as:

> Skip test if we are exporting as we can't tell whether a table name is an alias (which would fail the test).

devenbansod:
I think since the issue #11679 is related to versions 4.5.0 to 4.5.2, and if this PR tries to fix it, it should target the QA_4_5 branch instead of master.

Achilles-96:
@devenbansod I changed the target branch.

devenbansod:
Great!
So, I am closing this now.

PR #11718, opened 2015-12-05
Fixes #11679

Achilles-96:
Signed-off-by: Raghuram Vadapalli raghuram.vadapalli@research.iiit.ac.in

PR #11738, opened 2015-12-10
Fixes #11679

### LINKED ISSUE #11679: Cannot export results of some queries

Achilles-96:
Modification for #11718

### DIFF
--- a/libraries/tbl_common.inc.php
+++ b/libraries/tbl_common.inc.php
@@ -48,6 +48,11 @@ $err_url = PMA_Util::getScriptNameForOption(
 
 /**
  * Ensures the database and the table exist (else move to the "parent" script)
+ * Skip test if we are exporting as we can't tell whether a table name is an alias (which would fail the test).
  */
-require_once './libraries/db_table_exists.lib.php';
+$URL = explode('/',$_SERVER['REQUEST_URI']);
+if ($URL[2])
+        $requested_function = explode('?',$URL[2])[0];
 
+if (!$URL[2] || $requested_function != 'tbl_export.php')
+        require_once './libraries/db_table_exists.lib.php';
--- a/tbl_export.php
+++ b/tbl_export.php
@@ -166,7 +166,10 @@ if (! empty($sql_query)) {
                         'value_str' => '.',
                     )
                 ),
-                array()
+                array(
+                    new SqlParser\Token($table),
+                    new SqlParser\Token('.',SqlParser\Token::TYPE_OPERATOR)
+                )
             );
         }
         $sql_query = SqlParser\TokensList::build($tokens);


lem9:
Thanks. Did you read my objection to your other patch, 5 days ago? This patch produces the same error message.

Achilles-96:
@lem9 
Yeah, I read the objection. I changed the patch. Now it doesn't produce same error for me. 
It exports properly. I assume this the command you are testing with:

SELECT a.product_id, b.price
 FROM product a, product_special b
 WHERE a.product_id = b.product_id
 LIMIT 1

lem9:
@Achilles-96 I tested this PR (#11738) and I get the same error when I want to export in CSV. I am testing with the scenario explained in #11679.

Achilles-96:
@lem9 
I followed the same scenario as in #11679. It is not rising any errors. Anyway, I'll confirm it once more.

lem9:
@Achilles-96 I mean this: With your fix I can reach the export dialog but I still cannot export the results. Let's say I choose CSV:
````
MySQL said: #1052 - Column 'product_id' in field list is ambiguous
````

Achilles-96:
@lem9 
I read that in my old PR. In my old PR, I changed only `libraries/tbl_common.inc.php` due to which we were able to go to export dialog. But the problem due to aliases is in `tbl_export.php`. I modified that file in this PR which resolved that MySQL error.

lem9:
@Achilles-96 I told you that in this PR (#11738) I still have the problem (MySQL error).

Achilles-96:
@lem9 I confirmed with one more person. This PR #11738 doesn't rise MySQL error for scenario in #11679 . Please recheck.

devenbansod:
Hi Marc, I tried to test this PR as Raghuram told me to and I am not able to reproduce the MySql error you noted.
The export goes through as expected for me!

lem9:
I rechecked and I can no longer reproduce the problem. It remains to be verified the full impact of the code removal in this PR.

lem9:
By the way, if you remove the code section that populates $aliases, there is another code section below which uses $aliases, that should be removed too. I'm still unsure of the goal of aliases removal in the original code.

Achilles-96:
Actually aliases are removed and are not replaced. The reason given in the comment is that alias will replaced. But ultimately they are not being replaced. So, I tried not removing them in first place and everything seemed to work fine. 
I didn't remove lower code section because I just wanted to make sure this didn't cause any malfunctioning and in that case I can update PR.

lem9:
"Everything seemed to work fine" until we find a test case that does not work fine. I'll ask the author of this code about it.

Achilles-96:
Actually there is no need to remove any code. I found bug in parser. I'll update PR.

lem9:
Standing by. Please reply on my message on the developers mailing list.

lem9:
Nice work, this looks fine.

PR #11748, opened 2015-12-15
Fix #11720

### LINKED ISSUE #11720: Message "An account already exists..." incorrectly displayed
In User accounts > Edit privileges (for one account) > Login Information, this message is displayed on entering the panel:
````
An account already exists with the same username but possibly a different hostname.
````
The goal is to warn about a conflict when creating an account, but here the user has just selected an account to possibly change login information.

devenbansod:
Check for duplication only, when username field is edited.

@lem9 : As we had discussed in #11720, do you think this behavior is fine for the warning display? 


lem9:
Nice work!

PR #11753, opened 2015-12-16
Refactor Server Plugins Page

madhuracj:


server_plugins.php
```diff
@@ -1,38 +1,32 @@
...
 
-$plugins = PMA_getServerPlugins();
+require_once 'libraries/common.inc.php';
 
-/**
- * Displays the page
- */
-$response->addHTML(PMA_getHtmlForSubPageHeader('plugins'));
-$response->addHTML(PMA_getPluginTab($plugins));
+$container = libraries\di\Container::getDefaultContainer();
```
nijel: I guess PMA\ prefix is missing here...

devenbansod: @nijel : I think that since Madhura has written on line 10 `namespace PMA`, the prefix wouldn't be required. What do you think ?

nijel: Indeed it's not required, though it would make it consistent with rest of the code.

PR #11768, opened 2015-12-21
Remove /*overload*/ prefix from mb_* calls

nijel:
I believe this was temporary flagging to see which code is migrated and
can be safely removed now.

Please correct me if I'm wrong...

### DIFF
--- a/db_operations.php
+++ b/db_operations.php
@@ -37,7 +37,7 @@ $sql_query = '';
 /**
  * Rename/move or copy database
  */
-if (/*overload*/mb_strlen($GLOBALS['db'])
+if (mb_strlen($GLOBALS['db'])
     && (! empty($_REQUEST['db_rename']) || ! empty($_REQUEST['db_copy']))
 ) {
     if (! empty($_REQUEST['db_rename'])) {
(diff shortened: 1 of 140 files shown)


Tithugues:
Hi,

From what I remember, it wasn't supposed to be temporary, but I agree that
may be removed.

H.

Le lun. 21 déc. 2015 à 12:17, Michal Čihař <notifications@github.com> a
écrit :

lem9:
The comment was supposed to remain there, so that someone looking at the call have a hint that these calls to mb_* functions may be overloaded. This makes me think: maybe we should remove mbstring as a requirement in the doc (because of the overloading done via libraries/string.lib.php).

nijel:
The problem with the comment is that it's not kept consistently - newly added code usually doesn't have it. Maybe because there is no documentation why it should be there and what it means and nobody enforces it. But current state when some mb_* calls have it and some not is really not good...

And for readability I prefer to avoid such things.

lem9:
Feel free to merge then.

nijel:
Thanks. I will keep this open for few more days (or rather till next week due to Christmas) to see if everybody is okay with this change.

nijel:
Okay merged as there was no offense here...

PR #11773, opened 2015-12-24
fix php notices

Gemorroj:
Notice in .\libraries\tbl_common.inc.php#54
Undefined offset: 2

Notice in .\libraries\tbl_common.inc.php#57
Undefined offset: 2

Signed-off-by: Gemorroj


devenbansod:
Hi @Gemorroj : I think this has already been fixed in a slightly better way in https://github.com/phpmyadmin/phpmyadmin/commit/3730d88215cc1c2db1d94cc33373f52647e2fe17

Gemorroj:
@devenbansod Need to backport this patch to branch 4.5

lem9:
We might even need to release 4.5.3.1, due to the high number of incidents, but do you know how to reproduce these notices?

PR #11782, opened 2015-12-27
Fixes #11538

Achilles-96:
If this looks fine, I'll add other options in 'Copy tables' form.

Signed-off-by: Raghuram Vadapalli raghuram.vadapalli@research.iiit.ac.in

Issue #11538, opened 2015-10-05
Copy multiple tables to database


mvl22:
As of PhpMyAdmin 4.5.0(.2), the option to copy multiple tables to another database seems to have disappeared:

![copy table gone](https://cloud.githubusercontent.com/assets/361423/10279329/be92dd2c-6b5c-11e5-8eb7-65b86721fdcb.png)

lem9:
In which version did you see this option?

mvl22:
Mea culpa - I could have sworn this was there, but it's not!

Perhaps treat this as a feature request? :)

madhuracj:
Had a look at this with the intention of using `Table::moveCopy` function to copy selected tables to another database. However, this seems to need quite some changes to the `mult_submits.lib.php` as all the multi submit functionality provided there are oriented towards changes done by executing SQL queries rather than calling functions.

Achilles-96:
So, is the implementation supposed be like this - 

* One more option under 'Table maintenance' called 'Copy table'.
* When that option is clicked target database is asked.
* On submit tables are copied.

devenbansod:
Seems correct, @Achilles-96. Just a few things.

One thing I would suggest is not to include it under 'Table Maintenance', but rather as a separate option because 'Table maintenance' on tbl_operations.php also has the same options like : Analyze table, Check table, Checksum table, Optimize table etc. while copy is a different option.

Another is, after selecting 'copy' from the drop-down, should we provide some subset of the options (which might be generic to all selected tables) from the 'Copy table' form on tbl_operations.php and then proceeding to actual copying may be ?

Achilles-96:
@devenbansod Yes, we can show those options when we ask for target database.

PR #11794, opened 2015-12-29
Improvements to Issue-#11538

Achilles-96:
As said in #11782, current db name is not shown and other options are added as in table Operations > Copy table to.
I didn't add modal because other options under prefix like 'Add prefix to table' are also shown as forms. To maintain consistency in page, they also must be shown as modals which I thought required many changes and can be a separate enhancement.


lem9:
Looking good, except for the messages that need to be translatable.

Achilles-96:
Sure, I thought words like 'database' is not translatable.

lem9:
They are. If you are using command-line git, try this:
````
git grep "__('Database')"
````

PR #11797, opened 2015-12-29
CloudFlare Flexible SSL support

laukstein:
Resource https://www.cloudflare.com/ssl/
Signed-off-by: Binyamin Laukstein 7raivis@inbox.lv

### DIFF
--- a/libraries/Config.php
+++ b/libraries/Config.php
@@ -1338,8 +1338,11 @@ class Config
                     return;
                 }
 
-                // If we didn't set port yet...
-                if (empty($url['port']) && PMA_getenv('SERVER_PORT')) {
+                if (in_array(PMA_getenv('HTTP_X_FORWARDED_PROTO'), array('http', 'https'), true)) {
+                    // CloudFlare's Flexible SSL port
+                    $url['port'] = PMA_getenv('HTTP_X_FORWARDED_PROTO') === 'http' ? 80 : 443;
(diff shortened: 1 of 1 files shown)


laukstein:
Added Signed-off-by.

lem9:
It would be slightly more efficient, in the same method, to keep in an intermediary variable the result of PMA_getenv('something') and reuse that. Plus I believe that the resulting code would be cleaner to read.

laukstein:
@lem9, check if it is what you wanted.

lem9:
Thanks; I squashed your commits and merged: https://github.com/phpmyadmin/phpmyadmin/commit/72cbfd920edb714269d440946638edb99282e1e9

laukstein:
Thanks!

benck:
I'm using version 4.6.4 and still having issues for cloudflare flexible SSL.

The login page looks good https://example.com/phpmyadmin/. But after entering username/password, the login redirects to https://example.com:80/phpmyadmin/.... which is wrong.

I used the solution on [serverfault](https://serverfault.com/questions/715247/phpmyadmin-uses-port-80-on-cloudflare-flexible-ssl) and it works. Thanks!

laukstein:
@benck, https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/Config.php doesn't use my proposed fix. I have noticed tons of phpMyAdmin bugs and lack of security over the years, and I recommend you to switch to [HeidiSQL](http://www.heidisql.com).

PR #11746, opened 2015-12-14
Fixes #11730

Achilles-96:
Signed-off-by: Raghuram Vadapalli raghuram.vadapalli@research.iiit.ac.in

PR #11803, opened 2016-01-02
Bug fix for #11746

Achilles-96:
Some times, the key 'TABLE_COMMENT' is not present instead of just being empty string. This fix is for handling such cases.

### DIFF
--- a/templates/table/structure/display_table_stats.phtml
+++ b/templates/table/structure/display_table_stats.phtml
@@ -2,7 +2,9 @@
     <?php $odd_row = false; ?>
     <fieldset>
         <legend><?php echo __('Information'); ?></legend>
-        <p> <strong> <?php echo __('Table comments: ') ?> </strong> <?php echo $showtable['TABLE_COMMENT']; ?></p>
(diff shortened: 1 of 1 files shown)


lem9:
Thanks, I have a few questions and remarks.
* You refer to #11746 which is not a bug
* Bug fixes should currently be done on the QA_4_5 branch
* Can you give a scenario to reproduce this problem?

Achilles-96:
Yes, I referred it wrongly. It is an update for #11746. I made that PR.
I can't say how exactly to reproduce the error because sometimes the key 'TABLE_COMMENT' isn't just an empty string. It is a null object. I didn't understand why is it a null object in some scenarios.

lem9:
Ok. Could you redo the fix in the QA_4_5 branch, this time using a better "if" with brackets or the ternary operator "?"?

Achilles-96:
But that PR is merged in master. So, the patch cannot be applied to QA_4_5 branch as the earlier patch is not present in it.

lem9:
Ok, we'll make this a master-only fix, for the upcoming 4.6.0 release. Can you improve the "if" ?

PR #11806, opened 2016-01-03
Fixes #11798

### LINKED ISSUE #11798: Table engine detection fails with dot in database name

Achilles-96:


devenbansod:
I think this is present in 4.5.3.1 too, and I perceive this as a bug in PMA, so we can target the next bug-fix release 4.5.4 (QA_4_5) instead of master.

Achilles-96:
I created a new PR. Thanks for reminding.

devenbansod:
Okay. So closing this.

PR #11807, opened 2016-01-03
Fixes #11798

### LINKED ISSUE #11798: Table engine detection fails with dot in database name

Achilles-96:

### DIFF
--- a/libraries/util.lib.php
+++ b/libraries/util.lib.php
@@ -24,14 +24,23 @@ if (! defined('PHPMYADMIN')) {
 function get($array, $path, $default = null)
 {
     if (is_string($path)) {
-        $path = explode('.', $path);
+        $path = str_getcsv($path,'.','`');
     }
+
     $p = array_shift($path);
     while (isset($p)) {
-        if (!isset($array[$p])) {
+        if (!isset($array[$p]) && !isset($array['`' . $p . '`'])) {
             return $default;
         }
-        $array = $array[$p];
+
+        if (isset($array[$p])) {
+            $array = $array[$p];
+        }
+
+        if (isset($array['`' . $p . '`'])) {
+            $array = $array['`' . $p . '`'];
+        }
+
         $p = array_shift($path);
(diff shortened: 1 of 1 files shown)


lem9:
Sorry but the problem described in #11798 is still present.

Achilles-96:
Can you please tell me how to reproduce it.

lem9:
It's described in the issue (second sentence).

Achilles-96:
I created a table named `some()+.=table`(in back ticks) and the relation view of that table is displaying 'Foreign key constraints'. Is there anything else which I'm missing?

lem9:
The issue does not mention a table name containing a dot.

Achilles-96:
Oh, it doesn't work for tables containing dot also which is why I got misled and fixed that. I'll try and fix the issue with databases as well and update the PR. Thank you.

Achilles-96:
This fix seems to be working fine for databases as well. I created a database with name `` `some.database` `` and a table in it. And the relation view showed Foreign key constraints detecting InnoDB engine.

lem9:
As I said above, 2d12475131c1d1a248b128f71bb531b15cedca29 does not fix the problem for me. Tried on MySQL 5.7.8 with a database named "some.database" and table t1 with one column c1. No "Foreign key constraints section" is shown.

Achilles-96:
Alright, the main problem here is joining database name and table name with dot like this https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/Table.php#L329 but dot is accepted character in database name and table names. Small modification may fix this particular bug but definitely there will be a problem in future. 
In order to avoid future bugs due, we have to change all those instances where db name and table name are joined with dot and also corresponding explode statements. The solution is that we should change separator to some invalid character for naming MySQL tables, but MySQL allows any character in names. 
Any suggestions are welcome.

lem9:
@Achilles-96 Please copy your last comment in the issue itself.

PR #11813, opened 2016-01-04
Fix #11752 : 4.5.2 breaks master-master cluster

### LINKED ISSUE #11752: 4.5.2 breaks MySQL Master Master Cluster
I have updated phpmyadmin from 4.2.12 to 4.5.2.

I have configured a MySQL Master Master Replication and I can login on both nodes from phpmyadmin, here is the phpmyadmin config:
/* Servers configuration */
$i = 0;

/* Server: hostname1 [1] */
$i++;
$cfg['Servers'][$i]['verbose'] = 'hostname1';
$cfg['Servers'][$i]['host'] = '1.1.1.1';
$cfg['Servers'][$i]['port'] = '3306';
$cfg['Servers'][$i]['socket'] = '';
$cfg['Servers'][$i]['connect_type'] = 'tcp';
$cfg['Servers'][$i]['extension'] = 'mysqli';
$cfg['Servers'][$i]['auth_type'] = 'cookie';
$cfg['Servers'][$i]['user'] = 'root';
$cfg['Servers'][$i]['password'] = '';

/* Server: hostname2 [2] */
$i++;
$cfg['Servers'][$i]['verbose'] = 'hostname2';
$cfg['Servers'][$i]['host'] = '2.2.2.2';
$cfg['Servers'][$i]['port'] = '3306';
$cfg['Servers'][$i]['socket'] = '';
$cfg['Servers'][$i]['connect_type'] = 'tcp';
$cfg['Servers'][$i]['extension'] = 'mysqli';
$cfg['Servers'][$i]['auth_type'] = 'cookie';
$cfg['Servers'][$i]['user'] = 'root';
$cfg['Servers'][$i]['password'] = '';

/* End of servers configuration */

$cfg['blowfish_secret'] = 'somelongstring';
$cfg['DefaultLang'] = 'en';
$cfg['ServerDefault'] = 1;
$cfg['UploadDir'] = '';
$cfg['SaveDir'] = '';
$cfg['VersionCheck'] = '0';
$cfg['DisplayServersList'] = 'true';
$cfg['VersionCheckProxyUrl'] = 'http://proxy.domain:8080'

After the update, my SLAVE on hostname2 breaks with the following errors:

Last_Error: Error 'Duplicate entry 'pma_test_host-mysql-pma_test_user' for key 'PRIMARY'' on query. Default database: ''. Query: 'INSERT INTO `mysql`.`db`(`host`, `Db`, `User`) VALUES("pma_test_host", "mysql", "pma_test_user")'

Last_Error: Error 'Duplicate entry 'pma_test_host-mysql-pma_test_user--' for key 'PRIMARY'' on query. Default database: ''. Query: 'INSERT INTO `mysql`.`columns_priv`(`host`, `Db`, `User`, `Table_name`, `Column_name`) VALUES("pma_test_host", "mysql", "pma_test_user", "", "")'

Last_Error: Error 'Duplicate entry 'pma_test_host-mysql-pma_test_user-' for key 'PRIMARY'' on query. Default database: ''. Query: 'INSERT INTO `mysql`.`tables_priv`(`host`, `Db`, `User`, `Table_name`) VALUES("pma_test_host", "mysql", "pma_test_user", "")'

Last_Error: Error 'Duplicate entry 'pma_test_host-mysql-pma_test_user--PROCEDURE' for key 'PRIMARY'' on query. Default database: ''. Query: 'INSERT INTO `mysql`.`procs_priv`(`host`, `Db`, `User`, `Routine_name`, `Routine_type`) VALUES("pma_test_host", "mysql", "pma_test_user", "", "PROCEDURE")'

The "mysql" database is not replicated over the cluster. 
replicate-ignore-db = mysql 
replicate-ignore-db = information_schema
replicate-ignore-db = performance_schema

I dont have configured the pma control user as I dont use the pma configuration storage.

It seems like pma is doing some permissions checks after I log in with my account which seems to break my master master replication.

devenbansod:
Temporarily remove the test for Insert privileges.

If many users complain of confusing UI, then we would make further changes if required.

### DIFF
--- a/libraries/check_user_privileges.lib.php
+++ b/libraries/check_user_privileges.lib.php
@@ -79,32 +79,6 @@ function PMA_checkRequiredPrivilegesForAdjust()
     }
 
     if ($privs_available) {
-        // unset the `foreign_key_checks` temporarily
-        $GLOBALS['dbi']->tryQuery('
-            SET @@session.foreign_key_checks = 0
-        ');
-
-        $insert_privs_available = $GLOBALS['dbi']->tryQuery(
-            'INSERT INTO `mysql`.`db`(`host`, `Db`, `User`) VALUES("pma_test_host", '
-            . '"mysql", "pma_test_user")'
-        );
-
-        // set the `foreign_key_checks` back
-        $GLOBALS['dbi']->tryQuery('
-            SET @@session.foreign_key_checks = 1
-        ');
-
-        // If successful test insert, delete the test row
-        if ($insert_privs_available) {
-            $GLOBALS['dbi']->tryQuery(
-                'DELETE FROM `mysql`.`db` WHERE host = "pma_test_host" AND '
-                . 'Db = "mysql" AND User = "pma_test_user"'
-            );
-        }
-        $privs_available = $insert_privs_available && $privs_available;
-    }
-
-    if ($privs_available) {
         $update_privs_available = $GLOBALS['dbi']->tryQuery(
             'UPDATE `mysql`.`db` SET `host` = "" WHERE `host` = "" AND '
             . '`Db` = "" AND `User` = ""'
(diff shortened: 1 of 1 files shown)


nijel:
I'm not sure this will fully resolve the issue, the other UPDATE queries IMHO have same problem as well...

devenbansod:
The main issue was insertion of the duplicate keys in a replication config, where the mysql database is not replicated. 
But IMO the update queries should not have the problem with duplicate keys since all the update queries would ideally be run on zero rows, as those conditions won't match any row.

nijel:
But what if they match something? BTW: There seems to be at least one DELETE left even with your patch.

Maybe we should figure out better way to find out privileges than directly messing up with the mysql database...

devenbansod:
yes. I agree that this is not the best way to do checking for the privileges. We should do something similar to `PMA_analyse grant()` (may add some code to it).

My original thinking was that: The `SELECT` would surely not be a problem, the `UPDATE` and `DELETE` statements are run on the rows where the fields which are a part of primary key are blank. But this *does not* guarantee that we would not affect the tables. 

So, closing this for now.

PR #11815, opened 2016-01-04
Fix handling of databases with dot in a name

### LINKED ISSUE #11798: Table engine detection fails with dot in database name

nijel:
We do not need to concatenate and parse the items, the cache backend
accepts array directly.

Fixes #11798

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -16,6 +16,7 @@ phpMyAdmin - ChangeLog
 - issue        Undefined index: host
 - issue #11810 'Add to central columns' (per column button) does nothing
 - issue #11727 SQL duplicate entry error trying to INSERT in designer_settings table
+- issue #11798 Fix handling of databases with dot in a name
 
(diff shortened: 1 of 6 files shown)


nijel:
I could not test the changes right now, so the commit is completely untested...

nijel:
I've fixed my test environment, and the patch seems to work fine. Still I'd like to hear feedback as this might look too intrusive for QA_4_5...

Achilles-96:
@nijel In https://github.com/phpmyadmin/phpmyadmin/blob/QA_4_5/libraries/util.lib.php#L27, as $path will be an array, I think explode can be removed.

nijel:
@Achilles-96 The explode is used only if it's not an array and there are still some places where the string is used in call to ``Util\\get``, but with safe values (eg. ``libraries/controllers/TableStructureController.class.php``).

PR #11832, opened 2016-01-08
Fix #11752

### LINKED ISSUE #11752: 4.5.2 breaks MySQL Master Master Cluster

devenbansod:
Move to analysing the grant strings to perform checks for privileges instead of running all the test queries on mysql db.

@nijel : could you please take a look/test this once if you get time ? I am not exactly sure if I have covered all the cases in checking for the user having required privileges.

### DIFF
--- a/libraries/check_user_privileges.lib.php
+++ b/libraries/check_user_privileges.lib.php
@@ -16,277 +16,69 @@ $GLOBALS['is_superuser'] = $GLOBALS['dbi']->isSuperuser();
 
 /**
  * Check if user has required privileges for
- * performing 'FLUSH PRIVILEGES' operation
- *
- * @return void
- */
-function PMA_checkRequiredPrivilegesForFlushing()
-{
-    if (PMA_Util::cacheExists('flush_priv')) {
-        $GLOBALS['flush_priv'] = PMA_Util::cacheGet(
-            'flush_priv'
-        );
-        return;
-    }
-
-    $GLOBALS['flush_priv'] = $GLOBALS['dbi']->tryQuery(
-            'FLUSH PRIVILEGES'
-    );
-
-    // must also cacheUnset() them in
-    // libraries/plugins/auth/AuthenticationCookie.class.php
-    PMA_Util::cacheSet('flush_priv', $GLOBALS['flush_priv']);
-}
-
-/**
- * Check if user has required privileges for
  * performing 'Adjust privileges' operations
  *
+ * @param string $show_grants_str     string containing grants for user
+ * @param string $show_grants_dbname  name of db extracted from grant string
+ * @param string $show_grants_tblname name of table extracted from grant string
+ *
  * @return void
  */
-function PMA_checkRequiredPrivilegesForAdjust()
-{
-    if (PMA_Util::cacheExists('db_priv')) {
-        $GLOBALS['db_priv'] = PMA_Util::cacheGet(
-            'db_priv'
-        );
-        $GLOBALS['col_priv'] = PMA_Util::cacheGet(
-            'col_priv'
-        );
-        $GLOBALS['table_priv'] = PMA_Util::cacheGet(
-            'table_priv'
-        );
-        $GLOBALS['proc_priv'] = PMA_Util::cacheGet(
-            'proc_priv'
-        );
-        return;
-    }
+function PMA_checkRequiredPrivilegesForAdjust(
+    $show_grants_str,
+    $show_grants_dbname,
+    $show_grants_tblname
+) {
+    // '... ALL PRIVILEGES ON *.* ...' OR '... ALL PRIVILEGES ON `mysql`.* ..'
+    // OR
+    // SELECT, INSERT, UPDATE, DELETE .... ON *.* OR `mysql`.*
+    if ($show_grants_str == 'ALL'
+        || $show_grants_str == 'ALL PRIVILEGES'
+        || (mb_strpos(
+                $show_grants_str, "SELECT, INSERT, UPDATE, DELETE"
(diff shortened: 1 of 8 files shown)


nijel:
I'll look at it next week. However I'm not sure if this is good fit for QA_4_5 as the change is quite intrusive. Maybe @lem9 can comment on this as well...

lem9:
It's a big patch but it contains mostly code removal, so I would be confident for QA_4_5.

zaverichintan:
@devenbansod
 how much efficient is this change? 
Will it affect the security part of pma?

devenbansod:
Hi @zaverichintan : this is slightly efficient as we are running only one query `SHOW GRANTS` now and actually a better way too, to check for the required privileges. 
Previously, we would run test queries once the user logged in, and if those queries were successful, we would assume that user has the required privileges. 
This patch would change that and we would now follow a procedure similar to `PMA_analyseShowGrant()` to determine the privileges.

Also, I don't see any changes to the security in phpMyAdmin due to this patch.
And no, this does not fix the issue #11834.

I hope I answered all your queries.

zaverichintan:
@devenbansod  Thank you for your response. 
What if someone fixes issue #11834 and then this fix #11752 is merged, the issue will again rise. Try to some that if possible. 
Its very efficient way of checking.

nijel:
@zaverichintan I don't think these are related. In #11834 it seems that the SQL we use for adjusting privileges is generated wrong, this code is about detecting whether we can adjust privileges (so in case it would be wrong, it would not work at all, not just in single case).

nijel:
I did some testing and this seems to work fine. However with current complexity, I think the ``PMA_analyseShowGrant`` really deserves separate testcase whether it works as expected...

devenbansod:
@nijel : I was also thinking on re-factoring some code of `PMA_analyseGrant()` and adding some tests.
But do you think the tests and the refactoring should be targetted to QA_4_5 or to master in a separate PR ?

nijel:
Refactoring should definitely go to the master and it indeed makes sense to add tests with it. I'll merge this and let's target refactoring and tests to master...

PR #11838, opened 2016-01-09
SSL validation flag

machavity:
I installed the latest PMA and had to make it work with a remote server running 5.7. The connection had to be encrypted and it conveniently came with its own self-signed certificates so I used them in my PMA setup and I couldn't get it to work. It always returned a 2002 error. I could connect over CLI and Workbench so I dug into it. After some searching I ran across [this PHP bug](https://bugs.php.net/bug.php?id=68344) which described the problem as being due to the following combination
- MySQL 5.6 or later
- PHP using mysqlnd
- A self-signed certificate (with the name not matching the server name)

PHP 5.6.16 introduced a fix as a result of that bug report. They added an additional flag (still not documented yet) that disables the mysqlnd validation. By replacing **MYSQLI_CLIENT_SSL** with **MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT** in my local PMA install I got PMA to finally work over SSL.

This patch adds a new configuration `$cfg['Servers'][$i]['ssl_verify']` which, when set to `false`, will tell PMA to use this new flag (if present).

Signed-off-by: Joel Hutchinson joel.hutchinson@onlinecommercegroup.com

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -289,6 +289,20 @@ Server connection settings
     :default: NULL
 
     List of allowable ciphers for SSL connections to the MySQL server.
+    
+.. config:option:: $cfg['Servers'][$i]['ssl_verify']
+
+    :type: boolean
+    :default: true
+
+    If your PHP install uses the MySQL Native Driver (mysqlnd), your
+    MySQL server is 5.6 or later, and your SSL certificate is self-signed, 
+    there is a chance your SSL connection will fail due to validation.
+    Setting this to ``false`` will disable the validation check.
+    
+    .. note::
+    
+        This flag only works with PHP 5.6.16 or later
 
 .. config:option:: $cfg['Servers'][$i]['connect_type']
 
--- a/libraries/dbi/DBIMysqli.php
+++ b/libraries/dbi/DBIMysqli.php
@@ -158,7 +158,16 @@ class DBIMysqli implements DBIExtension
                 $cfg['Server']['ssl_ca_path'],
                 $cfg['Server']['ssl_ciphers']
             );
-            $client_flags |= MYSQLI_CLIENT_SSL;
+            $ssl_flag = MYSQLI_CLIENT_SSL;
+            /*
+             * disables SSL certificate validation on mysqlnd for MySQL 5.6 or later
+             * @link https://bugs.php.net/bug.php?id=68344
+             * @link https://github.com/phpmyadmin/phpmyadmin/pull/11838
+             */
+            if(!$cfg['Server']['ssl_verify'] && defined('MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT')) {
+                $ssl_flag = MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT;
+            }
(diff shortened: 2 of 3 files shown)


ibennetch:
Very interesting. Thanks for your research and pull request. Would you mind squashing your commits and pushing the result back to your branch for a cleaner merge history?

libraries/dbi/DBIMysqli.php
```diff
@@ -158,7 +158,13 @@ public function connect(
                 $cfg['Server']['ssl_ca_path'],
                 $cfg['Server']['ssl_ciphers']
             );
-            $client_flags |= MYSQLI_CLIENT_SSL;
+            $ssl_flag = MYSQLI_CLIENT_SSL;
+            if(!isset($cfg['Server']['ssl_verify'])) $cfg['Server']['ssl_verify'] = true;
+            /* disables SSL certificate validation on mysqlnd for MySQL 5.6 or later */
```
ibennetch: I think I would add links here referencing both this pull request and the PHP bug report.

machavity:
@ibennetch Added comments and squashed as requested

ibennetch:
Looks good to me, thanks. I'd like to get at least one other developer to agree with adding another configuration directive before merging, but I think this is worthwhile.

lem9:
Adding this directive looks fine to me too.

libraries/dbi/DBIMysqli.php
```diff
@@ -158,7 +158,17 @@ public function connect(
                 $cfg['Server']['ssl_ca_path'],
                 $cfg['Server']['ssl_ciphers']
             );
-            $client_flags |= MYSQLI_CLIENT_SSL;
+            $ssl_flag = MYSQLI_CLIENT_SSL;
+            if(!isset($cfg['Server']['ssl_verify'])) $cfg['Server']['ssl_verify'] = true;
```
ibennetch: Sorry if I'm missing the obvious here, but is there a reason for this line? `config.default.php` is always read and so `ssl_verify` should always be set to *something*, so I'm not sure why there's a need to test it again here.

machavity: Ah. I wasn't sure if it was read or not. So this test is redundant then. I'll remove it

doc/config.rst
```diff
@@ -289,6 +289,20 @@ Server connection settings
     :default: NULL
 
     List of allowable ciphers for SSL connections to the MySQL server.
+    
+.. config:option:: $cfg['Servers'][$i]['ssl_verify']
+
+    :type: boolean
+    :default: true
+
+    If your PHP install uses the MySQL Native Driver(mysqlnd), your
```
ibennetch: Since you're making changes now anyway, could you add a space here before the `(` ? Then squash once more and it's all set.

ibennetch:
I'll take care of the rest this, thanks @machavity for your contribution!

doc/config.rst
ibennetch: Nevermind, I'll do it for you.

ibennetch:
ff014ac34deb44c277c52744758479bb937b3c4a merged. Thanks again!

Issue #11231, opened 2015-04-28
Do not auto-redirect to login page after timeout


pma-import:
This should not happen and is really anoying!
If you have many windows with queries and doing some implementation not all windows have regular reloads.
Keep the page open - you might show a hint that this session ended - but please only a small hint.

---
- **Original URL**: https://sourceforge.net/p/phpmyadmin/bugs/4877/
- **Original author**: mokraemer
- **Found in version**: 4.4.4

udan11:
Some tick box on login screen labeled "Remember me" would work?

lem9:
@udan11 Due to PHP's session.gc_maxlifetime the "24 hours or something like that" delay could not be guaranteed. Also, I don't believe that we can override this setting, see comments in http://www.webhostingtalk.com/showthread.php?t=489764

udan11:
Alright. I understand now.

What if an ajax call was made every couple of minutes that refreshed the session? As far as I can understand all sessions that are older than `gc_maxlifetime` seconds are removed. If that Ajax call would increase a counter or something like that and save it in the session, the last modified time would be updated of the session would be updated and the session would be active for other `gc_maxlifetime` seconds.

lem9:
I believe this is what the UpdateIdleTime() function is doing, see js/functions.js. By the way, it must respect the LoginCookieValidity directive.

4wk-:
Hey guys. Sorry to re-open this issue, but I still get the "auto-refresh page => login" when my pma session expired. As one guy said, it's really annoying!

I work with a lot of PMA tabs, with many databases, and even my session expires I don't want my browser to refresh the page then redirect me to the login page if I don't do anything.

Is there a way to disable this auto-refresh ? 

I **think** this is a new behaviour, cause I never noticed this behaviour in previous version of PMA (I'm now running Version : 4.5.5.1)

CaptainMisery:
I am also annoyed by the auto-logout. I now have to remember to copy queries to a text editor before taking a long break. I could increase the timeout, but I don't want to. I'm ok with the session expiring, I just don't want to lose work because the system decided to send 10 tabs back to the login screen. I would even be ok with the current setup being the default if a setting is added to disable that redirect "feature".

nijel:
I've adjusted the code to better deal with most frequent situations:

- when PHP's `session.gc_maxlifetime` is shorter than LoginCookieValidity
- when there are multiple phpMyAdmin windows open

[...]

PR #11844, opened 2016-01-11
fixing of bug #11231

sunnygkp10:
By making a ajax call to server after a certain time (10 min) . we can fix the problem of session timeout.

Kindly give reviews about the change i have made.
Thanks.

### DIFF
--- a/js/functions.js
+++ b/js/functions.js
@@ -809,6 +809,20 @@ AJAX.registerOnload('functions.js', function () {
     function SetIdleTime() {
         _idleSecondsCounter++;
     }
+    //function for auto-redirect to login page after timeout #11231 bug
+    //It make a ajax call to server in every 10 minute since session expire in 24 minutes.
+    var refreshTime = 600000; // every 10 minutes in milliseconds
+    window.setInterval( function() {
+        $.ajax({
+            cache: false,
+            type: "GET",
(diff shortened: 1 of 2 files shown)


lem9:
Thanks for the suggestion. Why do you assume that an interval of 10 minutes is correct in all situations? The timeout can vary based on the LoginCookieValidity directive and php.ini's garbage collection setting. Also, please have a look in js/functions.js, there is a function called UpdateIdleTime() and I thought that its purpose was what you're trying to do.

sunnygkp10:
thank you sir . 
can we take an interval less than the LoginCookieValidity for solution ??

lem9:
I updated my comment, have a look at UpdateIdleTime().

sunnygkp10:
Sir , in my solution timeout never occur, even if user is inactive from a long time and browser window is open .

lem9:
Did you have a look at the function I talked about? It comes from this commit: https://github.com/phpmyadmin/phpmyadmin/commit/42223e1bf658e721359ccfb8b19fbad292b7c4e8

lem9:
Also, removing the feature provided by the LoginCookieValidity directive needs to have good reasons, as I can see good reasons to keep this feature. Therefore it should be discussed on the mailing list.

PR #11858, opened 2016-01-14
Refactor check_user_privileges.lib.php and add test cases for 2 functions

devenbansod:
Refactor check_user_privileges.lib.php and add test cases for 2 functions


devenbansod:
Some changes still required. I will make those and merge later.

PR #11862, opened 2016-01-15
Add support for Ctrl + C

Achilles-96:
Go to database structure of any database, select some tables and press `ctrl + c`. If you like this functionality, we can implement this wherever possible to enhance usability. Shortcuts do increase speed of work.

### DIFF
--- a/js/db_structure.js
+++ b/js/db_structure.js
@@ -182,6 +182,27 @@ function PMA_fetchRealRowCount($target)
     });
 }
 
+/* Support Ctrl + C event for copying multiple tables */
+$(document).ready(function() {
+    var ctrlDown = false;
+    var ctrlKey = 17, cKey = 67;
+
+    $(document).keydown(function(e) {
+
+        if (e.keyCode == ctrlKey) {
+            ctrlDown = true;
+        }
+        if(ctrlDown==true&&e.keyCode==cKey) {
+            $('select[name=submit_mult]').val('copy_tbl');
+            $('select[name=submit_mult]').trigger('change');
+        }
+    }).keyup(function(e) {
+        if (e.keyCode == ctrlKey) {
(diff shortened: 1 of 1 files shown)


lem9:
Please read this thread: https://lists.phpmyadmin.net/pipermail/developers/2015-October/thread.html

Achilles-96:
@lem9 Is there any scheme that PMA follows for not clashing with browser's shortcuts(I mean, is any scheme followed for coming up with `ctrl+enter`). If there isn't anything, then I'll try to come up with reasonable shortcuts myself.

lem9:
@Achilles-96 I could not find any doc related to this on wiki.phpmyadmin.net, but you are right, it's needed. It's a tricky problem, because of the difference between browsers for shortcut implementation, and also what they will do in the future. That's why we never used many shortcuts. You may restart the discussion on the developers mailing list, or ask there for more participation here.

guillesalazar:
Hi all! I saw @Achilles-96 invitation on the developers mailing list. I think we could start with a basic list/chart/table of well known browser shortcuts to start with a list of them that we *can't use* and then take it from there. I think implementing them is an amazing idea.

ibennetch:
I think it's a good idea to make a list in the wiki. Please see https://wiki.phpmyadmin.net/pma/Shortcut_keys for the beginnings of that page.

I'm slightly in favor of having some shortcut keys, but these should be discussed and care should be used to avoid overriding browser functionality.

lem9:
Stealing CTRL+C is IMO one of the worst things you can do. It's a common convention to copy text that has been selected.

Achilles-96:
I didn't consider clashing with browser while making this PR. I'll update once new shortcuts are decided.

Achilles-96:
Please refer #11865.

Issue #11864, opened 2016-01-16
.noSelect is not a function


lem9:
$(...).css(...).noSelect is not a function

This report is related to user submitted report [#18361](https://reports.phpmyadmin.net/reports/view/18361) on the phpmyadmin error reporting server.

lem9:
.noSelect is only called from makegrid.js, and is defined in functions.js. I don't understand why functions.js is sometimes included directly, as it's part of _addDefaultScripts(), but evidently this did not happen in the case of this bug. Maybe the .noSelect definition could be moved to makegrid.js.

zixtor:
A user comment indicating an update prior to this error ( from a similar report https://reports.phpmyadmin.net/reports/view/17026 ) :- 
"Manual upgraded to 4.5.2, restarted the wamp server and clicked on home button in phpMyadmin, this error popped up."

zenineasa:
I would like to get started with the development of phpMyAdmin. Can I take up this bug?

lem9:
@zenineasa You just did! :)

i-nik:
Can I take up this bug??

ibennetch:
@zenineasa any report?

shuklank2709:
I am new to phpMyAdmin development. Please tell me if I am correct what I understood about this error :  .noSelect function is defined in functions.js but it is used in makegrid.js only. So we have to move .noSelect definition to makegrid.js. Am I right?

ibennetch:
@i-nik Since @zenineasa hasn't responded in a while, you may go ahead and work on this.

@shuklank2709, currently @i-nik seems to be working on this particular bug, so you probably should find something else to look at.

i-nik:
Okay and thanks for the opportunity @ibennetch

PR #11871, opened 2016-01-18
Solution for ".noSelect is not a function #11864"

zenineasa:
I think I just came up with a solution for ".noSelect is not a function #11864". Just move the function named .noSelect from function.js to makegrid.js.

### DIFF
--- a/js/functions.js
+++ b/js/functions.js
@@ -4274,48 +4274,6 @@ AJAX.registerOnload('functions.js', function () {
         $(this).data('val-hash', AJAX.hash($(this).is(":checked")));
     });
 });
-/**
- * jQuery plugin to cancel selection in HTML code.
- */
-(function ($) {
-    $.fn.noSelect = function (p) { //no select plugin by Paulo P.Marinas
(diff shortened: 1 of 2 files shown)


lem9:
@zenineasa You did not come up with this solution, I was the one to suggest it in the issue's comments.

zenineasa:
@lem9 Yes, I agree with you. I agree with your DCO too. Where and in what name should the file containing the certificate be added?

lem9:
@zenineasa There is no file involved, you just have to follow the advice given above and use the Signed-Off-By tag when you do your commit. So please redo your commit (just one commit is needed) and sign it. See http://stackoverflow.com/questions/13457203/how-to-add-the-signed-off-by-field-in-the-git-patch for an example.

zenineasa:
@lem9 Do I have to make a fresh fork and pull request in order to do this?

lem9:
@zenineasa Make a new branch and pull request. Anyway it's better to name your branch according to the fix you are doing, see http://wiki.phpmyadmin.net/pma/Git#Working_on_new_features.

PR #11880, opened 2016-01-23
Issue#11872 Resolved

007durgesh219:


ibennetch:
You currently have three commits, could you squash them to make one? In the process, it would be helpful to change your commit message to something like "Fix HTML rendering in error message, issue 11873" or something like that. You can force push your amended commits back to the same branch which will update this pull request, there's no need to create a new one.

007durgesh219:
Thanks @ibennetch I have done require changes.

ibennetch:
Thanks, that looks great. Someone should complete a final review in the next few days.

By the way, normally a fix like this would target the QA branch (see https://wiki.phpmyadmin.net/pma/Release_Targets), which in this case would be QA_4_5.

007durgesh219:
So, Do I have to change the target or its fine

nijel:
The fix is wrong, as the error message has to be escaped in some cases (eg. when coming from MySQL server or from URL parameters).

007durgesh219:
I've corrected it. Please review again.

nijel:
Can't it be done rather at time generating the error message? Regexp parsing of HTML will break at some point. (in this case what if $param contains it).

007durgesh219:
Yup you are right.  I will generate bbcode at time of generating error then.  It will execlude $param from it.

007durgesh219:
Should I make a different function( let's say ```showDocuBBCode('faq', 'faqmissingparameters')```  ) for generating bbcode of HTML to make things clean?? In that case I will use ```showDocuBBCode('faq', 'faqmissingparameters')``` insted of ```showDocu('faq', 'faqmissingparameters')``` while generating error message.

nijel:
I think passing it parameter indicating output type (html/bbcode, defaulting to html) would be better.

007durgesh219:
OK, will do that.

007durgesh219:
I've done require changes

007durgesh219:
@nijel can you please review it again?

libraries/Util.php
```diff
@@ -549,16 +549,47 @@ public static function getDocuLink($page, $anchor = '')
     /**
      * Displays a link to the phpMyAdmin documentation
      *
-     * @param string $page   Page in documentation
-     * @param string $anchor Optional anchor in page
+     * @param string  $page   Page in documentation
+     * @param string  $anchor Optional anchor in page
+     * @param boolean $bbcode Optional flag to output bbcode
```
ibennetch: Can I suggest making this comment "Optional flag whether to output bbcode"? -- it's a small thing that IMHO makes it more readable.

ibennetch: Can I suggest making this comment "Optional flag indicating whether to output bbcode"? -- it's a small thing that IMHO makes it more readable.

007durgesh219:
Thanks @ibennetch I've changed the message.

libraries/Util.php
```diff
@@ -549,16 +549,47 @@ public static function getDocuLink($page, $anchor = '')
...
      *
      * @return string  the html link
      *
      * @access  public
      */
-    public static function showDocu($page, $anchor = '')
+    public static function showDocu($page, $anchor = '', $bbcode = false)
     {
-        return self::showDocLink(self::getDocuLink($page, $anchor));
+        $doc_link = self::showDocLink(self::getDocuLink($page, $anchor));
```
nijel: How about passing bbcode further down, so that we directly generate it instead of doing preg_replace magic?

007durgesh219: You mean generating bbcode in ```showDocLink``` instead of ```showDocu```.

nijel: Yes - generating the code directly instead of doing preg_replace on it...

007durgesh219: Thanks I've done that. Please review it again

nijel:
Thanks, now it looks good! Merged to QA_4_6 and master.

PR #11882, opened 2016-01-24
Add note about passing HTTP auth headers when running as FastCGI

djc:


ibennetch:
Thanks for your contribution. It looks good.  I assume you've tested this and had success with it?

nijel:
Thanks, indeed this setting is needed to make HTTP authentication work.

nijel:
(Pretty much same would be for nginx, but we don't document this case at all...)

PR #11888, opened 2016-01-26
Update release verification docs for Isaac

nijel:
As we now have new release manager, let's update the docs.

TODO: 
- [x] @ibennetch should register keybase profile
- [x] Update keyring on fileserver


ibennetch:
Maybe we should leave Marc listed a little while longer as in https://github.com/phpmyadmin/phpmyadmin/pull/11889 because of older releases; of course I'll only be signing new releases and people might still need to see his key.

nijel:
Good point!

Issue #11896, opened 2016-01-28
Remove hard dependency on phpseclib


robert-scheck:
Please remove the hard dependency on phpseclib again. Before (4.0.10.13), 4.4.15.3 and 4.5.4 it was possible to use phpMyAdmin without phpseclib as long as php-openssl was available (since at least 4.3.0). This has been changed in order to fix some security flaws, however without making phpseclib conditionally, see also #11892 and #11891.

nijel:
Having optional dependency on phpseclib is indeed good thing. On the other side, I'm not really sure if we want to fully go into this. In the end we would need wrapper that would wrap phpseclib, which itself is a wrapper on top of PHP modules (openssl/mcrypt). Using standard library for crypto stuff is usually called best practice as less things can go wrong there.

We already have [quite a lot of code handling this case](https://github.com/phpmyadmin/phpmyadmin/pull/11897/files) and this would just add few more conditions on top of that.

robert-scheck:
While I understand your aspects until a certain level, breaking a stable release series such as 4.4.x and 4.5.x is not great. Please use something like the following patch to unbreak at least for the current stable release series >= 4.3.

    --- phpMyAdmin-4.4.15.3-all-languages/libraries/session.inc.php	2016-01-28 04:37:03.000000000 +0100
    +++ phpMyAdmin-4.4.15.3-all-languages/libraries/session.inc.php	2016-01-28 16:44:51.000000000 +0100
    @@ -13,7 +13,10 @@
         exit;
     }
     
    -require PHPSECLIB_INC_DIR . '/Crypt/Random.php';
    +if (! function_exists('openssl_random_pseudo_bytes')
    +    || PHP_VERSION_ID < 50304) {
    +    require PHPSECLIB_INC_DIR . '/Crypt/Random.php';
    +}
     
     // verify if PHP supports session, die if it does not
     
    @@ -108,7 +111,12 @@
      * (we use "space PMA_token space" to prevent overwriting)
      */
     if (! isset($_SESSION[' PMA_token '])) {
    -    $_SESSION[' PMA_token '] = bin2hex(phpseclib\Crypt\Random::string(16));
    +    if (! function_exists('openssl_random_pseudo_bytes')
    +        || PHP_VERSION_ID < 50304) {
    +        $_SESSION[' PMA_token '] = bin2hex(phpseclib\Crypt\Random::string(16));
    +    } else {
[...]

PR #11897, opened 2016-01-28
Remove direct support for openssl in cookie auth [WIP]

nijel:
By this code we're duplicating phpseclib functionality with only benefit
of making it's dependency optional.

**This is not for merge right now, it just shows what cleanup could be done if we decide to make phpseclib non optional.**

Signed-off-by: Michal Čihař <michal@cihar.com


nijel:
See also #11896.

nijel:
In the end, I think it's worth keeping support for this feature.

PR #11900, opened 2016-01-28
Fix #11896 Remove hard dependency on phpseclib

### LINKED ISSUE #11896: Remove hard dependency on phpseclib

madhuracj:

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,6 +3,7 @@ phpMyAdmin - ChangeLog
 
 4.4.15.4 (2016-01-28)
 - issue #11892 Error with PMA 4.4.15.3
+- issue #11896 Remove hard dependency on phpseclib
 
 4.4.15.3 (2016-01-28)
 - issue        [Security] Multiple full path disclosure vulnerabilities, see PMASA-2016-1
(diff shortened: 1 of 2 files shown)


madhuracj:
I skipped the version checks suggested in https://github.com/phpmyadmin/phpmyadmin/issues/11896#issuecomment-176244475 since PHP 5.3.7 is the minimum version for 4.4.x branch.

ibennetch:
That makes sense

nijel:
I think you should also pass true as second parameter to openssl_random_pseudo_bytes. Same for #11899.

madhuracj:
Looks to me like it's a variable filled in by the function to specify whether the generated random value is cryptographically strong rather than to direct the function to generate a cryptographically strong one.

madhuracj:
Yes, didn't see your edit in time.

PR #11901, opened 2016-01-29
Add 'pageselector' below navigation tree

fidelt:
This is simply to make it easier to navigate between pages of tables
instead of scrolling up again in order to change the page.

### DIFF
--- a/libraries/navigation/NavigationTree.php
+++ b/libraries/navigation/NavigationTree.php
@@ -867,6 +867,7 @@ class NavigationTree
                 }
             }
         }
+        $retval .= $this->_getPageSelector($this->_tree);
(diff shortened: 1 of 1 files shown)


zixtor:
@fidelt This would add another selector unconditionally which seems undesirable, as in the case of database pagination, the top selector is always visible. In case of tables, adding a selector to bottom might help only when the no. of tables exceed a certain number to go out of sight.

Another way to remove the need to scroll through table pages is to suitably decrease the value of http://docs.phpmyadmin.net/en/latest/config.html#cfg_MaxNavigationItems

fidelt:
I understand, thank you for your response.

PR #11902, opened 2016-01-29
Fix issue 11834

### LINKED ISSUE #11834: Adjust privileges fails if database name contains underscores
Using phpMyAdmin 4.5.3.1, MariaDB 10.1.9, PHP 7.0.1.

If database-specific privileges are assigned to a user account on a database that has underscores in its name, Adjust privileges fails when copying or renaming a database. However, if the original database name does not contain underscores, Adjust privileges does work when copying or renaming to a database with underscores in its name. Further changes are also possible because the privilege type is changed from "database-specific" to "wildcard".
- Create a database called my_test, and create a dedicated user account on my_test with all privileges except Grant.
- In the Operations tab of my_test, copy the database to a new database called "another", leaving Adjust privileges selected.
- Select the "another" database, and select the Privileges tab. The original user account is not listed.
- Edit the privileges for the original user account to grant the same privileges on "another".
- In the Operations tab of "another", copy the database to "something_else", leaving Adjust privileges selected.
- Select the "something_else" database's Privileges tab. The user account is listed, but the type is "wildcard". Copying something_else or renaming it now correctly adjust's the user account's privileges.

devenbansod:
Fix #11834 

### DIFF
--- a/libraries/operations.lib.php
+++ b/libraries/operations.lib.php
@@ -588,6 +588,8 @@ function PMA_AdjustPrivileges_moveDB($oldDb, $newname)
             && $GLOBALS['is_reload_priv']
         ) {
             $GLOBALS['dbi']->selectDb('mysql');
+            $newname = str_replace("_", "\_", $newname);
+            $oldDb = str_replace("_", "\_", $oldDb);
 
             // For Db specific privileges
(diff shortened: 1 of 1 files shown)


nijel:
Looks good, thanks.

PR #11903, opened 2016-01-29
Search and replace with nothing

zalog:
:)

### DIFF
--- a/templates/table/search/search_and_replace.phtml
+++ b/templates/table/search/search_and_replace.phtml
@@ -1,7 +1,7 @@
 <?= __('Find:'); ?>
 <input type="text" value="" name="find" required />
 <?= __('Replace with:'); ?>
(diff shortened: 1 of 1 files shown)


zixtor:
Thanks, sounds like an acceptable change to me. Could you please redo and sign the commit, thereby agreeing to our DCO: https://github.com/phpmyadmin/phpmyadmin/blob/master/DCO ?

templates/table/search/search_and_replace.phtml
```diff
@@ -23,4 +23,4 @@
     false,
     false,
     'useRegex'
-); ?>
\ No newline at end of file
+); ?>
```
ibennetch: Perhaps your editor automatically added a newline character at the end of the file; by our coding guidelines this is not needed.

PR #11921, opened 2016-02-06
Issue fix 11688

zaverichintan:
By this code we're making wrapper functions for setting and getting cookies from array.

Signed-off-by:Chintan zaveri zaverichintan5@gmail.com

### DIFF
--- a/libraries/Config.php
+++ b/libraries/Config.php
@@ -20,6 +20,13 @@ require_once './libraries/vendor_config.php';
 $GLOBALS['pma_config_loading'] = false;
 
 /**
+ * Array to store information temporarily and later set cookie(see setcookie function).
+ */
+$GLOBALS["pmaUser"] = array();
+$GLOBALS["pmaAuth"] = array();
+$GLOBALS["pmaTest"] = array();
+
+/**
  * Configuration class
  *
  * @package PhpMyAdmin
@@ -1868,7 +1875,80 @@ class Config
         // cookie has already $value as value
         return true;
     }
+    /**
+     * sets the cookie
+     * @param string      $cookieArr Cookie array name
+     * @param string      $cookie    sub cookie name
+     * @param mixed       $value     sub cookie value
+     * @param int         $validity validity of cookie in seconds (default is one month)
+     * @param bool        $httponly whether cookie is only for HTTP (and not for scripts)
+     */
+    public function setCookieArr($cookieArr,$cookie, $value,
+            $validity = null, $httponly = true
+        ) {
+            // set cookie with new value
+            /* Calculate cookie validity */
+            if ($validity === null) {
+                $validity = time() + 2592000;
+            } elseif ($validity == 0) {
+                $validity = 0;
+            } else {
+                $validity = time() + $validity;
+            }
+            if (defined('TESTSUITE')) {
+                $_COOKIE[$cookie] = $value;
+                return true;
+            }
 
+            switch ($cookieArr) {
+            case 'pmaUser':$size = 2 ;
+                break;
+            case 'pmaAuth':$size = 2 ;
+                break;
+            default:$GLOBALS[$cookieArr] = array();
+                break;
+            }
+
+
+            if (sizeof($GLOBALS[$cookieArr])!=$size) {
+                $GLOBALS[$cookieArr][$cookie] = $value ;
+
+            }
+
+            if (sizeof($GLOBALS[$cookieArr])==$size){
+
+                $json_encoded = json_encode($GLOBALS[$cookieArr]);
+
+                return setcookie(
+                    $cookieArr,
+                    $json_encoded,
+                    $validity,
+                    $this->getCookiePath(),
+                    '',
+                    $this->isHttps(),
+                    $httponly
+                );
+            }
+        }
+
+    /**
+     * gets the cookie array
+     * @param  string $cookieArr name of cookie array to be retrieved
+     * @param  string $cookie    name of the subcookie
+     * @return string           value of subcookie
+     */
+    public function getCookieArr($cookieArr,$cookie){
+        if (isset($_COOKIE[$cookieArr])){
+
+          $json_decoded = json_decode($_COOKIE[$cookieArr]);
+            if (isset($json_decoded->$cookie)){
+                return( $json_decoded->$cookie);
+            }
+            else{
+                return null;
+            }
+        }
+    }
 
     /**
      * Error handler to catch fatal errors when loading configuration
--- a/libraries/plugins/auth/AuthenticationCookie.php
+++ b/libraries/plugins/auth/AuthenticationCookie.php
@@ -31,6 +31,10 @@ if (! empty($_REQUEST['target'])) {
 require './libraries/plugins/auth/swekey/swekey.auth.lib.php';
 
 /**
+ * To use getCookieArr function
+ */
+require_once './libraries/Config.php';
+/**
  * Handles the cookie authentication method
  *
  * @package PhpMyAdmin-Authentication
@@ -405,14 +409,14 @@ class AuthenticationCookie extends AuthenticationPlugin
         }
 
         // check cookies
-        if (empty($_COOKIE['pmaUser-' . $GLOBALS['server']])
-            || empty($_COOKIE['pma_iv-' . $GLOBALS['server']])
+        if (getCookieArr('pmaUser','pmaUser-' . $GLOBALS['server'])== null
+            || getCookieArr('pmaAuth','pma_iv-' . $GLOBALS['server'])==null
         ) {
             return false;
         }
 
         $GLOBALS['PHP_AUTH_USER'] = $this->cookieDecrypt(
-            $_COOKIE['pmaUser-' . $GLOBALS['server']],
+            getCookieArr('pmaUser','pmaUser-' . $GLOBALS['server']),
             $this->_getEncryptionSecret()
         );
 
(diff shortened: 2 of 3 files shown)


zaverichintan:
@zixtor kindly help in Over view

libraries/plugins/auth/AuthenticationCookie.php
```diff
@@ -405,14 +405,14 @@ public function authCheck()
         }
 
         // check cookies
-        if (empty($_COOKIE['pmaUser-' . $GLOBALS['server']])
-            || empty($_COOKIE['pma_iv-' . $GLOBALS['server']])
+        if (getCookieArr('pmaUser','pmaUser-' . $GLOBALS['server'])== null
```
zixtor: PHP Fatal error:  Call to undefined function libraries\plugins\auth\getCookieArr() in libraries/plugins/auth/AuthenticationCookie.php on line 408

libraries/plugins/auth/AuthenticationCookie.php
```diff
@@ -405,14 +405,14 @@ public function authCheck()
...
+            || getCookieArr('pmaAuth','pma_iv-' . $GLOBALS['server'])==null
         ) {
             return false;
         }
 
         $GLOBALS['PHP_AUTH_USER'] = $this->cookieDecrypt(
-            $_COOKIE['pmaUser-' . $GLOBALS['server']],
+            $getCookieArr('pmaUser','pmaUser-' . $GLOBALS['server']),
```
zixtor: typo: $getCookieArr

libraries/Config.php
```diff
@@ -1868,7 +1875,85 @@ public function setCookie($cookie, $value, $default = null,
...
+    /**
+     * sets the cookie
+     * @param string      $cookieArr Cookie array name
+     * @param DocBlockr   $cookie    sub cookie name
+     * @param mixed       $value     sub cookie value
+     * @param string      $default  default value
+     * @param int         $validity validity of cookie in seconds (default is one month)
+     * @param bool        $httponly whether cookie is only for HTTP (and not for scripts)
+     */
+    public function setCookieArr($cookieArr,$cookie, $value, $default = null,
```
zixtor: setCookieArr seems to duplicate the logic in setCookie, which should not be necessary.

zixtor:
I was not able to test the PR.
As for qualitative help, in general, volunteers will readily review what you wish to get merged and may also provide some suggestions in flow with you own work direction; but beyond that it poses limitations and concerns to provide thorough help, more so if you wish to apply for GSoC, as it will violate fair play for other candidates.

ibennetch:
Hi, when making changes to a pull request you don't need to close it to start a new one; you can just push in to your repository and the pull request will automatically be updated. Sometimes you'll have to `--force` the push, such as if you squash the commit history, but this still generally works.

zaverichintan:
How do I test the three types of authentication on local? 
And should I import the libraries/config.php in authentication.php?

zixtor:
Please refer to https://wiki.phpmyadmin.net/pma/Auth_types. 
I don't understand the second question about "import", but you can simply do $GLOBALS['PMA_Config']->getCookie(...) if that is what you meant.

libraries/Config.php
```diff
@@ -20,6 +20,13 @@
 $GLOBALS['pma_config_loading'] = false;
 
 /**
+ * Array to store information temporarily and later set cookie(see setcookie function).
+ */
+$GLOBALS["pmaUser"] = array();
+$GLOBALS["pmaAuth"] = array();
+$GLOBALS["pmaTest"] = array();
```
zixtor: Why do you need these globals, just for local use?

libraries/plugins/auth/AuthenticationCookie.php
zaverichintan: @zixtor To use getCookieArr in libraries/plugins/auth/AuthenticationCookie.php how should I include the config file (getCookieArr Function) .
Kindly suggest the method .

libraries/plugins/auth/AuthenticationCookie.php
zaverichintan: Fixed

libraries/Config.php
zaverichintan: I agree with you , but this is done as setCookieArr function may replace setCookie function in future.

libraries/Config.php
zaverichintan: The use of cookies didn't had wrapper function for retrieving cookies, so by making it global array , we can access it anywhere.

libraries/plugins/auth/AuthenticationCookie.php
zixtor: @zaverichintan Please see my previous comment https://github.com/phpmyadmin/phpmyadmin/pull/11921#issuecomment-180753203, I did answer it.

zaverichintan:
should I include the config.php in authenticationcookie?
Kindly help me

libraries/plugins/auth/AuthenticationCookie.php
```diff
@@ -31,6 +31,10 @@
 require './libraries/plugins/auth/swekey/swekey.auth.lib.php';
 
 /**
+ * To use getCookieArr function
+ */
+require './libraries/Config.php';
```
nijel: At most it should be require_once. But I think it should be rather used through autoloading...

libraries/Config.php
```diff
@@ -1868,8 +1875,81 @@ public function setCookie($cookie, $value, $default = null,
...
+
+                return setcookie(
+                    $cookieArr,
+                    $json_encoded,
+                    $validity,
+                    $this->getCookiePath(),
+                    '',
+                    $this->isHttps(),
+                    $httponly
+                );
```
nijel: Why don't you either replace existing setCookie or use it here? Copying the code has never done anything good.

libraries/plugins/auth/AuthenticationCookie.php
zaverichintan: Should i make a change in autoloader ?

libraries/Config.php
zaverichintan: I made two functions right now so other developers can easily use the existing function and also to solve cookie related errors.

poush: I too believe that it is a bad idea to duplicate the code. And after all phpmyadmin codebase should be more improved to OO not to make it more procedural

zaverichintan: I agree with you , but this code is a copy to replace the function once everyone agrees to the array json concept.

poush:
One my first look there many places where $_COOKIE is directly used and to make it work i think there is need to update those lines like ```isset($_COOKIE['...']``` , etc from whole codebase.

zaverichintan:
@poush I am also thinking the same but the cookies are set when http header is sent, I only replaced the usage with the wrapper function already used. 
Replacing leads to different issues.

poush:
But then how it will work if half of the code expecting old style of cookies when half portion is setting up cookies in JSON

ibennetch:
Near the bottom of the CI report we can see the error output:

> [exec] PHP Fatal error:  Cannot declare class PMA\libraries\Config, because the name is already in use in /home/travis/build/phpmyadmin/phpmyadmin/libraries/Config.php on line 34

Which I haven't investigated fully, but am guessing is related to `+require './libraries/Config.php';` which @nijel has already mentioned above. Try fixing that and then check the test again.

zaverichintan:
@ibennetch thank you for your response, trying that. 
@poush there are different parameters which are same, only one parameter has changed (value of the cookie is now saved as json)

poush:
Did you tested your code ? Because what i am saying is lets suppose you have

```[pmaConfig] => [pma_lang,pma_fontSize,....]```

but in LanguageManager.php
```
  725:         if (! empty($_COOKIE['pma_lang'])) {
  726:             if (isset($langs[$_COOKIE['pma_lang']])) {
  727:                 return $langs[$_COOKIE['pma_lang']];
  728              }
```
so it expects the old way

and about this
```
/phpmyadmin/import_status.php:
   45  
   46      session_name('phpMyAdmin');
   47:     session_id($_COOKIE['phpMyAdmin']);
   48  }
   49   */
```

Above that, whenever you call setCookieArr(), a new "set-cookie" header will be made. Having multiple header for same name,path has undefined behaviour. For example my chrome and safari is saving first one only while firefox is saving N times the first one header.

**Comment by zaverichintan** [comment]
@poush could you please suggest a workaround as there is no proper wrapper function for getting cookies .

File: libraries/Config.php (hunk 4 above)
Comment (nijel) [inline libraries/Config.php]: It's better to store within some singleton, most likely Response class. This way you can also properly handle setting them once when sending response to client.

File: libraries/Config.php [hunk 7]
```diff
@@ -1868,8 +1875,81 @@ public function setCookie($cookie, $value, $default = null,
...
     }
+    /**
+     * sets the cookie
+     * @param string      $cookieArr Cookie array name
+     * @param string      $cookie    sub cookie name
+     * @param mixed       $value     sub cookie value
+     * @param int         $validity validity of cookie in seconds (default is one month)
+     * @param bool        $httponly whether cookie is only for HTTP (and not for scripts)
+     */
+    public function setCookieArr($cookieArr,$cookie, $value,
```
nijel: Thinking more about this, the cookie handling really doesn't belong to the Config class, but rather to Response one.

zaverichintan: Should I change the file for the function? As I too feel the same. It should be with other wrapper  functions. @nijel  kindly suggest the file.

bhagyeshworld:
@zaverichintan You are missing class object reference while calling getCookieArr() in /home/travis/build/phpmyadmin/phpmyadmin/libraries/plugins/auth/AuthenticationCookie.php on line 412.

Try calling that functoin using $GLOBALS['PMA_Config']->setCookie(... and it should build well !!

Issue #11930, opened 2016-02-09
Can't change AUTO_INCREMENT


HLeithner:
I its not possible to set the auto increment value in phpmyadmin 4.5.4.1 on a MyISAM table.

After changing the value in the Operations tab, the page reloads and says SQL query was ok.
But some options are missing in "Table options" box and "Table maintenance" box. also the value has not changed.

Here are some screenshots

![before](https://cloud.githubusercontent.com/assets/1497730/12912518/6b27594e-cf19-11e5-8c0c-03246947c92d.jpg)

![after](https://cloud.githubusercontent.com/assets/1497730/12912517/6b25fc7a-cf19-11e5-96b4-17f55902d7e8.jpg)

nijel:
Same happens with InnoDB on QA_4_5, QA_4_6 and master.

007durgesh219:
This is because of the case issue in ```setGlobalVariablesForEngine``` in libraries/operations.lib.php:1873. We need to convert $tbl_storage_engine to upper case.

PR #11931, opened 2016-02-09
Fix Can't change AUTO_INCREMENT, Issue #11930

007durgesh219:
Signed-off-by: Durgesh 007durgesh219@gmail.com


nijel:
I don't think this is fix in the right place - there are more places where the code assumes this is already upper case.

Issue #11688, opened 2015-11-24
Proposal: decrease cookies usage.


netroby:
Current phpmyadmin using cookies for auth type by default.
But phpmyadmin take lot's of cookies. it will use at least 10 cookies.

That's too much. because the per domain cookie limit only 20.

```
pma_lang
pma_iv-1
pma_console_mode
pma_console_height
pma_console_config
pma_collation_connection
pmaUser-1
pmaPass-1
pmaCookieVer
phpMyAdmin
```

We can decrease cookie by merge cookie into json style value or other ways to do this.

2 - 3 cookies will be good.

netroby:
For example: cookie named pma_data
```json
{"pma_lang":"en","pma_iv-1":"xxxxx","pma_console_mode":"coolapse","pma_console_height":"92"}
  
```

lem9:
You are quoting an older cookie spec. The current one AFAIK (https://tools.ietf.org/html/rfc6265) mentions "At least 50 cookies per domain."

netroby:
I am using a domain with other my php apps. and also using phpmyadmin (on a special port, private access only). I meet a problem with login with phpmyadmin. at last i figure out it was cookies problem. I cleared other cookies. Then  i can normally login into phpmyadmin.

lem9:
Starting with version 4.5.1, there is code to clear cookies on version change.

lem9:
@netroby Was it a case of updating phpMyAdmin? Which version are you using?

nijel:
Indeed the cookie usage could be reduced. I think it could go down to:

* ``pmaUser-N`` (we want longer validity here than other cookies)
* ``pmaAuth-N`` (session cookie containing IV and encrypted password)
* ``pmaConfig`` containing rest of the configuration including it's version to handle upgrades

zaverichintan:
@lem9 is there any need to create multiple cookies? 
As it seems the creator of this logic mother have thought about merging it.

nijel:
I think the reason is mostly historical - nobody really thought about cookie usage and when needed to store something in cookie, he added new one...

lem9:
@zaverichintan There might be a reason: not deleting all cookies at once. Try "git grep removeCookie" to get an idea about when which cookies are deleted and why.

zaverichintan:
Where should I start working on this issue? 
Kindly help me.

lem9:
[...]

PR #11936, opened 2016-02-10
setcookie and getcookie wrapper function for JS #11688 

zaverichintan:
Wrapper function for setting and getting cookies in JS.
Signed-Off-By: chintan zaveri zaveri_chintan@rediff.com

### DIFF
--- a/js/functions.js
+++ b/js/functions.js
@@ -80,6 +80,11 @@ var fulltext_indexes = [];
 var spatial_indexes = [];
 
 /**
+ * @var cookieObj object to hold cookies.
+ */
+var cookieObj = {};
+
+/**
  * Make sure that ajax requests will not be cached
  * by appending a random variable to their parameters
  */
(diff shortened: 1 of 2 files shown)


libraries/Config.php
```diff
@@ -18,6 +18,7 @@
  * Indication for error handler (see end of this file).
  */
 $GLOBALS['pma_config_loading'] = false;
+//chintanzaveri issue11688
```
nijel: I guess this should not be included...

js/functions.js
```diff
@@ -4967,3 +4972,39 @@ AJAX.registerOnload('functions.js', function(){
...
+    var jsonDecoded = JSON.parse(arr);
+    var value = jsonDecoded[String(key)];
+}
+
+if (typeof value =='undefined'){
+    return(null);
+}
+else
+    return(value);
+}
```
nijel: Can you please properly indent the code?

zaverichintan:
@nijel kindly over view this, This change does not produce any errors and is tested.

nijel:
This really makes no sense without the PHP side of the patch, which is not yet in state it could be merged (or work).

zaverichintan:
IMHO both the parts can be done separatly as in JS part there was no wrapper function to set cookie and get cookie which made it clear to develop. 
PHP part fails due to no usage of existing wrapper function,which can't be done using my commit even. 
So we can just decrease the JS set Cookies.

nijel:
Closing as there is no matching PHP code for this.

zaverichintan:
Wrapper function for cookies not being used everywhere. Cookies are set and accessed directly using php setcookie function, which makes it difficult to combine multiple cookies.

PR #11961, opened 2016-02-14
Issue Fix #11947 | Resolved undefined variable error

### LINKED ISSUE #11947: Notice in sql.lib
Undefined index: is_affected in ./libraries/sql.lib.php:1349

Happens with 4.6.0-alpha1

This report is related to user submitted report [#19829](https://reports.phpmyadmin.net/reports/view/19829) on the phpmyadmin error reporting server.

flash1452:


nijel:
Sorry, but this fix is wrong - it fixes just once case and only for few variables.

PR #11963, opened 2016-02-15
Fixes #11865 partially

### LINKED ISSUE #11865: Support for shortcuts
The discussion here can be continued from #11862.

Achilles-96:
Implemented shortcuts
- k - Toggle console
- h - Go to home page
- s - Open settings
- d + s - Go to database structure (Provided you are in database related page)
- d + f - Search database (Provided you are in database related page)
- t + s - Go  to table structure (Provided you are in table related page)
- t + f - Search table (Provided you are in table related page)

Update: backspace - Takes you to older page.

### DIFF
--- a/js/shortcuts_handler.js
+++ b/js/shortcuts_handler.js
@@ -1,0 +1,106 @@
+/* vim: set expandtab sw=4 ts=4 sts=4: */
+/**
+ * @fileoverview    Handle shortcuts in various pages
+ * @name            Shortcuts handler
+ *
+ * @requires    jQuery
+ * @requires    jQueryUI
+ */
+
+/**
+ * Register key events on load
+ */
+$(document).ready(function() {
+    var databaseOp = false;
+    var tableOp = false;
+    var keyD = 68;
+    var keyT = 84;
+    var keyK = 75;
+    var keyS = 83;
+    var keyF = 70;
+    var keyE = 69;
+    var keyH = 72;
+    var keyC = 67;
+    var keyBackSpace = 8;
+    $(document).keyup(function(e) {
+        if( e.target.nodeName === 'INPUT' || e.target.nodeName === 'TEXTAREA' || e.target.nodeName === 'SELECT' ) {
+            return;
+        }
+
+        if(e.keyCode === keyD) {
+            setTimeout(function() {
+                databaseOp = false;
+            }, 2000);
+        }
+        else if(e.keyCode === keyT) {
+            setTimeout(function() {
+                tableOp = false;
+            }, 2000);
+        }
+    });
+    $(document).keydown(function(e) {
+        if ( e.ctrlKey && e.altKey && e.keyCode === keyC ) {
+            PMA_console.toggle();
+        }
+
+        if( e.target.nodeName === 'INPUT' || e.target.nodeName === 'TEXTAREA' || e.target.nodeName === 'SELECT' ) {
+            return;
+        }
+
+        var isTable;
+        var isDb;
+        if(e.keyCode === keyD) {
+            databaseOp = true;
+        }
+        else if(e.keyCode === keyK) {
+            PMA_console.toggle();
+        }
+        else if(e.keyCode === keyS) {
+            if(databaseOp === true) {
+                isTable = PMA_commonParams.get('table');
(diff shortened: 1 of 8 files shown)


zaverichintan:
Kindly use camel case

nijel:
I think console activation should be also moved from it's scrip to this one...

Achilles-96:
@nijel Do you mean toggle function?

nijel:
I mean just the keyboard handling for that: https://github.com/phpmyadmin/phpmyadmin/blob/master/js/console.js#L130

Achilles-96:
Any update on this?

nijel:
I think this looks good. The only missing piece now is the documentation. I'd probably put that to the user guide section...

Achilles-96:
Sure.

Achilles-96:
By the way, I forgot to mention that backspace is for going back to earlier page.

libraries/config.default.php
```diff
@@ -3133,3 +3133,10 @@
     'internal' => 50500,
     'human' => '5.5.0'
 );
+
+/**
+ * Disable shortcuts
+ *
+ * @global array $cfg['DisableShortcuts']
+ */
+$cfg['DisableShortcuts'] = false;
```
ibennetch: I think this might be better as DisableShortcutKeys, as it's a bit more readable.

ibennetch:
I'm not sure what was intended here, but to get to a combination page (for example, database structure), I must press both keys at the same time (`d`+`s`). What I was expecting was to press first `d`, let go, then press `s`. Which one is better is debatable, but my preference is the individual sequence.

By the way, I like this feature more than I thought I would and find it quite handy already.

js/shortcuts_handler.js
```diff
@@ -0,0 +1,120 @@
...
+ * @name            Shortcuts handler
+ *
+ * @requires    jQuery
+ * @requires    jQueryUI
+ */
+
+/**
+ * This function is for parsing URL and getting required parameter
+ */
+function getUrlParameter(sParam) {
```
zixtor: @Achilles-96 It seems that getUrlParameter is used just to get current database and table name values? then you could instead use PMA_commonParams.get('db') and PMA_commonParams.get('table').

zixtor:
Though I would be okay with either way of key pressing, subsequent or simultaneous; but I would personally prefer simultaneous as that would be similar to usual Ctrl, Alt usage .

zaverichintan:
I agree with @zixtor , as our users can easily adopt with ctrl alt system. As per my logic.

Achilles-96:
Yes, what if they press d and they forget and press s for settings and they go to structure of database. I don't think that's a very good idea.

js/shortcuts_handler.js
Achilles-96: Sure, I didn't know you can do that.

nijel:
There needs to be some (short) timeout. Try it on Github where I think it works within 2 seconds. If somebody forgets what he has pressed 2 seconds ago then it's different problem I think ;-).

Achilles-96:
Fine. Implementing that is not a problem but are we sure we want to use subsequent instead of simultaneous.

Achilles-96:
Please give an update.

zixtor:
I think it will be great if both ways of using the key combination work, subsequent as well as simultaneous. For example in github also both ways seem to work.

Achilles-96:
I agree. This is a better option.

Achilles-96:
Is there any more objection to be addressed regarding this PR?

nijel:
What is still missing is user documentation, otherwise I think it's good to merge.

Achilles-96:
@nijel But aren't docs supposed to be in localized_docs repository

nijel:
No, docs are in this repository, the other one (as it's name states) is for localized documentation only.

Achilles-96:
Done.

nijel:
Thanks! And my last request for you will be to improve commit message. It should describe the change in first, referencing fixed issue is nice to have, but this is something what will not help to identify reason for the change when browsing history...

Achilles-96:
Is this commit message fine.

nijel:
Thanks, it's better now.

Issue #11962, opened 2016-02-15
Switch dependency handling to composer


nijel:
After discussion on the mailing list, I think we should remove embedded PHP libraries and install those by composer.

This should include:

- sql-parser
- tcpdf
- phpseclib

Maybe the php-gettext could be used from it as well (there are few forks on packagist).

On creating release all those would be include in the tarball, so the tarball would contain complete code, but it would be easier for downstream distributions to detect deps and replace them with libraries.

zaverichintan:
Will this be for users too? 
If yes, them we can have multiple options with proper documentation.

PR #11976, opened 2016-02-17
Move dependencies handling to composer

nijel:
Implementation of #11962.

This is for review only right now, there are probably some missing pieces.
- [x] Replace tcpdf with package
- [x] Replace sql-parser with package
- [x] Replace phpseclib with package
- [x] Replace recaptcha lib with package
- [x] Use composer autoloader instead of own
- [x] Support for creating releases
- [x] Strip down vendor directory on creating release (we require only few parts of the libraries)
- [x] Add tests for stripped down releases if they still work correcty
- [x] Documentation how to install phpMyAdmin from Git


nijel:
I think this is ready to merge, reviews welcome :-)

ibennetch:
Can you step me through how this will work since I'm a bit new to the idea?

For developers testing different phpMyAdmin versions, I guess the proper Composer dependencies (sql-parser in particular) would be installed by running Composer after git checkout, which would ensure we have the correct version of the dependencies.

For users working off of the git version, I guess they'll need to occasionally run the Composer update function to make sure they are using the correct version of the libraries. If my understanding is right, we might want to build version checks in to the code to warn users to update their Composer dependencies  in case of a version mismatch.

Finally, for official releases, I suppose the release manager will use Composer to make sure the proper dependencies are installed, then build the release which will include those libraries directly, so that end users do not end up with a new dependency on Composer to install the official tarball.

Am I close?

nijel:
Yes, that pretty much describes it. That reminds me that this should be documented :-).

I'm quite unsure about version check, maybe there is way to ask composer if versions do match, I will look at it.

As for release manager, you just need to have composer installed, the create-release script still does all the hard work of creating proper tarballs :-).

PR #11980, opened 2016-02-18
Fix #11979 DECLARE not accepted as valid SQL

### LINKED ISSUE #11979: DECLARE not accepted as valid SQL
It appears as if DECLARE is not recognized as valid SQL. Related: #11776 and #11191

This is valid at the MySQL command line:
```
DELIMITER |
CREATE PROCEDURE SPTEST()
BEGIN
  DECLARE V INT UNSIGNED DEFAULT 1;
END
|
```

But trying 
`DECLARE V INT UNSIGNED DEFAULT 1;`
in the Routines window gives an immediate red x syntax error and this message on submission:
One or more errors have occurred while processing your request:

> The following query has failed: "CREATE PROCEDURE `ads`() NOT DETERMINISTIC NO SQL SQL SECURITY DEFINER DECLARE V INT UNSIGNED DEFAULT 1; "
>
> MySQL said: #1064 - You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'DECLARE V INT UNSIGNED DEFAULT 1' at line 1

Tested with QA_4_5 and master.

madhuracj:
This is one way of handling the bug. Here I try to pass the complete SQL for the creation of routine/trigger/event to the Linter by adding dummy values for the options.

### DIFF
--- a/js/codemirror/addon/lint/sql-lint.js
+++ b/js/codemirror/addon/lint/sql-lint.js
@@ -31,7 +31,8 @@ CodeMirror.sqlLint = function(text, updateLinting, options, cm) {
         data: {
             sql_query: text,
             token: PMA_commonParams.get('token'),
-            server: PMA_commonParams.get('server')
+            server: PMA_commonParams.get('server'),
+            options: options.lintOptions,
         },
         success: handleResponse
     });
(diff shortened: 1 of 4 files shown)


nijel:
Are parse errors reported correctly after this change? (Validating another string than is in the input)

madhuracj:
The body of a stored procedure does not seem to be parsed at all.
```
CREATE PROCEDURE SPTEST()
BEGIN
  FROM a SELECT *;
END
```
Above SQL on the SQL window does not generate any Lint errors.

nijel:
True,  then this is definitely safe...

PR #11992, opened 2016-02-19
Rephrase and reposition the error report submission form

atul516:
Signed-off-by: Atul Pratap Singh atulpratapsingh05@gmail.com


nijel:
I think we should target this to 4.6.0 as well...

zixtor:
Committed to QA_4_6 and merged to master.

nijel:
Thanks!

Issue #11990, opened 2016-02-19
 Refactor url parmeters generating


nijel:
The ``./libraries/url_generating.lib.php`` is currently using functions and thus can not be used with autoloading, which would simplify the code.

The task is to convert this file to object with static method and change all it's usages to this new way.

I think the best approach is to create URL class which would provide such static methods based on current functions.

007durgesh219:
@nijel till we resolve issues with the parser I can't work on the tracker so, I want to work on this.

nijel:
Sure, go ahead, nobody expressed interest so far.

007durgesh219:
@nijel current format of the function names is ```PMA_URL_<name>```. I think we don't need URL in the prefix because now we are moving functions to URL class, So, should I change it to ```PMA_<name>```

nijel:
I think URL::name could be enough, the PMA prefix is not needed as well.

007durgesh219:
Also, Name of the file. It will be more convenient to change it to URL.php

nijel:
Definitely (that's mostly needed for autoloading).

007durgesh219:
@nijel I moved all functions to the class But Now I am getting class not found error.  Do I have to put ```use PMA\libraries\URL``` everywhere? Kindly help

nijel:
Either reference it with full path PMA\libraries\URL... or add use as you've written.

007durgesh219:
OK, I think use will be better, Don't have to write full path everywhere.  Also, I think now we don't need ```include 'libraries\URL.php'```. Should I remove it from everywhere

nijel:
Yes, the goal is to get rid of explicit include.

007durgesh219:
@nijel Please review my PR and let me know if everything is as you were expecting

madhuracj:
Implemented with https://github.com/phpmyadmin/phpmyadmin/pull/11996

Issue #11988, opened 2016-02-19
Refactor javascript escaping


nijel:
The ``./libraries/js_escape.lib.php`` is currently using functions and thus can not be used with autoloading, which would simplify the code.

The task is to convert this file to object with static method and change all it's usages to this new way.

I think the best approach is to create Sanitize class which would provide such static methods together with text sanitisation (see #11989).

nijel:
Note if working on this issue, please coordinate with #11989.

wadhwasahil:
I am working on it.

madhuracj:
Implemented with https://github.com/phpmyadmin/phpmyadmin/pull/11996

PR #11994, opened 2016-02-19
Refactored url parmeters generating, Issue #11990

007durgesh219:
Before submitting pull request, please check that every commit:
- [x] Has proper Signed-Off-By
- [x] Has commit message which describes it
- [x] Is needed on it's own, if you have just minor fixes to previous commits, you can squash them

Signed-off-by: Durgesh 007durgesh219@gmail.com

PR #11996, opened 2016-02-20
Issue#11989, #11988, #11990

007durgesh219:


nijel:
Is the patch here different from one in #11994?

libraries/URL.php
```diff
@@ -0,0 +1,255 @@
...
+/**
+ * URL/hidden inputs generating.
+ *
+ * @package PhpMyAdmin
+ */
+namespace PMA\libraries;
+
+if (! defined('PHPMYADMIN')) {
+    exit;
+}
```
nijel: This is not needed here.

test/libraries/PMA_generateCommonUrl_test.php
```diff
@@ -1,18 +1,17 @@
 <?php
 /* vim: set expandtab sw=4 ts=4 sts=4: */
 /**
- * tests for PMA_URL_getCommon()
+ * tests for URL::getCommon()
```
nijel: I think this file should be renamed to test/classes/URL.php

test/libraries/PMA_js_escape_test.php
```diff
@@ -9,7 +9,7 @@
 /*
  * Include to test.
  */
-require_once 'libraries/js_escape.lib.php';
+use PMA\libraries\Sanitize;
```
nijel: I think this file should be renamed to test/classes/Sanitize.php

test/libraries/PMA_sanitize_test.php
```diff
@@ -1,20 +1,17 @@
 <?php
 /* vim: set expandtab sw=4 ts=4 sts=4: */
 /**
- * tests for PMA_sanitize()
+ * tests for Sanitize::sanitize()
```
nijel: I think this file should be renamed to test/classes/Sanitize.php (and merged with js based tests)

nijel:
This looks good, but I've some minor comments, see above.

007durgesh219:
Thanks @nijel  I will update PR shortly

nijel:
PS: There seems to be also merge conflict meanwhile, it would be great if you could rebase it on current master.

007durgesh219:
@nijel I've done required changes

nijel:
What you've not yet done is suggested file renaming ;-)

test/classes/URL.php
```diff
@@ -1,18 +1,17 @@
 <?php
```
007durgesh219: changed test/libraries/PMA_generateCommonUrl_test.php → test/classes/URL.php

test/classes/Sanitize.php
```diff
@@ -1,20 +1,17 @@
 <?php
```
007durgesh219: this one also  test/libraries/PMA_sanitize_test.php → test/classes/Sanitize.php

007durgesh219:
@nijel I think you suggested only those two. also, I've merged PMA_js_escape_test to Sanitize.php .  Is there something else I am missing??

007durgesh219:
This seems a bug in the github. If we change only file name then it won't show it as outdated diff

nijel:
Okay, I somehow trusted GitHub in this, sorry. Merging the changes right now...

Issue #12004, opened 2016-02-21
Templating getWithSelected() and getCheckbox() in libraires/Util.php


zixtor:
In libraires/Util.php, make templates from getWithSelected and getCheckbox functions which return HTML and are used in various scripts and other templates as well.
See https://wiki.phpmyadmin.net/pma/Templates
Also, here is an example commit doing similar templating: https://github.com/phpmyadmin/phpmyadmin/commit/39761819c31b89c93bd312a9454178f9f9abb041

007durgesh219:
I'd like to work on it

flash1452:
@zixtor i had like to work on this issue.

flash1452:
@007durgesh219 Either of us should work on issue #12005

zixtor:
Merged https://github.com/phpmyadmin/phpmyadmin/pull/12008

PR #12008, opened 2016-02-21
Templating getWithSelected() and getCheckbox(), Issue #12004

007durgesh219:

Signed-off-by: Durgesh 007durgesh219@gmail.com


007durgesh219:
This fails because of following.  I think here we need to change the tests.
```
[exec] --- Expected
     [exec] +++ Actual
     [exec] @@ @@
     [exec] -'<input type="checkbox" name="test_display_html_checkbox"
     [exec] -                    id="test_display_html_checkbox"
     [exec] -        <label              for="test_display_html_checkbox">text_label_for_checkbox</label>
     [exec] -'
     [exec] +'<input type="checkbox" name="test_display_html_checkbox" id="test_display_html_checkbox" /><label for="test_display_html_checkbox">text_label_for_checkbox</label>'
```

007durgesh219:
@zixtor Kindly help

zixtor:
Indeed, the tests might be failing due of some indentation/formatting changes, and need to be updated accordingly.

007durgesh219:
Thanks, I will update test and PR shortly

zixtor:
You may also choose to assert for something less than full html tags.

007durgesh219:
So, Can I use DOMDocument for asserting only attributes in which we are interested so that in future if we change the position of attributes it won't fail??. Or please suggest something similar to assert only attributes.

007durgesh219:
Or what about assertTag.  [https://phpunit.de/manual/3.7/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.assertions.assertTag](url)

zixtor:
Yes, we should tend to avoid recurring of such failures, asserting for attributes and their values as essential in the generated html seems fine to me. Using DOMDocument for that seems promising to me at first look, I think it would also validate the generated html? Also, assertTag doesn't seem to be a reliable choice and is deprecated, see https://github.com/sebastianbergmann/phpunit/issues/1292

007durgesh219:
OK, In that case using assertEqualXMLStructure seems a better option to me [https://phpunit.de/manual/3.7/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.assertions.assertEqualXMLStructure](url)  Also, domDocument provide method to validate HTML [http://php.net/manual/en/domdocument.validate.php](url)

zixtor:
That seems good and sufficient to me as well, please feel free to proceed.

007durgesh219:
Analyzing composer.json/composer.lock files to determine dependencies of the root package. 
I didn't get why it failed here. Kindly help

007durgesh219:
@zixtor Kindly look into this issue and help me figure out what I did wrong here

zixtor:
I think Scrutinizer error is not due to your changes, and the tests seem to be passing now.

templates/getCheckbox.phtml
```diff
@@ -0,0 +1,13 @@
...
+        <?php }
+        if (isset($checked) && $checked) {      ?>
+            checked="checked"
+        <?php }
+        if (isset($onclick) && $onclick) {      ?>
+            class="autosubmit"
+        <?php } ?> />
+<label  <?php if (isset($html_field_id)) {      ?>
+            for="<?php echo $html_field_id ; }  ?>">
+        <?php echo $label ; ?> </label>
```
zixtor: This file should be named 'checkbox.phtml'

007durgesh219:
OK, Thanks, Please review and let me know if any changes required. I've changed test also, Now we don't have to maintain test if  the relative position of attributes changed.

templates/getWithSelected.phtml
```diff
@@ -0,0 +1,7 @@
+<img class="selectallarrow" src="<?php echo $pmaThemeImage . 'arrow_' . $text_dir . '.png' ; ?>"
+    width="38" height="22" alt="<?php echo __('With selected:') ; ?>" />
+<input type="checkbox" id="<?php echo $formName . '_checkall' ;   ?>" class="checkall_box"
+    title="<?php echo __('Check all') ;      ?>" />
+<label for="<?php echo $formName . '_checkall' . __('Check all') ;?>" </label>
+<i style="margin-left: 2em">
+    <?php echo __('With selected:') ;                             ?> </i>
```
zixtor: For this, maybe 'select_all.phtml'

templates/getCheckbox.phtml
```diff
@@ -0,0 +1,13 @@
+<input type="checkbox" name="<?php echo $html_field_name; ?>"
+        <?php if (isset($html_field_id)) {      ?>
+            id="<?php echo $html_field_id ;     ?>"
```
zixtor: use <?= in all places where you would do <?php echo

templates/getCheckbox.phtml
```diff
@@ -0,0 +1,13 @@
+<input type="checkbox" name="<?php echo $html_field_name; ?>"
+        <?php if (isset($html_field_id)) {      ?>
```
zixtor: avoid curly braces, see http://php.net/manual/en/control-structures.alternative-syntax.php

test/libraries/common/PMA_getCheckbox_test.php
```diff
@@ -30,11 +30,13 @@ function testGetCheckbox()
...
-            PMA\libraries\Util::getCheckbox($name, $label, false, false, $name),
-            '<input type="checkbox" name="' . $name . '" id="' . $name
+        // assertXmlStringEqualsXmlString require both inputs to be a valid xml string
+        // dummy <root> tag will make input a valid xml string
+        $this->assertXmlStringEqualsXmlString(
+            '<root>' . PMA\libraries\Util::getCheckbox($name, $label, false, false, $name) . '</root>',
+            '<root> <input type="checkbox" name="' . $name . '" id="' . $name
             . '" /><label for="' . $name . '">' . $label
-            . '</label>'
+            . '</label> </root>'
```
zixtor: This seems better, thanks!

zixtor:
As an additional improvement, the functions get getWithSelected() and getCheckbox() can be removed now, and in all places where they are used, the extracted templates can be called instead.

007durgesh219:
OK, will do all those changes and update PR

007durgesh219:
Also, I think it will be more convenient to change test file names, Like test/libraries/common/PMA_getCheckbox_test.php -> test/libraries/common/checkbox_test.php and for other one also

zixtor:
Yes, makes sense.

007durgesh219:
@zixtor I've done all changes you suggested. Please review PR

zixtor:
Thanks @007durgesh219 for the contribution!

Issue #11476, opened 2015-09-13
Errors instead of git info when PHP has no gzip support


ibennetch:
When PHP is compiled without gzip support, I get this instead of the git revision information:
<pre>
Fatal error: Uncaught Error: Call to undefined function gzuncompress() in /var/www/pma-dev/fork/libraries/Config.class.php:510
Stack trace:
#0 /var/www/pma-dev/fork/libraries/display_git_revision.lib.php(26): PMA_Config->checkGitRevision()
#1 /var/www/pma-dev/fork/index.php(92): PMA_printGitRevision()
#2 {main} thrown in /var/www/pma-dev/fork/libraries/Config.class.php on line 510
</pre>

ibennetch:
I've searched for a few workarounds, but I think the only way to fix this is to test whether PHP has gzip support and not show the detailed git revision data when we're unable to `gzuncompress`.

lem9:
Makes sense.

lem9:
@ibennetch Are you working on this?

ibennetch:
Unfortunately I haven't had a chance to finish my work yet, I'm unassigning myself for now.

devenbansod:
Do you think 
`((function_exists('gzuncompress') && ini_get('zlib.output_compression'))` 
work for this ?

ibennetch:
I don't know that the `init_get` portion is needed; since zlib provides the function `gzuncompress` isn't it enough to just test for that?

madhuracj:
`zlib.output_compressio` serves a different purpose and unnecessary here. See http://php.net/manual/en/zlib.configuration.php#ini.zlib.output-compression

PR #12037, opened 2016-02-25
Fix parse git data without gz support (bug 12030)

muhakh:


nijel:
Sorry, it seems this error has been already fixed in 9117700 (issue #11476). The if condition is already there, so your change does nothing...

Issue #12006, opened 2016-02-21
Templating getRadioFields() and getDropdown() in libraries/Util.php


zixtor:
In libraires/Util.php, make templates from getRadioFields and getDropdown functions which returns HTML and are used in various scripts and other templates as well.
See https://wiki.phpmyadmin.net/pma/Templates
Also, here is an example commit doing similar templating: https://github.com/phpmyadmin/phpmyadmin/commit/39761819c31b89c93bd312a9454178f9f9abb041

flash1452:
I had like to work on this issue.

flash1452:
PMA_GetRadioFieldsTest::testGetRadioFieldsWithoutBR

     [exec] Failed asserting that two strings are equal.

     [exec] --- Expected

     [exec] +++ Actual

     [exec] @@ @@

     [exec] -'                            <input type="radio" name="test_display_radio"

     [exec] -    id="test_display_radio_value_1"

     [exec] -    value="value_1"

     [exec] -    /><label for="test_display_radio_value_1">choice_1</label>

     [exec] -            

     [exec] -                <input type="radio" name="test_display_radio"

     [exec] -    id="test_display_radio_value&_&lt;2&gt;"

     [exec] -    value="value&amp;_&amp;lt;2&amp;gt;"

     [exec] -    /><label for="test_display_radio_value&_&lt;2&gt;">choice_2</label>

     [exec] -            

     [exec] +'<input type="radio" name="test_display_radio" id="test_display_radio_value_1" value="value_1" />

     [exec] +<label for="test_display_radio_value_1">choice_1</label>

     [exec] +<input type="radio" name="test_display_radio" id="test_display_radio_value&_&lt;2&gt;" value="value&amp;_&amp;lt;2&amp;gt;" />

     [exec] +<label for="test_display_radio_value&_&lt;2&gt;">choice_2</label>

     [exec]  '

     [exec] 

     [exec] /home/travis/build/flash1452/phpmyadmin/test/libraries/common/PMA_getRadioFields_test.php:168

This is the error am getting during build.Can someone please point out the error.

poush:
@flash1452  I would like to do this if you are no longer working on it.

flash1452:
@poush I am working on it and i will open a PR soon

flash1452:
@zixtor Please review my PR

vishpandey2014:
@zixtor I would like to work on this issue if noone is working on it.

ibennetch:
[...]

PR #12043, opened 2016-02-28
Drop down radio field templates refactoring. Issue #12006

flash1452:

Signed-Off-By: Vishal Pandey vishapndey2014@gmail.com


flash1452:
Signed-Off-By: Vishal Pandey vishapndey2014@gmail.com

zixtor:
- Signed-Off-By is missing from your commit, please use -s while committing.
- It still contains not needed merge commits.

libraries/DbSearch.php
```diff
@@ -377,7 +377,8 @@ private function _getResultsRow($each_table, $newsearchsqls, $odd_row, $res_cnt)
      */
     public function getSelectionForm()
     {
-        $html_output = '<a id="db_search"></a>';
+        $html_output  ='<main>';
```
zixtor: Why adding main tag here?

flash1452: I had to add main tag because I think in order to render a XML script,all tags should be enclosed within a head/root tag that is parent to all tag.

zixtor: I don't see its applicability here. If you need that for testing, it should be added in the test itself.

nijel:
I don't se this fit for merging at least due to <main> and <root> tags.

PR #12113, opened 2016-03-20
Add new server variables

devenbansod:
Added new server variables for version 5.7.
Added a list of Dynamically editable variables.
Fixed the broken byte-formatting.

### DIFF
--- a/libraries/controllers/server/ServerVariablesController.php
+++ b/libraries/controllers/server/ServerVariablesController.php
@@ -205,15 +205,23 @@ class ServerVariablesController extends Controller
                 . Util::sqlAddSlashes($_REQUEST['varName'])
                 . '";', 'NUM'
             );
-            $this->response->addJSON(
-                'variable',
-                htmlspecialchars(
-                    $this->_formatVariable(
-                        $_REQUEST['varName'],
-                        $varValue[1]
-                    )
-                )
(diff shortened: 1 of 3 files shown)


libraries/controllers/server/ServerVariablesController.php
```diff
@@ -301,8 +319,8 @@ private function _getHtmlForServerVariables($serverVars, $serverVarsSession)
     private function _getHtmlForServerVariablesItems(
         $serverVars, $serverVarsSession
     ) {
-        // list of static system variables
-        $static_variables = $this->_getStaticSystemVariables();
+        // list of dynamic (i.e. editable) system variables
+        $dynamic_variables = $this->_getDynamicSystemVariables();
```
madhuracj: The issue of introducing a list of dynamic (editable) variables is that all the other variables are considered static. This is troublesome when MySQL adds new dynamic variables and phpMyAdmin is not (yet) updated with those. Further, MariaDB might add their own dynamic variables and phpMyAdmin dynamic variable list might not (yet) include them. So, IMO, it's best to keep a list of static variables and consider the rest as dynamic variables. In the end, if a variable is not editable the database server will throw an error, so the user will be informed anyway.

devenbansod: Thanks for pointing that out @madhuracj, I did not realise this before, but it makes sense. 
I will update the PR to revert those changes.

PR #12127, opened 2016-03-25
Fixup for missing glossary terms

ibennetch:
When compiling the documentation for a release, this should fix notices about missing glossary terms for `mod_proxy_fcgi` (which I added) and `Authorization` (which I removed).


nijel:
Thanks.

PR #12128, opened 2016-03-25
Update location of tcpdf library in FAQ

ibennetch:
tcpdf moves for 4.7.0, I think I got the new location correct.


nijel:
Thanks.

PR #12136, opened 2016-03-27
Fix offering JSON datatype in incompatible MySQL versions

### LINKED ISSUE #12135: JSON datatype is only valid starting with MySQL 5.7.8
We show the JSON data type in the create table dropdown even on servers where it is not valid. I believe we have been hiding invalid data types so we should also hide JSON for MySQL older than 5.7.8.

ibennetch:
Fixes #12135


madhuracj:
Looks good.

ibennetch:
Thanks for the review

PR #12147, opened 2016-03-31
Fix #12143 Cannot login with certain password

### LINKED ISSUE #12143: Cannot login with certain password
phpMyAdmin is treating password 0 (one zero) as empty password so when AllowNoPassword is set (the default), you cannot login.

madhuracj:
This fixes the bug. However, I wonder whether this commit has any security implications. Further discussion is appreciated before merging.

### DIFF
--- a/libraries/common.inc.php
+++ b/libraries/common.inc.php
@@ -788,7 +788,7 @@ if (! defined('PMA_MINIMUM_COMMON')) {
 
         // is a login without password allowed?
         if (! $cfg['Server']['AllowNoPassword']
-            && $cfg['Server']['password'] == ''
+            && $cfg['Server']['password'] === ''
         ) {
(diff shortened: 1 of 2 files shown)


nijel:
I think it's correct, just the testsuite needs to be adjusted.

The only thing I wonder about is \xff(blank) value, maybe it's needed to avoid empty cookie string? It's there since 11b0c087...

ibennetch:
That was my thought as well, that `\xff(blank)` was to avoid an empty string, but I'm not very clear about it either.

I'm not 100% confident of my assessment of part of the old code, but I also think this looks okay.

nijel:
It was probably only problem with previous ciphers. With current block one, it always produces some content as it's padded...

nijel:
Merged and adjusted tests to not fail.

PR #12157, opened 2016-04-05
Simplify SSL setup

nijel:
Simplify setup to follow current PHP documentation:
- the MYSQLI_CLIENT_SSL flag should not be needed
- there is mysqli_option flag to turn of certificate verification
- remove check for defined constants, all should be there in PHP
  versions we support

This is for testing only right now.


ibennetch:
This works fine in my testing. Thanks.

nijel:
Thanks @ibennetch for testing, I'm merging this.

Issue #12129, opened 2016-03-25
Slow database list on 4.6


JFDionne:
Hi, when I click on database at the left panel, it take like 10-12 second to load if I have "SkipLockedTables" to true, when I set it to false it take like 2-3 second to load

### Server configuration
**Operating system**:

**Web server:**
    Apache/2.2.15 (Scientific Linux)
    Database client version: libmysql - mysqlnd 5.0.11-dev - 20120503 - $Id: 76b08b24596e12d4553bd41fc93cccd5bac2fe7a $
    PHP extension: mysqliDocumentation curlDocumentation mbstringDocumentation
    PHP version: 5.6.19

nijel:
SkipLockedTables needs additional SQL query for every database, so it indeed slows down things. Not sure if there is way to improve it...

nijel:
...another problem that we do all this twice.

nijel:
Should be better in git. BTW: Did you really need SkipLockedTables? I'm wondering if this configuration is really needed these days...

JFDionne:
Hmm good question, but sometime when I "lock" a table with a big update or delete of data I can't use anymore phpmyadmin on other table, it "wait" for the lock table to be unlocked, no?

nijel:
You mean using two phpMyAdmin windows? I don't think it should block (it did block in past due to session locking). I'm just wondering what usecase it is supposed to solve as the code comes from MySQL 3.23.30 times :-).

JFDionne:
Hmm, 
Lets say I do a locking table from the mysql gui, what happend if a table is locked and you try to go on the table list of phpmyadmin?

nijel:
It seems to work without any problems in my tests (I'm using MariaDB 10.0.24).

nijel:
The only thing which seems to be blocked in insertion to the locked tables or changing structure, anything else works just fine.

nijel:
I've create PR to remove this setting: https://github.com/phpmyadmin/phpmyadmin/pull/12163

PR #12163, opened 2016-04-06
Remove SkipLockedTables settings

nijel:
**This needs further testing before merging!**

I've tested with following servers and there doesn't seem to be any problem with listing locked tables, so this settings just add code complexity and the only visible effect is that:
1. Loading of tables is way slower due to getting status for every non locked table separately (see #12129)
2. Locked tables do not show any stats.

When this setting is not enabled, the stats for locked tables are shown without any visible blocking.

I've tested with:
- MariaDB 10.0.24
- MariaDB 10.0.22
- MySQL 5.5.46

### DIFF
--- a/libraries/Util.php
+++ b/libraries/Util.php
@@ -4644,105 +4644,86 @@ class Util
         /**
          * information about tables in db
          */
-        $tables = array();
-
         $tooltip_truename = array();
         $tooltip_aliasname = array();
 
-        // Special speedup for newer MySQL Versions (in 4.0 format changed)
-        if (true === $cfg['SkipLockedTables']) {
-            $db_info_result = $GLOBALS['dbi']->query(
-                'SHOW OPEN TABLES FROM ' . Util::backquote($db) . ';'
+        // Set some sorting defaults
+        $sort = 'Name';
+        $sort_order = 'ASC';
+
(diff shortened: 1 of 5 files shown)


ibennetch:
With your patch, I get some different behavior that I don't think is desired. I'll admit that locking tables isn't something I ever do so my testing may be a bit off.

I'm testing with MySQL 5.5.47-0+deb8u1.

From another connection I lock a table (`lock tables foo write`), then try to browse to that database.

Without this path and with `$cfg['SkipLockedTables']` set to either true or false and starting from the main page, I'm able to load the Structure tab of the database. Of course, I can't do anything with the table, but I expect that.

After applying this pull request, loading the database Structure tab waits/hangs, so I can't even work in another table of that same database.

However, if I lock the table with `lock tables foo read` then all three methods (with SkipLockedTables true, false, and this pull request) all seem to behave the same.

nijel:
Thanks for feedback, however it seems a bit strange. All this patch does is to remove branch where ``$cfg['SkipLockedTables']`` is true, so the outcome you describe is not expected at all. It however might indicate that the condition when listing blocks might be more complex...

nijel:
I was able to reproduce this now as well, so it seems this setting is still needed in some cases...

PR #12170, opened 2016-04-07
Remove undocumented caching for table status

nijel:
This feature is not documented, not in the default config, so probably
not used at all.

This was introduced in revert commit
19432a02cc218e7469dc8c123f73b10ab47c872c, so I believe it was not
intentional.

Alternatively we could properly document it, so feedback is welcome. Especially from @madhuracj :-).


nijel:
As there is no negative feedback, I'm merging this.

Issue #12146, opened 2016-03-31
Issue connecting to MySQL using SSL


LucasNN:
### Steps to reproduce
1) Configure the SSL connection in the config.inc.php

```
/**
 * MySQL hostname or IP address
 *
 * @global string $cfg['Servers'][$i]['host']
 */
$cfg['Servers'][$i]['host'] = '104.XXX.XXX.XX';

/**
 * MySQL port - leave blank for default port
 *
 * @global string $cfg['Servers'][$i]['port']
 */
$cfg['Servers'][$i]['port'] = '3306';

/**
 * MySQL extension
 *
 * @global string $cfg['Servers'][$i]['extension']
 */
$cfg['Servers'][$i]['extension'] = 'mysqli';

/**
 * Path to the socket - leave blank for default socket
 *
 * @global string $cfg['Servers'][$i]['socket']
 */
$cfg['Servers'][$i]['socket'] = '';

/**
 * Use SSL for connecting to MySQL server?
 *
 * @global boolean $cfg['Servers'][$i]['ssl']
 */
$cfg['Servers'][$i]['ssl'] = true;

/**
 * Path to the key file when using SSL for connecting to the MySQL server
 *
 * @global string $cfg['Servers'][$i]['ssl_key']
 */
$cfg['Servers'][$i]['ssl_key'] = '/opt/bitnami/apps/phpmyadmin/conf/certs/google-cloud-sql-client-key.pem';

/**
 * Path to the cert file when using SSL for connecting to the MySQL server
 *
 * @global string $cfg['Servers'][$i]['ssl_cert']
 */
$cfg['Servers'][$i]['ssl_cert'] = '/opt/bitnami/apps/phpmyadmin/conf/certs/google-cloud-sql-client-cert.pem';

/**
 * Path to the CA file when using SSL for connecting to the MySQL server
 *
 * @global string $cfg['Servers'][$i]['ssl_ca']
 */
$cfg['Servers'][$i]['ssl_ca'] = '/opt/bitnami/apps/phpmyadmin/conf/certs/google-cloud-sql-server-ca.pem';

/**
 * Directory containing trusted SSL CA certificates in PEM format
 *
 * @global string $cfg['Servers'][$i]['ssl_ca_path']
 */
$cfg['Servers'][$i]['ssl_ca_path'] = '/opt/bitnami/apps/phpmyadmin/conf/certs/';

/**
 * List of allowable ciphers for SSL connections to the MySQL server
 *
[...]

PR #12193, opened 2016-04-18
Do not hide errors from database connection attempts

nijel:
Without these errors it's impossible to figure out what went wrong in
the SSL negotiation. All you get is error 2002 without any means to
diagnose and all important information is issued as PHP warnings.

Issue #12146

I'm not sure this is best approach as the errors look quite ugly in the end, but we really need to find way to pass the information to the user. What is however annoying is that the SSL setup error leads to at least four warnings:

phpmyadmin_-_2016-04-18_12 25 45


nijel:
I think the reporting is now reasonably verbose:

phpmyadmin_-_2016-04-18_13 30 24

PR #12235, opened 2016-05-07
Fixes #11983 pop out SQL window

### LINKED ISSUE #11983: Feature-Request: Pop-Out SQL Window
In earlier versions of phpmyadmin it was possible to pop out the SQL window and exectute commands in it. As I found out this feature was replaced by the Console Window at the bottom.

When using multiple screens it would be nice the old windows was more practical as you could move it to another screen. Opening a whole new window takes too much space however.

I don't really see why the two features are mutually exclusive, would it be possible to bring the popup window back or (probably even better) have a button to open the console in a new window?

Achilles-96:

Signed-off-by: Achilles-96 raghuram.vadapalli@research.iiit.ac.in


ibennetch:
This shows promise.

I'm not immediately sure what complaints Scrutinizer and codecov have, I'll have to look in to that.

I doubt we need to have the console link at the bottom of the pop-out window and wonder if it's possible to hide that. Further, I don't see that the Console is updated immediately with the new query.

Finally, when submitting a query, I seem to be taken to a new browser tab instead of the existing tab my full screen PMA window was using. If I run a lot of queries from the pop up SQL box, I get a lot of browser tabs which isn't great. Is there a way to prevent that?

Achilles-96:
I thought we may not want to change the current content of full screen window. But since more queries result in too many tabs, this is a problem. How about opening a new tab for the first query and using the same for any of the following queries.

Achilles-96:
Any update on this?

nijel:
Just tried that and I have few remarks:

- the query window is not properly sized to contain the content
- it doesn't feel integrated with rest much as it submits query into new tab
- the SQL console doesn't keep executed queries (because whole page is reloaded every time)
- there is SQL console in the query window which makes little sense

In summary what I'd like to see is the window to communicate with main window by javascript, submit queries by AJAX, keep history, currently selected database and so on.

Achilles-96:
* By size, do you mean the height?
* I tried to replicate the functionality which we had earlier. The query is submitted to a new tab in that version. So, I just followed that. I will modify that since you pointed out.
* I removed console at bottom now.

nijel:
I meant height and width, at least for me it showed just part of the textarea and I had to resize the window to see the submit button.

As for the submission to new tab, I don't think this is expected behavior. I want the results to appear in the same window where I opened the query window. What might be used is detect if there are any pending changes in that window (we already have this to prevent accidental closing) and warn user in such case.

nijel:
Closing for now as this is not ready to merge. Once you have updated the code, please open new pull request, thanks.

PR #12236, opened 2016-05-07
fixed use zh_CN lead to error _lang_failed_cookie or _lang_failed_cfg or _lang_failed_request

pkujhd:
i use phpMyadmin 4.6.1 on CentOS6.4, in setup/index.php i can't change language to zh_CN, it almost jump to en, i check the source code, find, $lang is be to lower in function availableLanguages(), but in function getCurrentLanguage() not use strtolower, it  will be cause error.

### DIFF
--- a/libraries/LanguageManager.php
+++ b/libraries/LanguageManager.php
@@ -827,7 +827,7 @@ class LanguageManager
         if ($accepted_languages && false === mb_strpos($accepted_languages, '<')) {
             foreach (explode(',', $accepted_languages) as $header) {
                 foreach ($langs as $language) {
-                    if ($language->matchesAcceptLanguage($header)) {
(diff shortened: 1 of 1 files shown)


feisuzhu:
Why closing this PR?
I've encountered the same issue and made exactly the same fix, and found here.

pkujhd:
it can not be merged, so i closed it, these days i am busying on work.. no time pull a new request

ibennetch:
Also, I believe this has already been fixed for the upcoming 4.6.2 release (you can test if you'd like by downloading the master branch from the git repository or downloading the .zip directly from https://github.com/phpmyadmin/phpmyadmin/archive/master.zip).

@pkujhd thanks for your pull request.

PR #12250, opened 2016-05-11
Make ThemeManager a singleton instead of storing it in the session

nijel:
I believe this is better approach and in my tests there is no performance difference between these two usages. Storing the data in session can cause problems on restoring in case API of the object has changed meanwhile.

Please review, if no negative feedback appears I'm going to merge this in a week or so.


ibennetch:
I haven't yet tested but the idea makes sense and code looks good. I don't see a strong reason to use the session so this should be fine.

nijel:
The motivation probably was to avoid some checks with every requests. But it shows that restoring object from session takes about same time as these checks...

PR #12270, opened 2016-05-21
Error message improved when user tries to rename database to same name.

ghost:
Signed-off-by: Gregory Buhler webmaster@gregorybuhler.com

### DIFF
--- a/js/db_operations.js
+++ b/js/db_operations.js
@@ -41,7 +41,7 @@ AJAX.registerOnload('db_operations.js', function () {
         var new_db_name = $('#new_db_name').val();
 
         if (new_db_name == old_db_name) {
-            PMA_ajaxShowMessage(PMA_messages.strDropDatabaseStrongWarning);
+            PMA_ajaxShowMessage(PMA_messages.strDatabaseRenameToSameName, false, "error");
             return false;
         }
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,7 @@ phpMyAdmin - ChangeLog
 ======================
 
 4.7.0.0 (not yet released)
+- patch #12233 [Display] Improve message when renaming database to same name
 - issue #11924 Remove caching of server information
 - issue #6146  Log authentication attempts to syslog
 - issue #11981 Remove support for Swekey authentication
--- a/js/functions.js
+++ b/js/functions.js
@@ -2209,11 +2209,17 @@ function PMA_updateCode($base, htmlValue, rawValue)
  * @param mixed   timeout     number of milliseconds for the message to be visible
  *                              optional, defaults to 5000. If set to 'false', the
  *                              notification will never disappear
+ * @return string type        string to dictate the type of message shown.
+ *                              optional, defaults to normal notification.
+ *                              If set to 'error', the notification will show message
+ *                              with red background.
+ *                              If set to 'success', the notification will show with
+ *                              a green background.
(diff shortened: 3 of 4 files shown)


js/db_operations.js
```diff
@@ -41,7 +41,7 @@ AJAX.registerOnload('db_operations.js', function () {
         var new_db_name = $('#new_db_name').val();
 
         if (new_db_name == old_db_name) {
-            PMA_ajaxShowMessage(PMA_messages.strDropDatabaseStrongWarning);
+            PMA_ajaxShowMessage("<div class=\"error\"><img src=\"./themes/dot.gif\" title alt class=\"icon ic_s_error\">Cannot rename database to the same name. Change the name and try again</div>", false);
```
ibennetch: Thanks for your pull request.

We'll have to make this string translatable by adding a new message like strRenameToSameName or something silly like that to js/messages.php.

Further, I'm not sure embedding style code in the function call here is the best way to improve the look. I could be wrong, for instance the same thing done in js/import.js:77, but my gut reaction is that there might be a better way.

Gregory-Buhler: Alright, give me a little bit and I'll add that message to the messages.

I'll see if the error styling is defined anywhere else in JS for a pull. If it's not defined would you like me to add it in to a specific file?

ibennetch: Thanks! I think @nijel might better be able to comment about whether it's worth changing the styling function.

js/db_operations.js
```diff
@@ -41,7 +41,7 @@ AJAX.registerOnload('db_operations.js', function () {
         var new_db_name = $('#new_db_name').val();
 
         if (new_db_name == old_db_name) {
-            PMA_ajaxShowMessage(PMA_messages.strDropDatabaseStrongWarning);
+            PMA_ajaxShowMessage("<div class=\"error\"><img src=\"./themes/dot.gif\" title alt class=\"icon ic_s_error\">" + PMA_messages.strDatabaseRenameToSameName + "</div>", false);
```
nijel: This will work only with themes using sprites, technically we still support those.

Also I think it would be better to do styling in the PMA_ajaxShowMessage, so that it can be reused in other parts (there are probably already places where it makes sense to display the message as an error).

ibennetch: @nijel would you suggest making that another parameter as part of `PMA_ajaxShowMessage` (`error` or `isError` defaulting in to false)?

So this function call might look like `PMA_ajaxShowMessage(PMA_messages.strDatabaseRenameToSameName, false, true)`

nijel: Yes something like this.

Gregory-Buhler: Maybe instead of straight checking for error, you would want to know whether it was a notification, a success, or an error, much like you have it set up in libraries/Message.php?

The call might look a little more like `PMA_ajaxShowMessage(PMA_messages.strDatabaseRenameToSameName, false, error);`

With the options of `error`, `notification`, `success`.

nijel: That's probably even better.

ibennetch:
Thanks for the improvements! This looks a lot better. I'm not sure what's going on with the travis-ci tests, I'll have to look at those later. 

Would you mind squashing your commits and pushing (with --force probably) one more time?

ibennetch:
To be clear, I suspect the code test failures are not related to your commit, I just haven't actually looked in detail yet at the report.

ibennetch:
Great work, thanks for your contribution!

PR #12271, opened 2016-05-24
Fixed checking user rights on Insert page

sfobiwahn:

pma only checked the user rights for global and table privileges and ignored the database privileges, this change fixes the behaviour

Signed-Off-By: Florian Steenblock florian.steenblock@gmail.com


nijel:
Can you please squash commits into one?

PR #12273, opened 2016-05-25
Remove mcrypt mention in setup documentation

slokhorst:
mcrypt is no longer mentioned in the 'require' section (removed in commit 692c555c3fd625f9efdee7abf8483cce2d0d6986), so this is confusing.

Signed-off-by: Sebastiaan Lokhorst sebastiaanlokhorst@gmail.com

### DIFF
--- a/doc/setup.rst
+++ b/doc/setup.rst
@@ -734,8 +734,8 @@ Cookie authentication mode
 * If you want to allow users to enter any hostname to connect (rather than only
   servers that are configured in :file:`config.inc.php`),
   see the :config:option:`$cfg['AllowArbitraryServer']` directive.
-* As mentioned in the :ref:`require` section, having the ``mcrypt`` or
-  ``openssl`` extension will speed up access considerably, but is not required.
+* As mentioned in the :ref:`require` section, having the ``openssl`` extension
+  will speed up access considerably, but is not required.
 
 .. index:: pair: Signon; Authentication mode
 


nijel:
AFAIK phpseclib still works fine with either mcrypt or openssl, so I think the documentation is correct here...

slokhorst:
As I understand it, mcrypt still is an option but is discouraged in favor of openssl. (see 692c555c3fd625f9efdee7abf8483cce2d0d6986)
My point is that this line says "the require section says that you can use mcrypt", which is not true because mcrypt is never mentioned anywhere in the docs except for here.

So I think we should either clarify the status of mcrypt, or remove it completely from the docs.

nijel:
I don't think it's much worse than openssl, but it's definitely better to have mcrypt than nothing. So in the end we should state this in the docs ;-).

slokhorst:
I'm not sure how bad it is, but mcrypt will be [deprecated](https://wiki.php.net/rfc/mcrypt-viking-funeral) in PHP 7.1, although it will still be possible to obtain it via PECL after deprecation.

The [page](http://blog.remirepo.net/post/2015/07/07/About-libmcrypt-and-php-mcrypt) that was linked in 692c555c3fd625f9efdee7abf8483cce2d0d6986 suggests that mcrypt is horrible and should be forgotten ASAP. In that case I don't think it is wise to keep advertising it in the phpMyAdmin docs...

But again; I'm not qualified to make any decision about mcrypt support in phpMyAdmin, I just noticed that the documentation was incoherent.

nijel:
Indeed it seems that mcrypt is not actively maintained, so it's probably better not to use it for crypto. For example see https://github.com/phpseclib/phpseclib/issues/918

Thanks to pointing this out, I'm merging your changes.

PR #12274, opened 2016-05-26
fixed checking of user rights to make it look for database privileges

sfobiwahn:
pma only checked the user rights for global and table privileges and ignored the database privileges, this change fixes the behaviour

Signed-off-by: Florian Steenblock florian.steenblock@gmail.com


nijel:
Thanks!

Issue #12262, opened 2016-05-17
Remove inline styles


nijel:
Inline stying doesn't play well with themes and also forces us to use permisive Content-Security-Policy.

The styling should be moved completely to themes and after this is done the CSP should remove ``'unsafe-inline'`` for styles.

flash1452:
@nijel Does fixing this error include even those inline style statements that are used in some of the  .js files??
 And also I wanted to ask if I am allowed to use any id or class-name for those tags that have yet not been given any class or id names.

nijel:
Yes to both answers. The definite goal is to remove inline usage from Conent-Security-Policy.

nijel:
Also this doesn't have to be done all in one step, incremental patches are welcome!

emanuelb:
here list of style attribute usage: (use class or css via id instead)
```
./gis_data_editor.php:echo '<div id="placeholder" style="width:450px;height:300px;'
./gis_data_editor.php:echo '<div id="openlayersmap" style="width:450px;height:300px;'
./gis_data_editor.php:echo '<div class="choice" style="float:right;clear:right;">';
./index.php:        echo '<p style="margin: 0.5em 1em 0.5em 1em">';
./prefs_manage.php:            <div class="click-hide-message group-cnt" style="display:none">
./prefs_manage.php:                <div style="padding-bottom:0.5em">
./templates/error/report_form.twig:    <fieldset style="padding-top:0">
./templates/columns_definitions/column_name.phtml:    <p style="font-size:80%;margin:5px 2px"
./templates/database/designer/having_query_panel.twig:<table id="query_having" style="display:none;" width="5%" cellpadding="0" cellspacing="0">
./templates/database/designer/aggregate_query_panel.twig:<table id="query_Aggregate" style="display:none;" width="5%" cellpadding="0" cellspacing="0">
./templates/database/designer/side_menu.phtml:    <div id="name-panel" style="overflow:hidden">
./templates/database/designer/side_menu.phtml:        <span id="page_name" style="border:none">
./templates/database/designer/side_menu.phtml:        <span id="saved_state" style="border:none">
./templates/database/designer/database_tables.phtml:           style="position:absolute; left:<?= (isset($tab_pos[$t_n])
./templates/database/designer/database_tables.phtml:                               style="margin: 0;"
[...]

PR #12275, opened 2016-05-26
Remove inline styles from elements with id='filterText' (partial fix …

andavies:

…of issue #12262)

Signed-off-by: Andy Davies contact@andy-davies.me


nijel:
Thanks!

PR #12283, opened 2016-05-30
Add support for SSL connection for controluser, issue 12218

ibennetch:


nijel:
I think that either all of these settings should be handled in common.inc.php or here, but not in both locations. See https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/common.inc.php#L756

nijel:
I've rewritten the connection handling code to be more generic and it allowed simpler implementation of this feature: 90498afe7671a088c64d304ed7cc59f95c2e060a

Issue #12070, opened 2016-03-05
Can we get rid of `odd` and `even` classes in tables?


nisargjhaveri:
Can we get rid of the `odd` and `even` classes we use to style table rows? We seem to spend a lot of efforts handling that throughout the codebase.

Realized while looking at https://github.com/phpmyadmin/phpmyadmin/commit/d798809215c3147150dad12a14a3f049db3bbea4#diff-ec96a27edeea8db227299965812df402R176.

Alternative is to use `:nth-child([odd|even])` pseudo selector. Looks like it is well supported too. http://caniuse.com/#search=nth-child

Sorry for not using the template, the template is not really a good fit for this kind of issues.

ibennetch:
I haven't looked through all the code on this, there might be an edge case or two where this wouldn't fit well — but short of anything like that, I think this is a great idea.

nijel:
I think it's reasonable move. This comes from times this was not possible to achieve using CSS.

madhuracj:
+1

OlafvdSpek:
Is alternating row coloring still a good idea?
I think it went out of fashing a decade ago. ;)

nijel:
I think it makes it easier to read, but this is something what can be done purely in CSS now.

vosgus:
Hi, I have developed a fix for this issue. At the moment the fix is implemented in the pmahomme theme. Should I also implement it in the other themes (metro, original etc.)? Also I have only made changes to the CSS code since I'm not that familiar with the PHP codebase yet.

nijel:
It would be great if you could port it to original theme (which is part of every release). The metro theme is separate, but fix for it is welcome as well.

nijel:
Mostly fixed by #12287, but there are still some places left (this just just a bit filtered `git grep` output, it's probably not complete and might contain false positives):

```
db_datadict.php:    $odd_row = true;
db_datadict.php:        echo $odd_row ? 'odd' : 'even'; $odd_row = ! $odd_row;
db_export.php:$odd = true;
db_export.php:    $multi_values .= '<tr class="' . ($odd ? 'odd' : 'even') . ' marked">';
db_export.php:    $odd = ! $odd;
js/functions.js:            last_clicked_row = last_click_checked ? $table.find('tr.odd:not(.noclick), tr.even:not(.noclick)').index($tr) : -1;
js/functions.js:                $tr.parent().find('tr.odd:not(.noclick), tr.even:not(.noclick)')
[...]

PR #12287, opened 2016-05-30
Remove CSS classses odd and even

vosgus:

The odd and even classes for tables are no longer needed. The same
functionality can be achieved using nth-child(). This commit is for the
pmahomme and original theme.

Signed-off-by: Patric Gustafsson patricjgustafsson@outlook.com


ibennetch:
It appears as if your editor has used tabs instead of spaces for indentation; the phpMyAdmin standard is to use all spaces. Would you mind changing the files and pushing again? (You'll probably have to use the `-force` option with git push).

ibennetch:
Thanks for your contribution.

To track the larger issue, we have:
[x] remove `.odd` and `.even` classes from .css files (implemented in this pull request)
[ ] remove `.odd` and `.even` classes and logic in .php files

vosgus:
Ouch, sorry for that. I'll fix it today or tomorrow.

nijel:
Your change breaks at least displaying tables with th inside:
localhost_localhost_bar_phpmyadmin_4 7 0-dev_-_2016-06-16_10 53 28

nijel:
An progress on missing bits?

* introducing this breaks current rendering in some cases
* it does not remove matching PHP code

vosgus:
Hi, I've submitted a patch for the rendering issues with the table headers, it should be fine now. If the patch is OK I'll start working on removing the PHP code.

nijel:
Thanks, the CSS now seems to work fine.

vosgus:
Since there will be many small commits in this patch I was wondering what would be the best way to do it. Should I make all the changes and then squash them together or is there some other preferred way?

nijel:
I don't mind with small commits either, just try to group similar things into the commit.

nijel:
I see there is quite good progress here. Any estimates when it will be ready for merge?

vosgus:
There are about ten files left that still needs to be worked on. About a week from now I think is a good estimate.

nijel:
Great, thanks for update!

vosgus:
I'm working on the last couple of files right now. One bug that I have found is on the tbl_search.php page and tbl_zoom_select.php page. There are some hidden rows in the tables there so the CSS will not look right. The hidden rows will get the background color instead. 

image

One solution would perhaps be to keep the old CSS code just for these pages. Also I'm a bit unsure what the hidden fields are doing. Are they just there to provide some spaces between the rows?

nijel:
I think these rows with hidden can be folded into previous ones in both cases:

* https://github.com/phpmyadmin/phpmyadmin/blob/master/templates/table/search/rows_normal.phtml
* https://github.com/phpmyadmin/phpmyadmin/blob/master/templates/table/search/rows_zoom.phtml

rpv-tomsk:
Hi!

JFYI:  I think newly reported #12611 may be related to your work in progress.

nijel:
@vosgus I've just fixed the two offending files in master. Are you still interested in completing this?

vosgus:
@nijel Yes I still am. I'll get back to working on it in a few days. If I recall correctly there should not be many files left to fix.

nijel:
That's true, it's almost complete.

vosgus:
@nijel Ok, I think this is done now, I've fixed all the files I could find at least. Please tell me if anything else still needs to be done that I might have missed.

nijel:
Well there are still some places left (see https://github.com/phpmyadmin/phpmyadmin/issues/12070#issuecomment-261484535), but I've decided to merge the code for now as what is done works okay. Thanks for your contribution!

PR #12296, opened 2016-06-05
Fix #12281 Copying a user resets SSL to None

### LINKED ISSUE #12281: Copying a user resets SSL to None
### Steps to reproduce
1. Create a user with the Require SSL radio button selected
2. Copy that user

### Expected behaviour
The new user should have Require SSL set

### Actual behaviour
The new user defaults to "None"

### Server configuration
**Database:**
MySQL  5.5.49-0+deb8u1 - (Debian) 
**phpMyAdmin version:**
git master
### Client configuration
**Browser:**
Firefox
**Operating system:**
MacOS

madhuracj:

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,7 @@ phpMyAdmin - ChangeLog
 - issue #12279 Fixed error reporting on connect problems
 - issue #12290 Fixed export of tables without explicitly set engine
 - issue #12285 Designer JavaScript error: Show/Hide tables list
+- issue #12281 Keep user attributes (privileges, authentication mode, etc) when copying a user
 
 4.6.2 (2016-05-25)
 - issue        [security] User SQL queries can be revealed through URL GET parameters, see PMASA-2016-14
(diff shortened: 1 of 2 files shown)


ibennetch:
Works great for me. Nice job.

ibennetch:
If I add a user to a usergroup then copy that user, it probably should keep the usergroup as well. Is that reasonable to fix in this patch or should I open a new issue?

ChangeLog
```diff
@@ -6,6 +6,7 @@ phpMyAdmin - ChangeLog
 - issue #12279 Fixed error reporting on connect problems
 - issue #12290 Fixed export of tables without explicitly set engine
 - issue #12285 Designer JavaScript error: Show/Hide tables list
+- issue #12281 Copying a user resets SSL to None
```
ibennetch: Given the expanded scope of this patch, I think I would change this to "Keep user attributes (privileges, authentication mode, etc) when copying a user"

madhuracj:
Given that user group is a phpMyAdmin feature rather than a MySQL privilege/feature, I am not sure whether this should be set. If you think it should be set for the new user, please open a separate ticket for it.

PR #12306, opened 2016-06-08
Shapefile lib

j0kan:

Signed-off-by: Johannes Vockeroth vockeroth@nextbike.net


ibennetch:
Hi, I think there may be a problem with your pull request. It looks like the only change that got committed was removing the last end-of-line character from `libraries/bfShapeFiles/ShapeFile.lib.php`.

What is it you're trying to fix here? Are you having specific git problems we can help with?

j0kan:
Sorry, you're right - I'll open a new request

PR #12318, opened 2016-06-11
Fix #12277 ENUM editor adds blank values

### LINKED ISSUE #12277: VIsual enum editor adds extra (blank) values
When using the visual enum editor to add less than four enum values, it will add blank values. For instance, `` `e_test` enum('a','b','','') NOT NULL``

ibennetch:

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -10,6 +10,7 @@ phpMyAdmin - ChangeLog
 - issue #12279 Fix MySQL connection error on version mismatch
 - issue #12281 Keep user attributes (privileges, authentication mode, etc) when copying a user
 - issue #12308 Fix division by zero in case of misconfigured MySQL server
+- issue #12277 ENUM editor adds extra blank values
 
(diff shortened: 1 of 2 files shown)


nijel:
Hmm, not sure this is good approach - what if user wants to have empty value as one of choices?

ibennetch:
I did consider that, but didn't come up with a good solution and don't think that scenario comes up often. In the case a user does want a blank value, they can always enter it in the text field on the main page once the visual editor is hidden. I don't really have an idea how we can tell the difference between a default `''` or an intentional one, unless we change the editor to only open one one ENUM value by default, forcing the user to add the rest.

nijel:
I still think empty value is something what can be used. More importantly if you edit value where blank value is already there, it would be removed causing data corruption (not sure how MySQL would behave in such case)...

OlafvdSpek:
Doesn't MySQL always allow empty values in enum columns?

PR #12328, opened 2016-06-17
Fixes #11983 pop out SQL window

### LINKED ISSUE #11983: Feature-Request: Pop-Out SQL Window

Achilles-96:


Achilles-96:
Sorry for the delay.
@nijel, I am showing the results in parent window as you said. Please review.

nijel:
There are still few bugs:

* once you close the query window, the link no longer opens new window, but rather replaces phpMyAdmin main window
* the opened window is still badly sized (at least the submit button should be visible, but ideally it would scale the textarea to fit into window making the controls visible all the time), see screenshot below
* there is no synchronization with main window, so there is really no way to change database used in querywindow, ideally it would change database in sync with main window

snimek z 2016-07-29 13-41-23

nijel:
Closing this as it's not ready for merge at this state and hasn't received updates for months.

PR #12329, opened 2016-06-19
Basic functionality for varchar(10) without Faker added

Achilles-96:

Signed-off-by: Achilles-96 raghuram.vadapalli@research.iiit.ac.in

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,8 @@
         "phpmyadmin/motranslator": "~0 || ~1.0",
         "tecnickcom/tcpdf": "~6.2",
         "phpseclib/phpseclib": "~2.0",
-        "google/recaptcha": "~1.1"
+        "google/recaptcha": "~1.1",
+        "fzaninotto/faker": "v1.6.0"
     },
     "suggest": {
         "ext-zlib": "For gz import and export",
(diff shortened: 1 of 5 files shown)


Achilles-96:
Not to merge. I made basic functionality for varchar(10). Please review interface. If it is fine, I'll integrate with Faker for all datatypes.

nijel:
Seems this really slipped of our attention, sorry.

Few comments:

- Such feature and dependency should be IMHO optional.
- I'm not really sure if placing it on insert/edit page is good idea, certainly it makes no sense for edit one. Maybe it would be better to do this on the operations page?
- Please avoid adding more spaghetti code to top level files, rather implement the code in a separate class which can be properly tested.

Achilles-96:
Thanks for the review. Am I encouraged to work on this now. [Here](https://github.com/phpmyadmin/phpmyadmin/wiki/GSoC_2017_applicant_guide#get-involved-with-phpmyadmin), it says otherwise.

nijel:
I'm not sure - you've already spent quite some time on it, but having it implemented before summer would force us to add new tasks to the GSoC idea. What does @ibennetch think?

ibennetch:
I think you can continue to work on this and we can find another project for the GSoC list (such as I've done at https://github.com/phpmyadmin/phpmyadmin/wiki/GSoC_2017_ideas_list/_compare/0d46ca50b68e2fcd360dbcf9d76085e786228cef...71ea3d2e0cde72cb70057f396afbbf17d30d8893)

Achilles-96:
It is fine with me if you close this for now.

PR #12331, opened 2016-06-20
"SET lc_messages" is not supported on older MySQL versions

bigio:
The sql command "SET lc_messages" is not supported on 5.1.x MySQL versions

### DIFF
--- a/libraries/DatabaseInterface.php
+++ b/libraries/DatabaseInterface.php
@@ -1409,15 +1409,17 @@ class DatabaseInterface
             );
         }
 
-        /* Locale for messages */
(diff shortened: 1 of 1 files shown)


ibennetch:
Hi and thanks for your contribution. However, MySQL 5.1 and older is only supported by phpMyAdmin 4.0 (which itself is only supported for security fixes). Since we no longer support connecting to such old MySQL versions, I don't see a need for this.

bigio:
There is a "(PMA_MYSQL_INT_VERSION >= 50100)" in db_operations.php, line 106, this should be removed as well if you want to drop support for  MySQL 5.1 then.

PR #12363, opened 2016-07-02
Fixes #12352 - password strength simulator

### LINKED ISSUE #12352: Adding Password Strength Simulator
1) Adding Password Strength Simulator

Add Password Strength simulator (indicator about how much the password is strong) for password input fields that user supply (for example: user creation process)
for example:

use zxcvbn library.
https://github.com/dropbox/zxcvbn
https://www.bennish.net/password-strength-checker/

or other solutions:
https://www.sitepoint.com/5-bootstrap-password-strength-metercomplexity-demos/

Motivations:
    \* Reduce the chance of using weak passwords by the user.

Achilles-96:


js/server_privileges.js
```diff
@@ -93,6 +93,28 @@ AJAX.registerOnload('server_privileges.js', function () {
...
+                    break;
+            case 1: $('#password_strength').html('Very weak');
+                    break;
+            case 2: $('#password_strength').html('Weak');
+                    break;
+            case 3: $('#password_strength').html('Good');
+                    break;
+            case 4: $('#password_strength').html('Strong');
+        }
+        console.log(zxcvbn_obj);
```
devenbansod: Hi, I believe this was for debugging purpose ?

devenbansod:
Also, I think this is not working as expected while I edit an existing user. I create a user, the password strength indicator works. 
I again go and try to copy that user by User Accounts -> Edit Privileges -> Login information, then it does not change and is stuck at 'Good' .
Also, if we are introducing this, we can do so for the 'Change password' fields as well ?

Achilles-96:
I checked it. That is because we are having two elements with same `id` on that page. I did not expect that.

Achilles-96:
https://github.com/blog/2141-squash-your-commits. I think we can use this.

libraries/display_change_password.lib.php
```diff
@@ -60,16 +60,20 @@ function PMA_getHtmlForChangePassword($mode, $username, $hostname)
...
         . '<input type="radio" name="nopass" value="0" id="nopass_0" '
-        . 'onclick="document.getElementById(\'text_pma_pw\').focus();" '
+        . 'onclick="document.getElementById(\'text_pma_change_pw\').focus();" '
         . 'checked="checked" />'
         . '<label for="nopass_0">' . __('Password:') . '&nbsp;</label>'
         . '</td>'
         . '<td>'
-        . '<input type="password" name="pma_pw" id="text_pma_pw" size="10" '
+        . __('Enter:') . '&nbsp;&nbsp;&nbsp;&nbsp;&nbsp'
+        . '<input type="password" name="pma_pw" id="text_pma_change_pw" size="10" '
```
nijel: Why are you changing ID of the inputs? It will most likely break other things (like password generating).

Achilles-96: @nijel , The form for changing the password and the form for editing login information are on the same page. It is operated by hiding one or the other at a time. Both the forms have the same id for the password field. We cannot have two elements with the same id in any HTML page. I changed the way generate password accesses input elements of form (i.e by name instead of id) as it makes sense to have two elements with same name in a HTML form provided they are in different forms.

nijel:
Somehow it doesn't work for me (tried on password change):

localhost_localhost_phpmyadmin_4 7 0-dev_-_2016-08-03_09 56 49
localhost_localhost_phpmyadmin_4 7 0-dev_-_2016-08-03_09 56 32

nijel:
Also is it possible to have non minified version of zxcvbn.js?

Achilles-96:
Please check now.

nijel:
Still same. I see you have CSS only for Webkit, isn't that the problem?

nijel:
Hmm, playing a bit with the upstream demo, I'm not really sure including this is a good idea. It might just give wrong feeling of security when there is none.

For example choosing password "phpMyAdmin123456" gains score 4 / 4...

Achilles-96:
I think we cannot handle all the cases. For example, if user uses his/her name with 123456, then also it would give a score of 4. This happens in many websites. Eventhough we cannot give best score possible, it sure does prevent really weak passwords.

nijel:
In my opinion using password phpMyAdmin123456 in phpMyAdmin **is** really weak password and this code marks it as strong leading to false assumption of security, what is IMHO worse than no indicator at all.

Either the code should be extended to cover easiest to guess words for phpMyAdmin's use case (at least phpMyAdmin, MySQL and MariaDB, but maybe some others) or not used at all.

Overall I'm not really convinced this such feature is good security wise, but it probably can do something good to less experienced users. See also https://nakedsecurity.sophos.com/2016/08/17/why-you-still-cant-trust-password-strength-meters/

js/server_privileges.js
```diff
@@ -31,6 +31,25 @@ function checkAddUser(the_form)
     return PMA_checkPassword($(the_form));
 } // end of the 'checkAddUser()' function
 
+
+function checkPasswordStrength(value, meter_obj, meter_object_label) {
+    var zxcvbn_obj = zxcvbn(value);
```
nijel: Please add second parameter here with our customized word list.

js/server_privileges.js
```diff
@@ -31,6 +31,25 @@ function checkAddUser(the_form)
...
+    var strength = zxcvbn_obj.score;
+    meter_obj.val(strength);
+    strength = parseInt(strength);
+    switch(strength){
+        case 0: meter_obj_label.html('Too short');
```
nijel: This is not just too short, but really weak password (eg. dictionary word), please adjust text.

js/server_privileges.js
```diff
@@ -31,6 +31,25 @@ function checkAddUser(the_form)
...
+                break;
+        case 1: meter_obj_label.html('Very weak');
+                break;
+        case 2: meter_obj_label.html('Weak');
+                break;
+        case 3: meter_obj_label.html('Good');
+                break;
+        case 4: meter_obj_label.html('Strong');
+    }
```
nijel: All the strings should be properly localized.

server_privileges.php
```diff
@@ -27,6 +27,7 @@
 $header   = $response->getHeader();
 $scripts  = $header->getScripts();
 $scripts->addFile('server_privileges.js');
+$scripts->addFile('zxcvbn.js');
```
nijel: Probably same needs to be done in index.php where password change form can be opened.

nijel (changes requested):
I think once these issues are fixed and good words dictionary is created this can be merged.

js/server_privileges.js
nijel: ...and username from the form.

js/server_privileges.js
```diff
@@ -31,6 +31,39 @@ function checkAddUser(the_form)
...
+                         return 0;
+        }
+        return 4;
+}
+
+function checkPasswordStrength(value, meter_obj, meter_object_label, username = null) {
+    var zxcvbn_obj = zxcvbn(value);
+    var strength = zxcvbn_obj.score;
+    strength = parseInt(strength);
+    strength = Math.min(strength, custom_pw_checker(value, username));
```
nijel: You can simply pass the customDict as second parameter to zxcvbn, no need to implement own checks.

nijel:
Thanks, I've merged your code with some minor fixes:

* 2de2eb7 Changelog entry for #12352
* 6145a84 Use username for password change strength indicator
* 4cfe911 Upgrade zxcvbn
* eb57b8d Improve black list
* 3bf24ad Correctly initialize scripts on loading password change form
* ac2cfef Move scripts loading to correct location

PR #12389, opened 2016-07-17
Fixes #12378 - Support for IPv6 proxies

### LINKED ISSUE #12378: Support IPv6 proxies
Currently PMA_getIp only validates IPv4 addresses. Probably we could use PHP validation for that (FILTER_VALIDATE_IP) instead of own regexp.

Achilles-96:

### DIFF
--- a/libraries/ip_allow_deny.lib.php
+++ b/libraries/ip_allow_deny.lib.php
@@ -35,14 +35,11 @@ function PMA_getIp()
     $matches = array();
     // the $ checks that the header contains only one IP address,
     // ?: makes sure the () don't capture
-    $is_ip = preg_match(
-        '|^(?:[0-9]{1,3}\.){3,3}[0-9]{1,3}$|',
-        $trusted_header_value, $matches
-    );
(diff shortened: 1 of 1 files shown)


libraries/ip_allow_deny.lib.php
```diff
@@ -35,14 +35,11 @@ function PMA_getIp()
     $matches = array();
     // the $ checks that the header contains only one IP address,
     // ?: makes sure the () don't capture
-    $is_ip = preg_match(
-        '|^(?:[0-9]{1,3}\.){3,3}[0-9]{1,3}$|',
-        $trusted_header_value, $matches
-    );
+    $is_ip = (filter_var($trusted_header_value, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6) || filter_var($trusted_header_value, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4));
```
nijel: I think that using just ``filter_var($trusted_header_value, FILTER_VALIDATE_IP)`` is better and does same.

libraries/ip_allow_deny.lib.php
```diff
@@ -35,14 +35,11 @@ function PMA_getIp()
...
 
-    if ($is_ip && (count($matches) == 1)) {
+    if ($is_ip) {
```
nijel: It's safer to compare ``$is_ip !== false`` (though I don't see specific value which could cause problem here, it's always better to avoid PHP's automatic type casting).

PR #12406, opened 2016-07-26
GPG sign Git release tags

nijel:
This makes it possible to verify that the tag matches what release
manager did release.

I don't think this could cause any problems (as gpg setup is anyway already required for signing releases), but still this is for @ibennetch as a release manager to merge.


ibennetch:
This looks fine to me. Thanks.

PR #12409, opened 2016-07-27
Restore Overhead column in Server > Databases > Enable Statistics.

rpv-tomsk:
Restore Overhead column in Server > Databases > Enable Statistics which was removed in 582b02262bcda6edb10ffc6df1b67c47468bbe59.

Similar can also be applied to earlier versions, but I use this on 4.6.3.


nijel:
Merged, thanks!

PR #12415, opened 2016-07-28
Remove token from GET requests

nijel:

The CSRF token really should be used only in POST requests. The reason
for that is that it's a bit harder to get to if it is in request body
(with POST) compared to GET request, where it is in the URL (being
easily available in server logs).

Also this will make the URLs look nicer ;-).

This change will definite break some functionality, but since #6297 most
of the code should be safe and remaining bugs can be fixed for upcoming
release.


madhuracj:
Indeed, it is better that this fix go in master.

devenbansod:
With this merged, does it mean that all the Ajax POST requests that we do should have a 'token' parameter in them now ? 
Like https://github.com/phpmyadmin/phpmyadmin/commit/9233365ca50c31c65a4c9da3097ed9a314f79f8f ?

nijel:
Yes, all POST requests should have token (otherwise it would be vulnerable to XSRF).

PR #12420, opened 2016-07-31
Fix advisory rule 'Slow launch time'

rpv-tomsk:
Issue string 'Slow_launch_thread' in advisory rule should be changed to 'Slow_launch_time' to match checked variable.


rpv-tomsk:
Rebased.

nijel:
Merged, thanks.

PR #12430, opened 2016-08-02
Partitions editor: Allow to edit partition names; Add 'COLUMNS Partitioning' support.

rpv-tomsk:
Fixed issues:
- Table partitions editor does not support 'COLUMNS Partitioning' ('RANGE COLUMNS', 'LIST COLUMNS'). Added in MySQL 5.5 (http://dev.mysql.com/doc/refman/5.5/en/partitioning-columns.html).
- Table partitions editor ignores exising partitions names and overwrites them to 'pXX' on saving.
- The 'Partition maintenance' snippet on 'Operations' tab has duplicate partition names in list when partitioning has subpartitions.
- Partition comment textarea adds extra spaces to value. Caused by word wrap/formatting.

Optimization:
- Optimize Partition::getPartitionMethod() - need only one row, so use 'LIMIT 1' in query to avoid much data transfer from database when table has many partitions/subpartitions.

### DIFF
--- a/templates/columns_definitions/partitions.phtml
+++ b/templates/columns_definitions/partitions.phtml
@@ -1,5 +1,5 @@
 <?php
-$partitionOptions = array('', 'HASH', 'LINEAR HASH', 'KEY', 'LINEAR KEY', 'RANGE', 'LIST');
+$partitionOptions = array('', 'HASH', 'LINEAR HASH', 'KEY', 'LINEAR KEY', 'RANGE', 'RANGE COLUMNS', 'LIST', 'LIST COLUMNS');
 $subPartitionOptions = array('', 'HASH', 'LINEAR HASH', 'KEY', 'LINEAR KEY');
 $valueTypeOptions = array('', 'LESS THAN', 'LESS THAN MAXVALUE', 'IN');
 ?>
@@ -88,9 +88,10 @@ $valueTypeOptions = array('', 'LESS THAN', 'LESS THAN MAXVALUE', 'IN');
         <?php foreach ($partitionDetails['partitions'] as $partition): ?>
             <?php $rowspan = ! empty($partition['subpartition_count']) ? ($partition['subpartition_count'] + 1) : 2; ?>
             <tr class="<?= ($odd ? 'odd' : 'even'); ?>">
-                <th rowspan="<?= $rowspan; ?>" class="vmiddle">
-                    <?= htmlspecialchars($partition['name']); ?>
-                </th>
+                <td rowspan="<?= $rowspan; ?>">
+                    <input type="text" name="<?= $partition['prefix']; ?>[name]"
+                       value="<?= htmlspecialchars($partition['name']); ?>" />
+                </td>
                 <?php if ($partitionDetails['value_enabled']): ?>
                     <td rowspan="<?= $rowspan; ?>" class="vmiddle">
                         <select class="partition_value"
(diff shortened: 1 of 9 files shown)


rpv-tomsk:
Possible TODO/future improvement: Add columns hint/selectbox for 'COLUMNS Partitioning'.

nijel:
I think this is something what is out of scope for maintenance branch (4.6), please target such changes to master branch.

rpv-tomsk:
@nijel , what do you think about this part:

* The 'Partition maintenance' snippet on 'Operations' tab has duplicate partition names in list when partitioning has subpartitions.
* Partition comment textarea adds extra spaces to value. Caused by word wrap/formatting. 

?

rpv-tomsk:
> I think this is something what is out of scope for maintenance branch (4.6),

As for me, I have put this to our production server.
I tried to make things backward-compatible, how it possible, so I see no much changes to forbid merging to 4.6. But last decision is on team, of course.

templates/columns_definitions/partitions.phtml
```diff
@@ -88,9 +88,10 @@ $valueTypeOptions = array('', 'LESS THAN', 'LESS THAN MAXVALUE', 'IN');
         <?php foreach ($partitionDetails['partitions'] as $partition): ?>
             <?php $rowspan = ! empty($partition['subpartition_count']) ? ($partition['subpartition_count'] + 1) : 2; ?>
             <tr class="<?= ($odd ? 'odd' : 'even'); ?>">
-                <th rowspan="<?= $rowspan; ?>" class="vmiddle">
-                    <?= htmlspecialchars($partition['name']); ?>
-                </th>
+                <td rowspan="<?= $rowspan; ?>">
+                    <input type="text" name="<?= $partition['prefix']; ?>[name]"
```
nijel: htmlspecialchars probably missing here.

templates/columns_definitions/partitions.phtml
```diff
@@ -123,7 +124,10 @@ $valueTypeOptions = array('', 'LESS THAN', 'LESS THAN MAXVALUE', 'IN');
             <?php foreach ($subpartitions as $subpartition): ?>
                 <tr class="<?= ($odd ? 'odd' : 'even'); ?>">
                     <?php if ($partitionDetails['can_have_subpartitions'] && $partitionDetails['subpartition_count'] > 1): ?>
-                        <th class="vmiddle"><?= htmlspecialchars($subpartition['name']); ?></th>
+                        <td>
+                            <input type="text" name="<?= $subpartition['prefix']; ?>[name]"
```
nijel: htmlspecialchars probably missing here.

templates/columns_definitions/partitions.phtml
```diff
@@ -135,9 +139,8 @@ $valueTypeOptions = array('', 'LESS THAN', 'LESS THAN MAXVALUE', 'IN');
                         ); ?>
                     </td>
                     <td>
-                        <textarea name="<?= $subpartition['prefix']; ?>[comment]">
-                            <?= htmlspecialchars($subpartition['comment']); ?>
-                        </textarea>
+                        <?php  //Please keep this at one line to avoid extra spaces in textarea value. ?>
+                        <textarea name="<?= $subpartition['prefix']; ?>[comment]"><?= htmlspecialchars($subpartition['comment']); ?></textarea>
```
nijel: htmlspecialchars probably missing here.

nijel:
I don't think this qualifies as bug fix, which is only thing which should be going into 4.6 branches.

rpv-tomsk:
> I don't think this qualifies as bug fix, which is only thing which should be going into 4.6 branches.

How I found this issue:

We have some table with partitions named 'year_2015', 'year_2016', 'year_max' and so on.
Our queries use partition names (https://dev.mysql.com/doc/refman/5.6/en/partitioning-selection.html)

After editing partitions all these names which are referenced in queries was lost. Is this not a bug?

rpv-tomsk:
>htmlspecialchars probably missing here.

You are likely to be right. I will continue the work on this patch on this week.

Thanks for code review.

madhuracj:
With @rpv-tomsk's  explanation, I tend to think this qualifies as a bug fix.

rpv-tomsk:
As per `libraries/tbl_partition_definition.inc.php` and `libraries/controllers/table/TableStructureController.php` code,  `partition['prefix']` value is built by code, in form of `partitions[X][subpartitions][Y]`  where X and Y are integers. So, no htmlspecialchars is required on `partition['prefix']` value. But if htmlspecialchars() is required by codestyle, I will update the patch.

While digging, found missed `htmlspecialchars` on `$subParition->getComment()` in `templates/table/structure/display_partitions.phtml`. Patch updated.

Also, please look into templates/columns_definitions/partitions.phtml:111, there is unescaped `$partition['value'];` :

```
<input type="text" class="partition_value"
                               name="<?= htmlspecialchars($partition['prefix']); ?>[value]"
                               value="<?= $partition['value']; ?>" />
```
I'm unsure if this variable requires htmlspecialchars().

nijel:
Okay, merged now. Thanks!

PR #12439, opened 2016-08-03
Fix html header for user properties editor

rpv-tomsk:
Header for user properties has no table name if table was selected from dropbox.

user-properties-table

Looking into PMA_getDataForDBInfo() code I found the cause: $tablename and $routinename can be set from dropdown box, then their values are set from $_REQUEST['pred_tablename'] /  $_REQUEST['pred_routinename'] in that case. So checks for $_REQUEST['routinename'] and $_REQUEST['tablename'] in PMA_getHtmlHeaderForUserProperties() are 
insufficient.

Changing only $_REQUEST['tablename'] to $tablename will break routinename display, so added new function argument.

Proposed patch fixes the issue.

### DIFF
--- a/libraries/server_privileges.lib.php
+++ b/libraries/server_privileges.lib.php
@@ -974,7 +974,9 @@ function PMA_getHtmlForRoutineSpecificPrivilges(
     $username, $hostname, $db, $routine, $url_dbname
 ) {
     $header = PMA_getHtmlHeaderForUserProperties(
-        false, $url_dbname, $db, $username, $hostname, $routine
+        false, $url_dbname, $db, $username, $hostname,
+        '', /* Tablename */
+        $routine
     );
 
     $sql = "SELECT `Proc_priv`"
@@ -1094,7 +1096,7 @@ function PMA_getHtmlForTableSpecificPrivileges(
         . '<input type="hidden" name="column_count" '
         . 'value="' . count($columns) . '" />' . "\n"
         . '<fieldset id="fieldset_user_priv">' . "\n"
-        . '<legend data-submenu-label="Table">' . __('Table-specific privileges')
+        . '<legend data-submenu-label="' . __('Table') . '">' . __('Table-specific privileges')
         . '</legend>'
         . '<p><small><i>'
         . __('Note: MySQL privilege names are expressed in English.')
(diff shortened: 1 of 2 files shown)


rpv-tomsk:
Hi!

What can I do about this issue to help merge it?

ibennetch:
Hi, and thanks for your submission. Probably no one has really looked at it yet. There's nothing obviously wrong (or we would have commented), we have just been directing our attention elsewhere for the past few weeks. Thanks for your patience.

libraries/server_privileges.lib.php
```diff
@@ -4573,7 +4575,8 @@ function PMA_getAddUserHtmlFieldset($db = '', $table = '')
  * @return string $html_output
  */
 function PMA_getHtmlHeaderForUserProperties(
-    $dbname_is_wildcard, $url_dbname, $dbname, $username, $hostname, $tablename
+    $dbname_is_wildcard, $url_dbname, $dbname, $username, $hostname, $tablename,
+    $routinename
 ) {
```
devenbansod: I don't have a strong opinion but don't you think having two parameters as `$itemname` (which would be the routine name or table name) and `$itemtype` which would default to 'table' and can take value as 'routine' would go better ?

I can be persuaded against it though.

devenbansod:
Otherwise, it works for me, and looks okay to merge.

libraries/server_privileges.lib.php
rpv-tomsk: This change was built as a patch to stable branch, so it tends to minify changes. 
I think such change is outside of this patch.

That change can be done, but I also see no strong reasons to do it.
Function called only from two places now, so it is not difficult to change these when required.

What is your reasons against it? I'm interesting to know them. :-)

devenbansod: As I said, I did not have a strong opinion and could be persuaded. I just felt that it would go with the general trend of function declarations in our codebase (having a default parameter value which can be over-riden, rather than forcing two values one of them which is just to indicate that it is blank).

For example, (not exactly same but similar parameter names and the sense in which they are used) 
1. https://github.com/phpmyadmin/phpmyadmin/blob/QA_4_6/libraries/server_privileges.lib.php#L2141
2. https://github.com/phpmyadmin/phpmyadmin/blob/QA_4_6/libraries/navigation/Navigation.php#L78

rpv-tomsk: Hi!

Should I update the patch to help merge it?

devenbansod:
Merged with some changes in https://github.com/phpmyadmin/phpmyadmin/commit/f80140c46e7e3df31dd5450b60136518b8ac1f0a

devenbansod:
Thanks!

PR #12457, opened 2016-08-13
Fixes #12380 Refactored http requests into one function "httpRequest" in Util.php

### LINKED ISSUE #12380: One code to handle HTTP requests
Currently we have at least three places which handle HTTP requests and those are inconsistent. 

* Config::checkHTTP
* VersionInformation::getLatestVersion
* PMA_sendErrorReport

This should be all moved to single place and share as much code as possible.

Seldar:

Changed below methods to call the same function for httpRequests, which resides in Util.php.  

Config::checkHTTP
VersionInformation::getLatestVersion
PMA_sendErrorReport

Signed-off-by: Volkan Ulukut arthan@gmail.com

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -23,6 +23,7 @@ phpMyAdmin - ChangeLog
 - issue        Removed MySQL connection retry without password
 - issue #12218 Allow to specify further parameters for control connection
 - issue #12162 Show charset for each table on Database structure page
+- issue #12330 Shortcut for closing console
 
 4.6.5 (not yet released)
 - issue        Remove potentionally license problematic sRGB profile


libraries/Util.php
```diff
@@ -4893,4 +4893,89 @@ public static function requestString($value)
...
+    /**
+     * Creates HTTP request
+     *
+     * @param string $url Url to send the request
+     * @param string $method HTTP request method (GET, POST, PUT, DELETE, etc)
+     * @param int $connection_timeout Timeout seconds for the HTTP request
+     * @param bool $return_only_status If set to true, the method would only return response status
+     * @param mixed $content Content to be sent with HTTP request
+     * @param string $header Header to be set for the HTTP request
+     * @param bool $use_curl Force using curl
```
nijel: Is forcing of curl really needed?

libraries/Util.php
```diff
@@ -4893,4 +4893,89 @@ public static function requestString($value)
...
+     *
+     * @return mixed
+     */
+    public static function httpRequest($url, $method, $connection_timeout, $return_only_status = false, $content = null, $header = "", $use_curl = false)
+    {
+        if (ini_get('allow_url_fopen') && !$use_curl) {
```
nijel: Please prefer curl as it always does SSL verification while alow_url_fopen wrappers failed to do that in older PHP versions. 

This has been changed in recent security release, see e749214b1681ce6af31df169f57b0c23d2a40232 (and is reason why your changes now can not be merged).

libraries/VersionInformation.php
```diff
@@ -51,44 +46,9 @@ public function getLatestVersion()
                 session_write_close();
```
nijel: I think session closing and reopening should go to the wrapper as well.

libraries/Config.php
```diff
@@ -716,41 +716,16 @@ public function checkGitRevision()
      */
     public function checkHTTP($link, $get_body = false)
```
nijel: Is this function still necessary? Maybe it's enough to have just one for handling the requests and convert checkHTTP users to the new API?

PR #12458, opened 2016-08-14
Translation improvements

rpv-tomsk:
- Improved Russian translation
- Added lost internationalization to column_definitions_form.phtml


nijel:
Please use our [translation server](https://hosted.weblate.org/projects/phpmyadmin/) for doing fixed in translation, that way it will automatically appear in all supported branches.

Anyway I've merged this for now.

PR #12459, opened 2016-08-14
When editing table records, display readonly inputs for fields without write access.

rpv-tomsk:
The problem:

While editing records, user with limited privileges does not see all fields, which he has read access.
Only fields with write access are displayed.

Expected behaviour:

The edit form should display all fields which user has read access. 
Fields, for which user have no write access, should be displayed as read-only.

Details:

We have users with limited privileges. They have SELECT privilege to all table fields, but only some of fields are accessible for insert/update. While table browsing and making decision to edit record, user see all available fields. When table row edit form opened, user see only fields, which he has insert/update access. In our case that means what user has only one field (while it has SELECT privilege to many fields) on the edit form, which makes is completely unuseable - other fields are the record editing context. We caught this after our Debian-based installation upgrade to 4.6.3.

Problem caused by rfe # 689 / ff1592483a9befbe93db30b766f4b38da4c775f1 commit.

### DIFF
--- a/libraries/insert_edit.lib.php
+++ b/libraries/insert_edit.lib.php
@@ -379,12 +379,13 @@ function PMA_getEnumSetAndTimestampColumns($column, $timestamp_seen)
  * @param integer $tabindex              tab index
  * @param integer $idindex               id index
  * @param boolean $insert_mode           insert mode or edit mode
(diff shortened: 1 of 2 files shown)


nijel:
Thanks, looks good!

PR #12465, opened 2016-08-17
Fixes #12380 Refactored http requests into one function "httpRequest" in Util.php

### LINKED ISSUE #12380: One code to handle HTTP requests

Seldar:
Changed below methods to call the same method for httpRequests, which resides in Util class.

Util::httpRequest - new method
Config::checkHTTP - removed
VersionInformation::getLatestVersion - updated
PMA_sendErrorReport - updated

Fixed a couple of issues according to comments in #12457.

Signed-off-by: Volkan Ulukut arthan@gmail.com

### DIFF
--- a/test/classes/ConfigTest.php
+++ b/test/classes/ConfigTest.php
@@ -14,6 +14,7 @@ use PMA\libraries\Theme;
 
 require_once 'libraries/relation.lib.php';
 require_once 'test/PMATestCase.php';
+require_once 'libraries/Util.php';
 
 /**
  * Tests behaviour of PMA\libraries\Config class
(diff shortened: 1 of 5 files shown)


test/classes/ConfigTest.php
```diff
@@ -936,32 +936,4 @@ public function testIsGitRevision()
...
-        );
-        $this->assertFalse(
-            $this->object->checkHTTP("https://www.phpmyadmin.net/test/nothing")
-        );
-        // Use rate limit API as it's not subject to rate limiting
-        $this->assertContains(
-            '"resources"',
-            $this->object->checkHTTP("https://api.github.com/rate_limit", true)
-        );
-    }
```
nijel: Can you please adjust these tests to work on the new code? It really should be covered by tests.

ibennetch:
Thanks for submitting a pull request. Could you please use your full/legal name for the signed-off-by?

test/classes/ConfigTest.php
Seldar: Added new test cases.

nijel:
Thanks, I've merged your changes, however some more fixes were necessary:

* 5a8666c Fix allow_url_fopen request handling
* 4382446 Consistent coding style
* d52e2cf Move Util::httpRequest tests to UtilTest.php

PR #12502, opened 2016-08-26
Proper type support for mysqli_real_connect() parameters

fcorriga:
HHVM is complaining about argument 4 and 5 of mysqli_real_connect() must be of type string and not boolean.
PHP should not actually be allowing false as a valid alternative for false.
Empty string or null are both valid. I preferred empty string.
This PR is for stable branch.


nijel:
Thanks for spotting this, however I think it's better to remove this unused parameter completely.

PR #12524, opened 2016-09-01
Check for token mismatches only if it's a POST request

### LINKED ISSUE #12474: Broken links
currently when visiting:
    https://demo.phpmyadmin.net/master-config/setup/
all links in menu return 404.

### LINKED ISSUE #12494: Can't add new user
### Steps to reproduce
1. Go to the "User accounts" tab
2. Scroll down and press "Add new user"

### Expected behaviour
I should get the dialog page to add a new user

### Actual behaviour
Nothing happens; the page appears to reload and remains on the user overview display.

Git says:
The first bad commit could be any of:
7f43348e6b77f12fc4669f7e5defb582bde49911
ea73fded7138038aa5a415c7081d838fc094eff7
599effd165623bcdd8624092cace4556bd2655fa
dae3390a02ca6687fd31ca784474d56240c6c538
238713b452ccb8a419ab13d8604d5576b9e81bcc

devenbansod:

Checking for Token mismatches for GET requests does not make sense
because they don't contain the Token parameter.

Review and suggestions welcome.

When merged this should:
Fix #12474 
Fix #12494 

### DIFF
--- a/libraries/common.inc.php
+++ b/libraries/common.inc.php
@@ -372,56 +372,31 @@ if (PMA_checkPageValidity($_REQUEST['back'], $goto_whitelist)) {
  * could access this variables before we reach this point
  * f.e. PMA\libraries\Config: fontsize
  *
+ * Check for token mismatch only if the Request method is POST
+ * GET Requests would never have token and therefore checking
+ * mis-match does not make sense
+ *
  * @todo variables should be handled by their respective owners (objects)
  * f.e. lang, server, collation_connection in PMA\libraries\Config
  */
-$token_mismatch = true;
-$token_provided = false;
-if (PMA_isValid($_REQUEST['token'])) {
-    $token_provided = true;
-    $token_mismatch = ! hash_equals($_SESSION[' PMA_token '], $_REQUEST['token']);
-}
 
-if ($token_mismatch) {
-    /**
-     *  List of parameters which are allowed from unsafe source
-     */
-    $allow_list = array(
-        /* needed for direct access, see FAQ 1.34
-         * also, server needed for cookie login screen (multi-server)
-         */
-        'server', 'db', 'table', 'target', 'lang',
-        /* Session ID */
-        'phpMyAdmin',
-        /* Cookie preferences */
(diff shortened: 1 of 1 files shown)


libraries/common.inc.php
```diff
@@ -372,56 +372,63 @@
...
+            /* Permit to log out even if there is a token mismatch */
+            'old_usr',
+            /* Permit redirection with token-mismatch in url.php */
+            'url',
+            /* Permit session expiry flag */
+            'session_expired',
+            /* JS loading */
+            'scripts', 'call_done',
+            /* Navigation panel */
+            'aPath', 'vPath', 'pos', 'pos2_name', 'pos2_value', 'searchClause', 'searchClause2'
```
nijel: The list will have to be cleaned up afterwards as well, maybe even completely removed as the whitelist generally was about allowing tokenless GET requests at certain places.

nijel:
I think it looks good, there is further potential for cleanup which I've commented above.

libraries/common.inc.php
devenbansod: So, removing the `$whitelist` would mean that in all the code here, we only set the value of `$token_mismatch` for use later in the file [here] (https://github.com/devenbansod/phpmyadmin/blob/c5cc3d9cf42b039cc51b6bf07100c8fef0be9500/libraries/common.inc.php#L913) which would handle returning 'Token mismatch' for Ajax requests, as the function call to `removeRequestVars` on L430 would not make much sense without the `$whitelist` defined.

What would be the case when someone does a *non-ajax POST request with invalid data* (may be a malicious user)? 

1. I suppose all files, before any operation, check if the `ajax_request` parameter is set in the request, and therefore would not affect the security even if we don't take any action. 
OR 
2. We can remove *all* the parameters using removeRequestVars.

nijel: I think we could remove all request vars in case of POST request without valid token as this is something what should not happen.

nijel:
I think it's now good to merge, however we should probably review remaining usages of token_mismatch and token_provided variables as they now do not have to be set.

PR #12532, opened 2016-09-03
Fix #12531 : Properly flag queries as DROP DATABASE

### LINKED ISSUE #12531: Properly flag queries as DROP DATABASE
Currently several SQL queries are not properly flagged as DROP DATABASE:

* DROP SCHEMA IF EXISTS DBNAME
* DROP /*! DATABASE */ DBNAME
* DROP DATABASE\`DBNAME\`

The SQL Parser recognizes all of these correctly, but the flags set by SQL parser are not used correctly while flagging these queries.

Moved here from https://github.com/phpmyadmin/sql-parser/issues/54

devenbansod:

Parser recognizes various DROP statements correctly, but the flags set by SQL Parser are incorrectly used.
In js/functions.js, removed the Regex-based matching, rather handle it in PHP as we have correct flags available from SQL Parser

### DIFF
--- a/js/functions.js
+++ b/js/functions.js
@@ -629,8 +629,7 @@ function confirmLink(theLink, theSqlQuery)
 } // end of the 'confirmLink()' function
 
 /**
- * Displays an error message if a "DROP DATABASE" statement is submitted
- * while it isn't allowed, else confirms a "DROP/DELETE/ALTER" query before
+ * Confirms a "DROP/DELETE/ALTER" query before
  * submitting it if required.
  * This function is called by the 'checkSqlQuery()' js function.
  *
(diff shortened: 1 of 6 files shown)


js/functions.js
```diff
@@ -648,25 +647,14 @@ function confirmQuery(theForm1, sqlQuery1)
         return true;
     }
 
-    // "DROP DATABASE" statement isn't allowed
-    if (PMA_messages.strNoDropDatabases !== '') {
-        var drop_re = new RegExp('(^|;)\\s*DROP\\s+(IF EXISTS\\s+)?DATABASE\\s', 'i');
-        if (drop_re.test(sqlQuery1.value)) {
-            alert(PMA_messages.strNoDropDatabases);
```
nijel: Probably strNoDropDatabases can be removed from javascript now...

js/messages.php
```diff
@@ -37,9 +37,6 @@ function () {
 );
 
 $js_messages['strNoDropDatabases'] = __('"DROP DATABASE" statements are disabled.');
```
nijel: ...remove this as well, as AFAIK there are no users left here.

libraries/import.lib.php
```diff
@@ -189,71 +184,83 @@ function PMA_importRunQuery($sql = '', $full = '', &$sql_data = array())
...
+
+        // parse sql query
+        include_once 'libraries/parse_analyze.lib.php';
+        list(
+            $analyzed_sql_results,
+            $db,
+            $table_from_sql
+        ) = PMA_parseAnalyze($sql_query, $db);
+        // @todo: possibly refactor
+        extract($analyzed_sql_results);
```
nijel: Can't we skip the extract here?

nijel:
Thanks, looks great now.

PR #12541, opened 2016-09-06
Fixes #12470 Stores tab_id hash into a variable instead of location.hash

### LINKED ISSUE #12470: Store 'tab_' + tab_id into a variable
as the location.hash is dynamic value, it's better to store:
'tab_' + tab_id
into a variable and use it in:
    https://github.com/phpmyadmin/phpmyadmin/blob/b24347ca8f388c5c8c832a2e3e9af91b7a4cb704/js/config.js#L591
for example:
    var h = 'tab_' + tab_id;
    location.hash = h;
    $this.parent().find('input[name=tab_hash]').val(h);

sudheesh001:
Signed-off-by: Sudheesh Singanamalla sudheesh1995@outlook.com


nijel:
Thanks!

PR #12544, opened 2016-09-06
Fix #12530: Show enabled links for Edit and Execute after proper priv checks

### LINKED ISSUE #12530: Function "Edit routine" crashes when the current user is not the definer, even if privileges are adequate
### Steps to reproduce
1. Create a database `foo` and two users, `bar1` and `bar2`. Grant all privileges on `foo.*` to each user.
2. Create two procedures as follows:
```
USE foo;
DELIMITER $$
CREATE DEFINER=bar1@localhost PROCEDURE do1 () 
BEGIN
    SELECT 'Hello, ' AS `Greeting`, USER() AS `user` FROM DUAL;
END$$
CREATE DEFINER=bar2@localhost PROCEDURE do2 () 
BEGIN
    SELECT 'Hello, ' AS `Greeting`, USER() AS `user` FROM DUAL;
END$$
DELIMITER ;
```
Verify that both users can call each procedure when logged in.
3. Log into phpMyAdmin as `bar1`, click on the database `foo` and on the tab `Routines`.

### Expected behaviour
There should be four links under the `Action` heading: `Edit`, `Execute`, `Export`, and `Drop` for both procedures `do1` and `do2`.

### Actual behaviour
pma_4 6 4_routines_foo

There is no `Execute` link for `do2` showing if logged in as bar1, and if logged in as bar2, it is missing on `do1`. However, both procedures are callable by either user as an SQL `CALL` statement in the SQL command window.

But if `Edit` is clicked for the procedure where the `Execute` link is missing, or if the link to the procedure is clicked in the navigation bar on the left, an unhandled exception is thrown which will be shown if you have your PHP environment set up to display errors and if PHP error_reporting = E_ALL.

See screenshot below.
pma_4 6 4_crash_routines_foo

Presumably, the `Edit` link should be missing instead of the `Execute` link. But I am not at all sure what the intended behavior should be for clicking on the navigation bar link.

If logged in as `root`, editing either routine works normally, as expected.

### Server configuration
**Operating system**: Linux Ubuntu 16.04 LTS (64-bit)

**Web server:** Apache2 (localhost / UNIX socket)

**Database:** MySQL 5.7.13-0ubuntu0.16.04.2 - (Ubuntu)

**PHP version:** 7.0.8-0ubuntu0.16.04.2

**phpMyAdmin version:** 4.6.4deb1+deb.cihar.com~xenial.1

### Client configuration
**Browser:** Mozilla Firefox 48.0

**Operating system:** Linux Ubuntu 16.04 LTS (64-bit)

devenbansod:

Fix #12530: Show enabled links for Edit and Execute after proper priv checks

### DIFF
--- a/libraries/rte/rte_export.lib.php
+++ b/libraries/rte/rte_export.lib.php
@@ -44,8 +44,7 @@ function PMA_RTE_handleExport($export_data)
     } else {
         $_db = htmlspecialchars(PMA\libraries\Util::backquote($db));
         $message  = __('Error in processing request:') . ' '
-                  . sprintf(PMA_RTE_getWord('not_found'), $item_name, $_db)
-                  . ' OR ' . PMA_RTE_getWord('no_view');
+                  . sprintf(PMA_RTE_getWord('no_view'), $item_name, $_db);
         $message = Message::error($message);
 
         if ($GLOBALS['is_ajax_request'] == true) {
--- a/libraries/rte/rte_export.lib.php
+++ b/libraries/rte/rte_export.lib.php
@@ -44,15 +44,17 @@ function PMA_RTE_handleExport($export_data)
     } else {
         $_db = htmlspecialchars(PMA\libraries\Util::backquote($db));
         $message  = __('Error in processing request:') . ' '
-                  . sprintf(PMA_RTE_getWord('not_found'), $item_name, $_db);
-        $response = Message::error($message);
+                  . sprintf(PMA_RTE_getWord('not_found'), $item_name, $_db)
+                  . ' OR ' . PMA_RTE_getWord('no_view');
+        $message = Message::error($message);
+
         if ($GLOBALS['is_ajax_request'] == true) {
             $response = PMA\libraries\Response::getInstance();
             $response->setRequestStatus(false);
             $response->addJSON('message', $message);
             exit;
         } else {
-            $response->display();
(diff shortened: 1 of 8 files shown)


libraries/rte/rte_export.lib.php
```diff
@@ -44,15 +44,17 @@ function PMA_RTE_handleExport($export_data)
     } else {
         $_db = htmlspecialchars(PMA\libraries\Util::backquote($db));
         $message  = __('Error in processing request:') . ' '
-                  . sprintf(PMA_RTE_getWord('not_found'), $item_name, $_db);
-        $response = Message::error($message);
+                  . sprintf(PMA_RTE_getWord('not_found'), $item_name, $_db)
+                  . ' OR ' . PMA_RTE_getWord('no_view');
```
nijel: This will look weird when localized. Try avoiding constructing sentences out of localized components.

libraries/rte/rte_routines.lib.php
```diff
@@ -161,6 +161,8 @@ function PMA_RTN_handleEditor()
                 ),
                 htmlspecialchars(PMA\libraries\Util::backquote($db))
             );
+            $message .= ' OR ' . PMA_RTE_getWord('no_edit');
```
nijel: ...Same here.

PR #12548, opened 2016-09-09
Fix Export and Copy of Tables with Generated/Virtual columns

### LINKED ISSUE #12221: Bad table export in case of generated columns
### Steps to reproduce
Assume we have generated column idGenerated in table1 and phpmyadmin 4.6.0 exports it as:
```
CREATE TABLE `table1` (
  `id` int(11) NOT NULL,
  `idGenerated` int(11) GENERATED ALWAYS AS ((`id` > 10)) STORED

INSERT INTO `table1` (`id`, `idGenerated`) VALUES
(1, 0),
(11, 1);
```
When importing the data, database says:
`#3105 - The value specified for generated column 'idGenerated' in table 'table1' is not allowed.`

Looks very close to #11909.

**MySQL 5.7.11**

### LINKED ISSUE #12518: Unable to copy database with virtual generated columns... / MySQL Error #3105
### Steps to reproduce
1. Create table
`CREATE TABLE Table_01 (
id int(11) NOT NULL,
name varchar(255) COLLATE utf8mb4_unicode_ci NOT NULL,
attributes json NOT NULL
) ENGINE=MyISAM DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;`

2. Alter table with virtual generated colum:
`ALTER TABLE Table_01 ADD COLUMN test VARCHAR(30) GENERATED ALWAYS AS (json_unquote(json_extract(attributes,'$.test'))) VIRTUAL;`

3. Use phpMyAdmin to copy the table to Table_02, choose structure **and** data. The following error will appear: `INSERT INTO test.Table_02 SELECT * FROM test.Table_01` / #3105 - The value specified for generated column 'test' in table 'Table_02' is not allowed.

### Expected behaviour
I just want a table to be copied...

### Actual behaviour
MySql gives an error... We're using virtually generated tables based on the blog found on [https://www.percona.com/blog/2016/03/07/json-document-fast-lookup-with-mysql-5-7/](url). Unfortunately, phpMyAdmin can't deal with this :((

### Server configuration

**Web server:** Apache/2.4.12 (Ubuntu)

**Database:** 5.7.12-1~exp1+deb.sury.org~wily+1 - (Ubuntu)

**PHP version:** 7.1.0-dev

**phpMyAdmin version:** 4.6.4 (latest version... which is said to be compatible with MySQL 5.5 and newer...)

### Client configuration
**Browser:** Google Chrome Versie 52.0.2743.116 m

**Operating system:** Windows 10 Pro

devenbansod:

We include only the _non-generated columns_ insert using `INSERT INTO .. SELECT .. FROM ..` while copying data.
In Export too, we export the data of only the non-generated columns
Fix #12221
Fix #12518

Similar [bug](https://bugs.mysql.com/bug.php?id=79148) was present in mysqldump's exports before MySQL version 5.7.9 too. MariaDB allows INSERT with generated columns but generates a warning saying those values were ignored. With the import of our dumps, that warning also would not be generated.

### DIFF
--- a/libraries/Table.php
+++ b/libraries/Table.php
@@ -1511,8 +1511,13 @@ class Table
             && $columns_meta_query_result !== false
         ) {
             foreach ($columns_meta_query_result as $column) {
+                $value = $column['Field'];
+                if ($backquoted === true) {
+                    $value = Util::backquote($value);
+                }
+
                 if (strpos($column['Extra'], 'GENERATED') === false) {
-                    array_push($ret, Util::backquote($column['Field']));
+                    array_push($ret, $value);
                 }
             }
         }
--- a/libraries/Table.php
+++ b/libraries/Table.php
@@ -1017,10 +1017,17 @@ class Table
             $GLOBALS['dbi']->query($sql_set_mode);
             $GLOBALS['sql_query'] .= "\n\n" . $sql_set_mode . ';';
 
-            $sql_insert_data = 'INSERT INTO ' . $target
-                . ' SELECT * FROM ' . $source;
-            $GLOBALS['dbi']->query($sql_insert_data);
-            $GLOBALS['sql_query'] .= "\n\n" . $sql_insert_data . ';';
+            $_old_table = new Table($source_table, $source_db);
+            $nonGeneratedCols = $_old_table->getNonGeneratedColumns(true);
+            if (count($nonGeneratedCols) > 0) {
+                $sql_insert_data = 'INSERT INTO ' . $target . '('
+                    . implode(', ', $nonGeneratedCols)
+                    . ') SELECT ' . implode(', ', $nonGeneratedCols)
+                    . ' FROM ' . $source;
+
+                $GLOBALS['dbi']->query($sql_insert_data);
+                $GLOBALS['sql_query'] .= "\n\n" . $sql_insert_data . ';';
+            }
         }
 
         PMA_getRelationsParam();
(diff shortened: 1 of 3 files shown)


libraries/Table.php
```diff
@@ -1017,10 +1017,17 @@ static public function moveCopy($source_db, $source_table, $target_db,
...
-            $sql_insert_data = 'INSERT INTO ' . $target
-                . ' SELECT * FROM ' . $source;
-            $GLOBALS['dbi']->query($sql_insert_data);
-            $GLOBALS['sql_query'] .= "\n\n" . $sql_insert_data . ';';
+            $_old_table = new Table($source_table, $source_db);
+            $nonGeneratedCols = $_old_table->getNonGeneratedColumns(true);
+            if (count($nonGeneratedCols) > 0) {
+                $sql_insert_data = 'INSERT INTO ' . $target . '('
+                    . implode(', ', $nonGeneratedCols)
+                    . ') SELECT ' . implode(', ', $nonGeneratedCols)
```
nijel: The columns need to be backquoted here, otherwise it will break and cause SQL injection.

libraries/Table.php
```diff
@@ -1485,6 +1492,35 @@ public function getColumnsMeta()
...
+        $columns_meta_query_result = $this->_dbi->fetchResult(
+            $columns_meta_query
+        );
+
+        if ($columns_meta_query_result
+            && $columns_meta_query_result !== false
+        ) {
+            foreach ($columns_meta_query_result as $column) {
+                if (strpos($column['Extra'], 'GENERATED') === false) {
+                    array_push($ret, Util::backquote($column['Field']));
```
nijel: ...ah it's backquoted here. So that's probably fine as well, just $backquoted parameter doesn't have any effect :-).

PR #12563, opened 2016-09-18
Fix error check of a routine execution

tfaller:
Because a routine execution is done by a multi query, all results need to be checked for errors.

Signed-off-by: Thomas Faller tfaller1@gmx.de

### DIFF
--- a/libraries/rte/rte_routines.lib.php
+++ b/libraries/rte/rte_routines.lib.php
@@ -1416,37 +1416,41 @@ function PMA_RTN_handleExecute()
 
                 $GLOBALS['dbi']->freeResult($result);
 
-            } while ($GLOBALS['dbi']->nextResult());
+            } while ($outcome = $GLOBALS['dbi']->nextResult());
 
-            $output .= "</fieldset>";
+            if ($outcome) {
+                $output .= "</fieldset>";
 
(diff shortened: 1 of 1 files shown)


libraries/rte/rte_routines.lib.php
```diff
@@ -1416,7 +1416,10 @@ function PMA_RTN_handleExecute()
 
                 $GLOBALS['dbi']->freeResult($result);
 
-            } while ($GLOBALS['dbi']->nextResult());
+            } while ($outcome = $GLOBALS['dbi']->nextResult());
```
nijel: $outcome will be always false at the end of this loop, is this really what you wanted to achieve?

tfaller: That is not the case.

https://github.com/tfaller/phpmyadmin/blob/037bddd7c3560c1f0e81730d815302f67f98d5bc/libraries/rte/rte_routines.lib.php#L1411

The loop will be left by a `break` if no more results are available.
$outcome will be only false if `$GLOBALS['dbi']->moreResults()` was true and an error occured.

nijel: Ah okay, sorry for confusion.

nijel:
Merged, thanks!

PR #12564, opened 2016-09-18
Properly escape strings in MySQL statement values

### LINKED ISSUE #12453: exported SQL is corrupted
### Steps to reproduce
1. create a new row with the following SQL: [insert.sql.txt](https://github.com/phpmyadmin/phpmyadmin/files/413081/insert.sql.txt)
2. copy the html content from [html-content.html.txt](https://github.com/phpmyadmin/phpmyadmin/files/413080/html-content.html.txt) and paste it into the column "html_content" of the row created in step 1 (click on the row in phpmyadmin and paste the content; do NOT use SQL for that)
3. export the row as SQL (or a whole DB) - for simplicity you can skip this step and use the file attached [update-after-export.sql.txt](https://github.com/phpmyadmin/phpmyadmin/files/413082/update-after-export.sql.txt)
4. try to feed the exported SQL back into phpmyadmin (click SQL tab and paste the SQL) - you can use the following update SQL: [update-after-export.sql.txt](https://github.com/phpmyadmin/phpmyadmin/files/413082/update-after-export.sql.txt)
5. the import will fail as the exported SQL is malformed, for both update and insert

Exporting DB/table with mysql cli will produce a valid SQL.
### Expected behaviour

Export should produce a valid SQL.
### Actual behaviour

**Error after running the SQL**
Static analysis:

1 errors were found during analysis.

```
Ending quote ' was expected. (near "" at position 3947)
```

MySQL said: Documentation

```
1064 - You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ''<!-- etm:B10:start -->\n<table data-etm-rname="B10" data-etm-rtype="row" class=' at line 1
```
### Server configuration

**Operating system**:
Debian latest stable
Redhat 6.8

**Web server:**
Apache/2.4.10
Apache/2.2.15

**Database:**
MySQL 5.6.30

**PHP version:**
PHP 7.0.9

**phpMyAdmin version:**
4.6.3
4.7.0-dev

**please see attached files required to reproduce the issue**

[insert.sql.txt](https://github.com/phpmyadmin/phpmyadmin/files/413081/insert.sql.txt)
[html-content.html.txt](https://github.com/phpmyadmin/phpmyadmin/files/413080/html-content.html.txt)
[update-after-export.sql.txt](https://github.com/phpmyadmin/phpmyadmin/files/413082/update-after-export.sql.txt)

devenbansod:

Use *_real_escape string functions provided by connectors to escape strings while exporting

Fix #12453

### DIFF
--- a/libraries/DatabaseInterface.php
+++ b/libraries/DatabaseInterface.php
@@ -489,10 +489,9 @@ class DatabaseInterface
                     $needAnd = false;
                     if ($table || (true === $tbl_is_group)) {
                         $sql .= " `Name` LIKE '"
-                            . 
-                            // Util::escapeMysqlWildcards(
+                            . Util::escapeMysqlWildcards(
                                 $this->escapeString($table, $link)
-                            // )
+                            )
                             . "%'";
                         $needAnd = true;
                     }
--- a/libraries/DatabaseInterface.php
+++ b/libraries/DatabaseInterface.php
@@ -428,7 +428,7 @@ class DatabaseInterface
             // comparison (if we are looking for the db Aa we don't want
             // to find the db aa)
             $this_databases = array_map(
-                'PMA\libraries\Util::sqlAddSlashes',
+                array($this, 'escapeString'),
                 $databases
             );
 
--- a/libraries/DatabaseInterface.php
+++ b/libraries/DatabaseInterface.php
@@ -307,13 +307,13 @@ class DatabaseInterface
             if (true === $tbl_is_group) {
                 $sql_where_table = 'AND t.`TABLE_NAME` LIKE \''
                     . Util::escapeMysqlWildcards(
-                        Util::sqlAddSlashes($table)
+                        $GLOBALS['dbi']->escapeString($table)
                     )
                     . '%\'';
             } else {
                 $sql_where_table = 'AND t.`TABLE_NAME` '
                     . Util::getCollateForIS() . ' = \''
-                    . Util::sqlAddSlashes($table) . '\'';
+                    . $GLOBALS['dbi']->escapeString($table) . '\'';
             }
         } else {
             $sql_where_table = '';
@@ -489,9 +489,10 @@ class DatabaseInterface
                     $needAnd = false;
                     if ($table || (true === $tbl_is_group)) {
                         $sql .= " `Name` LIKE '"
-                            . Util::escapeMysqlWildcards(
-                                Util::sqlAddSlashes($table, true)
-                            )
+                            . 
+                            // Util::escapeMysqlWildcards(
+                                $this->escapeString($table, $link)
+                            // )
                             . "%'";
                         $needAnd = true;
                     }
@@ -851,7 +852,7 @@ class DatabaseInterface
             // get table information from information_schema
             if (! empty($database)) {
                 $sql_where_schema = 'WHERE `SCHEMA_NAME` LIKE \''
-                    . Util::sqlAddSlashes($database) . '\'';
+                    . $this->escapeString($database, $link) . '\'';
             } else {
                 $sql_where_schema = '';
             }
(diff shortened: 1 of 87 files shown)


libraries/dbi/DBIDummy.php
```diff
@@ -1256,4 +1256,17 @@ public function fieldFlags($result, $i)
...
+     * returns properly escaped string for use in MySQL queries
+     *
+     * @param resource $link the connection object
+     * @param string   $str  value (as a string)
+     *
+     * @return string escaped string from $str
+     */
+    public function escapeString($link, $str)
+    {
+        return '';
```
nijel: IMHO it return either original or MySQL quoted string here, otherwise it's pretty confusing.

libraries/DatabaseInterface.php
```diff
@@ -2649,6 +2649,19 @@ public function fieldFlags($result, $i)
...
+     * returns properly escaped string for use in MySQL queries
+     *
+     * @param resource $link the connection object
+     * @param string   $str  value (as a string)
+     *
+     * @return string escaped string from $str
+     */
+    public function escapeString($link, $str)
+    {
+        return $this->_extension->escapeString($link, $str);
```
nijel: I'd prefer to have link optional and use $link = $this->getLink($link);. It's really not needed in most cases...

libraries/plugins/export/ExportSql.php
```diff
@@ -2411,10 +2411,13 @@ public function exportData(
...
-                            $search,
-                            $replace,
-                            Util::sqlAddSlashes($row[$j])
+                        . $GLOBALS['dbi']->escapeString(
+                            $GLOBALS['dbi']->getLink(),
+                            str_replace(
+                                $search,
+                                $replace,
+                                Util::sqlAddSlashes($row[$j])
+                            )
```
nijel: Is it intentional to use escapeString and Util::sqlAddSlashes? IMHO it will produce double escaped slashes.

nijel (changes requested):
Overall I'm not really sure how newly introduced DBI::escapeString relates to Util::sqlAddSlashes. They both IMHO do pretty much same thing.

libraries/DatabaseInterface.php
devenbansod: Changes done in the latest commit.

libraries/plugins/export/ExportSql.php
devenbansod: Indeed it would have been wrong. I have removed the extra call in the new commit.

devenbansod:
I feel that `sqlAddSlashes` does not take care of double quotes inside a SQL statement properly, while use of this inbuilt function(s) would do that.

nijel:
So probably all uses of sqlAddSlashes should be turned into this new function and sqlAddSlashes can be removed?

devenbansod:
I think so but I would welcome a second opinion here. 

The page [here] (http://php.net/manual/en/mysqli.real-escape-string.php) says it encodes the characters `NUL (ASCII 0)`, `\n`, `\r`, `\`, `'`, `"` and the backtick, while `sqlAddSlashes` is just meant to add slashes before the `'` and `\` characters and we also provide an option to encode `\r` and `\n`.

One of the [comment](http://php.net/manual/en/mysqli.real-escape-string.php#46339) on the same page provides some more info.

Also, do you think moving finally to usage of prepared statements might help us in tackling all this issues (as reported by [this person] (http://php.net/manual/en/mysqli.real-escape-string.php#102639) )?

nijel:
Using prepared statements would be way better than constructing the SQL queries manually as we do right now. AFAIK it would mean dropping support for the mysql extension, but that really should not matter these days...

As for escaping, the null byte probably should be escaped as well, but as it won't appear in the user input, it's not properly handled by our code. For the newlines I think it should be there as well automatically. It was introduced by 280336063.

devenbansod:
Would that mean, for the time till we move completely to prepared statements, we can use this instead of Util::sqlAddSlashes ? 
Also, if we decide to replace all sqlAddSlashes calls with this inbuilt-function, would it be okay to do it in QA_4_6 (like would it be a major change that may break some other thing ?) or should we include this small change for fixing the export bug in QA_4_6 and then target master for the complete replacement ?

nijel:
I think using mysql(i)_real_escape_string instead of sqlAddSlashes is correct approach (until we use prepared statements). I'm fine with doing this change in QA_4_6.

devenbansod:
Okay. Thanks for the comments. 
I will go ahead and make the changes across other places and would let you know once I feel the PR is ready.

devenbansod:
Changes on the way. 
Many tests will fail for now, as I would have to add the mock and stubs with the Mock object of `dbi`.

devenbansod:
Finally the tests are adjusted. :-) 
I think now it's ready for a review.

nijel:
I think it looks good, there are still some things to fix:

* There is still one usage left in DatabseInterface::getTablesFull (called by string)
* Another ones in libraries/rte/rte_list.lib.php
* I guess test/libraries/common/PMA_quoting_slashing_test.php should be migrated to test for this new code
* Why not remove sqlAddSlashes code?

devenbansod:
I have removed the old code (old tests for now as well) and fixed the remaining calls.

But I am not really sure how to cover the `escapeString` function in the tests, since we would have to again add a mock test-double for `GLOBALS['dbi']` to handle the `escapeString` call to the databaseInterface and define the escaped strings that it should return (using may be a `returnValueMap`). 

Any help would be appreciated.

nijel:
Hmm, it's true that it's now external code and hard to test...

nijel:
Okay, merging now. It would be great if you could review it in master, there were quite some fixes needed.

PR #12616, opened 2016-10-04
Refuse to work with mbstring.func_overload enabled

nijel:

In several places our code assumes that strlen does return length in
bytes and not in chars. This can be different with
mbstring.func_overload and will lead to several hard to debug breakages.
Therefore it's safer to simply deny using phpMyAdmin with this setting.

### DIFF
--- a/libraries/common.inc.php
+++ b/libraries/common.inc.php
@@ -530,6 +530,20 @@ if ($GLOBALS['PMA_Config']->error_config_default_file) {
     trigger_error($error, E_USER_ERROR);
 }
 
+/**
+ * As we try to handle charsets by ourself, mbstring overloads just
+ * break it, see bug 1063821.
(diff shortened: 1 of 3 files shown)


libraries/common.inc.php
```diff
@@ -530,6 +530,20 @@
     trigger_error($error, E_USER_ERROR);
 }
 
+/**
+ * As we try to handle charsets by ourself, mbstring overloads just
+ * break it, see bug 1063821.
+ */
+if (@extension_loaded('mbstring') && @ini_get('mbstring.func_overload') != '0') {
```
nijel: Note that this is intentionally stricter than previous check. I really don't think anybody should enable this setting system wide.

nijel:
Merging as there were no objectsion.

Issue #12234, opened 2016-05-06
Designer Export Schema 414


sukhbirgs:
### Steps to reproduce
1. Create relations in Designer, save page
2. Designer Menu -> Select  Export schema 
3. From the popup window select desired options 
    Type : PDF
    Show Color : Check
    Orientation : Landscape
    Paper Size: A4

Any option combination can be used.   
Please see attached images.

### Expected behaviour
Expecting a pdf file with colored relationship er diagram.

### Actual behaviour

On hitting GO the resulting window throws 414 Request-URI Too Long. 
Please see attached images.

### Server configuration
**Operating system**:
Centos 7

**Web server:**
Apache 2.4.6

**Database:**
5.5.47-MariaDB

**PHP version:**
5.6.5
**phpMyAdmin version:**
4.5.5.1  & 4.6.1

### Client configuration
**Browser:**
Chrome 50.0.2661.94

**Operating system:**
Windows 8
![1](https://cloud.githubusercontent.com/assets/74910/15084827/5904bb1c-1393-11e6-85c3-3d0203c1c0ef.png)

![2](https://cloud.githubusercontent.com/assets/74910/15084831/5c0d0a1c-1393-11e6-8197-eab91de8b650.png)

![3](https://cloud.githubusercontent.com/assets/74910/15084832/602a6da6-1393-11e6-98ae-06f5a6cc1b5e.PNG)

PR #12626, opened 2016-10-09
Fix bug #12234 No page indicator in view

madhuracj:

### DIFF
--- a/libraries/DisplayResults.php
+++ b/libraries/DisplayResults.php
@@ -1077,12 +1077,12 @@ class DisplayResults
 
         $maxRows = $_SESSION['tmpval']['max_rows'];
         $onsubmit = 'onsubmit="return '
-            . ($_SESSION['tmpval']['pos']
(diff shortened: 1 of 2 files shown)


devenbansod:
It works well normally. But I could find some glitches while using this (with default values for `MaxExact*` in config). 
* The 'Go to Last page' button gives  an error in SQL syntax as it tries `LIMIT -25, 25`. 
* My View has 200 rows, once I reach the end (i.e. currently showing 176-200) and press `Next page` button (which in a Table's scenario we don't show), I get an empty result which might get slightly incomprehensible for the user. 

I am not sure if we can actually get over these glitches with the constraints, but if we can it would be great.

madhuracj:
The first issue was due to an existing bug and is fixed with 088d2c3. I do not think we can get over the second.

nijel:
Thanks, I think having this is better than nothing, we can improve user experience on last page later...

PR #12631, opened 2016-10-12
Fix order of testing DBConnection with different parameters in Config method

### LINKED ISSUE #12472: Setup issues
in SETUP when config authentication is selected [with password] & "Connect without password" is selected
the index page of setup will show DSN:
    mysqli://root@localhost    
instead of:
    mysqli://root:***@localhost
(as the retry without password is occur after trying to use password)

devenbansod:

In setup, try to follow what the [docs](https://docs.phpmyadmin.net/en/latest/config.html#cfg_Servers_nopassword) say.
'Password login is still tried first, but as fallback, no password method is tried.'

Fix #12472


nijel:
Merged, thanks! (The retry is gone in master, I've adjusted the changes to it there)

PR #12647, opened 2016-10-18
Use justBrowsing only if outputting the results from a single table

### LINKED ISSUE #12365: Error on displaying total number of records when executing a select query on a table which contains large number of records
### Steps to reproduce
1. Create a database with several number of tables, some of which contains large number of records. (something about 300,000)
**Example**:
image

2. Execute a select query on a table or joined tables  (say T), which will produce a resulting table (say R) with large number of records but less than the number of records in T.
**Example**:
image

### Expected behaviour
Assume the number of records in T is N, number of records in R is M.
M should be less than N.
**Example**:
For the above query given, the expected total number of entries in the output is 178044. (Checked and verified with HeidiSQL client)

### Actual behaviour
Instead of showing total number of records in R, it shows total number of records in T (i.e: shows N instead of M) in the output.
**Example**:
output
300024 is the total number of records in table employees.
### Server configuration
**Operating system**:
Windows 8.1
**Web server:**
Apache/2.4.12 (Win32)
**Database:**
MySQL
**PHP version:**
PHP/5.6.8
**phpMyAdmin version:**
4.6.3 (latest)
### Client configuration
**Browser:**
Google chrome (Version 51.0.2704.103 m)
**Operating system:**
Windows 8.1

### LINKED ISSUE #12533: Incorrect selected record statement counts for complex queris using group by and derived tables
### Steps to reproduce

1. Create an INNODB table with duplicate values that can be grouped by. Run MYSQL 5.7 in DEFAULT SQL MODE (strict).

2. Run Query like this, where `Id` is the duplicate value column:

SELECT `Id`, FROM `table` GROUP BY `Id`

Note the number of reported selected rows at top (in the green box).

3. Run same query as derived table like this:

SELECT 'Id' FROM (SELECT `Id`, FROM `table` GROUP BY `Id`) as `test`

Note the number of reported selected rows at top  (in the green box) is now the full number of rows in that table.

4. Run query again, but using COUNT(*)

SELECT COUNT(*) FROM (SELECT `Id`, FROM `table` GROUP BY `Id`) as `test`

Same number of rows reported in the actual returned table COUNT(*) column as was returned in step 2 query.

NOTE: Try running queries in steps 2 and 3 with a COUNT column added in. You will see that the counted values are the same. 

### Expected behavior
Selected records statement should report original subquery selected result number.

### Actual behavior
Selected records statement reports full number of rows PRE group by instead.

### Server configuration

**Web server:** Apache/2.4.18 (Ubuntu)

**Database:**  MYSQL 5.7.13 - default sql_mode set (strict)

**PHP version:**  7.0

**phpMyAdmin version:**  4.5.4.1deb2ubuntu2

### Client configuration
**Browser:** Firefox 48

**Operating system:** Kubuntu 16.04

devenbansod:

The issue #12533 happens because we use countRecords on the parent table since it is incorrectly set,
while actually we should be looking at SQL_CALC_FOUND_ROWS instead.

Fix #12533
Fix #12365 

### DIFF
--- a/libraries/sql.lib.php
+++ b/libraries/sql.lib.php
@@ -655,7 +655,7 @@ function PMA_isJustBrowsing($analyzed_sql_results, $find_real_end)
         && empty($analyzed_sql_results['union'])
         && empty($analyzed_sql_results['distinct'])
         && $analyzed_sql_results['select_from']
-        && (count($analyzed_sql_results['select_tables']) <= 1)
+        && (count($analyzed_sql_results['select_tables']) === 1)
         && (empty($analyzed_sql_results['statement']->where)
             || (count($analyzed_sql_results['statement']->where) == 1
                 && $analyzed_sql_results['statement']->where[0]->expr ==='1'))


nijel:
Looks good.

PR #12657, opened 2016-10-22
Update ckb.po

meirusfandi:

fixing some error

### DIFF
--- a/po/ckb.po
+++ b/po/ckb.po
@@ -875,15 +875,15 @@ msgstr ""
 
 #: js/messages.php:73
 msgid "Do you really want to revoke the selected user(s) ?"
-msgstr "ئایا تۆ دڵنیایت لە جێبەجێکردنی \"%s\"؟"
+msgstr "ئایا تۆ دڵنیایت لە جێبەجێکردنی \'%s\'؟"
 
 #: js/messages.php:75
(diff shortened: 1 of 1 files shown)


po/ckb.po
```diff
@@ -2849,7 +2849,7 @@ msgstr ""
 
 #: libraries/Advisor.php:536
 #, php-format
-msgid "Unexpected character on line %1$s. Expected tab, but found \"%2$s\"."
+msgid "Unexpected character on line %1$s. Expected tab, but found \%2$s\."
```
nijel: If you want to change the source string, please edit the corresponding source file (libraries/Advisor.ph, line 536 in this case). Also what you've produced seems to be wrong (escaping % and . with \).

po/ckb.po
```diff
@@ -875,15 +875,15 @@ msgstr ""
 
 #: js/messages.php:73
 msgid "Do you really want to revoke the selected user(s) ?"
-msgstr "ئایا تۆ دڵنیایت لە جێبەجێکردنی \"%s\"؟"
+msgstr "ئایا تۆ دڵنیایت لە جێبەجێکردنی \'%s\'؟"
```
nijel: Anyway this translation is wrong as it uses format string which is not present in source....

PR #12658, opened 2016-10-22
Update ca.po

meirusfandi:
- [ x] Has proper Signed-Off-By
- [ x] Has commit message which describes it
- [ x] Is needed on it's own, if you have just minor fixes to previous commits, you can squash them
- [ x] Any new functionality is covered by tests

fix some error


nijel:
If you want to change the source strings, edit them in `libraries/advisory_rules.txt`. Anyway what error do you see here? These are names of MySQL variables and they are lowercase...

PR #12676, opened 2016-11-01
Create upgrade_column_info_4_6_5+.sql

danielgp:

- [x ] Has proper Signed-Off-By
- [x ] Has commit message which describes it
- [x ] Is needed on it's own, if you have just minor fixes to previous commits, you can squash them
- [x ] Any new functionality is covered by tests

Bookmark id that is an auto_increment column can only take positive values starting from 0, therefore should be unsigned otherwise half of allowed interval is wasted from start, and this is the fix for that

Signed-off-by: Daniel Popiniuc danielpopiniuc@gmail.com

### DIFF
--- a/sql/create_tables.sql
+++ b/sql/create_tables.sql
@@ -37,7 +37,7 @@ USE phpmyadmin;
 --
 
 CREATE TABLE IF NOT EXISTS `pma__bookmark` (
-  `id` int(11) NOT NULL auto_increment,
+  `id` int(10) unsigned NOT NULL auto_increment,
   `dbase` varchar(255) NOT NULL default '',
   `user` varchar(255) NOT NULL default '',
   `label` varchar(255) COLLATE utf8_general_ci NOT NULL default '',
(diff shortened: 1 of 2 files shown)


danielgp:
#12676 I have added the Sign-Off-By line and now everything should be ok

nijel:
Why this change? Is int limiting number of your bookmarks?

danielgp:
@nijel 
This change is meant to bring consistency for all "auto_increment" fields in the database which do have the "unassigned" flag except the bookmark one.
Having the "unassigned" flag for an "auto_increment" field is quite normal as the first value will be 1 which is giving a full use of the interval, without it 50% of the interval will be wasted right from the start, see https://dev.mysql.com/doc/refman/5.7/en/integer-types.html .
In conclusion even if the limitation of int(11) without unassigned is quite high, to 2+ billion records, it's more about consistency with all the other "auto_increment" fields, and avoid wasting 50% of the interval right from the start than bringing the limit to 4+ billion.

nijel:
Okay, as this is rather cleanup than bug fix, it should rather go to master branch instead of QA_4_6.

danielgp:
@nijel So, should I close this request and apply the merge request against master, right?

nijel:
Yes, please rebase your changes on master and open PR for that.

PR #12683, opened 2016-11-05
Fix #12681

### LINKED ISSUE #12681: Symbol $ in table names passed incorrect from "designer" to "table structure editor"
### Steps to reproduce
1. Go to designer
2. Find table with "$" in table name
3. Press cog-icon in table header

### Expected behaviour
Go to table with $ in name

### Actual behaviour
Go to table with %24 in name

devenbansod:

If used in encoded form, they are shown to be non-existent while exporting or moving to the structure of particular table from the designer view.

It would be great if someone can double-check and confirm that this change would not lead to a security flaw. 

Fix #12681 

Signed-off-by: Deven Bansod <devenbansod.bits@gmail.com>


nijel:
I agree that urlencode is not needed here...

PR #12688, opened 2016-11-09
Bookmark id that is an auto_increment column can only take positive v…

danielgp:

Bookmark ID from MySQL database is an auto_increment column, therefore will only have positive values starting from 1, therefore should be flagged as "unsigned" otherwise half of allowed interval is wasted from start, see [MySQL Integer Types (Exact Value)](https://dev.mysql.com/doc/refman/5.7/en/integer-types.html).
Current changes provide a fix for that and aligned bookmark table with other tables reaching a high consistency with MySQL database being used internally by phpMyAdmin.

Signed-Off-By: Daniel Popiniuc danielpopiniuc@gmail.com

### DIFF
--- a/sql/create_tables.sql
+++ b/sql/create_tables.sql
@@ -37,7 +37,7 @@ USE phpmyadmin;
 --
 
 CREATE TABLE IF NOT EXISTS `pma__bookmark` (
-  `id` int(11) NOT NULL auto_increment,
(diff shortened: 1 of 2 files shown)


danielgp (commented):
Signed-Off-By: Daniel Popiniuc danielpopiniuc@gmail.com

nijel:
Thanks merged with some fixes (see commits above).

PR #12690, opened 2016-11-11
Fixes #12514

### LINKED ISSUE #12514: OpenBaseDIR BUG

osaid-r:

Changed Encoding.php and File.php such that it takes temp directory from configfile.php


ragnerok:
I have a few doubts 
1. Do I have to sign off somewhere ? Cause I haven't . Although I don't plan to not agree with the agreement
2. I do not get the error that has been pointed out by the bot !

ibennetch:
Hello @ragnerok and thanks for your contribution.

Yes, for legal purposes we require agreement with the [DCO document](https://github.com/phpmyadmin/phpmyadmin/blob/master/DCO); the best way to do so is when committing in git with something like `git commit -s`. This automatically adds the proper signed-off-by line to the commit message.

The bot is stating that your commit adds a newline character at the end of libraries/config/ConfigFile.php which was probably done automatically by your editor. If I recall correctly, our template system does not deal well with newlines at the end of certain files, so the bot looks for this.

Finally, you seem to have added a few files that you have added some files that seem to be unnecessary (certainly .DS_Store, but probably also a few others), so it would be good if you could remove those from your commit.

I and the others might have some comment on specific parts of the code, but that should at least answer your questions so far. In order to fix the signed-off-by message and extra files, you can amend your existing commit and push (with --force) the changes back to your repository on github (git and the manuals warn about rewriting history, but since your change hasn't been merged such rewrites are currently harmless). That will update this pull request automatically with the changes.

Thanks for your enthusiasm and work.

ragnerok:
Hi, I am relatively new to git so I don't what to do to remove the extra files using git . What do I have to do exactly ?

libraries/config/ConfigFile.php
```diff
@@ -537,4 +537,24 @@ public function getConfigArray()
...
+    {
+        $tmp_subdir = null;
+        if (! empty($GLOBALS['cfg']['TempDir']) && @is_writable($GLOBALS['cfg']['TempDir'])) {
+        $tmp_subdir = $GLOBALS['cfg']['TempDir'];
+        } else {
+            $tmp_subdir = ini_get('upload_tmp_dir');
+            if (empty($tmp_subdir)) {
+                $tmp_subdir = sys_get_temp_dir();
+            }
+        $tmp_subdir = rtrim($tmp_subdir, DIRECTORY_SEPARATOR);
```
nijel: Can you please fix indentation in this function?

nijel (changes requested):
There is another copy of similar code in prefs_manage.php, can you please replace it with new function as well?

libraries/config/ConfigFile.php
ragnerok: yeah ok i'll work on both of these

ragnerok:
@nijel i am not able to find the similar code in prefs_manage.php

nijel:
It's slightly different, and starts here: https://github.com/phpmyadmin/phpmyadmin/blob/master/prefs_manage.php#L73

ragnerok:
I kind of get how this works , correct me if I am wrong
we'll first call the function that I have added and then if it's returned value ==$GLOBALS['cfg']['TempDir']  then the code will be same from line 83-86

nijel:
Actually it should share even more code, I've adjusted it in df66e4fa080e2aa938bc415a5be9bfe7af611301.

nijel:
Thanks, your contributions have been merged as b8a9b0f30d3d83746967ef457cca641e014c32a5 (after rebasing and squashing the commits). Additional fix was needed to make it work: 84971eaf4f1b47a148e60a0042291e5ccb45ea3d

PR #12693, opened 2016-11-13
Support TokuDB Maintainance Operations

shtse8:


shtse8:
What is Current Coverage? I have no ideas.

ibennetch:
Thanks for submitting your pull request.

The codecov-io bot is referring to coverage in unit tests (which are in the `tests` subdirectory); it always chimes in on pull requests to tell us what percentage of code is currently covered by tests. This is particularly useful if there's a large code contribution that doesn't include tests because then the bot complains and someone fixes it by writing tests. In this case, it's probably okay how it is (ultimately we should have a full test suite against the TokuDB engine, but it doesn't look like such tests currently exist).

shtse8:
Thanks for your explanation.

phpMyAdmin still can't support TukoDB. We are hard to find a way (except SQLs) to maintain TokuDB tables using phpMyAdmin. Creating Clustered Indices, Analyze Single Table, Change Compression Format. Will it support this storage engine in the future?

shtse8:
Should I update the test script to improve the coverage?

libraries/controllers/table/TableIndexesController.php
```diff
@@ -146,8 +146,18 @@ public function doSaveDataAction()
                     )
             );
         } elseif (!$error) {
-
-            $this->dbi->query($sql_query);
+            
+            /* Support For Multiple Querys returned by getSqlQueryForIndexCreateOrEdit */
+            if ( substr_count($sql_query, ";") > 1 ) {
+                while( ( $pos = strpos($sql_query, ";") ) !== false) {
```
nijel: I think this will break when the table name contains `;`. Maybe it would be better to let `getSqlQueryForIndexCreateOrEdit` return array of queries?

libraries/operations.lib.php
```diff
@@ -1869,10 +1869,12 @@ function PMA_setGlobalVariablesForEngine($tbl_storage_engine)
     $is_innodb = ($tbl_storage_engine == 'INNODB');
     $is_berkeleydb = ($tbl_storage_engine == 'BERKELEYDB');
     $is_pbxt = ($tbl_storage_engine == 'PBXT');
+    $is_tokudb = ($tbl_storage_engine == 'TOKUDB');
 
     return array(
         $is_myisam_or_aria, $is_innodb, $is_isam,
-        $is_berkeleydb, $is_aria, $is_pbxt
+        $is_berkeleydb, $is_aria, $is_pbxt,
+        $is_tokudb
```
nijel: This is really becoming ridiculous, but it's not your fault, created #12703 to track this.

nijel (changes requested):

libraries/controllers/table/TableIndexesController.php
shtse8: I have the same thought too, but I don't know will it affect others if I've changed the return of `getSqlQueryForIndexCreateOrEdit`. Let me check.

libraries/operations.lib.php
shtse8: Yes, I'm agree. It should return `$engine` instead. More and more engines are coming out and phpMyAdmin should work for supporting all.

libraries/controllers/table/TableIndexesController.php
nijel: I think having it return array in all cases and fix consumers to deal with it is cleanest approach.

libraries/operations.lib.php
nijel: BTW: This is being addressed in https://github.com/phpmyadmin/phpmyadmin/pull/12719

nijel:
Any chance to get this updated on current master?

nijel:
I've applied most of this patch in fa051ff3e8 and 2b6e5f892966302281d623e429cd61945392bbb7. The only missing are changes to libraries/Table.php, which duplicated quite a lot of code and I'm not really sure what is desired outcome there.

PR #12698, opened 2016-11-15
Added 'view changelog' to newer version alert

jameswolensky:


nijel:
The local changelog.php shows changelog for current version, not the latest one, so I think such link is pretty useless...

PR #12719, opened 2016-11-19
Fixes #12703

### LINKED ISSUE #12703: Refactor PMA_setGlobalVariablesForEngine
Currently `PMA_setGlobalVariablesForEngine` returns long list of variables which are then passed around the code. This is really bad design patter, it should be made more transparent to the users. One of possibilities is to move the code to the Table class and use it from there (with caching).

See also #12567

arimourao:
This commit refactors the over use of global variables to identify tables storage engine.


arimourao:
I noticed an obvious error in the isEngine method of the Table class. Please, close this PR, I will push another one later. Although it is interesting to notice that I was unable to detect this error running tests in my machine and also the checks of this PR was unable to detect it.

arimourao:
Ok, I think it is good to go. Let me now if I have to do any changes.

libraries/Table.php
```diff
@@ -160,6 +160,30 @@ public function getFullName($backquoted = false)
...
+     * array of engine strings or a single string
+     *
+     * @return bool True, if $engine matches the storage engine for the table,
+     * False otherwise.
+     */
+    public function isEngine($engine)
+    {
+        if (is_array($engine)){
+            foreach($engine as $e){
+                if($e == $this->isEngine()){
```
nijel: This seems to be wrong, you should probably compare with tbl_storage_engine as on line 183.

libraries/Table.php
```diff
@@ -160,6 +160,30 @@ public function getFullName($backquoted = false)
...
+                    return true;
+                }
+            }
+            return false;
+        }else{
+            return $this->tbl_storage_engine == $engine ? true : false;
```
nijel: The ternary operator is not necessary here.

arimourao: I think the ternary is necessary because I have to return a boolean informing if it is the engine informed or not. Otherwise I would have to place this logic when calling the function ie `if(Table::isEngine() == 'INNODB'){//Do Something}`. If you prefer that way I can change it.

nijel: But the comparison already returns boolean, so it can be just: 

```php
return $this->tbl_storage_engine === $engine;
```

arimourao: I will make the corrections as soon as I get home

arimourao:
Done nijel. Let me know if you need anything else.

tbl_operations.php
```diff
@@ -126,12 +122,8 @@
         && mb_strtoupper($_REQUEST['new_tbl_storage_engine']) !== $tbl_storage_engine
     ) {
         $new_tbl_storage_engine = mb_strtoupper($_REQUEST['new_tbl_storage_engine']);
-        // reset the globals for the new engine
-        list($is_myisam_or_aria, $is_innodb, $is_isam,
-            $is_berkeleydb, $is_aria, $is_pbxt
-        ) = PMA_setGlobalVariablesForEngine($new_tbl_storage_engine);
```
nijel: Now that PMA_setGlobalVariablesForEngine is not used it could be removed.

tbl_operations.php
```diff
@@ -142,12 +134,12 @@
     }
 
     $table_alters = PMA_getTableAltersArray(
-        $is_myisam_or_aria, $is_isam, $create_options['pack_keys'],
+        $pma_table->isEngine('MYISAM', 'ARIA'), $pma_table->isEngine('ISAM'), $create_options['pack_keys'],
```
nijel: IMHO it would be better to pass `$pma_table` as parameter and evaluate the condition when needed later in the code. This applies to all function calls.

nijel (changes requested):
It looks good so far. However I think we could make the code even better :-).

nijel:
This turned out to be less complete than I hoped for, have you even tested the code? It used non existing property of Table class.

I've merged it with following fixes:

* 6c93914 Rewrite PMA_getTableOptionDiv to use Table for engine flags
* 58f6e84 Make PMA_getTableAltersArray accept table instead of flags
* 1a99e00 Document todo
* 2f97016 Simplified parameters for PMA_getHtmlForTableMaintenance
* ffda016 Actually make isEngine work
* 1c337f4 Remove no longer used function

PR #12741, opened 2016-11-26
make PMA_isAllowedDomain case insensitive, fix links on home page

remicollet:
On home page, external link are broken.

Ex: http://localhost/phpMyAdmin/url.php?url=https%3A%2F%2Fwww.phpMyAdmin.net%2F

phpMyAdmin is not in the allowed domain list, only phpmyadmin.net.

Signed-off-by: Remi Collet <fedora@famillecollet.com>

### DIFF
--- a/libraries/core.lib.php
+++ b/libraries/core.lib.php
@@ -765,7 +765,7 @@ function PMA_isAllowedDomain($url)
             return false;
         }
     }
-    $domain = $arr["host"];
+    $domain = strtolower($arr["host"]);
     $domainWhiteList = array(
         /* Include current domain */
         $_SERVER['SERVER_NAME'],


nijel:
There is no such link in QA_4_6, so I really don't see reason to add this.

remicollet:
On the phpMyAdmin home page...

phpMyAdmin
* Version information: 4.6.5.1 (up to date)
* Documentation
* Official Homepage => http://localhost/phpMyAdmin/url.php?url=https%3A%2F%2Fwww.phpMyAdmin.net%2F
*  Contribute => http://localhost/phpMyAdmin/url.php?url=https%3A%2F%2Fwww.phpMyAdmin.net%2Fcontribute%2F
*  Get support => http://localhost/phpMyAdmin/url.php?url=https%3A%2F%2Fwww.phpMyAdmin.net%2Fsupport%2F
* List of changes 
*  License

nijel:
The link was changed to lowercase together with the change to not convert string to lowercase on domain validation, see https://www.phpmyadmin.net/security/PMASA-2016-66/

remicollet:
indeed, I see the PMA_linkURL have been fixed in QA_4_6

BTW just released 4.6.5.1 is broken

nijel:
BTW: The contribute and support link never used camelcase (it was introduced in a6fd4c2e), the homepage was fixed in dac36c3c.

And 4.6.5.1 is not broken, I've just checked the release and it does not contain these links. Not sure where they come from in your case...

nijel:
PS: Here is diff from the release to current QA_4_6: https://github.com/phpmyadmin/phpmyadmin/compare/RELEASE_4_6_5_1...QA_4_6

remicollet:
Indeed oficial archive is ok... (was an old downstream patch... which need to be dropped)

PR #12770, opened 2016-12-01
Fix #12769 : Current user might vary from the actual defined user

### LINKED ISSUE #12769: Edit/Export links are not clickable under Routines tab
### Steps to reproduce
1. Create any routine (Procedure)
2. Click Routines
3. Only Drop link is working. Edit and Export does not work

### Expected behaviour
Edit and Export links should work as User has all privileges to Database.

### Actual behaviour
Edit and Export are not clickable

### Server configuration
**Operating system**:
Centos5

**Web server:**
Apache 2.4

**Database:**
MySQL 5.6

**PHP version:**
7.0

**phpMyAdmin version:**
4.6.5.1 (4.6.4 is OK)

### Client configuration
**Browser:**
Chrome

**Operating system:**
Windows 10

devenbansod:

Fix #12769 

The current user might vary from the actual defined user, so check for the defined user instead, as the `Definer` field stores the defined user too.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -8,6 +8,7 @@ phpMyAdmin - ChangeLog
 - issue #12588 ReCaptcha now works without allow_url_fopen
 - issue #12699 Show no local storage warning only on settings tab
 - issue #12765 Fixed SQL export with newlines
+- issue #12769 Edit/Export links are not clickable under Routines tab
 
 4.6.5.1 (2016-11-25)
 - issue #12735 Incorrect parameters to escapeString in Node.php
--- a/libraries/DatabaseInterface.php
+++ b/libraries/DatabaseInterface.php
@@ -2135,7 +2135,7 @@ class DatabaseInterface
         if (Util::cacheExists('mysql_cur_user')) {
             return Util::cacheGet('mysql_cur_user');
         }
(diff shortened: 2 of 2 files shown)


libraries/rte/rte_list.lib.php
```diff
@@ -206,7 +206,7 @@ function PMA_RTN_getRowForList($routine, $rowclass = '')
     $query = "SELECT `DEFINER` FROM INFORMATION_SCHEMA.ROUTINES WHERE $where;";
     $routine_definer = $GLOBALS['dbi']->fetchValue($query, 0, 0, $GLOBALS['controllink']);
 
-    $curr_user = $GLOBALS['dbi']->getCurrentUser();
+    $curr_user = $GLOBALS['dbi']->fetchValue('SELECT CURRENT_USER()');
```
nijel: I'm not sure if it won't be better to adjust `getCurrentUser` to use `CURRENT_USER()` instead of `USER()`. It's used only in one more place and I think `CURRENT_USER()` is better fit there as well.

devenbansod: True, that would make sense. 
Have made the change just now.

nijel:
Merged

PR #12797, opened 2016-12-09
PMA_getRealSize - Performance and Readability

jrchamp:

- [N/A] Any new functionality is covered by tests

The "upstream" Moodle code was updated in MDL-39524; rather than looping through the options, preg_match allows us to directly select the correct multiplication factor with a single function call. Note that the default behavior (integer coercion) handles both bare integers and numbers of bytes (factor 1).


nijel:

PR #12799, opened 2016-12-10
Added SignonCookieParams parameter

Ashus:
This is needed for PMA signon auth to be able to access correct session in case you have a specific session cookie params set.


nijel (changes requested):
Can you please add documentation for the new option?

ibennetch:
Thanks for providing tests, that's very helpful.

nijel:
Merged into master, thanks for your contribution!

Issue #12801, opened 2016-12-12
Custom header shown twice with cookie login


ibennetch:
When using a `config.header.inc.php` file and auth_type cookie, the header is displayed twice on the login page

Tested with 4.6.5.2 and master.

nijel:
Fixed by https://github.com/phpmyadmin/phpmyadmin/pull/12803

Issue #12802, opened 2016-12-12
Custom footer not shown with auth_type http failed login


ibennetch:
When using a `config.footer.inc.php` file and auth_type http, if a login fails (for me the easiest way to reproduce is to press cancel at the login prompt), the resulting error page does not show the custom footer.

Tested with 4.6.5.2 and master.

nijel:
Fixed by https://github.com/phpmyadmin/phpmyadmin/pull/12803 and  7d36180

ibennetch:
This is a much cleaner solution, thanks!

PR #12803, opened 2016-12-12
Fix display of custom header and footer in certain edge cases.

ibennetch:

Fixes issues #12801 and #12802

### DIFF
--- a/libraries/plugins/auth/AuthenticationHttp.php
+++ b/libraries/plugins/auth/AuthenticationHttp.php
@@ -90,7 +90,7 @@ class AuthenticationHttp extends AuthenticationPlugin
         $response->addHTML('</h3>');
 
         if (@file_exists(CUSTOM_FOOTER_FILE)) {
-            include CUSTOM_FOOTER_FILE;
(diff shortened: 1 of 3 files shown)


libraries/plugins/auth/AuthenticationHttp.php
```diff
@@ -90,7 +90,7 @@ public function authForm()
         $response->addHTML('</h3>');
 
         if (@file_exists(CUSTOM_FOOTER_FILE)) {
-            include CUSTOM_FOOTER_FILE;
+            $response->addHTML(file_get_contents(CUSTOM_FOOTER_FILE));
```
nijel: With this change we stop interpreting the footer as PHP, but include it as is in output....

ibennetch: It seems to work with my for a `config.footer.inc.php` like

`<h2>Test: <?php echo date('h:i:s');?></h2>`

I get the expected interpreted output, not literal PHP code.

ibennetch:
Rebased and force-pushed to fix merge conflicts, I think that conflicts with the merge you pushed earlier so if you pull again, you'll probably have an easier time deleting the local branch and pulling again.

Issue #12434, opened 2016-08-03
Improve suhosin documentation


nijel:
Our documentation could better cover how to configure suhosin to run with phpMyAdmin.

Reported by Emanuel Bronshtein:

1.3.1) PHP Suhosin configuration

in documentation:
https://docs.phpmyadmin.net/en/latest/faq.html#can-i-use-phpmyadmin-on-a-server-on-which-suhosin-is-enabled

I will recommend to note setting configuration directives that increase security:

NOTE: all the below was checked against code in master & phpmyadmin 4.6.3

1.3.1.1) enable include.max_traversal (mitigation against LFI)

suhosin.executor.include.max_traversal
https://suhosin.org/stories/configuration.html#suhosin-executor-include-max-traversal

suggest setting the value to 2.
as ../ is used in ReCaptcha library for require_once call:
https://github.com/google/recaptcha/blob/master/src/autoload.php#L34

1.3.1.2) enable transparent cookie encryption

suhosin.cookie.encrypt
https://suhosin.org/stories/configuration.html#suhosin-cookie-encrypt

transparent cookie encryption is possible in PMA because PHP don't read cookies set by JS & JS don't read cookies set by PHP.

1.3.1.3) enable disable_emodifier

suhosin.executor.disable_emodifier
https://suhosin.org/stories/configuration.html#suhosin-executor-disable-emodifier

PMA doesn't use /e modifier inside preg_replace function.

nijel:
Another suggestions from Emanuel:

2) improve compatible directives in Suhosin FAQ
 
in:
    https://docs.phpmyadmin.net/en/latest/faq.html#can-i-use-phpmyadmin-on-a-server-on-which-suhosin-is-enabled
 
I suggest adding information about additional directives for compatible
 
2.1) suhosin.sql.union needs to be disabled (the default)
 
suhosin.sql.union can block queries that contain UNION
https://suhosin.org/stories/configuration.html#suhosin-sql-union
 
UNION is used in the below files:
 
https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/dbi/DBIDummy.php#L45
https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/DatabaseInterface.php#L2115
https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/normalization.lib.php#L682
https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/server_privileges.lib.php#L2592
    
[...]

PR #12805, opened 2016-12-13
Improve documentation for servers with Suhosin; issue #12434

ibennetch:


ibennetch:
Maybe this should target master instead because a lot of the links to www.hardened-php.net were only fixed in master, not QA_4_6. Or maybe 118682f4 should be backported. I'm not sure what is best here.

nijel:
I think it's better to target this on 4.6, I've just backported the link fixes as c72b1563d1.

PR #12539, opened 2016-09-06
Fix numeric errors in PHP7.1 

devenbansod:
Fix numeric errors in PHP 7.1.

Before submitting pull request, please check that every commit:
- [x] Has proper Signed-Off-By
- [x] Has commit message which describes it
- [x] Is needed on it's own, if you have just minor fixes to previous commits, you can squash them
- [ ] Any new functionality is covered by tests

PR #12806, opened 2016-12-13
PHP 7.1 fixes for QA_4_6

nijel:

I hoped there will be phpMyAdmin 4.7.0 by time PHP 7.1 is out, but
obviously that's not the case, so we should fix 4.6 branch as well.


nijel:
This will pretty much contain backports from #12539.

PR #12815, opened 2016-12-15
Fix #12814 : Allow length parameter to be specified for DATETIME and TIME

### LINKED ISSUE #12814: DateTime won't allow to input length in Routine
### Steps to reproduce
1. Create Routine (or Edit Routine)
2. Set Parameter Type to DateTime

### Expected behaviour
Allow user to specify length of DateTime inputs for assigning milliseconds/microseconds inputs

### Actual behaviour
Hides Length field

### Server configuration
**PHP version:** 
5.6.14 

**phpMyAdmin version:**
4.6.5.2

### Client configuration
**Browser:**
Firefox, Chrome

**Operating system:**
WIndows 8

devenbansod:

MySQL allows precision to be specified for DATETIME, TIME type fields too, we should allow it to be specified.

Ref: http://dev.mysql.com/doc/refman/5.7/en/fractional-seconds.html

Fix #12814

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -20,6 +20,7 @@ phpMyAdmin - ChangeLog
 - issue #12800 Fixed various issues with PHP 7.1
 - issue #11816 Fixed operation with lower_case_table_names=2
 - issue #12813 Fixed stored procedure execution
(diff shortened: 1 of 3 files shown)


nijel:

PR #12823, opened 2016-12-17
Remove inline styles from elements with id='serverstatusquerieschart_…

clancelotti:
…data'(partial fix of issue #12262)

Signed-off-by: Cristiano Lancelotti <lancelot.info@gmail.com>


clancelotti:
Please, helpe me. It's my first Pull Request. I see I made mistakes  using tab for identation instead of spaces although I've read the developer guidelines. What should I do to fix it? Whatever sugestion is welcome...

clancelotti:
I changed the files and push again without tab for identation. I used git amend command and push with --force option. If there is something else that I can do, say me.

clancelotti:
nijel, thank you for tip.  I made the changes, but probably I made some mistakes. One check was not successful.

nijel:
Yes, the test needs to be adjusted to new output...

nijel:
Have you tested that charts still work? It does not work for me (with some JSON parsing errors).

clancelotti:
@nijel, I tested after seeing your message and it does not work for me too. I will try to fix it. If you have some suggestion, please, say me.

nijel:
My original intention was to store the value in the data attribute instead of element content. In this case I think it's some content being added to the element and it tries to parse it instead.

clancelotti:
I fixed the problem and it works for me (commit 0972666...), chart is normally rendered. But there was one conflict after push command, I resolved it and merge on GitHub. Then it generated another commit and the tests occurred on it (commit d3d973f...). I don't know if it is right. Take a look, please.

js/server_status_queries.js
```diff
@@ -16,9 +16,10 @@ AJAX.registerOnload('server_status_queries.js', function () {
     // Build query statistics chart
     var cdata = [];
     try {
-        $.each(JSON.parse($('#serverstatusquerieschart_data').text()), function (key, value) {
+        $.each(jQuery.parseJSON($('#serverstatusquerieschart').text()), function (key, value) {
```
nijel: Is this change really necessary? `jQuery.parseJSON` is deprecated....

clancelotti: I tested and it is not necessary. I'm going to fix it now.

clancelotti:
@nijel, when  I have tested the exhibition of chart I saw that its legend is not properly located with 2 colunms. The same problem occurs on demo server. Do you think it's a good idea to open a PR to fix it?

nijel:
Yes, feel free to go ahead.

nijel:
Sorry, but your code still did not work for me. I've changed it to use data attribute in 9503e0f6aaec8729869b347af1b2cff41e53f1ad and now it works fine.

clancelotti:
Anyway, tank you very much for deploying the correct solution for this problem instead of just closing the PR. I'm learning with this example. I go ahead and try again.

PR #12824, opened 2016-12-17
Fixes #12079

### LINKED ISSUE #12079: Consistent header manipulation and testing
Currently we need runkit to do testing of some `header()` calls and intentionally skip some of them when running in testsuite.

These should be replaced by `Response::header()` and tests should be adjusted to mock this method.

For example this has been already done for cookie auth in 2148e7cad879a9edbc6a82ab03115cd0957f9bee

osaid-r:

Added mocking of response object methods to the test

Signed-off-by: Osaid  osaid.nasir@gmail.com

### DIFF
--- a/setup/lib/form_processing.lib.php
+++ b/setup/lib/form_processing.lib.php
@@ -7,6 +7,7 @@
  */
 use PMA\libraries\config\FormDisplay;
 use PMA\libraries\URL;
+use PMA\libraries\Response;
 
 /**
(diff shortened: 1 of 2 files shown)


ragnerok:
@nijel do I need to install the response mock twice to remove this error ?

nijel:
You need to rewrite `setup/lib/form_processing.lib.php:76` to use Response::header instead of the PHP function.

nijel:

PR #12825, opened 2016-12-18
Fixes #12802 custom fo…

### LINKED ISSUE #12802: Custom footer not shown with auth_type http failed login

ghost:
…oter now shown with auth_type http failed login

Signed-off-by: Gregory Buhler <Gregory@GregBWebDev.Tech>

- [N/A] Any new functionality is covered by tests

### DIFF
--- a/libraries/Footer.php
+++ b/libraries/Footer.php
@@ -304,6 +304,14 @@ class Footer
         if ($this->_isEnabled) {
             if (! $this->_isAjax) {
                 $retval .= "</div>";
+                if (file_exists(CUSTOM_FOOTER_FILE)) {
+                    $retval .= '<div id="pma_footer">';
+                    ob_start();
+                    include CUSTOM_FOOTER_FILE;
+                    $retval .= ob_get_contents();
+                    ob_end_clean();
+                    $retval .= '</div>';
+                }
             }
             if (! $this->_isAjax && ! $this->_isMinimal) {
                 if (PMA_getenv('SCRIPT_NAME')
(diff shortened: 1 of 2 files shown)


ibennetch:
I can't fully test this at the moment due to #12828.

At the moment, it appears that with auth_type cookie, the footer is displayed twice with your patch.

Gregory-Buhler:
@ibennetch I realized that there were two different calls for the file, this works on my end, let me know if you're still ending up with the same issue on yours.

Cheers!

libraries/Footer.php
```diff
@@ -304,6 +304,14 @@ public function getDisplay()
...
             if (! $this->_isAjax) {
                 $retval .= "</div>";
+                if (file_exists(CUSTOM_FOOTER_FILE)) {
+                    $retval .= '<div id="pma_footer">';
+                    ob_start();
+                    include CUSTOM_FOOTER_FILE;
+                    $retval .= ob_get_contents();
+                    ob_end_clean();
+                    $retval .= '</div>';
+                }
```
nijel: This will lead to footer being render twice on normal pages, it is rendered for second time here: https://github.com/phpmyadmin/phpmyadmin/pull/12825/files#diff-80f5c35c397813b410d4ab117d9b0ac2L352

nijel (commented):
I've rather rewritten the logic to be same in all places in 7d361806c353911b2828053dc2e92fa270852838, this looks like cleaner approach to me.

Issue #12348, opened 2016-06-29
Missing web.config in test folder


ibennetch:
1) missing web.config file

there is usage of `.htaccess` in `test` directory in order to disallow access to the directory.
on IIS web server there is no .htaccess mechanism but there is a web.config mechanism that can be used instead.

fix:
    add web.config file where .htaccess file is used, more information:
        https://www.iis.net/learn/application-frameworks/install-and-configure-php-applications-on-iis/translate-htaccess-content-to-iis-webconfig

ibennetch:
(Initially reported by Emanuel Bronshtein)

ibennetch:
Some time ago, we removed `.htaccess` files from the rest of phpMyAdmin. I suspect that we simply neglected to remove the one from `./test` and suggest that we do so now.

ibennetch:
See also f7065ee and df17ff7448f1fbba85803556292830f2f04d6bd2

ibennetch:
We probably should remove the test directory entirely in `create-release.sh`, shouldn't we?

nijel:
I think it's good to be able to execute testsuite on downloaded code. Most likely some distributions are using that.

On the other side, there are clearly many of users who are not interested in many files with ship:

- source code for javascript libraries
- .po files
- tests
- documentation source

Pretty much this list should match what is being removed in the Docker container (minus setup):

https://github.com/phpmyadmin/docker/blob/master/Dockerfile#L39

Not sure if it won't be better to provide additional download to cover this and offer is as default (as I think this is what really majority of users want - download phpMyAdmin and run it).

In the end we could have something like:

- english - stripped down version without .mo files
- all languages - stripped down version with all .mo files
- source - what we currently ship as all languages version

ibennetch:
I'm not sure that the benefits of smaller download size would be worth the added confusion for new users trying to decide which package they want. With five different compression types, a user on the download page already has a lot of choices available.

ibennetch:
So for now, to close out this ticket, I think the solution is to leave `./test` and simply remove `.htaccess` from the test directory, to match the removal of `.htaccess` in the rest of the repository some time ago.

nijel:
I think  new users just hit the green download button, whatever we provide there :-). (If I look at download stats for 4.6.5.2, 85% downloads the zip we promote this way).

As for compression types, we might want to reduce them, what IMHO makes sense is:

* .tar.gz - standard for any unix environment
* .zip - standard for Windows
[...]

PR #12836, opened 2016-12-19
Add source release, strip current ones

nijel:

See https://github.com/phpmyadmin/phpmyadmin/issues/12348 for discussion which lead to writing this code.

This add source release and removes many files from the standard releases, the outcome is:

* smaller default download size (around 2 MB saved on every single file)
* less chance for exploits for normal users as they don't get testsuite (they don't need)
* there is one more download kit including sources, it is quite big as it includes po files we did not ship before at all (strictly speaking this violates GPL)
* to reduce number of download options, I've disabled 7z and tar.bz2 compressions

### DIFF
--- a/scripts/create-release.sh
+++ b/scripts/create-release.sh
@@ -17,7 +17,7 @@ set -u
 # Fail on failure
 set -e
 
-KITS="all-languages english"
+KITS="all-languages english source"
 COMPRESSIONS="zip-7z txz tgz"
 
 # Process parameters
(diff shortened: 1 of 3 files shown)


nijel:
I think this is ready for review, comments welcome from all @phpmyadmin/developers :-)

ibennetch:
Looks good to me.

nijel:
Probably this should be merged after 4.6.6 is out to make the change live for 4.7.0....

nijel:
@ibennetch  I think it would be optimal to merge this after 4.6.6 (so that it's release is unaffected), but prior to 4.7.0-beta1 (so that it gets the changes and follows new structure).

ibennetch:
Thanks for stating that, I was also considering that such an idea was probably the best way to handle it.

ibennetch:
Thanks!

PR #12852, opened 2016-12-24
Removed PMA_TEST_HEADERS checking as they have been defined

osaid-r:

Signed-off-by : Osaid <osaid.nasir@gmail.com>

### DIFF
--- a/test/PMATestCase.php
+++ b/test/PMATestCase.php
@@ -28,7 +28,7 @@ class PMATestCase extends PHPUnit_Framework_TestCase
      *
      * @return void
      */
-    public static function mockResponse($param)
+    public function mockResponse($param)
     {
         $restoreInstance = PMA\libraries\Response::getInstance();
 
--- a/test/PMATestCase.php
+++ b/test/PMATestCase.php
@@ -49,5 +49,6 @@ class PMATestCase extends PHPUnit_Framework_TestCase
         $attrInstance = new ReflectionProperty('PMA\libraries\Response', '_instance');
         $attrInstance->setAccessible(true);
         $attrInstance->setValue($mockResponse);
+        $attrInstance->setValue($restoreInstance);
     }
 }
--- a/test/PMATestCase.php
+++ b/test/PMATestCase.php
@@ -8,8 +8,8 @@
 
 class PMATestCase extends PHPUnit_Framework_TestCase
 {
-	public $restoreInstance;
-	public $attrInstance;
+    public $restoreInstance;
+    public $attrInstance;
     /**
      * This method is called before the first test of this test class is run.
      */
--- a/test/PMATestCase.php
+++ b/test/PMATestCase.php
@@ -21,4 +21,33 @@ class PMATestCase extends PHPUnit_Framework_TestCase
         require 'libraries/config.default.php';
         $GLOBALS['cfg'] = $cfg;
     }
+    /**
+     * Creates mock of Response object
+     *
+     * @param string $param parameter for header method
+     *
+     * @return void
+     */
+    public static function mockResponse($param)
+    {
+        $restoreInstance = PMA\libraries\Response::getInstance();
+
+        $mockResponse = $this->getMockBuilder('PMA\libraries\Response')
+            ->disableOriginalConstructor()
+            ->setMethods(array('header', 'headersSent'))
+            ->getMock();
+
+        $mockResponse->expects($this->once())
+            ->method('header')
+            ->with($param);
+
+        $mockResponse->expects($this->any())
+            ->method('headersSent')
+            ->with()
+            ->will($this->returnValue(false));
+
+        $attrInstance = new ReflectionProperty('PMA\libraries\Response', '_instance');
+        $attrInstance->setAccessible(true);
+        $attrInstance->setValue($mockResponse);
+    }
 }
--- a/test/PMATestCase.php
+++ b/test/PMATestCase.php
@@ -47,4 +47,18 @@ class PMATestCase extends PHPUnit_Framework_TestCase
         $this->attrInstance->setAccessible(true);
         $this->attrInstance->setValue($mockResponse);
     }
+    /**
+     *Tear down function for mockResponse method
+     *
+     *@return void
+     */
+    protected function tearDown()
+    {
+        if(isset($this->attrInstance, $this->restoreInstance))
+        {
+            $this->attrInstance->setValue($this->restoreInstance);
+            unset($this->restoreInstance);
+            unset($this->attrInstance);
+        }
+    }
 }
--- a/test/PMATestCase.php
+++ b/test/PMATestCase.php
@@ -6,13 +6,10 @@
  * @package PhpMyAdmin-test
  */
 
-/**
- * Base class for phpMyAdmin tests.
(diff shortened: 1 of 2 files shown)


ragnerok:
@nijel  why is this causing an error ? All I've done is remove the if conditions which were useless as the header tests were defined and hence not needed

nijel:
You did not mock the Response::getInstance, so the function is working on real Response object instead of the mocked one...

ragnerok:
So basically I have to mock response object in both functions for this to work right like done [here](https://github.com/phpmyadmin/phpmyadmin/commit/2148e7cad879a9edbc6a82ab03115cd0957f9bee#diff-7cd1872ccc0c19cfaf814d634989604aR370)

nijel:
Yes

ragnerok:
I still do not get why it wasn't showing an error for this earlier when the if conditions were still there but anyways thanks

nijel:
Prior to the change the PHP's `header()` function has been mocked by runkit (or this was not tested on PHP 7 where runkit is not available), so it didn't actually sent headers...

ragnerok:
Also I think I have written same response mock code in 2-3 files and it would have been better if I had made some kind of function for it

ragnerok:
ok then I'll add a function responseMock with 2 arguments 1 is an array of strings for arguments for header() and $n which is the number of calls of header function to ConfigFile.php

nijel:
I think it would better fit to test/PMATestCase.php...

ragnerok:
Yeah makes sense
Thanks

test/libraries/core/PMA_headerLocation_test.php
```diff
@@ -65,16 +66,14 @@ public function testSendHeaderLocationWithoutSidWithIis()
 
         $testUri = 'https://example.com/test.php';
 
-        $header = array('Location: ' . $testUri);
         PMA_sendHeaderLocation($testUri); // sets $GLOBALS['header']
-        $this->assertEquals($header, $GLOBALS['header']);
+        PMATestCase::mockResponse('Location: ' . $testUri);
```
nijel: I think it's better to base the test class on PMATestCase and use it as $this->mockResponse....

ragnerok: you mean PMA_HeaderLocation_Test should extend PMATestCase ?
but it already inherits from phpunit and multiple inheritance isn't allowed in php I think

nijel: But PMATestCase does extend PHPUnit_Framework_TestCase, so there is no problem with that :-)

ragnerok: oh I forgot about that nice !

test/PMATestCase.php
```diff
@@ -21,4 +21,33 @@ public static function setUpBeforeClass()
...
+    /**
+     * Creates mock of Response object
+     *
+     * @param string $param parameter for header method
+     *
+     * @return void
+     */
+    public static function mockResponse($param)
+    {
+        $restoreInstance = PMA\libraries\Response::getInstance();
```
nijel: You do not restore the instance anywhere...

test/libraries/core/PMA_headerLocation_test.php
```diff
@@ -61,29 +61,18 @@ public function setUp()
...
 
-            //reset $GLOBALS['header'] for the next assertion
-            unset($GLOBALS['header']);
+        $testUri = 'https://example.com/test.php';
 
-            $header = array('Refresh: 0; ' . $testUri);
-            PMA_sendHeaderLocation($testUri, true); // sets $GLOBALS['header']
-            $this->assertEquals($header, $GLOBALS['header']);
+        PMA_sendHeaderLocation($testUri); // sets $GLOBALS['header']
+        $this->mockResponse('Location: ' . $testUri);
```
nijel: The mock should be before function call....

test/libraries/core/PMA_headerLocation_test.php
```diff
@@ -171,4 +150,4 @@ public function testSendHeaderLocationIisLongUri()
 
         $attrInstance->setValue($restoreInstance);
```
nijel: The value is not set here.

test/libraries/core/PMA_headerLocation_test.php
```diff
@@ -61,29 +61,18 @@ public function setUp()
...
+        PMA_sendHeaderLocation($testUri); // sets $GLOBALS['header']
+        $this->mockResponse('Location: ' . $testUri);
 
-        } else {
-            $this->markTestSkipped(
-                'Cannot redefine constant/function - missing runkit extension'
-            );
-        }
+        //reset $GLOBALS['header'] for the next assertion
+        unset($GLOBALS['header']);
```
nijel: No longer needed...

nijel (changes requested):

test/libraries/core/PMA_headerLocation_test.php
ragnerok: sorry I just realised my mistake

test/libraries/core/PMA_headerLocation_test.php
ragnerok: this wasn't done by me but I'll correct it nonetheless

nijel: Ah, sorry, it works here because you're not using mockResponse in this method...

ragnerok: ok, but I was wondering is there a way to make use of mockResponse here as there is only the disable method that needs to be added

nijel:
Now the problem is that you're not restoring the original Response object after tests. Probably the mockResponse should return it and there should be method to restore it at the end of test...

ragnerok:
So the function should return $mockRespnse and then attrInstance 's value should be set to $mockResponse

nijel:
You need to call `$attrInstance->setValue($restoreInstance);` in the end. Probably it should be again wrapped by some method. Maybe the mockResponse should store both parameters as attributes. Or even better, if these attributes are set, the mock response would be automatically restored in tearDown method.

ragnerok:
why not make `$attrInstance` and `$mockResponse` as properties of PMATstCase then both are accessible using `$this`

nijel:
Yes, that is what I was thinking about:

* add these properties to the class
* set them in mockResponse
* add `tearDown` method (which is executed after each test) which will restore original instance and clear the attributes if they were set

This way it will work properly even in case of test failure.

test/libraries/core/PMA_headerLocation_test.php
```diff
@@ -171,4 +149,18 @@ public function testSendHeaderLocationIisLongUri()
 
         $attrInstance->setValue($restoreInstance);
     }
-}
+    /**
+     *Tear down function for mockResponse method
+     *
+     *@return void
+     */
+    public function tearDown()
```
nijel: Shouldn't the method rather go to the PMATestCase class?

ragnerok: Then the method will have to be protected but yeah that will be better

nijel:

PR #12857, opened 2016-12-28
added required checking if constant PHPMYADMIN is defined

pjona:
https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/plugins/export/README


nijel:
I'm afraid the README is just wrong here - these files define only class and thus execute no code when loaded directly...

pjona:
Yes, I think you have right. I will be prepare PR with changes only in README file.

Issue #12859, opened 2016-12-29
Feature Request: Add global configuration to remove DELETE button from SQL tab


seattledeveloper:
I would appreciate a feature that allows to control some User Interface functionality from phpMyAdmin.  In particular, I would like to be able to globally remove the DELETE button from the SQL tab when exploring a table.  It seems too easy to accidentally click that button and press Go before realizing what you're doing.  I don't wish to have "DELETE FROM `table` WHERE 1" ever displayed on our production systems.

![phpmyadmin](https://cloud.githubusercontent.com/assets/9749337/21554423/40aa3132-cdc4-11e6-8d3d-06be5fd9dc87.png)

Maybe something like $cfg['ShowSqlDeleteButton']?

Thanks in advance for your consideration.

**PHP version:**
7.0.14

**phpMyAdmin version:**
4.6.5.2

### Client configuration
**Browser:**
Firefox

**Operating system:**
Windows 7

bobhairgrove:
The question of whether the current user can delete something or not should be handled at the user privilege level, not in the GUI (IMHO). For users with DELETE privileges, there is always a confirmation dialog, so I don't see the need for this.

ervin210:
ok  where i need to  add? 
On Dec 29, 2016, at 09:42 AM, seattledeveloper <notifications@github.com> wrote:

I would appreciate a feature that allows to control some User Interface functionality from phpMyAdmin. In particular, I would like to be able to globally remove the DELETE button from the SQL tab when exploring a table. It seems too easy to accidentally click that button and press Go before realizing what you're doing. I don't wish to have "DELETE FROM table WHERE 1" ever displayed on our production systems.

Maybe something like $cfg['ShowSqlDeleteButton']?
Thanks in advance for your consideration.
PHP version:
7.0.14
phpMyAdmin version:
4.6.5.2
Client configuration
Browser:
Firefox
Operating system:
Windows 7

seattledeveloper:
Unfortunately, @ervin210 I do not know where to add.  But to answer @bobhairgrove , one other option may be to have the default query not be valid, similar to the UPDATE query.  An example might be: "DELETE FROM table WHERE `field-1`=[value-1]".  This might actually be good for all the default queries -- as anything that says WHERE 1 is really dangerous.

ibennetch:
Indeed, removing the privilege outright from those users you don't trust is probably the best solution.

[...]

PR #12867, opened 2017-01-02
Changed the SQL window suggestion tool to have DELETE...WHERE 0 inste…

ibennetch:
…ad of 1.

This should help avoid accidental data loss.

Issue #12859


ibennetch:
Even though this is a behavior change, I think it's important enough to push to QA_4_6 instead of how we would typically keep such changes in master. Other opinions are welcome.

PR #12882, opened 2017-01-09
$param was not being used in mockResponse()

CViniciusSDias:
There was a parameter ($param) not being used in the mockResponse() method.

https://scrutinizer-ci.com/g/phpmyadmin/phpmyadmin/inspections/173bc2d8-f521-4adf-ad46-0c5f65514a89/issues/files/test/PMATestCase.php?status=new&orderField=path&order=asc&honorSelectedPaths=0


nijel:
Thanks, indeed it's not needed since we use `func_get_args()`.

PR #12889, opened 2017-01-12
Row coloring issue

ShreyasSinha:

screen shot 2017-01-11 at 11 21 48 pm
I have tried to fix the issue by changing the rowspan to 1 and by keeping all the attributes (type,function,null,value) of a column in a single row and not two different row as was done in case of longtext.

### DIFF
--- a/libraries/insert_edit.lib.php
+++ b/libraries/insert_edit.lib.php
@@ -585,9 +585,10 @@ function PMA_getValueColumn($column, $backup_field, $column_name_appendix,
     } elseif ($GLOBALS['cfg']['LongtextDoubleTextarea']
         && mb_strstr($column['pma_type'], 'longtext')
     ) {
-        $html_output = '&nbsp;</td>';
-        $html_output .= '</tr>';
-        $html_output .= '<tr>' . '<td colspan="5" class="right">';
+    //    $html_output = '&nbsp;</td>';
+    //    $html_output .= '</tr>';
+    //    $html_output .= '<tr>' .
+    //        $html_output = '<td colspan="5" class="right">';
         $html_output .= PMA_getTextarea(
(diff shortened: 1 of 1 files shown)


ShreyasSinha:
I think there is some problem signing off. I will close and submit another pull request.

libraries/insert_edit.lib.php
```diff
@@ -2663,7 +2664,7 @@ function PMA_getHtmlForFunctionOption($column, $column_name_appendix)
         . '<td '
         . ($longDoubleTextArea
             && mb_strstr($column['True_Type'], 'longtext')
-            ? 'rowspan="2"'
+            ? 'rowspan="1"'
```
OlafvdSpek: Why not drop the rowspan attribute?

libraries/insert_edit.lib.php
```diff
@@ -585,9 +585,10 @@ function PMA_getValueColumn($column, $backup_field, $column_name_appendix,
     } elseif ($GLOBALS['cfg']['LongtextDoubleTextarea']
         && mb_strstr($column['pma_type'], 'longtext')
     ) {
-        $html_output = '&nbsp;</td>';
-        $html_output .= '</tr>';
-        $html_output .= '<tr>' . '<td colspan="5" class="right">';
+    //    $html_output = '&nbsp;</td>';
+    //    $html_output .= '</tr>';
+    //    $html_output .= '<tr>' .
+    //        $html_output = '<td colspan="5" class="right">';
```
OlafvdSpek: Should be deleted instead of commented I think.

ShreyasSinha: hey @OlafvdSpek, This is a closed pull request.
Look at #12890 instead.
I closed this one as I had not properly signed off the commit.

Issue #12831, opened 2016-12-19
Row coloring issue


OlafvdSpek:
Shouldn't the geolocation field have a dark background too? This looks weird.

![pma](https://cloud.githubusercontent.com/assets/2218936/21306027/359ec8f4-c5cf-11e6-95d4-7d18046d0618.PNG)

nijel:
Is this master or 4.6 release?

OlafvdSpek:
It's master

nijel:
Probably outcome of doing rows coloring purely in CSS, something we should fix before 4.7.0.

OlafvdSpek:
The geolocations bit has rowspan=2 I think, not sure why.

ibennetch:
It appears that the same root cause also makes the display strange on QA_4_6, however the shading is correct. It looks like something isn't right with the \<tr\> and \<td\> output with `longtext`.

ShreyasSinha:
I have tried to fix the issue #12889 @nijel @OlafvdSpek @ibennetch  please have a look.

ibennetch:
Closed thanks @ShreyasSinha  #12890 and 4a2695c

ibennetch:
I backported the fix to 4.6 as well because the underlying issue exists in that code as well. It was less visible due to how we handled the highlighting, but the table display wasn't optimal there.

PR #12890, opened 2017-01-12
Row coloring issue

ShreyasSinha:
Issue: #12831
Signed-off-by: Shreyas Sinha <shreyas.sinha14@gmail.com>

screen shot 2017-01-11 at 11 21 48 pm

I have tried to fix the issue by changing the rowspan to 1 and by keeping all the attributes (type,function,null,value) of a column in a single row and not two different row as was done in case of longtext.

### DIFF
--- a/libraries/insert_edit.lib.php
+++ b/libraries/insert_edit.lib.php
@@ -2658,11 +2658,6 @@ function PMA_getHtmlForFunctionOption($column, $column_name_appendix)
     $longDoubleTextArea = $GLOBALS['cfg']['LongtextDoubleTextarea'];
     return '<tr class="noclick">'
         . '<td '
-        . ($longDoubleTextArea
-            && mb_strstr($column['True_Type'], 'longtext')
-            ? 'rowspan="1"'
-            : ''
-        )
         . 'class="center">'
         . $column['Field_title']
         . '<input type="hidden" name="fields_name' . $column_name_appendix
--- a/libraries/insert_edit.lib.php
+++ b/libraries/insert_edit.lib.php
@@ -585,10 +585,6 @@ function PMA_getValueColumn($column, $backup_field, $column_name_appendix,
     } elseif ($GLOBALS['cfg']['LongtextDoubleTextarea']
         && mb_strstr($column['pma_type'], 'longtext')
     ) {
-    //    $html_output = '&nbsp;</td>';
-    //    $html_output .= '</tr>';
-    //    $html_output .= '<tr>' .
-    //        $html_output = '<td colspan="5" class="right">';
         $html_output .= PMA_getTextarea(
             $column, $backup_field, $column_name_appendix, $onChangeClause,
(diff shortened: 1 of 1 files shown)


libraries/insert_edit.lib.php
```diff
@@ -585,9 +585,10 @@ function PMA_getValueColumn($column, $backup_field, $column_name_appendix,
     } elseif ($GLOBALS['cfg']['LongtextDoubleTextarea']
         && mb_strstr($column['pma_type'], 'longtext')
     ) {
-        $html_output = '&nbsp;</td>';
-        $html_output .= '</tr>';
-        $html_output .= '<tr>' . '<td colspan="5" class="right">';
+    //    $html_output = '&nbsp;</td>';
+    //    $html_output .= '</tr>';
+    //    $html_output .= '<tr>' .
+    //        $html_output = '<td colspan="5" class="right">';
```
ibennetch: If this is dead code, it should be removed rather than just commented out.

libraries/insert_edit.lib.php
```diff
@@ -2939,10 +2940,8 @@ function PMA_getHtmlForInsertEditFormColumn($table_columns, $column_number,
             $no_support_types, $gis_data_types, $extracted_columnspec, $readOnly
         );
     }
-    $html_output .= '</td>'
-        . '</tr>';
-
-    return $html_output;
+//   $html_output .= '</td>'.'</tr>';
```
ibennetch: Same thing here, if this is dead code it should be removed outright.

libraries/insert_edit.lib.php
```diff
@@ -2939,10 +2940,8 @@ function PMA_getHtmlForInsertEditFormColumn($table_columns, $column_number,
...
+  return $html_output;
```
ibennetch: I think this accidentally got moved over, it looks like there should be four spaces to match the indentation in the rest of the function.

ibennetch (changes requested):
Your fix looks good and seems proper, however the dead code should be removed instead of simply commented.

I think this can safely be merged in to QA_4_6, and with the removal of lines 588-591 it should be a simple cherry-pick for us from your branch in to QA_4_6. In the future, though, it's better to target the proper branch when creating the pull request (at least, I don't think there's a way to change it from GitHub once the pull request is opened).

Can you remove the meaningless lines? Thanks.

ShreyasSinha:
Thank you @ibennetch for reviewing my pull request.
I have made the changes as requested.
Next time I will be careful which branch to target while creating pull request.

ibennetch (approved):

nijel (approved):
Looks good.

ibennetch:
It was tough to tell which branch you should have targeted, since the bug report lists 4.7 you technically did the right thing. However, the underlying problem with displaying the table cell was broken in 4.6, so I merged it there as well. When merging, I also squashed your commit in to one as 4a2695cf5b209d18374ed6c339cefd33abfea15b

Nice work, thanks.

Issue #12861, opened 2016-12-30
Renaming tables does not respect `lower_case_table_names` setting


MPLew-is:
This issue is a sort of follow-up to #11816.

### Steps to reproduce ###
1. On a case-insensitive file system (macOS or Windows when using the OS's default settings) and a MySQL installation using a `lower_case_table_names` setting of `2`, create or import a table that has at least one upper-case letter (the example used in this issue is `Testing`).
2. Through `phpmyadmin`'s "Operations" tab (either the "Move table" form or "Rename table to" input under "Table options"), rename the table to any other name also containing at least one upper-case letter (the example used in this issue is `Testing 2`).

### Expected behaviour ###
The table should be renamed to `Testing 2`, respecting the input name, since lower-case table names are not being forced by the server installation.

### Actual behaviour ###
The table is renamed to the lower-case version of the input table name, `testing 2`.

When renaming the table using SQL commands (both from the command line and in the "SQL" tab), the table is renamed as expected. This issue is only present when renaming through the "Operations" tab's "Move table" or "Rename table to" forms. This issue is also **not** present when copying the table through the "Operations" tab's "Copy table" form.

### Server configuration ###
**Operating system**: macOS 10.12.2

**Web server:** Apache 2.4.23

**Database:** MySQL 5.6.34

**PHP version:** 5.6.29

**phpMyAdmin version:** 4.6.5.2

### Client configuration ###
**Browser:** Safari 10.0.2

**Operating system:** macOS 10.12.2

nijel:
Does it happen with patch from #11816 applied?

MPLew-is:
@nijel yes, this occurs both with and without the patch applied.

nijel:
Fixed by 3bf78af, cherrypicked from #12891.

PR #12891, opened 2017-01-13
Only convert table name to lower-case when installation forces it

MPLew-is:
Issue #12861

A `lower_case_table_names` setting of `2` does not mean that the table needs to be forced to lower-case.
This preserves case unless the `lower_case_table_names` setting is `1`, implying that lower-case should be forced.

I left the last checkbox above blank because I have no idea if you all want tests for table renaming under different `lower_case_table_names` settings (currently, from [your table testing file](https://github.com/phpmyadmin/phpmyadmin/blob/b40e9c13f285be626de972e6414416a4f047f757/test/classes/TableTest.php#L745) you only run your tests with `lower_case_table_names=0`). I can definitely add some tests to this PR if you want or need them.

### DIFF
--- a/libraries/Table.php
+++ b/libraries/Table.php
@@ -1316,7 +1316,7 @@ class Table
                 );
             }
         );
-        if ($lowerCaseTableNames) {
+        if ($lowerCaseTableNames === '1') {
             $new_name = strtolower($new_name);
         }
 


MPLew-is:
And just to confirm, I've tested this code on installations with `lower_case_table_names` of `1` and `2`, and confirmed that this change works on both. I can't speak for installations with a value of `0`, but since the change is exceedingly minor, it should not affect the behavior with regards to those installations at all.

nijel:
Thanks! I've rather merged your fix to QA_4_6 for inclusion in upcoming 4.6.6 release, it has landed as 3bf78afeca.

PR #12898, opened 2017-01-15
Update DBIMysqli.php

aliber4079:
I think u forgot to put $client_flags


nijel:
Thanks for pointing out! However when looking at the code, it's simply unnecessary complex, I've done bigger cleanup in eafd5a5d93.

Issue #12764, opened 2016-11-30
Consider Refactor raw HTML adding to DOM to use createElement instead


emanuelb:
The createElement API has 2 advantages:
1. main advantage:
Using createElement instead of appending HTML that contain input appended to it, such as:
https://github.com/phpmyadmin/phpmyadmin/blob/master/js/functions.js#L526
https://github.com/phpmyadmin/phpmyadmin/blob/master/js/functions.js#L540
will be safe from attribute injection without using escapeHTML, such as:
http://stackoverflow.com/questions/2202269/best-way-to-add-dom-elements-with-jquery/2203063#2203063
and from XSS if current method is used (in above link text instead of html) , this code pattern will decrease the chance of introducing XSS vulnerabilities.

2. it's might be faster (probably, at least many comments on SO suggest so, but other says otherwise)

ShreyasSinha:
I have made the changes in the 2 above mentioned locations #12907.
@emanuelb If there are any other location where this change is suitable please post there link here.

nijel:
Fxied by https://github.com/phpmyadmin/phpmyadmin/pull/12907.

PR #12907, opened 2017-01-20
Using CreateElement for adding HTML

ShreyasSinha:
#12764 

### DIFF
--- a/js/functions.js
+++ b/js/functions.js
@@ -522,11 +522,12 @@ function PMA_current_version(data)
         var latest = parseVersionString(data.version);
         var url = 'https://web.phpmyadmin.net/files/' + escapeHtml(encodeURIComponent(data.version)) + '/';
         var version_information_message = document.createElement('span');
-        version_information_message.class = 'latest';
+        version_information_message.className = 'latest';
         var version_information_message_link = document.createElement('a');
         version_information_message_link.href = url;
-        version_information_message_link.class = 'disableAjax';
-        version_information_message_link.innerHTML = escapeHtml(data.version);
+        version_information_message_link.className = 'disableAjax';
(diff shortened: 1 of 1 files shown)


js/functions.js
```diff
@@ -521,10 +521,15 @@ function PMA_current_version(data)
...
-        var version_information_message = '<span class="latest">' +
-            PMA_messages.strLatestAvailable +
-            ' <a href="' + url + '" class="disableAjax">' + escapeHtml(data.version) + '</a>' +
-            '</span>';
+        var version_information_message = document.createElement('span');
+        version_information_message.class = 'latest';
+        var version_information_message_link = document.createElement('a');
+        version_information_message_link.href = url;
+        version_information_message_link.class = 'disableAjax';
+        version_information_message_link.innerHTML = escapeHtml(data.version);
```
nijel: I think it would be better to use text here to avoid need for escapeHtml....

js/functions.js
```diff
@@ -537,14 +542,22 @@ function PMA_current_version(data)
...
+            document.getElementById("maincontainer").appendChild(maincontainer_div);
         }
         if (latest === current) {
-            version_information_message = ' (' + PMA_messages.strUpToDate + ')';
+            version_information_message = document.createTextNode(' (' + PMA_messages.strUpToDate + ')');
         }
         var $liPmaVersion = $('#li_pma_version');
         $liPmaVersion.find('span.latest').remove();
-        $liPmaVersion.append(version_information_message);
+        document.getElementById("li_pma_version").appendChild(version_information_message);
```
nijel: Why getting element again, while we already have it as `$liPmaVersion`?

nijel (changes requested):

nijel (approved):

nijel:

PR #12912, opened 2017-01-21
Fix failing tests

ibennetch:

### DIFF
--- a/test/libraries/common/PMA_checkbox_test.php
+++ b/test/libraries/common/PMA_checkbox_test.php
@@ -44,8 +44,8 @@ class PMA_GetCheckboxTest extends PHPUnit_Framework_TestCase
                 )
             ) . ' </root>',
             '<root> <input type="checkbox" name="' . $name . '" id="' . $name
-            . '" /><label for="' . $name . '">' . $label
-            . '</label> </root>'
+            . '" />' . "\n" . '<label for="' . $name . '">' . $label
+            . '</label>' . "\n" . ' </root>'
         );
     }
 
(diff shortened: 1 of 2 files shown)


nijel:
Whitespace should not be significant in XML strings, so adding whitespace to assertXmlStringEqualsXmlString should not be needed....

ibennetch:
That's what I thought as well, but this fixes failing tests for me. I wonder if this is again a problem with my old version of PHPUnit.

ibennetch:
This isn't making much sense to me.

`which phpunit`: `/usr/bin/phpunit`
`phpunit --version` is 4.2.6
`./vendor/phpunit/phpunit/phpunit --version` is 4.8.31

So I have two PHPUnits, one is the package manager version 4.2.6, the other is Composer-installed 4.8.31. Makes sense. But when I run tests from within my phpMyAdmin folder, it calls the Composer version (apparently in spite of my system path):

```
phpunit --exclude-group selenium -c phpunit.xml.nocoverage
PHPUnit 4.8.31 by Sebastian Bergmann and contributors.
```

Which is strange, but even stranger, if I run `phpunit --exclude-group selenium -c phpunit.xml.nocoverage`, the tests in this pull request fail; if I run `./vendor/phpunit/phpunit/phpunit --exclude-group selenium -c phpunit.xml.nocoverage` they pass! Both of those commands show running with the same version, 4.8.31. 

Also, when running with the vendor path, I get two additional notices before the tests run:
```
Warning:        Deprecated configuration setting "strict" used
Warning:        Deprecated configuration setting "selenium" used
```

Since this seems to have *something* to do with my environment, so we can close this pull request, but I'm baffled by this.

Thanks.

nijel:
Maybe it's using phpunit binary from path, but libraries from composer? That could explain some of the confusion...

ibennetch:
Good suggestion, I suppose that makes the most sense. Anyway, I've entirely removed the Debian version and I'll just use Composer to maintain it where needed, since that's the simplest solution.

Thanks for the effort you've put in to these issues.

nijel:
BTW: Something similar has just happened on Travis CI as well: We were using system installed phpunit which picked phpunit libraries from vendor leading to failure:

https://travis-ci.org/phpmyadmin/motranslator/jobs/194429435

Issue #12901, opened 2017-01-17
Moving table reflects input case, not installation case setting


MPLew-is:
### Steps to reproduce
1. On a MySQL installation with a `lower_case_table_names` setting of `1`, create or import a table (the example used in this issue is `testing`).
2. Through the "Operations" tab, move the table (using the "Move table" form) to a table name with at least one upper-cease letter (the example used in this issue is `Testing`).

### Expected behaviour

The resulting page, including both the response message and the refreshed "Operations" tab, should reflect the actual case of the table. PHPMyAdmin should also prevent the table from being moved to the same name, as it does when you directly input the same name in the same case.

### Actual behaviour

The response message and the refreshed page both include the input-cased version of the table name, which could lead to some confusion. PHPMyAdmin also does not prevent you from moving the table (which is actually `testing`, but is shown as `Testing` in the interface) to `testing`, which results in a move to the same name. That doesn't itself pose a problem, but it's just generally weird behavior since a table move to itself is prevented already.

![screen shot 2017-01-17 at 14 02 21 pm](https://cloud.githubusercontent.com/assets/11891085/22035580/7422d43e-dcbe-11e6-8e93-fa92eba8d810.png)

### Server configuration
**Operating system**: macOS 10.12.2

**Web server:** Apache 2.4.25

**Database:** MySQL 5.6.34

**PHP version:** 5.6.29

**phpMyAdmin version:** 4.6.5.2

### Client configuration
**Browser:** Safari 10.0.2

**Operating system:** macOS 10.12.2

I realize this is a minor cosmetic issue, but I figured I'd report it anyways. I also was going to make a pull request for part of this (specifically, the response message that is retrieved via AJAX upon successful moving), but I have absolutely no clue how you want to approach changing the JS side of this. The two methods I could think of were to either have the AJAX calls somehow return the real table name (rather than just use the table name input by the user, as [it currently does](https://github.com/phpmyadmin/phpmyadmin/blob/179a86b2d38daab725445c648f5c180063165d9b/js/tbl_operations.js#L58-L64)), or have the JS code be aware of the server's `lower_case_table_names` setting, which it currently is not.

nijel:
Actually the new table and database name should be already part of the AJAX response in _params array, so that's probably easiest approach to deal with this.

MPLew-is:
Well, in that case I'll definitely look into a bit more, and try to have a PR fixing it later today. Thanks!

MPLew-is:
[...]

PR #12914, opened 2017-01-21
Use correct-case table name in message for successful rename

MPLew-is:
Previously, the name returned was the name used in the request, rather than the name after the correcting for the server’s `lower_case_table_names` setting.

Following up on issue #12901 and @b2e2e77.


nijel:

PR #12917, opened 2017-01-22
Create .user.ini

ghost:
Signed-off-by: Edward Dekker <github@edwarddekker.nl>

### DIFF
--- a/.user.ini
+++ b/.user.ini
@@ -1,0 +1,2 @@
+post_max_size=250MB;
+upload_max_filesize=256MB;


edwarddekker:
Signed-off-by: Edward Dekker <github@edwarddekker.nl>

ibennetch:
I think we should leave this up to the user to manage as we can't guess reasonable defaults not make configuration files that apply to all server configurations.

nijel:
Some time ago, we're removed `.htaccess` files as this approach applies only to one of webservers and it's not really possible for us to maintain configurations for all webservers people use.

Also imposing arbitrary limits here might be confusing for users having this properly configured at system level.

Overall I don't think we should include configuration for any webserver, this is really something person installing phpMyAdmin should take care of (depending on expected usage of phpMyAdmin and web server resources).

PS: Technically speaking, setting lower limit on post_max_size than upload_max_filesize is wrong, it should be other way, see http://php.net/manual/en/ini.core.php#ini.post-max-size.

PR #12919, opened 2017-01-23
Replace Util::whichCrlf() with PHP_EOL

nijel:
It's available since PHP 5.0.2 so it's really time to use it.


ibennetch (approved):
I don't suppose it makes much sense to continue testing here, since this is now a PHP constant any test failures would mean a bug with PHP itself.

ibennetch:
Looks like a good plan to me.

Issue #12707, opened 2016-11-16
Incorrect protocol number at 401/303 headers


emanuelb:
Issue similar to #12646 "Incorrect protocol number at 500 header"
in: (401)
https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/plugins/auth/AuthenticationHttp.php#L68
```
        $response->header('HTTP/1.0 401 Unauthorized');
```

in: (303)
https://github.com/phpmyadmin/phpmyadmin/blob/master/setup/config.php#L52-L53
https://github.com/phpmyadmin/phpmyadmin/blob/master/setup/config.php#L75-L76
https://github.com/phpmyadmin/phpmyadmin/blob/master/setup/config.php#L85-L86
https://github.com/phpmyadmin/phpmyadmin/blob/master/setup/config.php#L93-L94
https://github.com/phpmyadmin/phpmyadmin/blob/master/setup/config.php#L100-L101
```
    header('HTTP/1.1 303 See Other');
```

in:
https://github.com/phpmyadmin/phpmyadmin/blob/master/setup/lib/form_processing.lib.php#L76
```
    header('HTTP/1.1 303 See Other');
```

fix:
1. use http_response_code in AuthenticationHttp.php
2. use PMA_generateHeader303 function in setup/config.php instead of calling header function (add parameter to add string to Location header)
3. use http_response_code in PMA_generateHeader303 function

nijel:
Related to #12079

ShreyasSinha:
I have tried to resolve the issue in #12920.

flash1452:
@nijel Is there something pending with this issue?

nijel:
Probably nothing, just forgot to close this.

PR #12920, opened 2017-01-23
Incorrect protocol number at 401/303 headers

ShreyasSinha:
#12707 
Signed-off-by: Shreyas Sinha <shreyas.sinha14@gmail.com>

### DIFF
--- a/libraries/Response.php
+++ b/libraries/Response.php
@@ -466,51 +466,49 @@ class Response
      */
     public function http_response_code($response_code)
     {
+        http_response_code($response_code);
         switch ($response_code) {
-            case 100: $httpStatusMsg = 'Continue'; break;
-            case 101: $httpStatusMsg = 'Switching Protocols'; break;
-            case 200: $httpStatusMsg = 'OK'; break;
-            case 201: $httpStatusMsg = 'Created'; break;
-            case 202: $httpStatusMsg = 'Accepted'; break;
-            case 203: $httpStatusMsg = 'Non-Authoritative Information'; break;
-            case 204: $httpStatusMsg = 'No Content'; break;
-            case 205: $httpStatusMsg = 'Reset Content'; break;
-            case 206: $httpStatusMsg = 'Partial Content'; break;
-            case 300: $httpStatusMsg = 'Multiple Choices'; break;
-            case 301: $httpStatusMsg = 'Moved Permanently'; break;
-            case 302: $httpStatusMsg = 'Moved Temporarily'; break;
-            case 303: $httpStatusMsg = 'See Other'; break;
-            case 304: $httpStatusMsg = 'Not Modified'; break;
-            case 305: $httpStatusMsg = 'Use Proxy'; break;
-            case 400: $httpStatusMsg = 'Bad Request'; break;
-            case 401: $httpStatusMsg = 'Unauthorized'; break;
-            case 402: $httpStatusMsg = 'Payment Required'; break;
-            case 403: $httpStatusMsg = 'Forbidden'; break;
-            case 404: $httpStatusMsg = 'Not Found'; break;
-            case 405: $httpStatusMsg = 'Method Not Allowed'; break;
-            case 406: $httpStatusMsg = 'Not Acceptable'; break;
-            case 407: $httpStatusMsg = 'Proxy Authentication Required'; break;
-            case 408: $httpStatusMsg = 'Request Time-out'; break;
-            case 409: $httpStatusMsg = 'Conflict'; break;
-            case 410: $httpStatusMsg = 'Gone'; break;
-            case 411: $httpStatusMsg = 'Length Required'; break;
-            case 412: $httpStatusMsg = 'Precondition Failed'; break;
-            case 413: $httpStatusMsg = 'Request Entity Too Large'; break;
-            case 414: $httpStatusMsg = 'Request-URI Too Large'; break;
-            case 415: $httpStatusMsg = 'Unsupported Media Type'; break;
-            case 500: $httpStatusMsg = 'Internal Server Error'; break;
-            case 501: $httpStatusMsg = 'Not Implemented'; break;
-            case 502: $httpStatusMsg = 'Bad Gateway'; break;
-            case 503: $httpStatusMsg = 'Service Unavailable'; break;
-            case 504: $httpStatusMsg = 'Gateway Time-out'; break;
-            case 505: $httpStatusMsg = 'HTTP Version not supported'; break;
+            case 100: $httpStatusMsg = ' Continue'; break;
+            case 101: $httpStatusMsg = ' Switching Protocols'; break;
+            case 200: $httpStatusMsg = ' OK'; break;
+            case 201: $httpStatusMsg = ' Created'; break;
+            case 202: $httpStatusMsg = ' Accepted'; break;
+            case 203: $httpStatusMsg = ' Non-Authoritative Information'; break;
+            case 204: $httpStatusMsg = ' No Content'; break;
+            case 205: $httpStatusMsg = ' Reset Content'; break;
+            case 206: $httpStatusMsg = ' Partial Content'; break;
+            case 300: $httpStatusMsg = ' Multiple Choices'; break;
+            case 301: $httpStatusMsg = ' Moved Permanently'; break;
+            case 302: $httpStatusMsg = ' Moved Temporarily'; break;
+            case 303: $httpStatusMsg = ' See Other'; break;
+            case 304: $httpStatusMsg = ' Not Modified'; break;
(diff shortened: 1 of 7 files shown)


libraries/plugins/auth/AuthenticationHttp.php
```diff
@@ -66,10 +66,12 @@ public function authForm()
         // remove non US-ASCII to respect RFC2616
         $realm_message = preg_replace('/[^\x20-\x7e]/i', '', $realm_message);
         $response->header('WWW-Authenticate: Basic realm="' . $realm_message . '"');
-        $response->header('HTTP/1.0 401 Unauthorized');
         if (php_sapi_name() !== 'cgi-fcgi') {
             $response->header('status: 401 Unauthorized');
         }
```
nijel: Shouldn't this logic go to the the http_response_code method?

setup/lib/form_processing.lib.php
```diff
@@ -69,16 +69,17 @@ function PMA_Process_formset(FormDisplay $form_display)
 /**
  * Generate header for 303
  *
+ * @param string $location will set location to redirect.
+ *
  * @return void
  */
-function PMA_generateHeader303()
+function PMA_generateHeader303($location)
```
nijel: I'd probably move this to Response classs as well, it's not really related to the form processing where it's placed...

nijel (changes requested):
Thanks, there are some problems I've commented inline, additionally some tests now fail and the code needs to be rebased on top of current code (we've merged fixes from security branch).

ShreyasSinha:
After merging fixes from security branch my phpmyadmin is not working.
It is giving a HTTP ERROR 500. How can I fix that?

nijel:
Have you looked at PHP/webserver server logs? You will find the error message there.

libraries/Response.php
```diff
@@ -456,4 +456,38 @@ public function headersSent()
...
+     *
+     * @param int $response_code will set the response code.
+     *
+     * @return void
+     */
+    public function http_response_code($response_code)
+    {
+        $httpStatusMsg  = 'Web server is down';
+        if (php_sapi_name() !== 'cgi-fcgi') {
+            header('status: ' . $response_code . $httpStatusMsg);
```
ShreyasSinha: Displaying a general 'Web server is down' for all response code in case of CGI and Fast CGI.

ShreyasSinha:
@nijel The test is failing because output is already started at  /phpmyadmin/vendor/phpunit/phpunit/src/Util/Printer.php:134 before I set the http_response_code. How can I fix this?

libraries/Response.php
nijel: You should use `$this->header` here.

nijel (commented):
You probably have to mock the header method in Response class for that test.

libraries/Response.php
```diff
@@ -456,4 +456,38 @@ public function headersSent()
...
+    /**
+     * Wrapper around PHP's http_response_code() function.
+     *
+     * @param int $response_code will set the response code.
+     *
+     * @return void
+     */
+    public function http_response_code($response_code)
+    {
+        $httpStatusMsg  = ' Web server is down';
```
nijel: The message should be sent based on the code.

test/classes/plugin/auth/AuthenticationHttpTest.php
```diff
@@ -148,9 +146,7 @@ public function testAuthHost()
 
         $this->doMockResponse(
             1, 1, 1,
-            array('WWW-Authenticate: Basic realm="phpMyAdmin hst"'),
-            array('HTTP/1.0 401 Unauthorized'),
-            array('status: 401 Unauthorized')
+            array('WWW-Authenticate: Basic realm="phpMyAdmin hst"')
```
nijel: This looks strange, shouldn't the header be sent even in tests?

libraries/Response.php
```diff
@@ -456,4 +456,38 @@ public function headersSent()
...
+    }
+
+   /**
+     * Generate header for 303
+     *
+     * @param string $location will set location to redirect.
+     *
+     * @return void
+     */
+    public function PMA_generateHeader303($location)
```
nijel: Can you please remove the `PMA_` prefix here?

test/classes/plugin/auth/AuthenticationHttpTest.php
ShreyasSinha: The size of the array which stores these messages ($param) is tested and should be equal to the number of times header() function is called.
` $header_method = $mockResponse->expects($this->exactly(count($param)))
                    ->method('header');
` PMATestCase.php line:55

Earlier header was called three time for these 3 messages but now it is being called just once. Rest of the time http_response_code() is called. I have not yet written test cases for http_response_code(). I am working on it.

libraries/Response.php
ShreyasSinha: Ok.

libraries/Response.php
ShreyasSinha: Ok I have added a unique message based on the code for every `$response_code`.

test/classes/plugin/auth/AuthenticationHttpTest.php
nijel: But the `php_sapi_name()` should be `cli` on tests and the new code should still use the header path, so the header should be IMHO issued.

libraries/Response.php
```diff
@@ -456,4 +456,77 @@ public function headersSent()
...
+            case 504: $httpStatusMsg = 'Gateway Time-out'; break;
+            case 505: $httpStatusMsg = 'HTTP Version not supported'; break;
+            default: $httpStatusMsg  = ' Web server is down'; break;
+        }
+        if (php_sapi_name() !== 'cgi-fcgi') {
+            $this->header('status: ' . $response_code . $httpStatusMsg);
+        }
+        else{
+            http_response_code($response_code);
+        }
```
nijel: Looking at the original code it did always use http_response_code and *additionally* call header, you have changed it to use one or other, was that intentional?

ShreyasSinha: Thanks for pointing it out. It was a mistake, I have corrected it.

test/classes/plugin/auth/AuthenticationHttpTest.php
ShreyasSinha: Yes, you are right `php_sapi_name()` is `cli` on tests and the new code uses the header path. Thus I have added back `array('status: 401 Unauthorized')` in the new commit.

ShreyasSinha:
The code coverage tests are failing even after I have added Test Cases for `http_response_code()`. How can I fix this?

test/PMATestCase.php
```diff
@@ -51,11 +51,22 @@ public function mockResponse()
             if (is_array($param[0])) {
                 if (is_array($param[0][0]) && count($param) == 1) {
                     $param = $param[0];
+                    if(count($param)>=2){
+                        $http_response_code_param = array(end($param));
+                        $param = array_slice($param, 0, -1);
+
+                        $header_method = $mockResponse->expects($this->exactly(count($http_response_code_param)))
```
nijel: This will be always one, right? So it's probably better to use `once` instead of `exactly`....

test/PMATestCase.php
```diff
@@ -51,11 +51,22 @@ public function mockResponse()
...
+                        ->method('http_response_code');
+
+                        call_user_func_array(array($header_method, 'withConsecutive'), $http_response_code_param);
```
nijel: ...and here the array will have just one member as well, so using `with` directly instead of call_user_func_array machinery sounds more reasonable.

test/PMATestCase.php
```diff
@@ -51,11 +51,22 @@ public function mockResponse()
             if (is_array($param[0])) {
                 if (is_array($param[0][0]) && count($param) == 1) {
                     $param = $param[0];
+                    if(count($param)>=2){
```
nijel: I'm not really sure this is good idea. How about checking whether last member is integer instead? That way it's still possible to use the method with only headers without status code.

```php
if (is_int(end($param))
```

nijel (commented):
Thanks for fixes. The coverage is fine, it's more like an indicator for us rather than strict requirement. I think the code is now good to merge, I've just few minor comments to tests.

test/PMATestCase.php
ShreyasSinha: Ok. That will be better.

test/PMATestCase.php
ShreyasSinha: Yes, that will be better.

test/PMATestCase.php
ShreyasSinha: Ok. Made the changes.

nijel:
Thanks! I've squashed your commits and merged.

PR #12926, opened 2017-01-25
FIX: TH were used instead of TD in table fields

MrElectroNick:
This break custom styles for TH as header cell and TD as data cell
And finally it's just inconsistent


nijel:
What is wrong on using `<th>` for field names? IMHO the CSS should rather look for `<thead>`...

MrElectroNick:
All is good with TH but it should be inside <thead> and not <tbody>

As you can see all columns are TD and only field name is TH which is not correct

nijel:
I don't see anything like that in the [HTML specification](https://www.w3.org/TR/html5/tabular-data.html#the-th-element). Actually their example uses `<th>` in same way as we do - as heading for the row.

MrElectroNick:
Ok

Issue #12567, opened 2016-09-19
Remove ./libraries/tbl_info.inc.php by moving it's code to Table class


nijel:
There is `./libraries/tbl_info.inc.php` which does some basic preprocessing of table information and stores it into global variables. 

This code should be integrated into `Table` class and all callers adjusted to use `Table` class directly instead of relying on the global variables `./libraries/tbl_info.inc.php` currently sets.

Lakshya1605:
Hello! I am new to open source and I would like to work on this issue and start contributing.Please can someone help me get started

nijel:
Look at the [mentioned file](https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/tbl_info.inc.php) and all places where it's used. You can see it currently sets dozen of global variables, where most of them are not really needed in all cases. 

The work can be done in few steps:

1. Move any functionality from this file to Table class.
2. Migrate all uses of the data to use existing or new methods on Table class.
3. Remove no longer needed file.

flash1452:
@nijel  If no one is working on this issue I would like to take up this task.

nijel:
Probably nobody is working on this, so feel free to look at it.

flash1452:
So what I am assuming is that we need to shift the code from tbl_info.inc.php and develop new methods to initialize the global variables like - $showtable, $tbl_is_view, $tbl_storage_engine, $show_comment, $tbl_collation,$table_info_num_rows, $auto_increment.
I need to know if I am getting this right.

nijel:
I'm not sure if global variables are really needed, you should look at their usage and probably move (some of) the method calls there.

flash1452:
@nijel I have found two files that are including tbl_info.inc.php. First one is in $tbl_operations.php and second one is in TableChartController. TableChartController doesn't seem to use any of the variables that are set in tbl_info.inc.php. On the other hand tbl_operations.php uses $tbl_is_view, $auto_increment, $show_comment etc.
So why was it included in TableChartController in the first place??? Or am I missing somethings.

nijel:
Maybe some of it's methods access it as globals? Or it can be also used in the templates, so rather look for the variables themselves.

PR #12931, opened 2017-01-27
Tabl info.inc.php moved to table class #12567

flash1452:

Signed-off-by: flash1452 vishpandey2014@gmail.com


flash1452:
@nijel kindly review my PR and suggest changes.

flash1452:
@nijel Also if you can help me with this code coverage problem.

libraries/Response.php
```diff
@@ -1,5 +1,5 @@
 <?php
-/* vim: set expandtab sw=4 ts=4 sts=4: */
+/* vim: set expandtab sw=4 ts=4 sts=4:  */
```
nijel: I guess this change is not intentional.

libraries/controllers/table/TableChartController.php
```diff
@@ -113,7 +114,60 @@ public function indexAction()
             );
             $url_params['back'] = 'tbl_sql.php';
             include 'libraries/tbl_common.inc.php';
-            include 'libraries/tbl_info.inc.php';
+            // include 'libraries/tbl_info.inc.php';
+            $GLOBALS['dbi']->selectDb($GLOBALS['db']);
+            //$tbl_object = new Table($GLOBALS['db'], $GLOBALS['table']);
+            $showtable = $GLOBALS['dbi']->getTable(
+                                        $GLOBALS['db'],
+                                        $GLOBALS['table']
```
nijel: You should really do the getTable call just once and then use the resulting class in subsequent calls.

nijel (changes requested):
I assume phpunit.phar file should not be there, please remove it.

Also the repeated getTable calls should be removed - it should be called just once and then use resulting Table object.

libraries/controllers/table/TableChartController.php
flash1452: @nijel Yeah, you are damn right there is no point in calling getTable function so many times.

flash1452:
@nijel I have made some of the changes that you suggested but code coverage still remains an issue.

flash1452:
@nijel Should I consider a new PR?? Because Code coverage is not passing.

libraries/Table.php
```diff
@@ -321,6 +322,104 @@ public function getStatusInfo(
     }
 
     /**
+    * Returns the Table storage Engine for current table.
+    * @param  array $showtable       Current table properties.
+    *
+    * @return   string               Return storage engine info if it is set for *                                the selected table else return blank.
+    */
+    public function getTableStorageEngine($showtable){
```
nijel: It really should use `$this->getStatusInfo('ENGINE', false, true)` instead of getting the data as parameter.

libraries/Table.php
```diff
@@ -321,6 +322,104 @@ public function getStatusInfo(
...
+    * Returns the comments for current table.
+    * @param  array $showtable  Current table properties.
+    *
+    * @return string  Return comment info if it is set for the selected table or return blank.
+    */
+    public function getShowComment($showtable){
+        if($showtable){
+
+            return isset($showtable['Comment']) ? $showtable['Comment'] : '';
+        }
```
nijel: Same here, you should rely on class data and not get it from elsewhere as this can lead to errors.

Use `$this->getStatusInfo('COMMENT', false, true)` instead.

libraries/Table.php
```diff
@@ -321,6 +322,104 @@ public function getStatusInfo(
...
+
+    /**
+    * Returns the collation for current table.
+    * @param Current table properties.
+    * @return Return blank if collation is empty else return the collation info from table info.
+    */
+    public function getTableCollation($showtable){
+        if($showtable){
+
+            return empty($showtable['Collation']) ? '' : $showtable['Collation'];
```
nijel: Same here, you should rely on class data and not get it from elsewhere as this can lead to errors.

Use `$this->getStatusInfo('COLLATION', false, true)` instead.

libraries/Table.php
```diff
@@ -321,6 +322,104 @@ public function getStatusInfo(
...
+        }
+    }
+
+    /**
+    * Returns the info about no of rows for current table.
+    * @param Current table properties.
+    * @return Return no of rows info if it is not null for the selected table or return 0.
+    */
+    public function getTableNumRowInfo($showtable){
```
nijel: Again, do not rely on passed parameters.

libraries/controllers/table/TableChartController.php
```diff
@@ -113,7 +114,34 @@ public function indexAction()
...
-            include 'libraries/tbl_info.inc.php';
+            $GLOBALS['dbi']->selectDb($GLOBALS['db']);
+            $table_class_object = $GLOBALS['dbi']->getTable(
+                $GLOBALS['db'],
+                $GLOBALS['table']
+            );
+            $showtable = $table_class_object->getStatusInfo(
+                null,
+                (isset($reread_info) && $reread_info ? true : false)
+            );
```
nijel: You really should not need this, the Table class should handle this internally.

libraries/controllers/table/TableIndexesController.php
```diff
@@ -63,8 +63,35 @@ public function indexAction()
...
+        $reread_info = true;
+        $GLOBALS['dbi']->selectDb($GLOBALS['db']);
+        $table_class_object = $GLOBALS['dbi']->getTable(
+            $GLOBALS['db'],
+            $GLOBALS['table']
+        );
+        $showtable = $table_class_object->getStatusInfo(
+            null,
+            (isset($reread_info) && $reread_info ? true : false)
+        );
```
nijel: ...and same here.

nijel:
Yes most of the code is not covered by tests, but that really does not matter for now as it was previously not covered as well

flash1452:
@nijel Made the changes you suggested.

flash1452:
@nijel Kindly review the updated PR. The code coverage has improved drastically.

nijel:
Thanks! I will get to it next week.

libraries/Menu.php
```diff
@@ -231,8 +231,19 @@ private function _getBreadcrumbs()
             if (strlen($this->_table) > 0
                 && ! (isset($_REQUEST['purge']) && $_REQUEST['purge'] == '1')
             ) {
-                include './libraries/tbl_info.inc.php';
-
+                // include './libraries/tbl_info.inc.php';
+                $reread_info = true;
```
nijel: This does not seem to be used anywhere now. You should instead forcing of reading the information again here, for example using:

```php
$table_class_object->getStatusInfo(null, true);
```

libraries/Table.php
```diff
@@ -321,6 +322,110 @@ public function getStatusInfo(
...
+
+    /**
+    * Returns the comments for current table.
+    * @param  array $showtable  Current table properties.
+    *
+    * @return string  Return comment info if it is set for the selected table or return blank.
+    */
+    public function getShowComment() {
+        $table_comment = $this->getStatusInfo('COMMENT', false, true);
+        if ($table_comment) {
```
nijel: You probably want to check this as `=== false` instead to avoid surprises with PHP automatic typecasting. Without this you would be ignoring comment consisting of `0`, try yourself:

```php

if ('0') {
    echo "HERE\n";
}
```

Same applies to other methods as well.

flash1452:
@nijel I have made some of the suggested changes in this PR and also squashed all the commits.

flash1452:
@nijel I forgot, Congratulations on getting selected for this years GSoC.

libraries/Menu.php
nijel: What about this variable? You set it in several places but it really doesn't seem to be used. In case the rereading is really necessary here, you should actually reread the information as mentioned in above comment.

flash1452:
I have made the changes related to rereading. But the code coverage seems to have taken a hit.

flash1452:
@nijel Please review the PR.

nijel:
I did quick review and this is really far from being complete.

Following files do still use tbl_info.inc.php which you've removed:

* libraries/controllers/table/TableIndexesController.php
* libraries/controllers/table/TableRelationController.php
* libraries/controllers/table/TableSearchController.php
* libraries/controllers/table/TableStructureController.php
* libraries/mult_submits.inc.php
* tbl_addfield.php
* tbl_export.php
* tbl_find_replace.php
* tbl_import.php
* tbl_operations.php
* tbl_select.php
* tbl_sql.php
* tbl_structure.php
* tbl_zoom_select.php

There are some other things to improve in the code which I've fixed (I thought I could merge the code with such changes prior to realizing above):

* ba0f9785c3 Document return types
* 4c4f851d7a Remove documentation for non existing param
* bbef205ca4 Simplify getting table create options
* 5dbd529e69 Simplify getting row format
* f92cc062d4 Simplify getting table rows
* 8f121757c6 Simplify getting table collation
* 2344b79f70 Simplify getting table comment
* ff1427020d Simplify getting storage engine
* 1ddd344fb5 Fix comments indentation
* a90f2f2a08 Remove commented out code
* a7e6c07188 Remove unused variables
* 6899e8f6da Simplify rereading of information

And most important question: Have you actually tested the code you've written? Even table structure does not work with your changes...

flash1452:
@nijel Thanks for you review. I will have to accept I have been careless while testing my code but yeah, I will  soon make the required changes.

flash1452:
@nijel Please correct me if I am wrong but libraries/controllers/table/TableIndexesController.php,
libraries/controllers/table/TableRelationController.php, libraries/controllers/table/TableSearchController.php doesn't seem to be using any of the variables being set in tbl_info.inc.php.

flash1452:
@nijel There are many such files where the lib_info.inc.php is included but the variables being set in lib_info.inc.php is not used. One such example is TableRelationController.php where these variables are passed as parameters to the constructor. What should be the way to handle these situations?

nijel:
Yes, it can be that the variables are not used there, I didn't look at the code, I've just noticed that it's still included. You need to go through these files and check whether they need some of the variables...

flash1452:
@nijel I have updated the code with your mentioned changes. I have tested my code this time but still if you can test it and review it.

flash1452:
@nijel There is a link broken in docs, that's why the TRAVIS build is failing.

 (line   41) broken    https://anonscm.debian.org/cgit/collab-maint/phpmyadmin.git/tree/debian/README.Debian - HTTPSConnectionPool(host='anonscm.debian.org', port=443): Max retries exceeded with url: /cgit/collab-maint/phpmyadmin.git/tree/debian/README.Debian (Caused by NewConnectionError('<requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0x2b233231d8d0>: Failed to establish a new connection: [Errno 113] No route to host',)).

Should we remove this url from the doc or replace it.

nijel:
It was temporary failure, I've just restarted the job.

flash1452:
@nijel Still the code coverage seems to drop. But, I guess this should not make much of difference?

flash1452:
@nijel What can be done to improve the code coverage? Any suggestions.

nijel:
Have you tested the code? I got several errors, fixed them in f88971e34045eff61db397a0f33cfd48f2d69d43, you can merge it at https://github.com/flash1452/phpmyadmin/pull/2

As for coverage, it's probably okay, but you should really test it manually as well as the automated tests still don't cover everything...

flash1452:
@nijel I am getting a Error 500 whenever I am trying to test my phpmyadmin branch. 
The apache error logs show - 
Class 'PhpMyAdmin\\MoTranslator\\Loader' not found in /usr/share/phpmyadmin/libraries/common.inc.php on line 90.

But the weird thing is my old version of the same branch works perfectly and in that version in common.inc.php instead of 'PhpMyAdmin\\MoTranslator\\Loader' only 'MoTranslator\Loader' has been used.

nijel:
You need to run composer, see https://docs.phpmyadmin.net/en/latest/setup.html#installing-from-git

flash1452:
@nijel I have tried and tested the current version  of code with my changes. So, if you can have a look at it and give it a run.

nijel:
Thanks, I've manually merged and rebased your changes.

PR #12935, opened 2017-01-30
Fix Designer JavaScript error relating to token

### LINKED ISSUE #12804: Designer is broken
### Steps to reproduce
1. Open Designer tab

### Expected behaviour
Designer should appear without errors.

### Actual behaviour
Designer opens, but there are errors hidden behind the window.

95064ed9fbf77e4f43ac438591d03998c6bb38c6 seems to be the first bad commit.

ibennetch:
Fixes #12804


ibennetch:
This works for me and looks correct, but I'd like confirmation that this is the correct solution.

nijel:
I've rather migrated the code to use same logic as we use elsewhere in bca2e1c588626298c69a2b795cec84c223fce593.

nijel:
But thanks for pointing out where the problem is!

ibennetch:
That's a much better solution, thanks.

PR #12940, opened 2017-02-01
enhancing server variables table view

### LINKED ISSUE #12899: Variables table unreadable
Matching the values on the left to the values on the right is next to impossible..

I don't think this table should be full-width.

pma

poush:
Fixes: #12899 

Set text-align to center, looks like this:
screen shot 2017-02-01 at 11 53 29 am

Please provide me any suggestions to improve it more.

Signed-off-by: Piyush Agrawal <poush12@gmail.com>

### DIFF
--- a/themes/original/css/common.css.php
+++ b/themes/original/css/common.css.php
@@ -1095,7 +1095,9 @@ div#logTable table {
     text-overflow: ellipsis;
     line-height: 2em;
 }
-
+#serverVariables .var-row > td:last-child {
+    text-align: center;
+}
 #serverVariables .var-header {
     font-weight:        bold;
     color:              <?php echo $GLOBALS['cfg']['ThColor']; ?>;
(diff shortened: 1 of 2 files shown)


OlafvdSpek:
How about left-aligning the values?

poush:
@OlafvdSpek 
screen shot 2017-02-01 at 11 53 29 am

looks good, not sure if whitespace at the right is fine. 
@ibennetch any suggestions?

OlafvdSpek:
Perhaps this table too shouldn't be full-width.

ibennetch:
I also had experimented with changing this, and a few of the lines are so long (at least on my system), that without wrapping or truncating them the table stretches out to nearly full width. My suggestion, which I did not test myself, is to force the table to a certain width which will then cause the longer cells to wrap. We have to be cautious about keeping happy users who have wide displays, but I think that's going to be a better solution than the full-width table we have now.

I also like the left alignment better than centering.

OlafvdSpek:
If the value is left-aligned you only have to limit the width of the key field..

Fenn-CS:
Hello @ibennetch and @OlafvdSpek  what about adding lines to table to ease tracing since reducing the length does not really look good.

nijel:
Thanks for your pull request, however I'm closing it as it was not seen as good solution to the problem.

PR #12949, opened 2017-02-02
token mismatch error after php session timeout

the0ne:
Make sure the security token given by the client is compared only against a valid session token. If the php session no longer exists, use a valid client-provided token.

Many people seem to keep their phpMyAdmin browser windows open, receiving the - really annoying - token error on the first action next day - only because the php session has timed out.
Applying this tiny change transparently creates the new session re-using the old token if the user is (still) correctly signed in.

Signed-off-by: Thomas Horner <github-phpmyadmin@t-horner.com>

### DIFF
--- a/libraries/session.lib.php
+++ b/libraries/session.lib.php
@@ -30,7 +30,9 @@ function PMA_secureSession()
  */
 function PMA_generateToken()
 {
-    if (class_exists('phpseclib\Crypt\Random')) {
+    if (32 == PMA_isValid($_REQUEST['token']) && ctype_xdigit($_REQUEST['token'])) {
+        $_SESSION[' PMA_token '] = $_REQUEST['token'];
+    } else if (class_exists('phpseclib\Crypt\Random')) {
         $_SESSION[' PMA_token '] = bin2hex(phpseclib\Crypt\Random::string(16));
     } else {
         $_SESSION[' PMA_token '] = bin2hex(openssl_random_pseudo_bytes(16));


libraries/session.lib.php
```diff
@@ -30,7 +30,9 @@ function PMA_secureSession()
  */
 function PMA_generateToken()
 {
-    if (class_exists('phpseclib\Crypt\Random')) {
+    if (32 == PMA_isValid($_REQUEST['token']) && ctype_xdigit($_REQUEST['token'])) {
+        $_SESSION[' PMA_token '] = $_REQUEST['token'];
```
nijel: This seems to be completely wrong as you're accepting token from request and storing it into session, what makes it's CSRF protection completely void.

Also ctype is not required by phpMyAdmin.

nijel (changes requested):

libraries/session.lib.php
the0ne: i see. what other solution do you have?
using phpmyadmin regularly the described behaviour is really really annoying.

nijel:
First of all the [$cfg['LoginCookieValidity']](https://docs.phpmyadmin.net/en/latest/config.html#cfg_LoginCookieValidity) setting should match session.gc_maxlifetime so that the login cookie expires at same time as the session. Or you get this on login form?

the0ne:
LoginCookieValidity is set to 28800 in config.inc.php
$cfg['LoginCookieValidity'] = 28800;

the PHP setting session.gc_maxlifetime is only set to 1440.

nijel:
Okay, then just set session.gc_maxlifetime to 28800 and they won't be deleted by PHP. Anyway phpMyAdmin already complais about this on home page:

localhost_localhost_phpmyadmin_4 7 0-beta1_-_2017-02-13_13 24 14

the0ne:
i still believe this will not fix the issue.
by the way, how would my change above be CSRF?
after all, it would only (re-)use the token if there was no valid session (with that token) - right?

the whole issue is that there's a browser window full of links with the old session token.
and the user can not simply continue to use that browser window after having paused a day or two.
and that's what's annoying to users.

nijel:
The problem with your would be that if there was not valid session, attacker would create one with known token.

Anyway the cause here is that your session gets deleted earlier than login cookie, this is exactly what changing session.gc_maxlifetime is going to fix. Why are you restraining this change? phpMyAdmin warns about this and it is [documented](https://docs.phpmyadmin.net/en/latest/config.html#cfg_LoginCookieValidity) for ages...

the0ne:
Hi @nijel, I'm currently testing your proposed fix. If it really solves the issue then I'm very, very happy with it!
In this case we should probably put a hint to this setting on the (many!) online results showing people with the same issue when searching for the "token error" error message.
https://www.google.com/search?hl=en&q='Error%3A+Token+mismatch'

nijel:
I really can not fix people not looking into big red warning they have on phpMyAdmin main page :-). Anyway if that does not help, please open separate issue for this as the approach in this PR is clearly insecure.

nijel:
I've tried to workaround this in dc3ef3ce7f55cd19d01192ccd029230833f28cd1 - phpMyAdmin should keepalive the session for login validity as long as the window is open. Obviously this is not as reliable as adjusting PHP settings, but should help in most cases.

the0ne:
actually the warning message you mention in https://github.com/phpmyadmin/phpmyadmin/pull/12949#issuecomment-279374277
is never shown on the system i work on.

nijel:
And what is your setting for session.gc_maxlifetime?

the0ne:
even adding

>     trigger_error(
>        __('foobar'),
>       E_USER_WARNING
>     );

to the very bottom of index.php does not create a warning message.

the0ne:
i had adjusted session.gc_maxlifetime to 28800 which is the same as the $cfg['LoginCookieValidity'].
but it turns out the error "Fehler: Token stimmen nicht überein" / Error: Token mismatch still occurs despite the values being equal.

until today only using the approach of this pull request (despite as mentioned above being problematic in terms of csrf) turns out to provide a solution that enables to continue work on the next day.

the0ne:
i have patched dc3ef3ce7f55cd19d01192ccd029230833f28cd1 into my 4.6.6 installation.
i'll provide the results here asap.

the0ne:
even on the pc which isn't shut down over night the patch didn't improve the problem.
token error still comes up next day.

nijel:
Okay, let's split this up and solve the issues separately :-).

* Warnings not shown: https://github.com/phpmyadmin/phpmyadmin/issues/13021
* Token error: https://github.com/phpmyadmin/phpmyadmin/issues/13022

PR #12950, opened 2017-02-02
Remove connect_type setting

nijel:
It is really not necessary as MySQL decides connection type rather based
on hostname than on anything else.

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -221,7 +221,7 @@ Server connection settings
         The hostname ``localhost`` is handled specially by MySQL and it uses
         the socket based connection protocol. To use TCP/IP networking, use an
         IP address or hostname such as ``127.0.0.1`` or ``db.example.com``. You
-        can configure socket address by
+        can configure the path to the socket with
         :config:option:`$cfg['Servers'][$i]['socket']`.
 
     .. seealso::
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -219,7 +219,9 @@ Server connection settings
     .. note::
 
         The hostname ``localhost`` is handled specially by MySQL and it
-        ignores :config:option:`$cfg['Servers'][$i]['port']` in this case.
+        ignores :config:option:`$cfg['Servers'][$i]['port']` in this case
+        and uses socket based connection. You can configure
+        socket address by :config:option:`$cfg['Servers'][$i]['socket']`.
 
     .. seealso::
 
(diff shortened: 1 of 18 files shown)


ibennetch:
If, for some reason, we don't merge this, we should address an interesting situation where a user must use connect_type='socket' in order for `$cfg['Servers'][$i]['socket']` to take effect. It makes sense why it was implemented this way, but if this doesn't get merged we can improve that.

ibennetch:
Anyway, I like this and think it's a good improvement. I don't see a downside of merging this and it works for me.

ibennetch (approved):

doc/config.rst
```diff
@@ -219,7 +219,9 @@ Server connection settings
     .. note::
 
         The hostname ``localhost`` is handled specially by MySQL and it
-        ignores :config:option:`$cfg['Servers'][$i]['port']` in this case.
+        ignores :config:option:`$cfg['Servers'][$i]['port']` in this case
```
ibennetch: I'd make this "The hostname ``localhost`` is handled specially by MySQL and it uses the socket based connection protocol. To use TCP/IP networking, use an IP address or hostname such as ``127.0.0.1`` or ``db.example.com``. You can configure socket address by :config:option:`$cfg['Servers'][$i]['socket']`."

doc/config.rst
```diff
@@ -418,6 +425,13 @@ Server connection settings
     :type: string
     :default: ``'tcp'``
 
+    .. deprecated:: 4.7.0
+
+       This settings is no longer used as of 4.7.0, the MySQL anyway decides
```
ibennetch: I suggest "This setting is no longer used as of 4.7.0, since MySQL decides the"

doc/config.rst
```diff
@@ -418,6 +425,13 @@ Server connection settings
...
+       connection type based on host, so it could lead to unexpected results.
+       Please set :config:option:`$cfg['Servers'][$i]['host']` accordingly
```
ibennetch: It's helpful how you direct users to 'host' here instead, good job.

ibennetch (changes requested):
I have a few suggestions for documentation language.

nijel:
Applied your suggestions in 34946a285b.

ibennetch:
Great, I think this looks ready.

Do you really want to apply it to QA_4_7? I don't object because the modifications are relatively minor, but in general I think we should avoid such changes to a QA branch.

nijel:
I'm not sure, but I thought it is still reasonable to do that for 4.7 now.

ibennetch:
Sure, let's go ahead with putting it in 4.7.

nijel:
Okay, merging.

ibennetch:
Great. Did you use `--author` to add my name to this commit or is that part of the new Github review feature? Either way, it's pretty helpful!

nijel:
I've used commandline with ``-author``

Issue #12315, opened 2016-06-10
Add UI to remove certain parts from profiling chart


nijel:
When profiling some more complex query, one item can really hide details of others. There could be checkbox next to each profiling part and unchecking would make it disappear from the chart.

This can be seen for example on `SELECT * FROM `customer_list`` on sakila database, where sending data is long (because there is lot of data), but you're more interested in other parts:

![demo phpmyadmin net_phpmyadmin_demo_-_mysql_sakila_customer_list_phpmyadmin_4 7 0-dev_-_2016-06-10_16 08 00](https://cloud.githubusercontent.com/assets/212189/15966981/ca756046-2f25-11e6-875e-4c16a109427b.png)

nijel:
Implemented by #12952.

PR #12952, opened 2017-02-04
New functionality to show or hide slices of pie chart

clancelotti:
issue #12315 
Signed-off-by: Cristiano Lancelotti <lancelot.info@gmail.com>

Summary of changes:
- Add plugin EnhancedPieLegendRenderer that allow extending the functionalities of the legend: clicking the legend items, you can show or hide the corresponding slices of pie chart.
- Add dependency on files
- Activate the plug-in in options 
- Adjust the position of legend placed on server status query pie chart.

The changes allow the server status query pie chart and the SQL query pie chart show or hide each slice. 
Current behaviour: when clicking on each legend item, the page remains loading and nothing happens.
After changes: when clicking on each legend item, the corresponding chart slice toggles between showing and hiding.

I have tested it on Internet Explorer, Mozilla Firefox and Google Chrome. It works fine here. Take a look, please.

### DIFF
--- a/js/functions.js
+++ b/js/functions.js
@@ -2516,7 +2516,6 @@ function PMA_createProfilingChart(target, data)
             '#729fcf',
             '#ad7fa8',
             '#ef2929',
-            '#eeeeec',
             '#888a85',
             '#c4a000',
             '#ce5c00',
--- a/js/functions.js
+++ b/js/functions.js
@@ -2502,7 +2502,7 @@ function PMA_createProfilingChart(target, data)
         },
         legend: {
             show: true,
-            location: 'e',
+            location: 'se',
             rendererOptions: {
                 numberColumns: 2
             }
--- a/themes/original/css/common.css.php
+++ b/themes/original/css/common.css.php
@@ -977,7 +977,7 @@ div#serverstatusquerieschart {
     float: <?php echo $left; ?>;
     width: 500px;
     height: 350px;
-    margin-right: 50px;
+    margin-<?php echo $right;?>: 50px;
 }
 
 div#serverstatus table#serverstatusqueriesdetails {
--- a/themes/original/css/common.css.php
+++ b/themes/original/css/common.css.php
@@ -974,10 +974,10 @@ caption a.top {
 }
 
 div#serverstatusquerieschart {
-    float:<?php echo $left; ?>;
-    width:500px;
-    height:350px;
-    padding-<?php echo $left; ?>: 30px;
+    float: <?php echo $left; ?>;
+    width: 500px;
+    height: 350px;
+    margin-right: 50px;
 }
 
 div#serverstatus table#serverstatusqueriesdetails {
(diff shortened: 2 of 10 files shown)


clancelotti:
Applying the above changes (commit 5781a00), we have a gray shadow pie chart (defined configuration before these proposed changes). I think it's not cool because it looks like a gray slice of the chart when hiding a slice of another color. 
The next images are with shadow yet and I hid some elements to see the problem:

server_status_pie_chart


sql_pie_chart

Then I set to false the value of shadow and it is not displayed anymore.
Another problem occurs just with status queries pie chart. There are a defined list of colors used to apply on it. When there is a exhibited white slice (without shadow), it's difficult to make difference between the slice color and the background color of the chart. Therefore, I deleted color #eeeeec. Besides that, I added a slice margin, 
See the images with those changes (commit 37d39179):

server_status_pie_chart_change

sql_pie_chart_change

themes/pmahomme/css/common.css.php
```diff
@@ -1296,7 +1296,7 @@
     float: <?php echo $left; ?>;
     width: 500px;
     height: 350px;
-    padding-<?php echo $left; ?>: 30px;
+    margin-right: 50px;
```
nijel: This will break in RTL languages.

themes/original/css/common.css.php
```diff
@@ -974,10 +974,10 @@
...
 
 div#serverstatusquerieschart {
-    float:<?php echo $left; ?>;
-    width:500px;
-    height:350px;
-    padding-<?php echo $left; ?>: 30px;
+    float: <?php echo $left; ?>;
+    width: 500px;
+    height: 350px;
+    margin-right: 50px;
```
nijel: This will break in RTL languages.

js/functions.js
```diff
@@ -2516,7 +2516,6 @@ function PMA_createProfilingChart(target, data)
             '#729fcf',
             '#ad7fa8',
             '#ef2929',
-            '#eeeeec',
```
nijel: Why removing one color here?

clancelotti: Thanks for the feedback. This color is very similar to the background color. See the image (other element). What do you think? 
removed color

themes/pmahomme/css/common.css.php
clancelotti: In this case, a minor change in 1299: padding-<?php echo $left; ?>: 30px; is enough to adjust the legend position. I tested changing 30px to 5 px and it's OK. Any suggestion of other solution that does not break RTL?

nijel: Use `margin-<?php echo $right;?>` instead, so that it's properly oriented in RTL languages.

nijel:
Thanks for fixing those. Still I'm not sure this is something what would help in the original report:

localhost_localhost_phpmyadmin_4 8 0-dev_-_2017-02-14_13 22 35

I'd rather expect the removed parts not showing up at all in the chart - the remaining pieces should IMHO expand to cover full pie instead of just part of it...

clancelotti:
I have thought about it and makes sense what you said. 
I'm working on it, thanks.

clancelotti:
@nijel, I think that it's working like you suggested:
filling all chart

nijel:

Issue #11951, opened 2016-02-12
Warning open_basedir restriction in effect


nijel:
is_writable(): open_basedir restriction in effect. File(/var/www/tmp) is not within the allowed path(s): (/home/:/var/www/html/:/tmp:/var/tmp:/usr/loc

This report is related to user submitted report [#19832](https://reports.phpmyadmin.net/reports/view/19832) on the phpmyadmin error reporting server.

nijel:
Param | Value 
 -----------|--------------------
 Error Type | Warning
 Error Message |is_writable(): open_basedir restriction in effect. File(/phptmp) is not within the allowed path(s): (:/clientdata/php/pear-5-6:/websites/pe/pedalprix.
 Exception Type |php
 Link | [Report#19532](https://reports.phpmyadmin.net/reports/view/19532)

*This comment is posted automatically by phpMyAdmin's [error-reporting-server](http://reports.phpmyadmin.net).*

nijel:
Param | Value 
 -----------|--------------------
 Error Type | Warning
 Error Message |is_writable(): open_basedir restriction in effect. File(C:\Windows\Temp) is not within the allowed path(s): (C:\inetpub\bes\)
 Exception Type |php
 Link | [Report#19205](https://reports.phpmyadmin.net/reports/view/19205)

*This comment is posted automatically by phpMyAdmin's [error-reporting-server](http://reports.phpmyadmin.net).*

nijel:
Param | Value 
 -----------|--------------------
 Error Type | Warning
 Error Message |is_writable(): open_basedir restriction in effect. File(\xampp\tmp) is not within the allowed path(s): (/www/~dev)
 Exception Type |php
 Link | [Report#18801](https://reports.phpmyadmin.net/reports/view/18801)

*This comment is posted automatically by phpMyAdmin's [error-reporting-server](http://reports.phpmyadmin.net).*

nijel:
Param | Value 
 -----------|--------------------
 Error Type | Warning
 Error Message |is_writable(): open_basedir restriction in effect. File(/var/data/temp/vds.wspro.ru) is not within the allowed path(s): (/data/server/vds.wspro.ru/:.)
 Exception Type |php
 Link | [Report#18568](https://reports.phpmyadmin.net/reports/view/18568)

*This comment is posted automatically by phpMyAdmin's [error-reporting-server](http://reports.phpmyadmin.net).*

nijel:
[...]

PR #12974, opened 2017-02-09
Better error message when Composer files are missing.

### LINKED ISSUE #12972: Mention composer in the error message
Perhaps detect the missing file before the require..

```
Fatal error: require_once(): Failed opening required './vendor/autoload.php' (include_path='.:/usr/share/php') in /srv/pma/libraries/common.inc.php on line 78
```

ibennetch:
Fixes #12972

### DIFF
--- a/libraries/common.inc.php
+++ b/libraries/common.inc.php
@@ -75,6 +75,13 @@ require_once './libraries/vendor_config.php';
 /**
  * Activate autoloader
  */
+if (! @is_readable('./vendor/autoload.php')) {
+    die(
+        'File <tt>./vendor/autoload.php</tt> missing or not readable. <br />'
+        . 'Most likely you did not run Composer to '
+        . '<a href="https://docs.phpmyadmin.net/en/latest/setup.html#installing-from-git">install library files</a>.'
+    );
+}
 require_once './vendor/autoload.php';
 
 /**
--- a/ChangeLog
+++ b/ChangeLog
@@ -59,6 +59,7 @@ phpMyAdmin - ChangeLog
 - issue #12954 Fixed export of tracking data
 - issue #12960 Enclose exports in transaction by default
 - issue #12966 After adding a column ADD INDEX option won't be displayed when enabling AI
+- issue #12972 Better error message when Composer has not been run
 
 4.6.6 (2017-01-23)
 - issue #12759 Fix Notice regarding 'Undefined index: old_usergroup'


ibennetch:
I believe we can't translate these strings because motranslator hasn't been loaded yet.

libraries/common.inc.php
```diff
@@ -75,7 +75,16 @@
...
+if (file_exists('./vendor/autoload.php') && is_readable('./vendor/autoload.php')) {
+    require_once './vendor/autoload.php';
+}
+else {
+    die(
+       'File <tt>./vendor/autoload.php</tt> missing or not readable. <br />'
+       . 'Most likely you did not run Composer to '
+       . '<a href="https://docs.phpmyadmin.net/en/latest/setup.html#installing-from-git">install library files</a>.'
+     );
+}
```
nijel: I'd rather do it `if (! file_exist || ! is _readable) { die() };` so that require_once is not behind condition (or it should be turned into include_once, what I think does not make sense here).

OlafvdSpek: file_exists seems redundant if you call is_readable..

nijel: True, also it probably should have @ in front to avoid showing PHP errors.

OlafvdSpek: What errors?
PHP doesn't error / warn for is_readable does it?

poush: Yes, from php.net "Upon failure, an E_WARNING is emitted."

OlafvdSpek: A non-readable file is not a failure for this function.

nijel: Non-readable file is not, but file when `stat()` fails, it emits warning.

OlafvdSpek: `printf("%x\n", error_reporting()); is_readable('a');` outputs 57ff and doesn't emit any warnings.

OlafvdSpek: When would `stat()` fail? And why would you want to surpress the warning?

ibennetch:
I've updated the pull request with all of your suggestions. It now suppresses warnings and only tests with `is_readable()`.

Thanks for the feedback.

nijel:
Thanks, merged.

libraries/common.inc.php
nijel: For example in case of open_basedir restrictions (see https://github.com/phpmyadmin/phpmyadmin/issues/11951). We want to suppress it to avoid full path disclosure vulnerability.

jublo:
I loved this when I got to set up a Git-powered instance just today. It helped me figure the error in an instant.

ibennetch:
Great, I'm glad it helped!

PR #12975, opened 2017-02-09
Remove tcpdf from composer suggest

gulaandrij:
tecnickcom/tcpdf present in require section and dont need to suggest it

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,6 @@
         "phpmyadmin/sql-parser": "^4.1",
         "phpmyadmin/motranslator": "^3.0",
         "phpmyadmin/shapefile": "^2.0",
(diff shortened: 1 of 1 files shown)


ibennetch:
Good catch. I rather think we should remove it from the required section and keep it as a suggestion. phpMyAdmin handles a missing tcpdf rather gracefully.

gulaandrij:
did it

nijel:
Indeed it should be just optional, however we still need at least 6.2, so appropriate conflict with older versions should be added as well.

Also it should be either added to require-dev or explicitly installed in Travis so that we test with it.

nijel:
Fixed in git by a7950fa.

PR #12987, opened 2017-02-12
Typographical quotes

mynetx:
Using typographical, localisable quotes (`“` and `”`) instead of plain ones.

bildschirmfoto 2017-02-11 um 16 58 31


nijel (changes requested):
Can you please look at the tests failure?

nijel:
Thanks for your contribution, I've merged it with minor fixes:

* 6e78d489ee Use unicode for quotes
* dfa7995451 Include quotes directly in the string

PR #12988, opened 2017-02-12
Home page: Hide languages if only 1 is available

mynetx:
This may happen on the development version when the mo files generator hasn't been run.

bildschirmfoto 2017-02-11 um 16 31 15

### DIFF
--- a/index.php
+++ b/index.php
@@ -237,11 +237,14 @@ echo '  <ul>';
 
 // Displays language selection combo
 if (empty($cfg['Lang'])) {
-    echo '<li id="li_select_lang" class="no_bullets">';
+    $selector = LanguageManager::getInstance()->getSelectorDisplay();
 
-    echo PMA\libraries\Util::getImage('s_lang.png') , " "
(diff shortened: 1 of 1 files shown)


ibennetch:
Thanks. 

Can you remind me, did you see an Issue posted about this? I started to make one recently but may have gotten distracted with another matter.

It doesn't really matter aside from wanting to link to it if one exists.

jublo:
@ibennetch No, fixed because I noticed directly in the code.

-----

Can you check your emails? There’s a message for you.

jublo:
The Travis build failed because Weblate is down for alleged planned maintenance – while its completion estimate always shows the next full hour – for 5 hours already @nijel.

nijel:
Yes, the Weblate maintenance took longer than expected, anyway just the docs test has failed due to this.

Anyway I've fixed your issue differently in 924aabf652f35a35ccfe66a6fa706bb434282081 as the very same fix should be applied to login page as well.

ibennetch:
Thanks!

PR #13005, opened 2017-02-15
Fix  #12979: Fixed column alignment on foreign keys table.

### LINKED ISSUE #12979: Foreign keys table layout
The vertical alignment here looks weird..
Especially in row 2 but also in row 1 fields column.

pma

souravbadami:
Signed-off-by: Sourav Badami souravbadami@gmail.com

### DIFF
--- a/themes/pmahomme/css/common.css.php
+++ b/themes/pmahomme/css/common.css.php
@@ -381,7 +381,7 @@ table th,
 table td {
     padding: .1em .3em;
     margin: .1em;
-    vertical-align: middle;
+    vertical-align: top;
     text-shadow: 0 1px 0 #fff;
 }
 


souravbadami:
@ibennetch -- Could you PTAL ?

ibennetch:
This fixes the reported problem for me, however I haven't had a chance to see if there are other unintended side-effects.

nijel:
As this changes quite generic CSS, there are side effects, for example table structure is now also top alligned:

localhost_localhost_bar2_phpmyadmin_4 7 0-beta1_-_2017-02-20_11 40 02

What IMHO looks worse than original middle aligned:

localhost_localhost_bar2_phpmyadmin_4 7 0-beta1_-_2017-02-20_11 40 54

nijel:
You probably need to add the class in the code which adds fields?

souravbadami:
I did. But the problem is -- Initially I set up the alignment and it fixes the issue. But there's another option below to add new rows. When a new row is added, It gets a bit out of sync. Still, I'll try to figure out a solution.

PR #13006, opened 2017-02-15
Align NULL values according to column alignment

### LINKED ISSUE #13000: NULL is right-aligned in a left-aligned column
I think left-aligned makes more sense, in this case.

pma

mynetx:
Fixes #13000.

image


ibennetch (approved):
This looks great to me.

nijel:
Thanks! I've applied this to QA_4_7 branch as 53eff5d47b664558fd02881cde14f8144d997cfb.

PR #13007, opened 2017-02-15
OpenDocument export generates invalid file

ibennetch:

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -77,6 +77,7 @@ phpMyAdmin - ChangeLog
 - issue #12958 Always show error reporting dialog on top
 - issue #12693 Improved support for TokuDB
 - issue #11231 Try harder to honor LoginCookieValidity setting
+- issue #13002 OpenDocument Spreadsheet and Text (ODS/ODT) exports an invalid file
 
 4.6.6 (2017-01-23)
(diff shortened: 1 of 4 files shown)


ibennetch:
Note that the trailing space is significant; I was torn on whether it more appropriately belongs as part of the OpenDocument::NS declaration or before the next line in every place where NS is used. Since it's needed everywhere I though it more appropriately belongs here.

jublo:
Will this fix OpenDocument text and sheets?

ibennetch:
Changed and rebased because I decided to move the trailing space from the constant to the next code in each place where it's used. I felt having the trailing space exposed as part of the constant was too vulnerable to accidental breakage.

ibennetch:
Yes, this fixes both ODT and ODS files.

ibennetch:
I didn't dig too much in to this, but it's just standard xml so it should pretty much not care about whitespace, which is why this bug is a bit strange for me.

nijel:
I've fixed this independently in 59e9856200726359298a68895dc558c03f127254. I just didn't touch the NS definition...

Sorry, next time I will look at pull requests first time before I start to work on the code.

ibennetch:
Thanks @nijel. What's important is that it got fixed.

PR #13009, opened 2017-02-17
Also submit inline SQL form when pressing the command key on macOS

MPLew-is:
Allows for the natural macOS key combination to submit a form (command + enter) to be used in addition to the Windows combination (control + enter).


nijel:
Thanks, I've applied your change as dcaeb652b07fe6704897c6a4d444302b2a24e483 in 4.7 branch.

PR #13019, opened 2017-02-19
Copy database: Fix SQL error for copying PMADB metadata

### LINKED ISSUE #13010: Error while copying database
While copying my one database, it's gives "1136 Column count doesn't match value count at row 1" error about "REPLACE INTO `phpmyadmin`.`pma_column_info` (db_name, table_name, column_name, comment)  VALUES('kitti','ayar','ayar')" sql. 

Sql query is wrong. It has appeared 4.5.4, before that update I could copy this database without issue.

I have latest phpmaydmin version 4.6.6 on Windows 10.

mynetx:
Fixes #13010.


nijel:

PR #13020, opened 2017-02-20
Remove ServerLibraryDifference_DisableWarning setting

nijel:
This is getting more and more problematic these days as most of Linux
distributions ship MariaDB libraries, while many people run MySQL
servers. As MariaDB claims the protocol compatibility should stay, we
should not warn in this case.

Of course there is option to add this exception to the warning (so far
the exception is there only other way - when using MySQL libraries to
connect to MariaDB server), but we would be ignoring most cases already
and I think complete removal is better than adding exceptions.

Overall I don't think this warning is doing any good these days.

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -108,6 +108,10 @@ Basic settings
     :type: boolean
     :default: false
 
+    .. deprecated:: 4.7.0
+
+        This setting was removed as the warning has been removed as well.
+
     A warning is displayed on the main page if there is a difference
     between the MySQL library and server version.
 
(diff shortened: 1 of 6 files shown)


nijel:
See also https://github.com/phpmyadmin/docker/issues/106

ibennetch (approved):
I don't have a strong feeling about this. It seems to me that recent versions haven't had much incompatibility and are pretty unaffected by version mismatches (but I haven't done tests to confirm this opinion).

It definitely seems to be doing more harm than good at this point, especially with many distributions switching to MariaDB.

PR #13024, opened 2017-02-20
Remove no longer needed code in authentication

### LINKED ISSUE #12478: Investigate if code is still needed
In the authentication classes there are two almost identical pieces of code:

- https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/plugins/auth/AuthenticationHttp.php#L198
- https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/plugins/auth/AuthenticationCookie.php#L449

Both these come from 0c8575c9. I don't think this code is ever used, but I'm really not sure, so another pair of eyes is welcome. If it is needed, it should be shared and behave consistently across authentication methods.

### LINKED ISSUE #13003: Unit test failure with AuthenticationCookieTest.php
I get a unit test error with current master:

```
1) AuthenticationCookieTest::testAuthSetUserWithHeaders
Undefined index: host

/var/www/pma-dev/fork/libraries/plugins/auth/AuthenticationCookie.php:453
/var/www/pma-dev/fork/test/classes/plugin/auth/AuthenticationCookieTest.php:718
```

nijel:
The code is incosistent between auth methods and I don't see any purpose
for it.

Fixes #12478
Fixes #13003

### DIFF
--- a/libraries/plugins/auth/AuthenticationCookie.php
+++ b/libraries/plugins/auth/AuthenticationCookie.php
@@ -446,23 +446,6 @@ class AuthenticationCookie extends AuthenticationPlugin
     {
         global $cfg;
 
-        // Ensures valid authentication mode, 'only_db', bookmark database and
-        // table names and relation table name are used
-        if (! hash_equals($cfg['Server']['user'], $GLOBALS['PHP_AUTH_USER'])) {
-            foreach ($cfg['Servers'] as $idx => $current) {
-                if ($current['host'] == $cfg['Server']['host']
-                    && $current['port'] == $cfg['Server']['port']
-                    && $current['socket'] == $cfg['Server']['socket']
-                    && $current['ssl'] == $cfg['Server']['ssl']
-                    && hash_equals($current['user'], $GLOBALS['PHP_AUTH_USER'])
-                ) {
-                    $GLOBALS['server'] = $idx;
-                    $cfg['Server']     = $current;
-                    break;
-                }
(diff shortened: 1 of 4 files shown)


nijel:
What I think this should do is to allow specify per user rules in configuration - if logged in user matches one stored in config, different set of server configuration is used. I still fail to see why the code should be there:

- it's not documented anywhere
- it can have surprising results for users mixing cookie and config auth
- the cookie code is slightly broken as it ignores differences in ssl settings
- the code in http auth is broken completely as it ignores port/socket setting making it possible to match to completely different server

ibennetch:
I won't pretend to understand why this bit of code is here, so since you think it's safe to remove that's fine with me.

Also, I'm not sure I follow what you propose in your comment about allowing per-user rules.

nijel:
I was trying to describe what the removed code does. I have no clue if I understand it correctly as the comment is completely misleading.

ibennetch:
Oh, I understand your comment now. At first I thought you were proposing what we should do in the future, so it makes much more sense now.

Later today I can try to test this further, but if you want to merge to master that's fine as well.

nijel:
I'd really like to hear another opinion here, the code and feature does not make much sense to me and is not documented, but that doesn't mean that it's useless, maybe I'm just missing something...

In any case we should either remove this code (what this PR does) or document it in both code and end user documentation, so that it's clear what it does and what it's purpose.

Issue #12961, opened 2017-02-06
Indexes shown at both tbl_structure.php and tbl_relation.php


OlafvdSpek:
Shouldn't it be on only one page?

jublo:
Thanks for noting this. Which page would you prefer as per logical reasoning?

OlafvdSpek:
Structure

madhuracj:
I believe this was intentional. See https://github.com/phpmyadmin/phpmyadmin/issues/6021

OlafvdSpek:
I'm not sure why automatically creating required indexes requires showing the indexes as well.

madhuracj:
If you read the full thread you'll note that the idea of adding indexes automatically has been abandoned and allowing the user to create indexes on the relations page has been implemented instead.

OlafvdSpek:
I did read the full page but apart from one summary update I didn't see that being mentioned..

madhuracj:
I feel the page will be way too long if we merge the two.

OlafvdSpek:
Would it?
I think the Internal relationships could easily be shown in the main table structure table. An action button to edit the internal relationship for a field would be simple as well..

OlafvdSpek:
> and allowing the user to create indexes on the relations page has been implemented instead.

Foreign keys require indexes on the *target* table, but this functionality applies to the *source* table. So it's useless.

nijel:
Looking at the structure page, it might be better to split it up more in second level tabs:

- structure
- indexes
- partitions
- foreign keys

That would probably be more consistent than current approach. What do you think?

OlafvdSpek:
I think fewer pages and less second-level navigation is better.

I see no reason to split structure and indexes..

nijel:
Clearly current state is inconsistent (indexes are show on both structure and foreign keys, partitions are shown collapsed at structure page), what really doesn't look good. I thought using consistently second level navigation for all these table structure features might be better.

OlafvdSpek:
Dropping indexes from Relation view seems like a trivial fix.

For table structure, IMO partitions is taking up too much space when partitions isn't used at all.
"Table comments: " is showing while there is no comment.
Adding columns should be a single button / link, same for adding indexes.
Print, propose table structure, track table and improve table structure could perhaps be somewhere else too. Perhaps Improve should just be called Normalize.
Like the little key icon, I'd like to see an icon for columns with foreign keys and relations.

ibennetch:
[...]

PR #13025, opened 2017-02-20
Hide comment field when no table comment is set

ibennetch:
On the table structure tab, this hide table comments if there is none. See https://github.com/phpmyadmin/phpmyadmin/issues/12961#issuecomment-281040584

### DIFF
--- a/templates/table/structure/display_table_stats.phtml
+++ b/templates/table/structure/display_table_stats.phtml
@@ -1,9 +1,11 @@
 <div id="tablestatistics">
     <fieldset>
         <legend><?= __('Information'); ?></legend>
-        <p> <strong> <?= __('Table comments:') ?> </strong>
-            <?= isset($showtable['TABLE_COMMENT']) ? htmlspecialchars($showtable['TABLE_COMMENT']) : '';?>
-        </p>
+        <?php if($showtable['TABLE_COMMENT'] !== '') { ?>
+            <p> <strong> <?= __('Table comments:') ?> </strong>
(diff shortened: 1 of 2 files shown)


templates/table/structure/display_table_stats.phtml
```diff
@@ -1,9 +1,11 @@
 <div id="tablestatistics">
     <fieldset>
         <legend><?= __('Information'); ?></legend>
-        <p> <strong> <?= __('Table comments:') ?> </strong>
-            <?= isset($showtable['TABLE_COMMENT']) ? htmlspecialchars($showtable['TABLE_COMMENT']) : '';?>
-        </p>
+        <?php if(! empty($showtable['TABLE_COMMENT'])) { ?>
```
nijel: Using `empty()` is probably not a good idea as `empty('0')` is true, this ti will not display such comment.

PS: Generally I tend to think that `empty()` should not be used at all in most situations.

ibennetch: Ah, yes. I guess I should instead test for isset (which I think is always true here anyway) and if `== ''`

nijel: Either use `=== ''` or `strlen(...) === 0`. I'm not sure if isset is necessary here...

ibennetch:
I've updated this with the requested changes.

I'm starting to wonder if this should target master because we're trying to freeze changes to QA_4_7.

nijel:
IMHO such minor changes are certainly okay (and this is anyway mostly bugfix).

PR #13039, opened 2017-02-28
Fix #13028 Include 100% font size in the selection

### LINKED ISSUE #13028: Include 100% font size in the selection
Currently font size selection is generated by some increments, but there is no way to quickly go to 100%. 

See on [Twitter](https://twitter.com/arongriffis/status/834063750953066496):

page-shot-2017-2-21 aron griffis arongriffis twitter

plartem:
Signed-off-by: Artem Hryn <plartem01@gmail.com>

### DIFF
--- a/libraries/Config.php
+++ b/libraries/Config.php
@@ -1497,6 +1497,9 @@ class Config
             $factors[] = 1;
             $factors[] = 5;
             $factors[] = 10;
+            if($value !== 100) {
+                $factors[] = 100 - $value;
+            }
         } elseif ($unit === 'em') {
             $factors[] = 0.05;
             $factors[] = 0.2;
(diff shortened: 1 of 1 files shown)


libraries/Config.php
```diff
@@ -1497,6 +1497,9 @@ protected static function getFontsizeOptions($current_size = '82%')
             $factors[] = 1;
             $factors[] = 5;
             $factors[] = 10;
+            if($value !== 100) {
+                $factors[] = 100 - $value;
+            }
```
nijel: Wouldn't it be better to add it directly to $options rather than calculating the needed factor here. Or at least this code should be documented as it's certainly not obvious what it is supposed to do.

plartem:
@nijel I tried to add it directly to $options, can it still be improved?

nijel:
Indeed that is way easier to understand, thanks!

Merged squashed commit as 6014898779.

Issue #13040, opened 2017-02-28
phpmyadmin 4.6.4 doesn't work under HHVM 3.18


alex-mashin:
**Operating system**: Ubuntu 16.10
**Web server:** nginx 1.10.3
**Database:** MariaDB 10.1.21
**PHP version:** 7.0.99-hhvm 
**phpMyAdmin version:** 4.6.4

Doesn't work but can be fixed.

To get to the login screen, replace all `Error::` in `/usr/share/phpmyadmin/libraries/Error.php` with `self::`.

To get past the login screen, replace all `false` in `$this->_realConnect (…)` calls in `/usr/share/phpmyadmin/libraries/dbi/DBIMysqli.php` with `null`.

After that the interface may hang HHVM service a couple of times, but after restarting HHVM it becomes responsitive.

nijel:
I've fixed the Error class, however the mysqli connection code has been already completely rewritten and I don't think it has problem's you've mentioned, see eafd5a5d9337fa812d70607d053d7bea348cdcb8.

nijel:
I'll also try to enable testing against more hhvm versions, see https://github.com/phpmyadmin/phpmyadmin/pull/13041

nijel:
I've added hhvm 3.18 to our CI tests, so it should now work fine.

There is one thing I've noticed though: the tests on 3.18 take way longer than they used to take on 3.3 (see https://github.com/phpmyadmin/phpmyadmin/issues/13046 and https://github.com/facebook/hhvm/issues/7705), so there might be some performance issues with using phpMyAdmin as well...

alex-mashin:
There definitely is. It looks like a whole lot of code has to be compiled before anything runs; effectively, no JIT.

nijel:
The compilation is not issue, it's the execution which is damn slow. The time goes up at least exponentially when adding more lines, so it really looks like some problem in JIT. Also it works just fine in HHVM 3.3.

PR #13041, opened 2017-03-01
Test against multiple HHVM versions

nijel:
This requires test environment upgrade to trusty.

Issue #13040

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,5 +91,3 @@ addons:
     packages:
     - git
     - mysql-server-5.6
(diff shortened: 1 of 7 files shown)


nijel:
This *mostly* works, however MySQL is not always running in the test what makes the whole thing fail, see https://github.com/travis-ci/travis-ci/issues/6842

nijel:
Great, it seems to work fine with `sudo: required` and there does not seem to be any noticeable performance penalty from this.

PR #13054, opened 2017-03-05
Fixes #13029

### LINKED ISSUE #13029: Table tracker - PMA_getUntrackedTables() broken for nested table groups
### Steps to reproduce
1. Create DB with 3 tables: "hello_world", "hello_lovely_world, "hello_lovely_world2".
2. Under Settings > Nav Panel > Tables tab, increase "Maximum table tree depth" to 2 (causing 2 levels of grouping: "hello" group, containing "hello_world" table and "hello_lovely" group, which in turn contains the 2 other tables).
3. Go to DB > Tracking tab - only "hello_world" will be listed under 'Untracked tables'.

### Expected behaviour
All 3 tables should be listed.

### Actual behaviour
Only "hello_world" listed, because libraries > tracking.lib.php > PMA_getUntrackedTables() doesn't cope with nested groups. Should be rewritten to recursively handle table groups instead assume max depth = 1.

### Server configuration
**Operating system**: CentOS v6.8

**Web server:** Apache v2.4.23

**Database:** MySQL v5.6.35

**PHP version:**  v5.6.25

**phpMyAdmin version:** v4.6.6

### Client configuration
**Browser:** Firefox v50.1.0

**Operating system:** Win 7 64-bit

Achilles-96:

### DIFF
--- a/libraries/tracking.lib.php
+++ b/libraries/tracking.lib.php
@@ -1524,6 +1524,33 @@ function PMA_displayOneUntrackedTable($db, $tablename, $url_query)
 }
 
 /**
+ * Helper function: Recursive function for getting table names from $table_list
+ *
+ * @param string $db current database
+ *
+ * @return array $untracked_tables
+ */
+function PMA_extractTableNames($table_list, $db, $testing=false) {
+    $untracked_tables = array();
+    $sep = $GLOBALS['cfg']['NavigationTreeTableSeparator'];
+
+    foreach ($table_list as $key => $value) {
+        if (is_array($value) && array_key_exists(('is' . $sep . 'group'), $value)
+            && $value['is' . $sep . 'group']
+        ) {
+            $untracked_tables = array_merge(PMA_extractTableNames($value, $db), $untracked_tables); //Recursion step
+        }
+        else {
+            if (is_array($value) && ($testing || Tracker::getVersion($db, $value['Name']) == -1)) {
+                $untracked_tables[] = $value['Name'];
+            }
+        }
+    }
+    return $untracked_tables;
+}
+
+
+/**
  * Get untracked tables
  *
  * @param string $db current database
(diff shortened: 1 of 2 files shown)


ibennetch:
Thanks for your contribution. I haven't examined the code, but it's generally better to use a more descriptive commit message. Something like:

> Fix table tracking interface with grouped tables
> 
> Fixes #13029 

Whichever developer reviews and merges this may be able to improve it when merging, so don't worry about fixing it just yet.

nijel:
It would be great if you could create testcase for this code as well.

Achilles-96:
Hi @nijel , I am new to writing tests. Do I have to create a fake db. If yes, how do I create? This is required because Tracker::getVersion($db, $table_name) actually executes an sql query to get the status. I think there must be a way without creating a db. I think I need some help here.

nijel:
You can use Dummy dbi driver, see for example [DatabaseInterfaceTest::testGetCurrentUser](https://github.com/phpmyadmin/phpmyadmin/blob/master/test/classes/DatabaseInterfaceTest.php#L42).

Achilles-96:
I checked your comment a bit late. I tried to avoid testing getUntrackedTables itself. I just wrote a test to check the helper recursive function. I will change it if you recommend otherwise.

libraries/tracking.lib.php
```diff
@@ -1524,50 +1524,46 @@ class="checkall" id="<?php echo $checkbox_id;?>"
...
- * Get untracked tables
+ * Helper function: Recursive function for getting untracked tables
  *
  * @param string $db current database
  *
  * @return array $untracked_tables
  */
-function PMA_getUntrackedTables($db)
-{
+function getList($table_list, $db, $testing=false) {
```
nijel: I guess it should now have the PMA prefix and probably also better name to indicate what it does.

nijel (changes requested):
No problem, that is good approach as well. However the function probably should get better name now. Also it would be great if you could provide reasonable description in the commit messages as @ibennetch has mentioned in https://github.com/phpmyadmin/phpmyadmin/pull/13054#issuecomment-284275619

Achilles-96:
Any update on this?

nijel:
Merged into QA_4_7 as ebc935a91514bb3ca2faa6447227eb971da1469f, thanks for your contribution!

neogic:
Thank you for the quick fix @Achilles-96 - fantastic! :)

PR #13060, opened 2017-03-08
Implement PSR-11

### LINKED ISSUE #13052: [DIC] Implement PSR-11
The [PSR-11](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-11-container.md) standardizes the way entries are retrieved from a container.

`PMA\libraries\di\Container` should implement `Psr\Container\ContainerInterface`.

- https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-11-container.md
- https://packagist.org/packages/psr/container

MauricioFauth:
Fixes #13052

### DIFF
--- a/libraries/di/Container.php
+++ b/libraries/di/Container.php
@@ -54,26 +54,16 @@ class Container implements ContainerInterface
      */
     public function get($name, $params = array())
     {
-        try {
-            if (!$this->has($name)) {
-                throw new NotFoundException("No entry was found for $name identifier.");
-            }
-        } catch (NotFoundException $e) {
-            echo $e->getMessage();
-            return null;
+        if (!$this->has($name)) {
+            throw new NotFoundException("No entry was found for $name identifier.");
         }
 
-        try {
-            if (isset($this->content[$name])) {
-                return $this->content[$name]->get($params);
-            } else if (isset($GLOBALS[$name])) {
-                return $GLOBALS[$name];
-            } else {
-                throw new ContainerException("Error while retrieving the entry.");
-            }
-        } catch (ContainerException $e) {
-            echo $e->getMessage();
-            return null;
+        if (isset($this->content[$name])) {
+            return $this->content[$name]->get($params);
+        } else if (isset($GLOBALS[$name])) {
+            return $GLOBALS[$name];
+        } else {
+            throw new ContainerException("Error while retrieving the entry.");
         }
     }
 
@@ -90,15 +80,7 @@ class Container implements ContainerInterface
      */
     public function has($name)
     {
-        if (isset($this->content[$name])) {
-            return true;
-        }
-
-        if (isset($GLOBALS[$name])) {
-            return true;
-        }
-
-        return false;
+        return isset($this->content[$name]) || isset($GLOBALS[$name]);
     }
 
     /**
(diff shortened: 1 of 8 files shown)


mauriciofauth:
I've never written tests before, so I do not know if the ones I've written are correct.

libraries/di/Container.php
```diff
@@ -46,19 +47,58 @@ public function __construct(Container $base = null)
...
     public function get($name, $params = array())
     {
+        try {
+            if (!$this->has($name)) {
+                throw new NotFoundException("No entry was found for $name identifier.");
+            }
+        } catch (NotFoundException $e) {
+            echo $e->getMessage();
+            return null;
+        }
```
nijel: I really don't get the code here:

- Why is there `try { throw } catch {}` block at all?
- The echo is there for something else than debugging purposes?

libraries/di/Container.php
```diff
@@ -46,19 +47,58 @@ public function __construct(Container $base = null)
      * @param string $name   Name
      * @param array  $params Parameters
      *
+     * @throws NotFoundException  No entry was found for **this** identifier.
+     * @throws ContainerException Error while retrieving the entry.
```
nijel: It doesn't seem to throw these as they are always handled inside...

libraries/di/Container.php
```diff
@@ -46,19 +47,58 @@ public function __construct(Container $base = null)
...
+            return true;
         }
 
         if (isset($GLOBALS[$name])) {
-            return $GLOBALS[$name];
+            return true;
         }
 
-        return null;
+        return false;
```
nijel: Probably better to write whole method as:

```php
    return isset($this->content[$name]) || isset($GLOBALS[$name]);
```

test/classes/di/ContainerTest.php
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    /**
+     * Test for get
+     * @expectedExceptionMessageRegExp #Right.*#
+     * @return void
+     */
+    public function testGetWithInvalidEntry()
+    {
+        $this->assertSame(null, $this->container->get('invalid'));
```
nijel: IMHO it should throw NotFoundExceptionInterface according to PSR-11.

nijel (changes requested):

nijel:
And now the code code which uses container needs to be adjusted to understand new behavior, see at least the test failures: https://travis-ci.org/phpmyadmin/phpmyadmin/jobs/210553712#L457

mauriciofauth:
I will do this @nijel 

And about the `get` method, I think it would not be appropriate to use `isset` since this already happens in the `has` method. I think it should check whether the value is valid or not. Maybe `!is_null`?

nijel:
Why should it check for null value? I think it's perfectly okay to return null if it has been set...

mauriciofauth:
Oh, of course. I was not thinking straight.
Looking at the code, I realized that all values are instances of `ValueItem`. Maybe check for it.
I'm wanting to check for something other than `isset`, otherwise `ContainerException` will never be thrown.

mauriciofauth:
I left `get` method the way it was and corrected the tests that were failing.
Is everything all right now, @nijel?

nijel:

Issue #12984, opened 2017-02-12
Show default collation when creating new DB


OlafvdSpek:
When creating a database:
The Collation dropdown shows a placeholder currently, I think it'd be more useful if it showed the actual default collation that'd be used if you didn't pick one manually.

nimeshmora:
i am a newbie here who is interested in open source development. Can i try one of these issues? (Already forked your master branch & trying to get familiar with the code)

ibennetch:
@nimeshmora Sure, you can work on this issue. It's currently not assigned and no one else has requested to work on it here in the comments, so you're welcome to do so.

nimeshmora:
in index page it shows default server collation. So when creating new DB, you need to  show the default server collation as it shows in index page?

devenbansod:
Hi @nimeshmora the collation dropdown in the create new database form does not have a pre-selected value (it has 'Collation' pre-selected, whose value is '').

The issue requests to show the default server collation pre-selected (which is anyway going to be used by MySQL/MariaDB under the hood, if none is specified) so that user might change it accordingly if he/she wants to.

manishbisht:
I want to work on this issue. I think no one is right now working on this.

Fenn-CS:
I have solved this issue, I will like my code to be merged thank you.

PR #13061, opened 2017-03-09
Solved-Show default collation when creating new DB #12984

nfebe:
#12984

### DIFF
--- a/libraries/Charsets.php
+++ b/libraries/Charsets.php
@@ -7,6 +7,7 @@
  */
 namespace PMA\libraries;
 use PMA\libraries\Util;
+use PMA\libraries\DatabaseInterface;
 
 /**
  * Class used to manage MySQL charsets
@@ -203,14 +204,8 @@ class Charsets
         . (empty($id) ? '' : ' id="' . htmlspecialchars($id) . '"')
         . ($submitOnChange ? ' class="autosubmit"' : '') . '>' . "\n";
         if ($label) {
-         //Find default collation value
-            $sql = 'SELECT DEFAULT_CHARACTER_SET_NAME, DEFAULT_COLLATION_NAME FROM INFORMATION_SCHEMA.SCHEMATA WHERE SCHEMA_NAME ='.'"information_schema"';
-            $res = $GLOBALS['dbi']->query($sql);
-            $default_collation ='';
-            while ($row = $GLOBALS['dbi']->fetchAssoc($res)) {
-                // $char_set_name = $row['CHARACTER_SET_NAME'];
-                $default_collation = $row['DEFAULT_COLLATION_NAME'];
-            }
+         
+            $default_collation = $GLOBALS['dbi']->getServerCollation();
 
             $return_str .= '<option value="">'
             . __($default_collation)
--- a/libraries/Charsets.php
+++ b/libraries/Charsets.php
@@ -7,7 +7,6 @@
  */
 namespace PMA\libraries;
 use PMA\libraries\Util;
-use PMA\libraries\DatabaseInterface;
 
 /**
  * Class used to manage MySQL charsets
--- a/libraries/Charsets.php
+++ b/libraries/Charsets.php
@@ -77,7 +77,7 @@ class Charsets
         $GLOBALS['dbi']->freeResult($res);
 
         sort(self::$_charsets, SORT_STRING);
-    }[ \t]+\n
+    }
 
     /**
      * Loads collation data from the MySQL server.
--- a/libraries/Charsets.php
+++ b/libraries/Charsets.php
@@ -7,7 +7,6 @@
  */
 namespace PMA\libraries;
 use PMA\libraries\Util;
-
 /**
  * Class used to manage MySQL charsets
  *
@@ -15,7 +14,6 @@ use PMA\libraries\Util;
  */
 class Charsets
 {
-
     /**
      * MySQL charsets map
      *
@@ -46,12 +44,10 @@ class Charsets
         'windows-1256' => 'cp1256',
         'windows-1257' => 'cp1257',
     );
-
     private static $_charsets = array();
     private static $_charsets_descriptions = array();
     private static $_collations = array();
     private static $_default_collations = array();
-
     /**
      * Loads charset data from the MySQL server.
      *
@@ -63,10 +59,8 @@ class Charsets
         if (count(self::$_charsets) > 0) {
             return;
         }
-
         $sql = 'SELECT * FROM information_schema.CHARACTER_SETS';
         $res = $GLOBALS['dbi']->query($sql);
-
         self::$_charsets = array();
         while ($row = $GLOBALS['dbi']->fetchAssoc($res)) {
             $name = $row['CHARACTER_SET_NAME'];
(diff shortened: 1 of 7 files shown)


ibennetch:
Something may have gone wrong, because there are 2,268 files changed with this pull request.

Some of it looks like permission problems, but often this can be caused by working from one branch (such as QA_4_7) and creating a pull request against a different branch (master, for instance).

Fenn-CS:
@ibennetch  they are permission problems I work on ubuntu I had to change the permission to 777 to have full access on my apache server...How do I fix this pull request?

Fenn-CS:
@ibennetch I have fixed every issue and only the file that needs to be changed to fix the error is marked changed now. Thanks

Fenn-CS:
changes

OlafvdSpek:
The whitespace changes shouldn't be there.

ibennetch:
Indeed, the whitespace changes make it difficult to see what this accomplishes (even though, it seems the switch/case indentation could be fixed, that should be a different pull request). 

Perhaps your editor was making whitespace "fixes" without your knowledge; many editors that do autoindentation can do that inadvertently.

Also, your commit is missing the "signed off by" comment that indicates you agree with and abide by the terms of the "Developer Certificate of Origin"; more information about this is linked by phpmyadmin-bot above.

ibennetch:
I just tested this by changing the default collation of my server and the displayed value did not update; I'm not sure from where exactly you're gathering that information but it may not be correct.

Fenn-CS:
@ibennetch does this mean I supposed to close this pull request in order to fix the issue and then create a new pr?

ibennetch:
No, you can keep this open and push more commits to your branch. When you do so, the pull request automatically updates.

I prefer that because it's easier and less noisy than opening a new pull request.

Fenn-CS:
@ibennetch ok thanks, on the test that failed this is how I get the information about default collation 
`
$sql='SELECT DEFAULT_CHARACTER_SET_NAME, DEFAULT_COLLATION_NAME FROM INFORMATION_SCHEMA.SCHEMATA WHERE SCHEMA_NAME ='.'"information_schema"'`;

Fenn-CS:
It simply gets the information from the information_schema table. Maybe you check whether actually the default collation was actually changed. I think I have been able to fix the indentation problems (white space changes) with the pull request also thanks.

ibennetch:
If I'm not mistaken, that only tells you the default collation for the information_schema database, not for the whole system.

libraries/Charsets.php
```diff
@@ -191,37 +191,46 @@ public static function getCharsetDropdownBox(
...
-            . htmlspecialchars($name) . '"'
-            . (empty($id) ? '' : ' id="' . htmlspecialchars($id) . '"')
-            . ($submitOnChange ? ' class="autosubmit"' : '') . '>' . "\n";
+        . htmlspecialchars($name) . '"'
+        . (empty($id) ? '' : ' id="' . htmlspecialchars($id) . '"')
+        . ($submitOnChange ? ' class="autosubmit"' : '') . '>' . "\n";
         if ($label) {
+         //Find default collation value
+            $sql = 'SELECT DEFAULT_CHARACTER_SET_NAME, DEFAULT_COLLATION_NAME FROM INFORMATION_SCHEMA.SCHEMATA WHERE SCHEMA_NAME ='.'"information_schema"';
+            $res = $GLOBALS['dbi']->query($sql);
```
devenbansod: @Fenn-CS I suggest you use the existing function `getServerCollation()` defined in `libraries/DatabaseInterface.php` for getting the server collation (which will act as a default for a newly created database, if nothing is specified), instead of separately querying it using a raw query (which is not returning the correct value that we should look at actually) in the code here.

devenbansod:
@ibennetch yes. That's right. 

Also, @Fenn-CS I can still see the whitespace changes in this PR, please remove them while committing with the requested changes.

devenbansod (changes requested):

Fenn-CS:
I am using sublime text and am having a hard time with these white space changes... Any help will be appreciated thanks.

Fenn-CS:
@devenbansod I have applied the changes your requested. Thnks

libraries/Charsets.php
```diff
@@ -7,6 +7,7 @@
  */
 namespace PMA\libraries;
 use PMA\libraries\Util;
+use PMA\libraries\DatabaseInterface;
```
devenbansod: Unused import ?

libraries/Charsets.php
```diff
@@ -45,7 +46,7 @@ class Charsets
         'windows-1252' => 'latin1',
         'windows-1256' => 'cp1256',
         'windows-1257' => 'cp1257',
-    );
+        );
```
devenbansod: It seems that the whitespace problem still persists. Please fix this.

libraries/Charsets.php
```diff
@@ -145,32 +146,32 @@ public static function getMySQLCollationsDefault()
     public static function getCharsetDropdownBox(
         $name = null, $id = null, $default = null, $label = true,
         $submitOnChange = false
-    ) {
+        ) {
         self::loadCharsets();
```
devenbansod: I believe that this is not the correct place to make this change. 
The function `getCharsetDropdownBox` already takes a default value as a parameter `$default` and we don't every charset dropdown to have a default value as the server collation.

For example, the table creation might use the database collation as the default value and not the server collation.

Fenn-CS:
@devenbansod I am having challenges fixing the white space issues...They seem to be created by sublime text how do I combat this?

devenbansod:
@Fenn-CS I am not sure what is the problem, since even I use Sublime text sometimes, but it does not create any issue at all. It might be some configuration issue, so you can better ask for a solution to your problem on their support forum or on Stackoverflow. Till then you may want to use a different text editor. 

Also, use `git commit -s` while commiting, to avoid the error regarding the missing Signed-Off-By line.

Fenn-CS:
@devenbansod I have moved the changes from the general function to the specific page where the performance was expected now the default collation value is passed like a parameter to the function I previously edited. I have also addressed the issue with Trailing white spaces as from my last commit. Thanks

libraries/Charsets.php
```diff
@@ -472,4 +472,4 @@ public static function getCollationDescr($collation)
 
         return $descr;
     }
-}
+}
```
devenbansod: There are just two extra space changes in unrelated files. Please remove them and then [squash](https://git-scm.com/book/id/v2/Git-Tools-Rewriting-History#_squashing) all the commits into one.

Fenn-CS:
Hello @devenbansod and @ibennetch I have made all the changes requested. I think every thing is ok now apart from the fact I have been unable to squash all the commits into one. I will need help on that I tried doing that and it seemed to revert my work bringing back old changes. Apart from that I think it solves the  problem as required right now...

devenbansod (approved):

devenbansod:
I am squashing and merging your commits.

Also a suggestion regarding your next PRs, please make a new branch while implementing a feature/fixing a bug, instead of directly working on master/QA_* branches on your fork.

Thanks for your contribution!

Fenn-CS:
@devenbansod thanks. My pleasure to work with phpMyAdmin! :)

PR #13070, opened 2017-03-12
Changed button string in optional actions -table structures

nfebe:
Change string on one of the optional action buttons-From 'Improve table structure' to 'Normalize'
#12961


nijel:

Fenn-CS:
@nijel my pleasure happy to be contributing to PMA.

Issue #13058, opened 2017-03-08
Improve Captcha error message if service unreachable


ibennetch:
If the Captcha service is unreachable (for instance, if phpMyAdmin is running on a machine without network access), the error message shown "Entered captcha is wrong, try again!" is misleading.

I don't see an obvious way to improve this, but it would be great if we could detect various failure modes (unable to reach the service vs failing the test) and inform the user according to the failure type.

edwarddekker:
Than you have to detect the internet connection to the Google server for this failure.
This needs a script to view if the internet connection will be alive at all times.

nijel:
Probably using something like [this](http://stackoverflow.com/a/39160653/225718) will do the trick.

modivivek:
hey there,
if phpmyadmin is running on a pc without internet connection the captcha service will not show up on the login page i think so.
Accordingly we need to check  if the captcha service is reachable when the pc is connected to the internet,right?

Fenn-CS:
Hello @ibennetch  this my version information: 4.8.0-dev, latest stable version: 4.6.6 . How can I reproduce this issue. I see a captcha when logging in. Thanks

ibennetch:
My use case is a laptop which is sometimes not connected to wifi, but in further testing, the issue gets more difficult.

If I'm offline but my browser cache is rather fresh, I seem to get the "I am not a robot" checkbox, and if I check it, I pass the test. This seems like good behavior to me.

Sometimes when I'm offline, I need to pass the image identification portion of the test. In this case, it seems to show me endless pictures without allowing me to pass the test.

[ ] We should detect this scenario (if possible) to show an error message so the user doesn't try several times and get frustrated.

If I'm offline but my browser cache is expired or if I'm using incognito mode, the test is never shown, the "Go" button remains disabled, and I am completely unable to log in.

[ ] We should provide some feedback to the user in this case.

If I'm online, everything functions as expected.

Fenn-CS:
So far I have not been able to reproduce this issue simple because I have not encountered a captcha on my version even if I was connected or not.

modivivek:
@ibennetch 
It did not show the entered captcha is wrong instead it showed this message.
![screenshot from 2017-03-11 12-23-39](https://cloud.githubusercontent.com/assets/20069594/23821350/aaf9878a-0655-11e7-9acc-8f2fdba996a8.png)

modivivek:
@ibennetch 
[...]

PR #13071, opened 2017-03-13
#13058 Improve Captcha error message if service is unreachable

modihere:

### DIFF
--- a/libraries/plugins/auth/AuthenticationCookie.php
+++ b/libraries/plugins/auth/AuthenticationCookie.php
@@ -219,7 +219,6 @@ class AuthenticationCookie extends AuthenticationPlugin
             echo '    <input type="hidden" name="server" value="'
                 , $GLOBALS['server'] , '" />';
         } // end if (server choice)
-
         // Add captcha input field if reCaptcha is enabled
         if (!empty($GLOBALS['cfg']['CaptchaLoginPrivateKey'])
             && !empty($GLOBALS['cfg']['CaptchaLoginPublicKey'])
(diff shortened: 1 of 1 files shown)


libraries/plugins/auth/AuthenticationCookie.php
```diff
@@ -219,19 +219,32 @@ public function auth()
...
         if (!empty($GLOBALS['cfg']['CaptchaLoginPrivateKey'])
             && !empty($GLOBALS['cfg']['CaptchaLoginPublicKey'])
         ) {
             // If enabled show captcha to the user on the login screen.
-            echo '<script src="https://www.google.com/recaptcha/api.js?hl='
-                , $GLOBALS['lang'] , '" async defer></script>';
-            echo '<div class="g-recaptcha" data-sitekey="'
-                , htmlspecialchars($GLOBALS['cfg']['CaptchaLoginPublicKey']) ,
-                '" data-callback="loginButtonEnable" data-expired-callback="loginButtonDisable" captcha="enabled"></div>';
+            $connected = @fsockopen("www.google.com",80, $errno, $errstr, 30);
```
nijel: This can easily fail if fsockopen is disabled...

modivivek: I think we can check every time that fsockopen is enabled or not.
And if fsockopen is disabled can it be enabled without changing in php.ini file?
Or what do you suggest as a replacement for this?

nijel: I think it would be really better to use existing ReCaptcha code to check if it works, otherwise you end up duplicating it's logic to behave same. Actually doing that on captcha verification might be good enough...

modivivek: But won't it be better not to touch the code written by google for ReCaptcha and do the network checking part differently just like I did?
But anyways for your suggestion which particular files should I focus on modifying if I want to check the network connection through recaptcha?

nijel: I'm not saying you should touch the Google code, but to use error their library returns to indicate the problem to user. Otherwise you end up two code paths behaving differently and sometimes reporting error when there is none (eg. when the proxy is used, which the library will handle while fsockopen will not).

modivivek: Oh got it..sorry for the trouble...I missed out on the proxy part. 
When there is no internet connection then there is no captcha shown and the GO button is disabled as pre-written in the code..if I can add the message with that "the captcha service is unavailable at the moment" that will be ok I think so?

ibennetch:
The message seems okay, but we should display it in the phpMyAdmin error styling, such as:

image

instead of

image

Also, this seems to make the page take a long time to submit as compared to master.

modivivek:
@ibennetch  
On it..will try to change it into PHPMyAdmin error style.
one more thing Michael said, " but to use error their library returns to indicate the problem to user".Can you please point me out which part of the code will I specifically find this returned error message so that i can work on it?

modivivek:
@ibennetch  
Is this ok?
And can you please clarify my doubt which i wrote in my last comment?
screenshot from 2017-03-18 11-32-24

nijel:
As written before, I think this approach is wrong - it uses completely different code path to connect to the service making it source of false positives.

I've rather implemented error message based on ReCaptcha library return values in 7d1b0ca1ed6be52087a5d09ed52f5dd519c48a8e.

PR #13072, opened 2017-03-13
Fixes #12393

### LINKED ISSUE #12393: Improve testsuite for cookie authentication
Currently the cookie authentication tests are quite poorly matching the functionality of the code. There should be separate tests for each authentication step (login, recall from cookies, password change and logout) for possible configurations (plain, arbitrary server or captcha). Some of these tests already exists, but they probably need cleanup as the underlaying code has been changed a lot in 4.6.4.

osaid-r:
Signed-off-by: osaid <osaid.nasir@gmail.com>

### DIFF
--- a/test/classes/plugin/auth/AuthenticationCookieTest.php
+++ b/test/classes/plugin/auth/AuthenticationCookieTest.php
@@ -1049,7 +1049,7 @@ class AuthenticationCookieTest extends PMATestCase
         );
     }
 
-    public function testPasswordChange($value='')
+    public function testPasswordChange()
     {
         $newPassword = 'PMAPASSWD2';
         $GLOBALS['cfg']['AllowArbitraryServer'] = true;
@@ -1079,7 +1079,7 @@ class AuthenticationCookieTest extends PMATestCase
             $GLOBALS['cfg']['LoginCookieStore']
         );
         $this->assertEquals(
-            $_COOKIE['pmaAuth-' . $GLOBALS['server']].
+            $_COOKIE['pmaAuth-' . $GLOBALS['server']],
             $_COOKIE['testPasswordCookie']
         );
     }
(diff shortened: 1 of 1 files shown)


test/classes/plugin/auth/AuthenticationCookieTest.php
```diff
@@ -1049,6 +1049,40 @@ public function testAESSecretSplit($secret, $mac, $aes)
...
+            'server' => 'b 2'
+        );
+        $method = new \ReflectionMethod(
+            'PMA\libraries\plugins\auth\AuthenticationCookie',
+            '_getSessionEncryptionSecret'
+        );
+        $method->setAccessible(true);
+
+        $result = $method->invoke($this->object, null);
+        $GLOBALS['PMA_Config']->setCookie(
```
ragnerok: Yeah but then I'll have to use assertContains() right ?

nijel: You just need to compare whether cookieEncrypt return value is same as the pmaAuth cookie, or am I missing something?

ragnerok: So you are saying I should simply assertEquals pmaAuthCookie with returned value of cookieEncrypt but I was saying that I could call cookiedecrypt and assert with$newPassword

nijel: The cookieDecrypt takes string as a parameter, there is no need to pollute $_COOKIES superglobal for the test purpose here.

ragnerok: Yeah ok

ragnerok:
I think that the build is failing because createIV() probably returns random strings to cookieEncrypt(), right ?

nijel:
Yes, createIV is supposed to return new IV every time.

ragnerok:
So if I set $_cookie_iv then this problem shouldn't be there

nijel:

PR #13079, opened 2017-03-14
Removed inline onmousedown javascript

### LINKED ISSUE #12261: Remove inline javascript

manishbisht:
Fixes : #12261 

### DIFF
--- a/js/pmd/move.js
+++ b/js/pmd/move.js
@@ -132,6 +132,20 @@ function MouseDown(e)
 {
     Glob_X = dx = isIE ? e.clientX + document.body.scrollLeft : e.pageX;
     Glob_Y = dy = isIE ? e.clientY + document.body.scrollTop : e.pageY;
+
+    if (e.target.tagName == "SPAN") {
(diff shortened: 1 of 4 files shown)


Fenn-CS:
@manishbisht great.. I thought or rather expected to see an additional js file containing the code removed from in-line or added to and existing js file. It looks like some functionality might be brick by simple removing the in-line js.

nijel:
With your change moving objects in designer no longer works as before. Before it was possible to drag by whole table header, now it works only on the text.

manishbisht:
Are you sure about this. Because I have cross checked the master branch and found that it only works when mouse was moved over the table text. Correct me if I am missing something.

nijel:
You don't have the extra space where it doesn't work in your example :-). You need table with longer column definitions:

page-shot-2017-3-15 localhost localhost gammu phpmyadmin 4 8 0-dev

Here moving works fine on the text, but when I try to move using the big title area without text, it no longer works after your changes.

manishbisht:
Thanks for the quick update. Just updated the PR according to it. :)

nijel:

PR #13093, opened 2017-03-17
Fixed time selection for TIME type

mohdluqman:
Fixed the automatic rounding off of time greater than 23. TIME type in MySql supports a long range of values. The PMA insert                                                      widget don't allow the values greater than 23 currently. So, it can be                                                      corrected by setting the timepicker function have a greater hourMax                                                     value than it have by default in the file functions.js.


js/functions.js
```diff
@@ -182,12 +183,15 @@ function addDateTimePicker() {
                     timeFormat = 'HH:mm:ss.l';
                 }
             }
+            if(type == 'time'){
```
nijel: Can you please add space here after if?

js/functions.js
```diff
@@ -182,12 +183,15 @@ function addDateTimePicker() {
...
+                hourMax = 99;
```
nijel: Why 99 when MySQL docs states that the maximal value is 838?

nijel (changes requested):

mohdluqman:
Sorry for the late reply as your comment was not showing on my dashboard. As per your first instruction I have added the space after if statement and for your second comment, I used 99 because

- The timepicker plugin doesn't work well with the 3 digit hour values (i.e. > 99) based on my input testing. Since I found 99 the greatest 2 digit hour value so i used it.

- As per @ibennetch comment which says "phpMyAdmin validation function doesn't like hour values of 24 through 99 and automatically converts them to 23" I derived the meaning that 99 should be the largest hour values.

If you want some other modifications or give some suggestions please describe.

mohdluqman:
Sorry for the late reply as your comment was not showing on my dashboard and also sorry for above closing mistake(being a newbie).  As per your first instruction I have added the space after if statement and for your second comment, I used 99 because

The timepicker plugin doesn't work well with the 3 digit hour values (i.e. > 99) based on my input testing. Since I found 99 the greatest 2 digit hour value so i used it.

As per @ibennetch comment which says "phpMyAdmin validation function doesn't like hour values of 24 through 99 and automatically converts them to 23" I derived the meaning that 99 should be the largest hour values.

If you want some other modifications or give some suggestions please describe.

mohdluqman:
Sorry for the late reply as your comment was not showing on my dashboard and also sorry for above closing mistake(being a newbie). As per your first instruction I have added the space after if statement and for your second comment, I used 99 because

- The timepicker plugin doesn't work well with the 3 digit hour values (i.e. > 99) based on my input testing. Since I found 99 the greatest 2 digit hour value so i used it.

- As per @ibennetch comment which says "phpMyAdmin validation function doesn't like hour values of 24 through 99 and automatically converts them to 23" I derived the meaning that 99 should be the largest hour values.

If you want some other modifications or give some suggestions please describe.

ibennetch:
The validation function seems to have some odd behavior which isn't the correct behavior. I was more documenting another layer of trouble with this functionality than explaining how it should work.

mohdluqman:
Then what should we do about the bug. IMO it would be better 

- to overload the method **'regstr'**  at L1188 in jquery-ui-timepicker-addon.js file to accept hour values of 3 digits or change its regex (according to my code reading).

- to use some other time checker plugin.

- to restrict hour values within two digit only.

ibennetch:
In my opinion, making the picker accept all possible values will make control of the scroller too unrefined to be useful. My preference is to have the picker go from 0 to 24 hours, 0 to 60 minutes, and 0 to 60 seconds combined with two fixes to the validator function. 1) it should not automatically correct values, instead it show just show the red warning text and the value is invalid to MySQL, and 2) it should accept valid values from 24-100.

That's my opinion of the best solution here.

nijel:

Issue #12881, opened 2017-01-09
Search in database fails if locale not set to "en"


emayoral:
### Steps to reproduce
1. Log in with a locale other than english
2. Search on the database. Select search in all tables. Any search term will do.

### Expected behaviour
Search returns results. You can see expected behaviour if your locale is "en"

### Actual behaviour
I get an error message (locale dependant):
Error al procesar la petición
Código de error: 500
Texto de error: Internal Server Error

Error in php error log:
[09-Jan-2017 09:42:56 Europe/Madrid] PHP Fatal error:  Uncaught exception 'InvalidArgumentException' with message 'Select_string only accepts integers: 2883' in /usr/share/php/gettext/gettext.php:354
Stack trace:
#0 /usr/share/php/gettext/gettext.php(388): gettext_reader->select_string('2883')
#1 /usr/share/php/gettext/gettext.inc(293): gettext_reader->ngettext('%1$s match in <...', '%1$s matches in...', '2883')
#2 /usr/share/phpMyAdmin/libraries/DbSearch.php(331): _ngettext('%1$s match in <...', '%1$s matches in...', '2883')
#3 /usr/share/phpMyAdmin/libraries/DbSearch.php(283): PMA\libraries\DbSearch->_getResultsRow('MD5FaMiLyBeEr5M...', Array, true, '2883')
#4 /usr/share/phpMyAdmin/db_search.php(57): PMA\libraries\DbSearch->getSearchResults()
#5 {main}
  thrown in /usr/share/php/gettext/gettext.php on line 354

Seems the problem is select_string only accepts integers but it is being passed a string because all mysql values are returned as strings and the value is not casted as integer. 

Suggested fix (at least works for me):
```
diff -u /usr/share/phpMyAdmin/libraries/DbSearch.php.patched /usr/share/phpMyAdmin/libraries/DbSearch.php
--- /usr/share/phpMyAdmin/libraries/DbSearch.php.patched	2017-01-09 09:37:48.659142822 +0100
+++ /usr/share/phpMyAdmin/libraries/DbSearch.php	2017-01-09 09:37:58.032210548 +0100
@@ -279,7 +279,7 @@
             $num_search_result_total += $res_cnt;
             // Gets the result row's HTML for a table
             $html_output .= $this->_getResultsRow(
-                $each_table, $newsearchsqls, $odd_row, intval($res_cnt)
+                $each_table, $newsearchsqls, $odd_row, $res_cnt
             );
[...]

PR #13098, opened 2017-03-18
Fix database search with newer php-gettext

schoonc:
Backport from 4.6.6 (see #12881)

Signed-off-by: Сорокин Александр <ale10059@yandex.ru>


nijel:

However it's quite likely that this fix will never be released (the EOL for 4.0.10.x is in less than two weeks and there is no release scheduled).

adpe:
Hello @schoonc 

Why does this issue it's only backported to MAINT_4_0_10, but not to the MAINT_4_4_15 branch?

PR #13100, opened 2017-03-18
Initial support for Twig template engine

MauricioFauth:
I changed the `PMA\libraries\Template` class to add support for Twig templates.

It looks for the template with the `.twig` extension, if not found, it uses the file with the `.phtml` extension.
I also wrote a Twig extension that makes the `PMA\libraries\URL` methods available in templates.
The other files extend the I18n extension to support the gettext custom functions

**Examples**
`<?= $var ?>`
`{{ var }}`

`<?= __('Text'); ?>`
`{% trans "Text" %}`

`<?= PMA\libraries\URL::getHiddenInputs($db); ?>`
`{{ URL_getHiddenInputs(db)|raw }}`

**Documentation**
[Twig](http://twig.sensiolabs.org/doc/1.x/)
[i18n Extension](http://twig-extensions.readthedocs.io/en/stable/i18n.html)


nijel:
IMHO there should be first consensus on such change - there should be clear reasoning why to choose Twig over other templating libraries...

mauriciofauth:
So should I create a new Issue before?

nijel:
Probably rather open this topic on out mailing list.

libraries/Template.php
```diff
@@ -31,6 +31,16 @@ class Template
      */
     protected $helperFunctions;
 
+    /**
+     * Twig loader
+     */
+    protected $loader;
```
nijel: This probably doesn't have to be stored as attribute...

libraries/Template.php
```diff
@@ -31,6 +31,16 @@ class Template
...
+
+    /**
+     * Twig environment
+     */
+    protected $environment;
```
nijel: I'd prefer something like `twig_environment`, or even just `twig` to make it clear what the attribute does, `environment` is quite generic name.

libraries/Template.php
```diff
@@ -45,6 +58,13 @@ protected function __construct($name, $data = array(), $helperFunctions = array(
         $this->name = $name;
         $this->data = $data;
         $this->helperFunctions = $helperFunctions;
+
+        $this->loader = new \Twig_Loader_Filesystem(static::BASE_PATH);
+        $this->environment = new \Twig_Environment($this->loader, array(
+            'cache' => static::BASE_PATH . 'cache',
```
nijel: This is probably not a good choice. At least it should be configurable by vendor_config.inc.php to allow easier integration in Linux distributions.

Also we might prefer to have generic writable directory whose subdirectory will be template cache. This could be used for other things such as `$cfg['TempDir']` as well.

libraries/twig/NodeTrans.php
```diff
@@ -0,0 +1,28 @@
...
+class NodeTrans extends Twig_Extensions_Node_Trans
+{
+    /**
+     * @param bool $plural Return plural or singular function to use
+     *
+     * @return string
+     */
+    protected function getTransFunction($plural)
+    {
+        return $plural ? '_ngettext' : '_gettext';
```
nijel: Is there chance to support pgettext as well?

nijel (commented):
I've added some comments to the code. I don't think you should spend much time on implementing fixes for these until we agree what templating engine to use.

libraries/Template.php
mauriciofauth: Like `$cfg['TempDir'] . 'cache'`?
Or `$cfg['TempDir'] . CACHE_DIR`?

And in these cases, how do I make the cache directory ignored by git?

nijel: It would then need default value (probably configurable in vendor config as well), which you can then ignore in .gitignore.

This probably needs more thinking, because current this variable is not used directly, but rather by ConfigFile::getDefaultTempDirectory() which can end up using system /tmp dir and we certainly don't want to use that for templates cache.

mauriciofauth: Initially I had thought of a `cache` directory in the root directory and then use something like `cache/twig` or `cache/template`.

libraries/twig/NodeTrans.php
mauriciofauth: Not in this method. I would need to work more on I18nExtension.

mauriciofauth:
Hi @nijel,
Only two people responded on this topic in a week. What will be the follow-up on this?

libraries/twig/NodeTrans.php
nijel: As we're using pgettext in templates, it's good to have at least roadmap how to support that.

nijel:
I'm personally in favor of Twig as the template language is pretty much same as what I'm used to from Django. As there was no objections to using Twig from others, it's probably a good choice.

ibennetch:
I don't really have enough information or experience to have an opinion.

mauriciofauth:
I added pgettext support using this syntax:
```
{% trans %}
Text
{% context %}
Text context
{% endtrans %}
```
`<?= _pgettext('Text context', 'Text'); ?>`

nijel:
Thanks! I see there is a bit discourse in the syntax for providing context among Twig users.

For example Drupal seems to use `{% trans with {'context': 'Lolspeak'} %}`, see https://www.drupal.org/node/2049241

And there is https://github.com/twigphp/Twig-extensions/pull/158 which is waiting to be merged for few months.

But I don't see this as blockers to merging.

nijel:
I think this is really almost ready to merge. I have few questions left:

Is there way to avoid need to specify raw filter for some functions?  For example there will be many copies of `{{ URL_getHiddenInputs(...)|raw }}`. 

How does work extracting text for translation? We need to get texts from the twig templates to Gettext pot file to allow translators to translate them. I've found https://github.com/umpirsky/Twig-Gettext-Extractor to do that, but I'm not sure if it's best tool or how to make it work with context extension. Probably it's invocation have to be added to scripts/update-po to make the localization work with new templates.

mauriciofauth:
Thanks for the feedback, @nijel.

I chose this syntax for pgettext because I thought it was the easiest way to implement it since I extended the `Twig_Extensions_Extension_I18n` extension

Twig automatically escapes the templates. We can easily disable autoescape, but then we will have to explicitly escape what is necessary. Disabled is closer to the way we have today
I believe leaving it enabled makes it easy to see what functions have HTML in the code to later remove that HTML from PHP.
This decision belongs to you.

On the work of extracting the texts for translation, I believe that the best way is to extract them from the cache files, since they are PHP files.
The [i18n extension documentation](http://twig-extensions.readthedocs.io/en/stable/i18n.html#extracting-template-strings) has an example code used to force a compilation of all the cache files to extract the texts.

nijel:
I don't want to disable escaping generally, just for some functions which are known to generate HTML. That seems to be achievable by passing `array('is_safe' => array('html'))` as options to Twig_SimpleFunction.

While looking at their docs, I've noticed that you're using 1.x while there is 2.x version. Is there any good reason to stick with the older branch?

mauriciofauth:
Nice. I didn't know that option. I'll try it on.

In Packagist, Twig 2 requires PHP 7.

nijel:
Thanks, PHP 7 requirement is definitely good reason.

libraries/twig/UrlExtension.php
```diff
@@ -0,0 +1,53 @@
...
+                array('is_safe' => array('html'))
+            ),
+            new Twig_SimpleFunction(
+                'URL_getCommonRaw',
+                'PMA\libraries\URL::getCommonRaw',
+                array('is_safe' => array('html'))
+            ),
+            new Twig_SimpleFunction(
+                'URL_getArgSeparator',
+                'PMA\libraries\URL::getArgSeparator',
```
nijel: Sorry for not spotting it earlier, but I don't think that getArgSeparator and getCommonRaw should be available for templates. The getArgSeparator is useful only for manually constructing query strings, what should not be done in templates and getCommonRaw is only unescaped version of getCommon, what doesn't make sense to use in templates.

nijel:
Merged, thanks for your contribution and patience with my feedback!

Issue #13032, opened 2017-02-22
Errors in designer when database contains special chars


nijel:
When the database name contains some special chars, the designer fails to operate on that. In my particual case the database name is `"'"><iframe onload=ale2`.

This report is related to user submitted report [#31501](https://reports.phpmyadmin.net/reports/view/31501) on the phpmyadmin error reporting server.

nijel:
This is not a regression, it happens for ages (probably since the designer exists).

nijel:
There is now 4119b5ca3d2af76543ccaa5ebe1a408d2cc4fb31, but it doesn't completely fix the issue, the javascript error still happens.

Achilles-96:
Did you try 160eb93

nijel:
To reproduce problems, create following database and table:

```mysql
CREATE DATABASE IF NOT EXISTS `"'"><iframe onload=alert(1)>` 
DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
USE `"'"><iframe onload=alert(1)>`;
CREATE TABLE `"'"><iframe onload=alert(1)>` (
  `"'"><iframe onload=alert(1)>` int(11) NOT NULL,
  `data` text NOT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
ALTER TABLE `"'"><iframe onload=alert(1)>`
  ADD PRIMARY KEY (`"'"><iframe onload=alert(1)>`);
ALTER TABLE `"'"><iframe onload=alert(1)>`
  MODIFY `"'"><iframe onload=alert(1)>` int(11) NOT NULL AUTO_INCREMENT;
```

nijel:
That doesn't seem to help either.

nijel:
Really the whole approach of handling names in designer is problematic - using table names as element IDs is not a good idea as there are are at least whitespace not allowed in element ID in HTML 5 (there were much more restrictions in older HTML standards). Also the code uses urlencode where probably htmlspecialchars would make more sense.

Achilles-96:
Do you mean to rewrite entirely the way tables are handled in designer?

nijel:
Probably, I think this code is really broken in regard to escaping and it needs to be fixed at that side and not workarounding broken design on the client side (what probably is not doable for all situations).

udan11:
[...]

PR #13102, opened 2017-03-19
Fixes error in designer #13032

Achilles-96:

### DIFF
--- a/js/pmd/move.js
+++ b/js/pmd/move.js
@@ -313,6 +313,7 @@ function Rezize_osn_tab()
     var max_X = 0;
     var max_Y = 0;
     for (var key in j_tabs) {
+        key = Get_html_string(key);
         var k_x = parseInt(document.getElementById(key).style.left, 10) + document.getElementById(key).offsetWidth;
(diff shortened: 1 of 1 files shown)


js/pmd/move.js
```diff
@@ -1298,9 +1305,18 @@ function Relation_lines_invert()
     Re_load();
 }
 
+function Get_html_string(str)
+{
+    var parser = new DOMParser;
+    var dom = parser.parseFromString('<!doctype html><body>' + str,'text/html');
```
nijel: This seems to open door to XSS.

nijel (changes requested):
I don't think this is good approach. We most likely need to encode the values rather than doing some transforms on it later.

nijel:
Sorry, but I still think this is wrong approach:

- trying to treat arbitrary value (database or table name) as identifier will never work reliably
- using something like md5 or similar hash is probably better approach

Achilles-96:
I am sorry I forgot about this. I will try to send a patch where tables will have their md5 sums as their ids. I already used that approach in my other work - #13342 . I will do this as soon as I am free if no one takes this up.

Issue #13099, opened 2017-03-18
Change default database name in Operations > Copy database to


yevgenkrylov:
### Steps to reproduce
1. Go to "Databases" tab, select your database
2. Go to "Operations" tab
3. Click "Go" under "Copy database to" section (without changing pre-populated database name)

### Expected behaviour
- get a copy of your database with a different name (append something unique to the database name by default) OR
- get a warning, that user is going to destroy the database by copying it to itself OR
- do not pre-populate database name in Copy to section OR
- your option...

### Actual behaviour
Get data in your database corrupted.
I really do not understand what might be a real use case for the default behavior of pma, when it populates current database name...

### Server configuration
**Operating system**: CentOS Linux release 7.3.1611 (Core)

**Web server:** Apache 2.4.25

**Database:** 10.1.21-MariaDB - MariaDB Server

**PHP version:** 5.6.30

**phpMyAdmin version:** 4.6.6

### Client configuration
**Browser:** Chrome

**Operating system:** Windows 10

Achilles-96:
It does throw an error.
![image](https://cloud.githubusercontent.com/assets/10742413/24078257/05ed775c-0c8f-11e7-8da0-77bd08da8070.png)

yevgenkrylov:
I forgot to metion one step: tick "drop table/view" box...

TheSaturn:
I'm fixing it. Just need to write test for it

Achilles-96:
@TheSaturn , I just made sure that default name is not same as database name. Try to look into why data is getting corrupted if the db name is same and fix it. I feel that is the major issue here :) .

ibennetch:
Related to #12233 which was about the "Rename" dialog, not copy.

ibennetch:
We should definitely catch cases when the user tries this and provide a reasonable error message, but I think we also should remove the default suggestion from the copy and rename fields.

Achilles-96:
@ibennetch , Any specific reason why you want defaults to be removed? I feel they are good to enable user to create copy of db within a single click (typical use of any default) .

TheSaturn:
So I tried to fix it. I wrote a message and check if new name is the same as current. But I don't know how to write selenium test for that

nijel:
The default value has been fixed in 553d1ff, the error check on same db name is pending in https://github.com/phpmyadmin/phpmyadmin/pull/13104.

OlafvdSpek:
[...]

PR #13103, opened 2017-03-19
Adds non-conflicting default new database name #13099

Achilles-96:

### DIFF
--- a/libraries/operations.lib.php
+++ b/libraries/operations.lib.php
@@ -81,7 +81,7 @@ function PMA_getHtmlForRenameDatabase($db)
 
     $html_output .= '<input id="new_db_name" type="text" name="newname" '
         . 'maxlength="64" size="30" class="textfield" required="required" '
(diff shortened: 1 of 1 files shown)


ibennetch:
Nice work, but I think a better solution is to remove the default value entirely.

ibennetch:
Note also that we have #13104 which catches the situation where a user does enter the same database name, so we'll try continue having your pull request address the default values.

nijel (changes requested):
I'd also prefer not to have default value. Imagine user has databases named db1...db1000, with this patch going to operations tab would lead to 1000 queries just to have default value which almost nobody will use....

Achilles-96:
Do you also feel default value must be removed for rename form?

nijel:
Yes, I think having the field empty by default is best approach.

nijel (approved):

nijel:
Merged as 553d1ff343 into QA_4_7 branch, thanks for your contribution!

PR #13104, opened 2017-03-19
Adds error if target db has the same name as current db #13099

SaturnTeam:
Adds error if target db has the same name as current db name to prevent data corrupt #13099

### DIFF
--- a/db_operations.php
+++ b/db_operations.php
@@ -48,6 +48,8 @@ if (strlen($GLOBALS['db']) > 0
 
     if (! isset($_REQUEST['newname']) || strlen($_REQUEST['newname']) === 0) {
         $message = PMA\libraries\Message::error(__('The database name is empty!'));
+    } else if($_REQUEST['newname'] === $_REQUEST['db']) {
+        $message = PMA\libraries\Message::error(__('The new database name is equal to the current one!'));
     } else {
         $_error = false;
         if ($move || ! empty($_REQUEST['create_database_before_copying'])) {
--- a/templates/database/designer/database_tables.phtml
+++ b/templates/database/designer/database_tables.phtml
@@ -27,7 +27,7 @@
                 <?php endif; ?>
                 <td class="small_tab"
                     title="<?= __('Show/hide columns'); ?>"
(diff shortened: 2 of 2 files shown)


db_operations.php
```diff
@@ -48,6 +48,8 @@
 
     if (! isset($_REQUEST['newname']) || strlen($_REQUEST['newname']) === 0) {
         $message = PMA\libraries\Message::error(__('The database name is empty!'));
+    } else if($_REQUEST['newname'] === $_REQUEST['db']) {
+        $message = PMA\libraries\Message::error(__('The new database name is equal to the current one!'));
```
ibennetch: I suggest changing this message to "Cannot copy database to the same name. Change the name and try again."

templates/database/designer/database_tables.phtml
```diff
@@ -27,7 +27,7 @@
                 <?php endif; ?>
                 <td class="small_tab"
                     title="<?= __('Show/hide columns'); ?>"
-                    id="id_hide_tbody_<?= $t_n_url; ?>"
+                    id="id_hide_tbody_<?= htmlspecialchars($t_n_url) ?>"
```
ibennetch: You should really split unrelated changes in to different pull requests, even though both of these changes are only one or two lines each.

I have a database named `p&?&` and this doesn't seem to fix the problem for that or with the database name in the original error report, either.

ibennetch (changes requested):

ibennetch:
Also note that there is #13103 which is working towards changing the defaults for the the copy/rename fields.

nijel (changes requested):
As written by Isaac - try to submit one pull request with one fix. That way it's easier to merge parts of it when they are ready to get merged...

templates/database/designer/database_tables.phtml
nijel: I've cherry picked this as 4119b5ca3d2af76543ccaa5ebe1a408d2cc4fb31.

nijel:
I've committed dbdd7bcb9db9c7a54221aec6344c8c889768aea6 based on your fix and @ibennetch wording.

PR #13122, opened 2017-03-23
Go button not clicked due to not being in view

Achilles-96:

### DIFF
--- a/test/selenium/PmaSeleniumDbEventsTest.php
+++ b/test/selenium/PmaSeleniumDbEventsTest.php
@@ -49,6 +49,8 @@ class PMA_SeleniumDbEventsTest extends PMA_SeleniumBase
     public function setUpPage()
     {
         $this->login();
+        $this->waitForElement('byPartialLinkText','Databases')->click();
+        $this->waitForElementNotPresent('byCssSelector', 'div#loading_parent');
         $this->waitForElement('byPartialLinkText', $this->database_name)->click();
         $this->waitForElement(
             "byXPath", "//a[contains(., 'test_table')]"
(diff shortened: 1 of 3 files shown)


test/selenium/PmaSeleniumDbEventsTest.php
```diff
@@ -66,7 +68,7 @@ public function setUpPage()
      */
     public function tearDown()
     {
-        if ($this->_mysqli != null) {
+        if (!isset($this->_mysqli)) {
```
nijel: Shouldn't this be otherwise? We want to execute this quire when the connection has been established.

Achilles-96: Yes. That makes sense. I didn't think about that. I just tried to fix the error.In `parent::tearDown()` also, queries are ran only after checking connection is present. Should I change it?

nijel: The original code didn't work for you?

test/selenium/PmaSeleniumDbEventsTest.php
```diff
@@ -66,7 +68,7 @@ public function setUpPage()
      */
     public function tearDown()
     {
-        if ($this->_mysqli != null) {
+        if (isset($this->_mysqli)) {
```
Achilles-96: The old code was giving me error `Undefined property: PMA_SeleniumDbEventsTest::$_mysqli`

nijel: That's defined in PMA_SeleniumBase, there should be no reason for that. Also it's set in setUp method, so you should get failure there as well...

Achilles-96: Yes. Even I thought the same thing. I convinced myself that somehow `_mysqli` is being unset before `tearDown` is called. But now I see that event scheduler is not being turned off after the test completes. The problem may be somewhere else.

nijel:
Anyway, I've just merged first commit (53770cf934f32edb8a9b3ba85338e2e6a8c884dd) as it definitely makes sense, the changes in events testing certainly need some more tweaking.

Achilles-96:
It was giving undefined property because it was private. I changed it to protected. Now, it works fine. Any other issues?

nijel:
Good point, merging...

PR #13126, opened 2017-03-27
Fix: Focus back to codemirror editor on selecting column

### LINKED ISSUE #13125: Return focus to textarea after inserting column name
pma

poush:
Fixes #13125 


nijel (changes requested):
Thanks the fix looks correct for the CodeMirror case, however @OlafvdSpek was not using CodeMirror in the reported issue, can you please fix that as well? It's just the other branches are missing focus call as well.

nijel:
Landed as f89aafe9bfb194e4463bcc3469291a4101c5e539 to QA_4_7, thanks!

Issue #12437, opened 2016-08-03
CSP improvements


nijel:
Usage of not needed CSP values (result in weaker CSP policy)
 
from ReCaptcha Documentation: https://developers.google.com/recaptcha/docs/faq#im-using-content-security-policy-csp-on-my-website-how-can-i-configure-it-to-work-with-recaptcha
The needed directives are:
        
        script-src https://www.google.com/recaptcha/, https://www.gstatic.com/recaptcha/
        frame-src https://www.google.com/recaptcha/
        style-src 'unsafe-inline'
    
so the below are probably not needed ($captcha_url in style-src & img-src)
    
*    https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/Header.php#L553
*    https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/Header.php#L560
*    https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/Header.php#L572
*    https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/Header.php#L585
*    https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/Header.php#L590
    
and it's better to use frame-src (deprecated) or child-src

*        https://developer.mozilla.org/en-US/docs/Web/Security/CSP/CSP_policy_directives#child-src
*        https://developer.mozilla.org/en-US/docs/Web/Security/CSP/CSP_policy_directives#frame-src
        
instead of default-src for $captcha_url in:
    
*    https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/Header.php#L565
*    https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/Header.php#L577
*    https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/Header.php#L590
 
12) CSP Improvements
 
12.1)
add reflected-xss CSP directive to Content-Security-Policy header only (as it's available only from CSP Level 2)
    reflected-xss block
    
12.2)
add object-src CSP directive such as:
    object-src 'none'
in order to disallow plugin usage (embed/object/applet) tags which isn't used by PMA.
 
more information:

*    https://groups.google.com/a/chromium.org/forum/#!msg/security-dev/UqCSmNUHhNg/XBlvV_E5eowJ
    
12.3)
 
add form-action CSP directive such as:
    form-action 'self'
[...]

PR #13134, opened 2017-03-30
Add `object-src 'none'` to Content Security Policy header

ngyikp:
There’s no need to embed Flash or Java, so just turn off the ability to embed them for security.

Ref: https://github.com/phpmyadmin/phpmyadmin/issues/12437


nijel:

Issue #13147, opened 2017-04-02
Internal Server Error Interface 'Psr\\Container\\ContainerInterface' not found.


Fenn-CS:
### Steps to reproduce
I simply force update with the master branch(upstream) after which  I could not access the local instance anymore it shows a complete blank screen. (I was working on issue 13130)

### Expected behaviour
Show  PMA normally

### Actual behaviour
Shows blank screen and show the following in error log

[Sun Apr 02 19:06:19.551638 2017] [:error] [pid 6450] [client 127.0.0.1:55882] PHP Fatal error:  Interface 'Psr\\Container\\ContainerInterface' not found in /var/www/html/phpmyadmin/libraries/di/Container.php on line 17

### Server configuration
**Operating system**:Ubuntu 16

**Web server**:Apache2

**Database**:MySql

**PHP version**:PHP 7.0.15-0ubuntu0.16.04.4 (cli) ( NTS )

**phpMyAdmin version**:4.7-dev

### Client configuration
**Browser**:FireFox 52

ibennetch:
Even if you've already run Composer to install the library dependencies,
you'll have to occasionally update the library files as they are updated.

I regularly run `composer update` to accomplish this. You may also have
to clear your browser's cache.

Issue #13130, opened 2017-03-28
Improve AJAX error handling


nijel:
### Steps to reproduce

Originally brought by user on twitter who complained about empty error message on network timeout:

![c71nwfgxqaesvbg jpg large](https://cloud.githubusercontent.com/assets/212189/24396277/deda16d2-13a2-11e7-9148-ca77c07b63d5.jpeg)

See https://twitter.com/the_lux/status/845945943421784064

I've tried to reproduce and the AJAX "Loading" popup seems to be struck even when the network requests immediately fails, what is also not sane behavior. To reproduce

1. Open phpMyAdmin window.
2. Stop the web server.
3. Click on any link.

### Expected behaviour
Error message indicating network problems.

### Actual behaviour
The "Loading" popup is there forever.

### Server configuration
**phpMyAdmin version:** QA_4_7, master

Fenn-CS:
Hello @nijel let work through this issue!

Progyan1997:
I will be glad to contribute to the issue @nijel

nijel:
Feel free to look at it!

PR #13148, opened 2017-04-02
Fixed everlasting loading of ajax progress bar

nfebe:
Signed-off-by: Fenn-CS <fenn25.fn@gmail.com>
#13130.
Even though I am having problems running the current update #13147. I have tested this with an earlier build and it works just fine.

### DIFF
--- a/js/ajax.js
+++ b/js/ajax.js
@@ -793,7 +793,5 @@ $(document).ajaxError(function (event, request, settings) {
             false
             );
         AJAX.active = false;
-    } else if(request.status == 0) {
-        PMA_ajaxShowMessage('An unexpected error happened!');
     }
 });
--- a/js/ajax.js
+++ b/js/ajax.js
@@ -781,7 +781,7 @@ $(document).on('submit', 'form', AJAX.requestHandler);
  * (e.g: 500 - Internal server error)
  */
 $(document).ajaxError(function (event, request, settings) {
-    if (request.status !== 0) { // Don't handle aborted requests
+     if (!(request.status == 200 && request.status < 300)) {
         var errorCode = PMA_sprintf(PMA_messages.strErrorCode, request.status);
         var errorText = PMA_sprintf(PMA_messages.strErrorText, request.statusText);
         PMA_ajaxShowMessage(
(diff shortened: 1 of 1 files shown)


Fenn-CS:
Hello @ibennetch ,@nijel, @devenbansod  please someone should review this.

nijel:
Don't worry, we will review this, it's not even submitted for day, just be a bit more patient.

js/ajax.js
```diff
@@ -781,7 +781,7 @@ $(document).on('submit', 'form', AJAX.requestHandler);
  * (e.g: 500 - Internal server error)
  */
 $(document).ajaxError(function (event, request, settings) {
-    if (request.status !== 0) { // Don't handle aborted requests
+     if (!(request.status == 200 && request.status < 300)) {
```
nijel: This condition seems wrong - the second part is not needed as you've already checked the status == 200.

Also I don't think you should get status 200 without error. Maybe the request.status check is not needed at all? It does come from 16a060e527ed2494f03e1ffe82fe9c47cfab83d0, but there is no good reasoning in the commit message...

js/ajax.js
```diff
@@ -791,7 +791,9 @@ $(document).ajaxError(function (event, request, settings) {
             '<div>' + escapeHtml(errorText) + '</div>' +
             '</div>',
             false
-        );
+            );
         AJAX.active = false;
+    } else if(request.status == 0) {
+        PMA_ajaxShowMessage('An unexpected error happened!');
```
nijel: What is reason to show different error message in this case?

Fenn-CS:
> This condition seems wrong - the second part is not needed as you've already checked the status == 200.

When server response has been complete/done `(xhr.readyState === 4)`. It then check if the HTTP status is between 200 to 299 (`if (xhr.status == 200 && xhr.status < 300))`, if yes, it means the requested has been successful otherwise it failed `(if !(xhr.status == 200 && xhr.status < 300))`.

> Also I don't think you should get status 200 without error. Maybe the request.status check is not needed at all? It does come from 16a060e, but there is no good reasoning in the commit message.


 I am not checking for status 200 `if(request.status==200)` it the reverse i`f(!(request.status==200))` indicating there was some error.

> What is reason to show different error message in this case?


Well I think its quite unnecessary but status zero errors are special because they happen on the client side making it hard to determine the specific error.

nijel:
What I wanted to say is that following two conditions are equivalent:

```
if (!(request.status == 200 && request.status < 300)) {
```
```
if (!(request.status == 200)) {
```

There is really no need for second condition.

Also you won't get request.status == 200 in the error handler.

Anyway I've spent enough time on this to understand root cause of the problem and I've fixed this myself in https://github.com/phpmyadmin/phpmyadmin/commit/799ac0d4a0d1d6f2cb393c06e7d85e14dd89fa3a

Issue #13154, opened 2017-04-05
PHP Fatal error:  Uncaught Error: Call to a member function isAjax() on null


clevelas:
### Steps to reproduce
1. User uses the Change Password link after logging into phpMyAdmin
2. Type in the password and press Go

### Expected behaviour

Should change password and return successfully.

### Actual behaviour
The password is successfully changed, but an error message is presented:

Error in processing request
Error code: 500
Error text: Internal Server Error

Looking at the server error log:

PHP Fatal error:  Uncaught Error: Call to a member function isAjax() on null in /usr/share/phpMyAdmin/user_password.php:87\nStack trace:\n#0 /usr/share/phpMyAdmin/user_password.php(195): PMA_getChangePassMessage(Array, 'SET password = ...')\n#1 /usr/share/phpMyAdmin/user_password.php(55): PMA_changePassword('xxx', Object(PMA\\libraries\\Message), Array)\n#2 {main}\n  thrown in /usr/share/phpMyAdmin/user_password.php on line 87

It's also concerning the new password shows up in the error log.

### Server configuration

**Operating system**:

CentOS 7.3

**Web server:**
Apache 2.4 / PHP-FPM

**Database:**
MariaDB 10.1.21

**PHP version:**
7.0.17 (from remi-php70)

**phpMyAdmin version:**

phpMyAdmin-4.7.0-1 (from remi repo)

Fenn-CS:
@clevelas this is true! This is caused by the fact that the `$response` variable is known to the `PMA_getChangePassMessage() ` function to be **Local**.
https://github.com/phpmyadmin/phpmyadmin/blob/master/user_password.php#L87.
Let me fix that!.

jdwx:
The fix for this issue ( commit a661c06fd63fc7db30cf9c51154aed5a028dba3c ) appears to duplicate the declaration of $response in PMA_getChangePassMessage().

The second/old one, on line 92, should probably be removed.

nijel:
Thanks for pointing it out, I've just fixed that.

PR #13157, opened 2017-04-06
Fix: PHP Fatal error: Uncaught Error: Call to a member function isAjax() on null

nfebe:
Fix to #13154

### DIFF
--- a/user_password.php
+++ b/user_password.php
@@ -84,6 +84,7 @@ exit;
  */
 function PMA_getChangePassMessage($change_password_message, $sql_query = '')
 {
+    global $response;
     if ($response->isAjax()) {
         /**
          * If in an Ajax request, we don't need to show the rest of the page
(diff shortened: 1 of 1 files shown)


user_password.php
```diff
@@ -84,6 +84,7 @@
  */
 function PMA_getChangePassMessage($change_password_message, $sql_query = '')
 {
+    global $response;
```
nijel: It probably rather should be `$response = Response::getInstance();`

Fenn-CS: @nijel  did not want to do that since I found `$response = Response::getInstance();` earlier in the file.
https://github.com/phpmyadmin/phpmyadmin/blob/master/user_password.php#L22
But if that's what is better let me do it that way.

nijel:
Thanks, merged as a661c06fd63fc7db30cf9c51154aed5a028dba3c into QA_4_7 branch.

PR #13165, opened 2017-04-07
add item in Last-translator

atefBB:


nijel:
As per Gettext specification, this entry contains only the last translator, so the only one, not a list. By including newline you are also breaking the headers syntax.

You can also edit the translations at https://hosted.weblate.org/languages/ar/phpmyadmin/, they will be properly credited and propagate to all supported branches.

PR #13168, opened 2017-04-08
Ignore coding standard errors in transformations plugins classes

### LINKED ISSUE #13128: Transformations plugin classes doesn't follow PSR-1 class names rule
Hi,

The rule for naming classes according to the [PSR-1](http://www.php-fig.org/psr/psr-1/) is as follows:
> Class names MUST be declared in `StudlyCaps`.

Classes that are not following this rule are:

- All classes under `PMA\libraries\plugins\transformations\input` namespace
- All classes under `PMA\libraries\plugins\transformations\output` namespace
- `PMA\libraries\plugins\transformations\Text_Plain_Link` class
- `PMA\libraries\plugins\transformations\Text_Plain_Longtoipv4` class
- `PMA\libraries\plugins\transformations\Text_Plain_PreApPend` class
- `PMA\libraries\plugins\transformations\Text_Plain_Substring` class

If this rule cannot be followed for some reason, this must be made explicit with a comment in the code.

For example:
`// @codingStandardsIgnoreLine`

MauricioFauth:
Fix #13128

### DIFF
--- a/libraries/plugins/transformations/Text_Plain_Link.php
+++ b/libraries/plugins/transformations/Text_Plain_Link.php
@@ -16,6 +16,7 @@ use PMA\libraries\plugins\transformations\abs\TextLinkTransformationsPlugin;
  * @package    PhpMyAdmin-Transformations
  * @subpackage Link
  */
+// @codingStandardsIgnoreLine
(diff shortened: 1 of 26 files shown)


nijel:
Thanks, makes sense for now.

Issue #13120, opened 2017-03-23
Dropdown when editing enum default value


OlafvdSpek:
A text field is currently used, a dropdown might be better.

Fenn-CS:
Hello @OlafvdSpek please a few steps to check out (reproduce) the error.

ibennetch:
I believe this would be the edit/change table structure page, where you can set a default value for the row. 

The problem with doing a drop down there is that a user could edit the possible values and then the drop down for the defaults would not have the proper values.

It might be possible to have a JavaScript/jQuery function that re-populates the drop down when the possible values field is changed, but I'm not sure how I feel about that.

PR #13171, opened 2017-04-09
Fixes editing default value for ENUM #13120

Achilles-96:

### DIFF
--- a/js/tbl_structure.js
+++ b/js/tbl_structure.js
@@ -75,6 +75,44 @@ function checkFirst() {
         $("input[name=field_where]").val('after');
     }
 }
+
+function defaultEnumPopulate() {
+    $default_field = $('#field_0_4');
+    $default_field.empty();
+    vals = $('#field_0_3').val().split(',');
+    $default_field.append($('<option>', {
+        value: 'NULL',
+        text: 'NULL'
+    }));
+    $.each(vals, function(index, val){
+        $default_field.append($('<option>', {
+            value: val,
+            text: val
+        }));
+    });
+}
+
+function defaultEnumUnpopulate() {
+    $default_field = $('#field_0_4');
+    $default_field.empty();
+    $default_field.append($('<option>', {
+        value: 'NONE',
+        text: 'None'
+    }));
+    $default_field.append($('<option>', {
+        value: 'USER_DEFINED',
+        text: 'As defined:'
+    }));
+    $default_field.append($('<option>', {
+        value: 'NULL',
+        text: 'NULL'
+    }));
+    $default_field.append($('<option>', {
+        value: 'CURRENT_TIMESTAMP',
+        text: 'CURRENT_TIMESTAMP'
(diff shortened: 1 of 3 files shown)


js/tbl_structure.js
```diff
@@ -100,6 +138,25 @@ AJAX.registerOnload('tbl_structure.js', function () {
     /**
      *Ajax action for submitting the "Column Change" and "Add Column" form
      */
+    url = window.location.href;
+    last_param = url.split('&')[url.split('&').length - 1];
```
nijel: The URL separator can be different on the server and it would break this code. Also isn't there other way to get this data instead of parsing URL?

Achilles-96: I can't find any other way to get it as the page is loaded based on parameter. Any suggestions?

nijel: It's probably to better to store the value in HTML on the server and the load it here. Parsing request string this way is not really good approach. Besides the delimiter, the ordering can change as well....

nijel:
I've just tested this and while the selection is there, the value is not really  used:

  ```ALTER TABLE `test` ADD `en` ENUM('d','dsa','dsa','dsadsa') NOT NULL AFTER `date`;```

It seems to submit field_default_type instead of field_default_value parameter.

Also it seems to work only when editing first enum in the table, others get the usual dropdown...

js/tbl_structure.js
```diff
@@ -75,6 +75,44 @@ function checkFirst() {
...
+function defaultEnumUnpopulate() {
+    $default_field = $('#field_0_4');
+    $default_field.empty();
+    $default_field.append($('<option>', {
+        value: 'NONE',
+        text: 'None'
+    }));
+    $default_field.append($('<option>', {
+        value: 'USER_DEFINED',
+        text: 'As defined:'
```
nijel: This should use translated messages, not hardcoded English ones...

js/tbl_structure.js
```diff
@@ -100,6 +138,26 @@ AJAX.registerOnload('tbl_structure.js', function () {
     /**
      *Ajax action for submitting the "Column Change" and "Add Column" form
      */
+    if ($('#page_type').length != 0) {
+        page_type = $('#page_type').val();
+        if (page_type == 'change_column') {
+            if ($('#field_0_2').val() == 'ENUM') {
```
nijel: This makes it work only for ENUM being on first position when editing.

templates/columns_definitions/table_fields_definitions.phtml
```diff
@@ -127,5 +127,8 @@
                 ->render($content_row); ?>
         </tr>
     <?php endforeach; ?>
+    <tr>
+        <input type="hidden" id="page_type" value="change_column"/>
```
nijel: What is reason for this flag?

nijel (changes requested):

PR #13190, opened 2017-04-14
Fixed the invalid link which was suggested after an import timeout.

### LINKED ISSUE #13189: A 404 error occurred while importing the data script timeout
### Steps to reproduce
1. Use the browser to open "[http://127.0.0.1/phpMyAdmin-4.7.0/]()" and log in
2. Click the "Import" button on the right navigation
3. Select the script file to import by format, click "GO"
4. Wait for the import timeout to see the problem

### Expected behaviour
Click the "[resubmit the same file]()" link and will not continue importing.

### Actual behaviour
php

### Server configuration
**Operating system**:
>Window 8.1

**Web server:**
>Apache/2.4.23 (Win32) OpenSSL/1.0.2j mod_fcgid/2.3.9

**Database:**
>MySQL 5.5.53

**PHP version:**
>PHP 5.6.27 nts

**phpMyAdmin version:**
>phpMyAdmin-4.7.0-all-languages.zip

### Client configuration
**Browser:**
>Chrome 57.0.2987.133

**Operating system:**
>Windows 8.1

I simply analyzed the problem and found that the problem exists in `./import.php` and `./libraries/URL.php`, `./import.php` calls the `URL::getCommon()` method in `./libraries/URL.php` in the `./import.php` file.

Because `URL::getCommon()` This method returns an empty, so the resulting `./import.php` in `$err_url` variable error ( `./import.php` file 613 lines) in the connection string.

udan11:
Fixes #13189.

The faulty line here is [this](https://github.com/phpmyadmin/phpmyadmin/blob/QA_4_7/import.php#L613). `&timeout_passed=1&offset=Z` is being appended to one the following possible values of `$err_url`: `tbl_import.php?db=X&table=Y`, `db_import.php?db=X` or `server_import.php` is appended. This causes an issue when `$err_url` is simply `server_import.php`.

### DIFF
--- a/import.php
+++ b/import.php
@@ -610,7 +610,8 @@ if (! empty($id_bookmark) && $_REQUEST['action_bookmark'] == 2) {
 
 // Did we hit timeout? Tell it user.
 if ($timeout_passed) {
-    $importUrl = $err_url .= '&timeout_passed=1&offset=' . urlencode(
+    $divider = parse_url($url, PHP_URL_QUERY) ? '&' : '?';
+    $importUrl = $err_url .= $divider . 'timeout_passed=1&offset=' . urlencode(
         $GLOBALS['offset']
     );
     if (isset($local_import_file)) {


nijel:
I don't like this change, it will make all URLs end with ? what looks weird.

nijel:
How about constructing the URL from the parameters? I've tried to do that in https://github.com/phpmyadmin/phpmyadmin/pull/13364

nijel:
Merged #13364 instead.

PR #13192, opened 2017-04-15
Fix errors detected by PHP_CodeSniffer

### LINKED ISSUE #13181: Some classes and methods doesn't follow the coding standard naming rule
PHP_CodeSniffer has detected errors in some class names and method names.

Classes:
- `PMA\libraries\engines\Mrg_Myisam`
- `PMA\libraries\engines\Performance_Schema`
- `PMA\libraries\plugins\export\PMA_ExportPdf`

Methods:
- `PMA\libraries\PDF::Download`
- `PMA\libraries\PDF::SetAlias`
- `PMA\libraries\plugins\schema\pdf\Pdf::NbLines`
- `PMA\libraries\plugins\schema\pdf\Pdf::Row`
- `PMA\libraries\plugins\schema\pdf\Pdf::SetWidths`
- `PMA\libraries\Response::http_response_code`
- `PMA\libraries\Response::set_http_response_code`

Should these errors be fixed or be ignored?

MauricioFauth:
Fix #13181


libraries/PDF.php
```diff
@@ -136,7 +136,7 @@ public function Error($error_message = '')
      *
      * @return void
      */
-    public function Download($filename)
+    public function download($filename)
```
nijel: As this class bases on TCPDF, I'd probably keep their coding style here...

libraries/plugins/export/ExportPdf.php
```diff
@@ -12,7 +12,7 @@
 use PMA\libraries\properties\plugins\ExportPluginProperties;
 use PMA\libraries\properties\options\groups\OptionsPropertyMainGroup;
 use PMA\libraries\properties\options\groups\OptionsPropertyRootGroup;
-use PMA\libraries\plugins\export\PMA_ExportPdf;
+use PMA\libraries\plugins\export\PdfExport;
```
nijel: Maybe this class should be placed elsewhere? Having PdfExport end ExportPdf at same level is not really intuitive...

libraries/plugins/schema/pdf/Pdf.php
```diff
@@ -277,7 +277,7 @@ public function Footer()
      *
      * @return void
      */
-    public function SetWidths($w)
+    public function setWidths($w)
```
nijel: Same as with libraries/PDF.php - this extends TCPDF, so maybe it's better to stay consistent within the file?

nijel (commented):
Most of the changes make sense, though I'm not really convinced about changing TCPDF subclasses as just moves inconsistency to another level...

libraries/PDF.php
mauriciofauth: Coding style is quite inconsistent in TCPDF.
However, it's more common for method names to use the camelCase format. So I switched to that format.

libraries/plugins/export/ExportPdf.php
mauriciofauth: Didn't know very well what to do with this file. I also believe that this class should be placed elsewhere, but I do not know where.
Maybe `PMA\libraries\plugins\ExportPdf` or `PMA\libraries\Pdf\PdfExport`.

libraries/plugins/schema/pdf/Pdf.php
mauriciofauth: Same as with libraries/PDF.php. Coding style is quite inconsistent in TCPDF. That is why I prefer to follow the PMA standard. What do you think, @nijel?

libraries/plugins/export/ExportPdf.php
mauriciofauth: Maybe `PMA\libraries\plugins\helpers\Pdf`

nijel: What do you think about `PMA\libraries\plugins\export\helpers\Pdf`? It's helper for the export plugin in the end...

libraries/PDF.php
nijel: You're right, I thought that TCPDF starts methods always with upper case, but that really isn't true these days.

libraries/plugins/schema/pdf/Pdf.php
nijel: I agree with your changes in the end, I had false memory of what coding style TCPDF uses...

nijel:

PR #13195, opened 2017-04-16
Remove inline styles from file templates/columns_atributes.phtml #12262

Juvenv:


nijel:
Same change was done in https://github.com/phpmyadmin/phpmyadmin/pull/13202, I've merged that one as it did have proper Signed-Off-By.

PR #13199, opened 2017-04-16
Reorder a few keys in the composer.json file

lpalgarvio:
Reorder a few keys to make the composer.json file feel a bit more organized:
- repositories key goes above all the dependencies
- pair homepage with support key and move them to position after name, description and keywords
- pair authors with license key and move them to position just before technical stuff


nijel:

Issue #13194, opened 2017-04-16
Column attribute have a different size


Juvenv:
There is a special reason for the field "attributes" on below image has a different size of others:
![image](https://cloud.githubusercontent.com/assets/10359154/25068239/b9811fe4-2233-11e7-8f65-ced457401315.png)

### Steps to reproduce
1. Select an existing database
2. Create a new table

### Expected behaviour
The field "attributes" should be have the same width of others

### Actual behaviour
The field "attributes" have a width different of other inputs.

### Server configuration
**Operating system**:
Ubuntu 16.04

**Web server:**
PHP 7.0.9 Development Server

**Database:**
10.1.19-MariaDB 

**PHP version:**
PHP 7.0.9

**phpMyAdmin version:**
4.8.0-dev

### Client configuration
**Browser:**
Firefox
**Operating system:**
Ubuntu 16.04

nijel:
Fixed by https://github.com/phpmyadmin/phpmyadmin/pull/13202

PR #13202, opened 2017-04-17
Fix:Column attribute have a different size #13194

nfebe:
Signed-off-by: Fenn-CS <fenn25.fn@gmail.com>
#13194


Fenn-CS:
Also am wondering if it will be better to add an overflow-x property in a wrapper for that table so it will have its own horizontal scrollbar to avoid dragging the entire from left to right.

nijel:

Issue #13200, opened 2017-04-17
Boilerplate


lpalgarvio:
Hello

I've created a boilerplate for phpMyAdmin to deploy new projects.
In case you find it useful, you can link it in the project README or fork it.

github: https://github.com/solsoft/phpmyadmin-boilerplate
packagist: https://packagist.org/packages/solsoft/phpmyadmin-boilerplate

Regards
- Luís

udan11:
Hello Luis,

It seems indeed useful, however I have one question: how is this different from the method documented [here](https://github.com/phpmyadmin/phpmyadmin/blob/master/doc/setup.rst#installing-using-composer)?

lpalgarvio:
Hey Dan,

I hadn't discovered this documented method before.

The differences are as follows:

1. `create-project` command against the phpmyadmin repo downloads the source code for the project to the root of a new directory. The resulting code is the same, but is not updatable via composer, which defeats the purpose of using composer. Results of `composer show`:
`codacy/coverage                  
doctrine/instantiator            
gitonomy/gitlib                  
google/recaptcha                 
guzzle/guzzle                    
phpdocumentor/reflection-common  
phpdocumentor/reflection-docblock
phpdocumentor/type-resolver      
phpmyadmin/coding-standard       
phpmyadmin/motranslator          
phpmyadmin/shapefile             
phpmyadmin/sql-parser            
phpseclib/phpseclib              
phpspec/prophecy                 
phpunit/php-code-coverage        
phpunit/php-file-iterator        
phpunit/php-text-template        
phpunit/php-timer                
phpunit/php-token-stream         
phpunit/phpunit                  
phpunit/phpunit-mock-objects     
phpunit/phpunit-selenium         
psr/cache                        
psr/log                          
satooshi/php-coveralls           
sebastian/comparator             
sebastian/diff                   
sebastian/environment            
sebastian/exporter               
sebastian/global-state           
sebastian/recursion-context      
sebastian/version                
squizlabs/php_codesniffer        
symfony/cache                    
symfony/config                   
symfony/console                  
symfony/debug                    
symfony/event-dispatcher         
symfony/expression-language      
symfony/filesystem               
symfony/polyfill-mbstring        
symfony/process                  
symfony/stopwatch                
symfony/yaml                     
tecnickcom/tcpdf                 
webmozart/assert
`
Meaning, you can update the dependencies, but not phpmyadmin itself AFAIK.

[...]

PR #13212, opened 2017-04-23
Create an autoload.php in the root directory and tell libraries/common.inc.php to require it

lpalgarvio:
Hello

Currently an **autoloader** is implemented via `vendor/autoload.php`, which is then loaded into `libraries/common.inc.php`.

Although easy to change where this autoloader lives, this involves patching `common.inc.php`, making it a little less easier than it could be if the autoloader was instead loaded by an additional file, which could be easily replaced or edited.

# Why?
To support installing the phpmyadmin vendor dependencies inside the root project directory, that is, outside of phpmyadmin directory structure, being no longer visible to the world, and being updatable by an upper composer project - currently detailed [here](https://github.com/solsoft/phpmyadmin-boilerplate) and on issue #13200 .

# How?
The provided patch solves this by creating an `autoload.php` at the root of the directory project which then includes the `vendor/autoload.php` file.
The file `libraries/common.inc.php` will load this root autoload.php instead of the one inside the vendor directory.

### DIFF
--- a/autoload.php
+++ b/autoload.php
@@ -8,4 +8,4 @@
  * @see libraries/common.inc.php
  */
 
-require __DIR__ . '/vendor/autoload.php';
+require_once __DIR__ . '/vendor/autoload.php';
--- a/autoload.php
+++ b/autoload.php
@@ -8,4 +8,4 @@
  * @see libraries/common.inc.php
  */
 
-return require __DIR__ . '/vendor/autoload.php';
+require __DIR__ . '/vendor/autoload.php';
--- a/autoload.php
+++ b/autoload.php
@@ -1,0 +1,11 @@
+<?php
+
+/**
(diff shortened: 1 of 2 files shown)


lpalgarvio:
New release published
https://github.com/solsoft/phpmyadmin-boilerplate/releases/tag/v1.1.0

And a project created for the autoload / vendor scope
https://github.com/solsoft/phpmyadmin-boilerplate/projects/1

mauriciofauth:
I believe it would be better to move this information to the `libraries/vendor_config.php` file instead of creating a new file. What do you think?

lpalgarvio:
That could work as well :)

I'll prepare a new patch

lpalgarvio:
New pull request submitted to replace this pull request.

PR #13213, opened 2017-04-24
Implement AUTOLOAD_FILE constant

lpalgarvio:
This change implements AUTOLOAD_FILE constant to allow changing the autoload.php file being used by phpMyAdmin.

See obsolete pull request #13212

### DIFF
--- a/libraries/vendor_config.php
+++ b/libraries/vendor_config.php
@@ -14,6 +14,12 @@ if (! defined('PHPMYADMIN')) {
 }
 
 /**
+ * Path to vendor autoload file. Useful when you want to
+ * have have vendor dependencies somewhere else.
+ */
+define('AUTOLOAD_FILE', './../vendor/autoload.php');
+
+/**
  * Directory where cache files are stored.
  */
 define('CACHE_DIR', './cache/');
--- a/libraries/vendor_config.php
+++ b/libraries/vendor_config.php
@@ -17,7 +17,7 @@ if (! defined('PHPMYADMIN')) {
  * Path to vendor autoload file. Useful when you want to
  * have have vendor dependencies somewhere else.
  */
-define('AUTOLOAD_FILE', './../vendor/autoload.php');
+define('AUTOLOAD_FILE', './vendor/autoload.php');
 
 /**
  * Directory where cache files are stored.
--- a/scripts/advisor2po
+++ b/scripts/advisor2po
@@ -50,8 +50,9 @@ function print_message($idx) {
 }
 
 define('PHPMYADMIN', 1);
-require_once 'vendor/autoload.php';
-require_once 'libraries/core.lib.php';
+require_once './../libraries/vendor_config.php';
+require_once './../' . AUTOLOAD_FILE;
+require_once './../libraries/core.lib.php';
 
 $rules = PMA\libraries\Advisor::parseRulesFile();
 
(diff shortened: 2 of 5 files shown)


libraries/vendor_config.php
```diff
@@ -14,6 +14,12 @@
 }
 
 /**
+ * Path to vendor autoload file. Useful when you want to
+ * have have vendor dependencies somewhere else.
+ */
+define('AUTOLOAD_FILE', './../vendor/autoload.php');
```
nijel: Is the path change intentional? Originally it was `./vendor/autoload.php` and now `./../vendor/autoload.php`.

lpalgarvio: Sorry, a mistake when I was doing tests

nijel:
Ah, and the tests didn't fail because you don't use the new constant in tests bootstrap, please adjust it as well: https://github.com/phpmyadmin/phpmyadmin/blob/master/test/bootstrap-dist.php#L62

lpalgarvio:
I was testing with the vendor directory and it's autoload.php outside phpmyadmin root, hence the reason it didn't failed.

There was a few more places mentioning autoload.php and i've fixed those as well.

PS: This project's require/include logic might be needing a major refactoring.

scripts/advisor2po
```diff
@@ -50,8 +50,9 @@ function print_message($idx) {
 }
 
 define('PHPMYADMIN', 1);
-require_once 'vendor/autoload.php';
-require_once 'libraries/core.lib.php';
+require_once './../libraries/vendor_config.php';
+require_once './../' . AUTOLOAD_FILE;
+require_once './../libraries/core.lib.php';
```
nijel: This will break execution from root directory (see scripts/update-po).

scripts/generate-twig-cache
```diff
@@ -1,8 +1,8 @@
 <?php
 /* vim: set expandtab sw=4 ts=4 sts=4 ft=php: */
 define('PHPMYADMIN', 1);
-require_once 'vendor/autoload.php';
-require_once 'libraries/vendor_config.php';
+require_once './../libraries/vendor_config.php';
+require_once './../' . AUTOLOAD_FILE;
```
nijel: Same here, why do you change paths which work?

test/bootstrap-dist.php
```diff
@@ -58,9 +58,9 @@
     }
 }
 
-require_once 'libraries/vendor_config.php';
-require_once 'vendor/autoload.php';
-require_once 'libraries/core.lib.php';
+require_once './../libraries/vendor_config.php';
+require_once './../' . AUTOLOAD_FILE;
+require_once './../libraries/core.lib.php';
```
nijel: ...and here, see test failures on Travis.

scripts/generate-twig-cache
lpalgarvio: these looked as mistakes, as for example, in generate-twig-cache, these lines:
`$tplDir = dirname(__FILE__) . '/../templates';
$tmpDir = dirname(__FILE__) . '/../' . CACHE_DIR . 'twig';`
are assuming a different base path.
I also assumed that these scripts are being loaded directly via url or full path in the cli

nijel: These use dirname(__FILE__)  as current dir, not the one from where script is executed.

lpalgarvio: exactly, it's what I didn't gave enough attention.

I'm creating new commits

nijel:
Squashed and merged, thanks for your contribution!

lpalgarvio:
great, thanks =)

PR #13221, opened 2017-04-28
Don't destroy the session if logins are still present. fixes #12301

### LINKED ISSUE #12301: Multiple Tabs, Database connections (auth_type cookie) : Error Token Mismatch
I've found other topics on this problem, this problem exists since version 4 of pma at least, and has yet to be fixed. 

There are some fix propositions but I don't understand, because there are 3 different fixes, all modifying different files, I have no idea which fix to apply, if not all 3.  

That is why i'm reporting this problem again in the hopes of getting a clear answer.   Also to be noted I get this error also in version 4.4 as well. 
### Steps to reproduce
1.  Define mutliple sql servers in config.inc.php with auth_type as cookie
2.  Open up browser, navigate, to PMA, connect to one of the sql servers, open up a second tab, connect to a different sql server 
3. Try to perform any action on the first tab, you should get error : token mismatch
### Expected behaviour

Tell us what should happen
Open up mulitple tabs, on each tab connect to a different mysql server with the same user account, when trying to perform any action on either tabs, should be no problems
### Actual behaviour

Tell us what happens instead
Open up multiple tabs, on each tab connect to a different mysql server with the same user account, when trying to perform any action on the other tabs,  you get Error: Token Mismatch
### Server configuration

**Operating system**:
Debian Jessie

**Web server:**
Apache2.4.10

**Database:**
SQL

**PHP version:**
PHP 7.0

**phpMyAdmin version:**
PMA 4.6
### Client configuration

**Browser:**
Chrome

**Operating system:**
Windows 7

WanWizard:
If you are logged-in into multiple database servers, logOut() logs you out of all of them. This fix will prevent that, and will only destroy the session when you log out of the last server.

### DIFF
--- a/libraries/plugins/AuthenticationPlugin.php
+++ b/libraries/plugins/AuthenticationPlugin.php
@@ -74,9 +74,17 @@ abstract class AuthenticationPlugin
         $PHP_AUTH_USER = '';
         $PHP_AUTH_PW = '';
 
+        /* Get a logged-in server count */
+        $servers = 0;
+        foreach ($GLOBALS['cfg']['Servers'] as $key => $val) {
+            if (isset($_COOKIE['pmaAuth-' . $key])) {
+                $servers++;
+            }
+        }
+
         /* delete user's choices that were stored in session */
-        $_SESSION = array();
-        if (!defined('TESTSUITE')) {
+        if ($servers === 0 and ! defined('TESTSUITE')) {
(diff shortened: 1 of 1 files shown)


nijel:
Landed as https://github.com/phpmyadmin/phpmyadmin/commit/cde775bd2b7b3d56ba7cd608bad870feba75432f

Issue #13218, opened 2017-04-26
phpMyAdmin logo .png file incorrect transparency


druplance:
### Steps to reproduce
1. go to the homepage /phpmyadmin/
2. press F12 in Chrome or Firefox
3. select the div behind the logo in the top left of the screen: <div id="pma_navigation_header"> 
4. add element style to the div:  style="background-color: black;"

### Expected behaviour
All images should have smooth edges.

### Actual behaviour
All images do have smooth edges, EXCEPT the phpMyAdmin logo, which looks jagged as the alpha channel seems missing.

for example Chromium eye-saving plugins, like 'dark reader', invert the background to a dark color.

### Server configuration
**Operating system**:
Ubuntu 14.04

**Web server:**
Apache/2.4.7 (Ubuntu)
Database client version: libmysql - 5.5.54
PHP extension: mysqli

**Database:**
Server: Localhost via UNIX socket
Server type: MySQL
Server version: 5.5.54-0ubuntu0.14.04.1 - (Ubuntu)
Protocol version: 10
Server charset: UTF-8 Unicode (utf8)

**PHP version:**
PHP Version 5.5.9-1ubuntu4.21
System	Linux envy 3.13.0-117-generic #164-Ubuntu SMP Fri Apr 7 11:05:26 UTC 2017 x86_64

**phpMyAdmin version:**
Version information: 4.0.10deb1

### Client configuration
**Browser:**
Chromium
Version 53.0.2785.143 Built on Ubuntu , running on Ubuntu 14.04 (64-bit)

**Operating system:**

nijel:
This is caused by the fact that the logo is oldest image we have and it comes from time where Internet Explorer didn't support alpha layer on PNG images. But indeed it's time to fix that.

PR #13223, opened 2017-04-28
Update logo images

nijel:
The old ones did not have proper transparency (due to lack of support
for that in some browsers by the time they were created).

As we don't seem to have sources for the original logo, I've used
slightly newer version which was used on t-shirts and stickers.

Fixes https://github.com/phpmyadmin/phpmyadmin/issues/13218


udan11:
Looks good to me.

PR #13227, opened 2017-04-29
Run Twig in degraded mode if cache is not available.

### LINKED ISSUE #13225: Twig error on read only filesystem
If the cache directory is not writable, Twig fails and user ends up with blank page. This is certainly not good outcome.

Possible solution:

* Make Twig work in degraded mode without cache. It will be probably slow, but working.
* If that is not possible we should at least properly report such error.

Affects master, seen on demo server.

```
PHP message: PHP Fatal error:  Uncaught RuntimeException: Unable to create the cache directory (./cache/twig/80). in /home/demo/.pma-demo/master-config/vendor/twig/twig/lib/Twig/Cache/Filesystem.php:57
Stack trace:
#0 /home/demo/.pma-demo/master-config/vendor/twig/twig/lib/Twig/Environment.php(452): Twig_Cache_Filesystem->write('./cache/twig/80...', '<?php\n\n/* prefs...')
#1 /home/demo/.pma-demo/master-config/vendor/twig/twig/lib/Twig/Environment.php(397): Twig_Environment->loadTemplate('prefs_autoload....')
#2 /home/demo/.pma-demo/master-config/libraries/Template.php(175): Twig_Environment->load('prefs_autoload....')
#3 /home/demo/.pma-demo/master-config/libraries/user_preferences.lib.php(288): PMA\libraries\Template->render(Array)
#4 /home/demo/.pma-demo/master-config/libraries/Header.php(441): PMA_userprefsAutoloadGetHeader()
#5 /home/demo/.pma-demo/master-config/libraries/Response.php(260): PMA\libraries\Header->getDisplay()
#6 /home/demo/.pma-demo/master-config/libraries/Response.php(273): 
```

udan11:
Fixes #13225.

### DIFF
--- a/libraries/Template.php
+++ b/libraries/Template.php
@@ -58,9 +58,10 @@ class Template
         $this->helperFunctions = $helperFunctions;
 
         $loader = new Twig_Loader_Filesystem(static::BASE_PATH);
+        $cache_dir = CACHE_DIR . 'twig';
         $this->twig = new Twig_Environment($loader, array(
             'auto_reload' => true,
-            'cache' => CACHE_DIR . 'twig',
+            'cache' => is_writable($cache_dir) ? $cache_dir : false,
             'debug' => false,
         ));
         $this->twig->addExtension(new I18nExtension());
--- a/ChangeLog
+++ b/ChangeLog
@@ -11,6 +11,7 @@ phpMyAdmin - ChangeLog
 - issue #13119 No database selected error when adding foreign key
 - issue #12388 Improved database search to allow search for exact phrase match
 - issue #13099 Report error when trying to copy database to same name
+- issue #13225 Run Twig in degraded mode if cache is not available
 
 4.7.1 (not yet released)
 - issue #13132 Always execute tracking queries as controluser


libraries/Template.php
```diff
@@ -58,9 +58,10 @@ protected function __construct($name, $data = array(), $helperFunctions = array(
         $this->helperFunctions = $helperFunctions;
 
         $loader = new Twig_Loader_Filesystem(static::BASE_PATH);
+        $cache_dir = CACHE_DIR . 'twig';
         $this->twig = new Twig_Environment($loader, array(
             'auto_reload' => true,
-            'cache' => CACHE_DIR . 'twig',
+            'cache' => is_writable($cache_dir) ? $cache_dir : false,
```
nijel: This will fail if the directory does not exist (what it always true on first run). It's probably better to try to create the dir if it does not exist (Twig would try to create it).

nijel (commented):
I think it should also warn on the homepage about this in a similar way we warn for other configuration problems.

nijel:
I've taken a bit more complex approach on this, it would be great if you could review #13237.

nijel:
Merged #13237 which addresses this in wider context.

Issue #13231, opened 2017-05-02
Twig error


OlafvdSpek:
After git pull && composer update:

`Uncaught RuntimeException: Unable to create the cache directory (./cache/twig/dc). in /srv/pma/vendor/twig/twig/lib/Twig/Cache/Filesystem.php:57`

Is there a (new) requirement for a writable directory?

mauriciofauth:
Related with #13225

udan11:
Run:

```bash
mkdir -p cache/twig
chmod 770 cache/twig/
chgrp www-data cache/twig/
```

Error will be fixed with 8b38ddb089ce4b47e010131c5f72988231c2f5f8. Documentation will be updated to include relevant instructions for installation.

OlafvdSpek:
If you run chmod and chgrp on the cache dir itself you won't have a problem the next time another cache dir is needed. ;)

nijel:
Indeed duplicate of #13225 and #13226.

PR #13237, opened 2017-05-04
Share TempDir for Twig and other uses

### LINKED ISSUE #13226: Missing documentation for creating twig cache dir
As we started to use twig templates, the cache dir needs to be properly set up. This is not really covered by the documentation it should at least cover:

* secure permissions for the cache dir
* blocking access from browser to the cache
* moving the directory out of www root

nijel:
I think setting up just one folder for all the features is better approach than having separate folder for Twig cache and other temporary data.

Fixes #13226 and #13225 (also #13231).

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -2955,13 +2955,15 @@ the files.
 .. config:option:: $cfg['TempDir']
 
     :type: string
-    :default: ``''``
+    :default: ``'./tmp/'``
 
-    The name of the directory where temporary files can be stored.
+    The name of the directory where temporary files can be stored. It is used
+    for several purposes, currently:
 
-    This is needed for importing ESRI Shapefiles, see :ref:`faq6_30` and to
-    work around limitations of ``open_basedir`` for uploaded files, see
-    :ref:`faq1_11`.
+    * The templates cache which speeds up page loading.
+    * ESRI Shapefiles import, see :ref:`faq6_30`.
+    * To work around limitations of ``open_basedir`` for uploaded files, see
+      :ref:`faq1_11`.
 
     This directory should have as strict permissions as possible as the only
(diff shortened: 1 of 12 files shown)


doc/config.rst
```diff
@@ -2955,13 +2955,15 @@ the files.
...
+    :default: ``'./tmp/'``
 
-    The name of the directory where temporary files can be stored.
+    The name of the directory where temporary files can be stored. It is used
+    for several purposes, currently:
 
-    This is needed for importing ESRI Shapefiles, see :ref:`faq6_30` and to
-    work around limitations of ``open_basedir`` for uploaded files, see
-    :ref:`faq1_11`.
+    * The templates cache what speeds up page loading.
```
ibennetch: I'd use "which" instead of "what" here.

doc/config.rst
```diff
@@ -2955,13 +2955,15 @@ the files.
...
+    * ESRI Shapefiles import, see :ref:`faq6_30`.
+    * As a work around limitations of ``open_basedir`` for uploaded files, see
```
ibennetch: And either "To work around limitations" or "As a work around of limitations with ``open_basedir``". I prefer the second version, but either is fine.

ibennetch (commented):
This change looks good to me, although what are the security pitfalls if a user leaves the cache directory visible by the world (because most will)?

nijel:
Thanks for feedback, I will adjust the docs tomorrow.

As for risk, it's not higher than it was before - Twig cache files are php classes, they do nothing when executed directly. It's more a security precaution as it's web server writable dir...

ibennetch:
Then the warning/hardening statement you wrote here seems great.

I approve of this PR.

Issue #13228, opened 2017-04-29
Drop support for PHP 5.5


mauriciofauth:
As everyone knows, [PHP 5.5](http://php.net/eol.php) has reached its end of life nine months ago and is no longer supported.

We should increase the PHP requirement to version 5.6, which will be supported until December 2018.

lem9:
ISPs are slow to upgrade. Have a look at https://w3techs.com/technologies/details/pl-php/5/all. I admit that they are also slow to upgrade phpMyAdmin :)

mauriciofauth:
We could set a milestone to do that. For example, version 4.8, 4.9 or even 5.0.

ibennetch:
Is there a particular reason you're eager to drop support for these old PHP versions? Some feature set that is difficult to maintain or something?

mauriciofauth:
My interest is to upgrade PHPUnit from version 4.x, which is no longer supported, to version 5.x.

udan11:
@mauriciofauth Are there any notable advantages of using PHPUnit 5.x instead of 4.x?

mauriciofauth:
There are some benefits in version 5.x, but nothing that cannot be done in some way in version 4.x.

This change isn't something that makes me anxious. But I believe it's something that can be done for a milestone like 5.0.
My idea with that issue is just that. Mark this in some milestone, since it is something that will eventually have to be done.
I don't know what your thoughts are on this subject.

OlafvdSpek:
@lem9 And? 
If they insist on running unsupported versions of PHP I'm sure they don't mind running unsupported versions of PMA.
Even bumping the requirement to PHP 7.0 (for pMA 4.8) doesn't seem unreasonable.

lem9:
@OlafvdSpek My point is that bumping the minimum PHP version would block _some_ ISPs from updating their phpMyAdmin.

OlafvdSpek:
Not your problem. ;)

If software keeps supporting older versions of dependencies they have even less incentive to upgrade, while it might introduce 'cost' for your own development.

lem9:
@OlafvdSpek I am aware of all that, see my implication in the GoPHP5 initiative (https://www.linux.com/news/php-projects-join-forces-go-php-5). But there are pros and cons, and in this case I believe that the 'cost' of handling bug reports for older phpMyAdmin versions is greater.

OlafvdSpek:
What's the cost of handling those bug reports?
[...]

Issue #12790, opened 2016-12-07
Q: What are the reasons for supporting 4.0 longer than 4.4?


Rikk:
While 4.0.x is "Supported for security fixes only, until April 1, 2017", 4.4.x was "Supported for security fixes only, until October 1, 2016."

Servers at Ubuntu 12.04 are still at PHP 5.3 and the best version for them is 4.4.x.

If you support 4.0.x until 2017, and also Ubuntu 12.04 is supported until 2017, why do you stopped supporting 4.4.x already?

nijel:
The reason is to have supported version for older MySQL versions. MySQL 5.0 is still [supported upstream](https://www.mysql.com/support/supportedplatforms/database.html) and we want to provide supported version for that. On the other side, PHP 5.3 is [not supported upstream for more than two years](http://php.net/supported-versions.php), so it's really up to Ubuntu if they want to stick with that version.

We really do not have manpower to support everything what any distribution has released.

Rikk:
If you support 4.0 (PHP 5.2 and MySQL 5.0), couldn't the same fixes easily be applied to 4.4 (PHP 5.3.7 and MySQL 5.5)?

Do you have usage statistics for these versions?
What features would be broken if using a newer version with old PHP?

>support everything what any distribution has released
Well, Ubuntu is the most used one.

nijel:
> If you support 4.0 (PHP 5.2 and MySQL 5.0), couldn't the same fixes easily be applied to 4.4 (PHP 5.3.7 and MySQL 5.5)?

Not really, there are two years of development in between. For about half of the recent patches it was non trivial work to make them work on all three versions.

> Do you have usage statistics for these versions?

No, all we collect are error reports, but that feature was not present in 4.0, so we really can not compare.

> What features would be broken if using a newer version with old PHP?

I have no clue. PHP is evolving language adding parameters to functions during the development cycle and it's not really easy to track. Anyway I've tried to add 5.4 and 5.3 to our tests in https://github.com/phpmyadmin/phpmyadmin/pull/12795 and you can see that it's broken quite a lot as PHP 5.5 introduced finally section for exceptions and we're using that (there are probably more such things).

AgoraSecurity:
Hello,

Is there a page, wiki entry or something with the supported versions and the date until they will be supported? Ideally support of new features & security updates (sometimes those dates can be quite different)

Thanks!

AgoraSecurity:
Thanks for the answer @nijel 
[...]

PR #13242, opened 2017-05-05
Drop support for PHP 5.5

nijel:
This is mostly to continue discussion started in #13228. Please share your feedback!


MauricioFauth (changes requested):
You forgot to change the PHP requirement in `composer.json` from `"php": ">=5.5.0",` to `"php": "^5.6 || ^7.0",`.

nijel:
Thanks! I've stayed with existing syntax and used `"php": ">=5.6.0"`.

Dummann:
I ended up here searching why PHP 5.4 was no longer supported in recent builds, and now I'm reading that support for 5.5 is getting dropped.

Seems to me that you are forgetting that in a lot of serious hosting environments (with this I mean non-hobby) PHP 5.4x is still used. Hell, it's even the standard version for RedHat and Centos 7, and security fixes will get backported until 2024.

Simply throwing all those users an error like 'PHP 5.5+ is required' (while it runs fine on lower PHP version as well) seems lazy to me.

mauriciofauth:
@Dummann You can still use phpMyAdmin 4.0.x if you use older versions of PHP.

The fact that CentOS 7 will receive support until 2024 doesn't mean that PHP 5.4 will receive support until 2024. It's not because CentOS ships with a 2013 version of PHP that has known bugs, that PMA also need to support this version.

As I said before, you can continue to use PMA 4.0.x if you use an older version of PHP, since this branch still receives security updates.

mauriciofauth:
You can check more discussions on this subject in issue #13228.

Dummann:
It does in fact mean that PHP 5.4 will receive security support for many, many years to come, you are underestimating this. This is the sole reason why most companies that need stability are going with either RedHat or Centos.

Even version 5.3 (from 2009?) is still being backported: https://access.redhat.com/security/updates/backporting

One could say that most RedHat/Centos users will prefer commandline over phpmyadmin, which is probably true, but still; I don't think certain distro users with default installations should be excluded from using the most recent version of phpmyadmin - especially when phpmyadmin doesn't rely on any new functionality introduced in those newer PHP versions.

Linux === freedom. The least you could do is make sure phpmyadmin works with all default installations on all popular distro's. I seriously don't get why you should drop support for PHP versions that are known to work perfectly fine with phpmyadmin. Not all phpmyadmin users are students/Ubuntu users.

mauriciofauth:
It's strange that you want to have the latest version of a software while you want the maximum stability of other software to the point that you don't mind having old versions.
As I said before, and I don't mind repeating myself, if you want a stable version that supports your system, you can use PMA 4.0.x.

Please take into consideration that everything I said is just my opinion, I have no authority to speak on behalf of phpMyAdmin team.

nijel:
Sorry, but we really can't keep compatibility with all PHP versions. It's evolving language and PHP 5.5 brought few important things, the most important for us is probably [finally block for exceptions](http://php.net/manual/en/migration55.new-features.php#migration55.new-features.finally), but there might be more things. See also https://github.com/phpmyadmin/phpmyadmin/issues/12790#issuecomment-265680575 for discussion on similar topic.

Note that we're not really exception in the PHP ecosystem, random examples I've found:

* [Drupal needs PHP greater or equal PHP Version 5.5.9](https://www.drupal.org/docs/8/system-requirements/web-server)
* [Joomla needs 5.6 + or 7 +](https://docs.joomla.org/J3.x:Installing_Joomla)
* [Moodle needs minimum PHP 5.6.5](https://docs.moodle.org/dev/Moodle_3.2_release_notes#Server_requirements)
* [MediaWiki needs PHP version 5.5.9 or later](https://www.mediawiki.org/wiki/Manual:Installation_requirements#PHP)

So I'd say we're quite on par with current requirement of 5.5.

Also if this pull request would be merged, it would change requirements to 5.6 for future 4.8.0 release, which will certainly not happen earlier than in Q4, but more likely early in 2018.

Until April we did also provide security support for 4.0 exactly to support older PHP versions. In case we will raise PHP requirements in 4.8, the 4.7 will also get extended period of security support.

JorisDebonnet:
Perhaps this could be an opportunity to introduce [Semantic Versioning](http://semver.org/), which would mean dropping php 5.5 support should require pma's major version to bump to 5.

Dummann:
I don't really get the comparison with Joomla and other projects. PhpMyAdmin is a database tool, which should be all about reliability - it's not some fancy platform that needs the latest functionalities to satisfy the masses.

Some sysadmins suggested to start a private fork, which seems the wrong direction to me. Come on, why would we drop support for PHP versions without proper reason? The newest version of phpMyAdmin also works on older PHP versions, so there is no reason to throw 'You need PHP 5.X or newer' right away. If you want me to test upcoming changes in older PHP versions - I'll do it - if that means older version will be supported and effort will be made to keep supporting them.

nijel:
Okay, this is probably too early to do this.

@Dummann phpMyAdmin does not work properly with older PHP than 5.5, there are several places which depend on PHP 5.5 syntax, see https://github.com/phpmyadmin/phpmyadmin/pull/13363 (I've done it again as build logs for #12795 seem to be lost) where I've tried what is broken.

PR #13248, opened 2017-05-06
Fix unwanted escaping

CommanderRoot:
Fixes unwanted double escaping as "tbl_url_query" is already escaped before going into the Twig templates.


nijel:

PR #13249, opened 2017-05-06
Fix dropdown selector

CommanderRoot:

The previous syntax escaped the " of the class part which resulted in html code like this `class=&quot;autosubmit&quot;`
I also added a check for the id part so we don't have things like `id=""`


nijel:

PR #13258, opened 2017-05-09
Correct the check for Amazon RDS

duncan3dc:
RDS now uses version specific paths (eg `mysql-5.7.18` instead of just `mysql`), so just match on `rdsdbbin` instead.

Related to #1205


nijel:
Thanks, I've merged your commit as ac3a35bc973a20d87289f262f94bc3093fe26b48 to 4.7 maintenance branch. I've also added tests for the code in 61b9418f67c55789ab9a7a2ff7fe01d4af5c2578.

PR #13260, opened 2017-05-09
Replace eval() in Advisor by Symfony\ExpressionLanguage

nijel:


libraries/Advisor.php
```diff
@@ -556,4 +566,66 @@ public static function parseRulesFile()
...
+
+    /**
+     * Formats interval like 10 per hour
+     *
+     * @param integer $num       number to format
+     * @param integer $precision required precision
+     *
+     * @return string formatted string
+     */
+    public static function bytime($num, $precision)
```
udan11: This is just a nitpick: should this be `byTime`?

nijel: Probably yes, but as the API for advisor rules is defined in lowercase, I've kept it consistent with the public API here. We might as well change the API as this was not really used outside phpMyAdmin contrary to original intentions...

Issue #13173, opened 2017-04-10
Update old JS libraries


udan11:
Many of the jQuery Migrate warnings are caused by old libraries that have not been updated in a long time. We should handle these somehow.

* [x] jqplot
https://github.com/jqPlot/jqPlot

* [x] jQuery Timepicker Addon
https://github.com/trentrichardson/jQuery-Timepicker-Addon

* [x] jQuery hashchange event
https://github.com/cowboy/jquery-hashchange

* [x] jquery.event.drag
https://github.com/threedubmedia/jquery.threedubmedia

* [x] jquery.mousewheel
https://github.com/jquery/jquery-mousewheel

* [x] jquery.tablesorter
https://github.com/christianbach/tablesorter

* [x] OpenLayers
https://github.com/openlayers/openlayers

udan11:
Related to #13166.

nijel:
We really should automate this, also see https://github.com/phpmyadmin/phpmyadmin/issues/12677

Fenn-CS:
What if for now we write a small script( better in php than js) to check for updates based on version numbers for popular dependences then download and replace them accurately.
However I fear for compatibility problems--what if a js update for example requires changing some things in other dependent code?

udan11:
@Fenn-CS We do not have that many dependencies. IMO, creating a script that does what we could manually and better do is a waste of time.

OMENSAH:
@mauriciofauth , Most of the issues of JavaScript have been closed. Is it because of the GSoC?

mauriciofauth:
@OMENSAH No.
For example, issue #14103 was closed because it was considered a duplicate of this issue.
You can fix some warnings generated by jQuery Migrate if you want.

WueXiang:
Please look into this issue seriously, I have the latest version not working and the console is showing these:

jQuery.expr[':'] is deprecated; use jQuery.expr.pseudos
jQuery.expr.filters is deprecated; use jQuery.expr.pseudos
jQuery.fn.click() event shorthand is deprecated

![screenshot 108](https://user-images.githubusercontent.com/18504731/39628601-2351857c-4fdc-11e8-9fb0-b3606afbb2d3.png)
only the icon is visible hence nowhere to get into the interface.

Zerefdev:
Same issue here
![pma](https://user-images.githubusercontent.com/25670700/40574070-295ccdc2-60cc-11e8-9612-1adb2e937682.JPG)

[...]

PR #13281, opened 2017-05-17
Add a JavaScript package manager

### LINKED ISSUE #12677: Consider using package manager for javascript libraries
We already use [Composer](https://getcomposer.org/) for managing PHP dependencies. Having something like this for javascript would be useful as well to avoid need for manual library updates, however such approach first need to be investigated and checked for feasibility.

Things to check:

* Can we cover all javascript libraries we use?
   * if we can't find most of the libs packaged, it's probably not worth of the effort
* What package manager to use? Some choices (
   * [Bower](https://bower.io/) - Node based package manager
   * [BowerPHP](https://bowerphp.org/) - Composer ready reimplementation of above
   * use javascript packages from components project (eg. [jQuery](https://packagist.org/packages/components/jquery) or [jQuery-cookie](https://packagist.org/packages/components/jquery-cookie))

MauricioFauth:
I created a `package.json` file using the [Yarn](https://yarnpkg.com/) package manager.

I chose Yarn because it's [compatible](https://yarnpkg.com/en/docs/migrating-from-npm) with [npm](https://www.npmjs.com/), which is the most popular package manager, and also because it's the manager [recommended by the Bower team](https://bower.io/blog/2016/using-bower-with-yarn/). In other words, you can use npm to install the dependencies if you want.

I don't know if the way I copied the files after they were installed is the most appropriate one. I chose this way because I thought it would be easier to later remove the `node_modules` directory when creating a release.

I doesn't included all the scripts, as most of them are scripts that seem abandoned by their maintainers and some are not in the package manager registries.

Fix #12677 


mauriciofauth:
Related to #13173.

mauriciofauth:
I think that in the future we should use a tool like [Browserify](http://browserify.org/) or [webpack](https://webpack.js.org/).

nijel:
Not that I'd be experienced in this area, but I have some generic questions:

* Couldn't it be done without postinstall? If not, will it work on Windows?
* Should be yarn.lock part of the repository?
* Does it make sense to use yarn now and switch to Browserify or webpack later?
* What about OpenLayers, those doesn't seem to be handled here?

mauriciofauth:
Thanks for the feedback, @nijel 

> Couldn't it be done without postinstall? If not, will it work on Windows?

Maybe there are other ways, I'm not the most experienced at it either.
npm's postinstall is common and probably works on Windows. But the shell script I don't know if works, because I don't have a Windows system to test on. Maybe it's a good time to start using a CI tool for Windows, like the [AppVeyor](https://www.appveyor.com/).

> Should be yarn.lock part of the repository?

According to the [Yarn documentation](https://yarnpkg.com/lang/en/docs/yarn-lock/#toc-check-into-source-control), the lock file should be included in version control.

> Does it make sense to use yarn now and switch to Browserify or webpack later?

Yes, because they are not the same thing. Webpack and Browserify are module loaders/bundlers and not package managers. They serve to concatenate js files in a smart way to use them in browsers. This helps to modularize the scripts.

> What about OpenLayers, those doesn't seem to be handled here?

We use OpenLayers 2.13.1 and this version is not present in the registry, only 3.0+, which causes some conflicts. So I decided to do the pull request without OpenLayers to save the work of updating the library, because I think we can do this gradually. Just like the other scripts that are missing.

nijel:
Thanks for clarifications! AppVeyor will certainly not answer this question as their environment has quite a lot of tools installed which are not there on normal Windows systems. Thinking about that it's probably not that big problem as Git does include shell anyway and most people wanting to run this will get phpMyAdmin from Git anyway...

mauriciofauth:
An alternative would be to include the files directly from `node_modules`, but I don't think that's a good idea.

I thought about using webpack this way:
We could require dependencies directly from `node_modules` and then bundle them with webpack.
That's a lot of work, but the result would be great. I think this would involve reviewing all js files and where they are used, since the files that would be included in the HTML would be those bundled.

mauriciofauth:
I think for now it would be better just to add the package manager and update all the libraries, what do you think?

nijel:
How about having the package manager to update the libraries, but store the in git (to include js/vendor in your patch)?

This way we would not enforce having yarn for anybody who downloads git (I'm quite worried about portability of the postinst script), but still would have automated way to update (most of) javascript libraries?

mauriciofauth:
This is a good idea, @nijel. So we can test the package manager without causing so much trouble.
I'll do it as soon as I have some free time.

mauriciofauth:
I had to revert jqPlot and jquery.event.drag because they were giving errors related to module.
In other words, for these cases, it will be necessary to use the webpack, which is something that I need to study a little more.

nijel:
I just realized there is additional topic to address - distributing source code. We currently do that in js/jquery/src/, but that's just for jQuery (which you can remove in the PR as it no longer matches shipped jQuery anyway). The good thing is that the yarn packages include source code, but the legal question is whether including link to download source code is good enough :-). I will raise this to Software Freedom Conservancy...

mauriciofauth:
I just removed the `js/jquery/src` directory, @nijel.

mauriciofauth:
I started to think that this isn't a good patch to merge, because this script isn't a good way to copy the files. I'm studying how to use webpack and it seems promising.

mauriciofauth:
I will close this PR without merging it, because I think it is better to use the webpack directly, and in the meantime I will prepare the code base with updates and necessary changes.

nijel:
Okay, thanks for investigating this!

PR #13284, opened 2017-05-17
Small refactoring in preparation to CSP.

### LINKED ISSUE #13269: setTimeout first argument is string instead of function (depend on unneeded unsafe-eval)
in:
https://github.com/phpmyadmin/phpmyadmin/blob/61f294062a51be76b3b56f8cca3bc50feae6ad11/templates/header_location.twig#L11
`setTimeout("window.location = decodeURI('{{ Sanitize_escapeJsString(uri) }}')", 2000);`
setTimeout first argument is string, which won't work if CSP script-src directive don't contain unsafe-eval.

more information:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#Unsafe_eval_expressions

fix:
	use function constructor instead of string literal.

udan11:
Fixes #13269.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -14,6 +14,7 @@ phpMyAdmin - ChangeLog
 - issue #13167 Themes now have to contain metadata in theme.json
 - issue #6363  phpMyAdmin no longer requires eval() in PHP
 - issue #12386 The mbstring dependency is now optional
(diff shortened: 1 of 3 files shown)


nijel:
Can you please adjust tests to changed code?

nijel:

PR #13291, opened 2017-05-22
Removed inline onsubmit

### LINKED ISSUE #12261: Remove inline javascript

manishbisht:
Fixes #12261


templates/table/structure/add_column.phtml
```diff
@@ -1,4 +1,4 @@
-<form method="post" action="tbl_addfield.php" id="addColumns" name="addColumns" onsubmit="return checkFormElementInRange(this, 'num_fields', '<?= str_replace('\'', '\\\'', __('You have to add at least one column.')); ?>', 1)">
+<form method="post" action="tbl_addfield.php" id="addColumns" name="addColumns">
```
nijel: Are you sure the removal without replacement is okay here?

manishbisht: @nijel Yes, I have checked the function `checkFormElementInRange`
https://github.com/phpmyadmin/phpmyadmin/blob/master/js/functions.js#L800-L832
It only check that the entered content is number and it should be greater than 0. But using `min=0` and `type=number` both the things are checked automatically.

nijel:

PR #13294, opened 2017-05-23
Removed inline onchange

### LINKED ISSUE #12261: Remove inline javascript

manishbisht:
Fixes #12261

### DIFF
--- a/js/tbl_select.js
+++ b/js/tbl_select.js
@@ -81,6 +81,13 @@ AJAX.registerOnload('tbl_select.js', function () {
             return false;
         });
 
+    var tableRows = $('#fieldset_table_qbe select');
+    $.each(tableRows, function(index, item){
+        $(item).on("change", function() {
+            changeValueFieldType(this, index);
+        });
+    });
+
     /**
      * Ajax event handler for Table search
      */
--- a/js/tbl_structure.js
+++ b/js/tbl_structure.js
@@ -503,6 +503,10 @@ AJAX.registerOnload('tbl_structure.js', function () {
             $.post(url, params, AJAX.responseHandler);
         });
     });
+
+    $(document).on('change', 'select[name=after_field]', function () {
+        checkFirst();
+    });
 });
 
 /** Handler for "More" dropdown in structure table rows */
--- a/templates/table/search/column_comparison_operators.twig
+++ b/templates/table/search/column_comparison_operators.twig
@@ -1,4 +1,3 @@
-<select name="criteriaColumnOperators[{{ search_index }}]"
-    onchange="changeValueFieldType(this, {{ search_index }})">
+<select id="ColumnOperator{{ search_index }}" name="criteriaColumnOperators[{{ search_index }}]">
     {{ type_operators|raw }}
 </select>
--- a/templates/table/structure/add_column.twig
+++ b/templates/table/structure/add_column.twig
@@ -9,7 +9,7 @@
     {{ 'Add %s column(s)'|trans|format(num_fields)|raw }}
     <input type="hidden" name="field_where" value="after"/>&nbsp;
     {# I tried displaying the drop-down inside the label but with Firefox the drop-down was blinking #}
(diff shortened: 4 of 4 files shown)


manishbisht:
@nijel Can you please restart the build. I think there is some timeout issue.

templates/table/search/column_comparison_operators.phtml
```diff
@@ -1,7 +1,11 @@
...
         $columnNullFlags[$column_index], $selected_operator
     ); ?>
-</select>
\ No newline at end of file
+</select>
+<script type="text/javascript">
+    $(document).on('change', '#ColumnOperator<?= $search_index; ?>', function () {
+        changeValueFieldType(this, <?= $search_index; ?>);
+    });
+</script>
```
nijel: This really doesn't help for CSP - both need unsafe-inline. Can you please move that to one of js files?

manishbisht:
@nijel I have updated the PR.

manishbisht:
@nijel @ibennetch Can we merge this PR ?

ibennetch:
I'd like to wait for Michal's comment since he's more familiar with the CSP issue than I am.

js/tbl_select.js
```diff
@@ -81,6 +81,13 @@ AJAX.registerOnload('tbl_select.js', function () {
             return false;
         });
 
+    var tableRows = $('.data tbody tr');
```
nijel: This selector looks too generic. Also it's probably better to use `$('...').each(...)` instead.

templates/table/search/column_comparison_operators.phtml
```diff
@@ -1,5 +1,4 @@
-<select name="criteriaColumnOperators[<?= $search_index; ?>]"
-        onchange="changeValueFieldType(this, <?= $search_index; ?>)">
+<select id="ColumnOperator<?= $search_index; ?>" name="criteriaColumnOperators[<?= $search_index; ?>]">
```
nijel: Unfortunately I've merged migration of those templates to Twig meanwhile, can you please update your PR?

nijel (changes requested):

manishbisht:
@nijel I have update the PR

js/tbl_select.js
```diff
@@ -81,6 +81,13 @@ AJAX.registerOnload('tbl_select.js', function () {
...
+    $(tableRows).each(function(index, item){
```
nijel: You don't need $() around tableRows here (it's already constructed as jQuery object list).

js/tbl_select.js
nijel: Have you checked this generic selector is actually good?

manishbisht:
@nijel Updated the PR with the suggested changes.

js/tbl_select.js
manishbisht: Yes, It will select the data which have `.data` class then the `tbody` tag and at the last all the `tr` tags inside it. And make an object of the result.

nijel: But the data class is used in more situations, won't it match too many?

Also shouldn't you rather target on select instead of tr? The tr certainly doesn't have onchange...

manishbisht:
@nijel Yes, You were right `select`should be used instead of `tr` I have updated the PR.

js/tbl_select.js
```diff
@@ -81,6 +81,13 @@ AJAX.registerOnload('tbl_select.js', function () {
             return false;
         });
 
+    var tableRows = $('select');
```
nijel: You can't just match all selects on the page. Probably adding class to the one you want to match is needed. Have you tested the change that it doesn't affect something else?

manishbisht:
@nijel I have added extra id selector `$('#fieldset_table_qbe select');` to get results more accurately. Yes I have tested the change by putting the `console.log()` statements before the function call and also verified that the length of `tableRows` should be equal to the number of columns in the table.

nijel (approved):

nijel:

PR #13295, opened 2017-05-24
token on export

fullbl:
added hidden input with pma token when exporting custom data


nijel:
This really shouldn't be needed since 0f1c011c289.

Issue #13187, opened 2017-04-13
#1066 - Not unique table/alias error only export


davidbouche:
I'm actually having an issue with PHPMyAdmin :

- I can execute my SQL query with nice result
- I get an error when I'm trying to export the same query

Here's the query :

```sql
SELECT pn.nomenclature_id as nomenclature_id, pn.lettre as lettre, a.reference as reference_article, at.nom as nom_article
FROM item a, item_translation at, item_nomenclature an, item_nomenclature pn, item p 
WHERE a.type = 1
AND a.id = at.id
AND at.lang = 'fr'
AND a.id = an.item_id
AND an.nomenclature_id = pn.nomenclature_id
AND pn.item_id = p.id
AND p.type = 2
AND p.marque_id = 2
```
Here's a video where I had the problem : https://youtu.be/Z5AAZhoX6W0

There's lot of thread on Not unique table/alias error but I didn't found any reason to explain why the query works in PHPMyAdmin "SQL" tab but not with export.

Thanks for your support, David.

### Server configuration
**Operating system**:
5.7.17-0ubuntu0.16.04.2-log - (Ubuntu)
**Web server:**
Apache/2.4.18 (Ubuntu)
**Database:**
libmysql - mysqlnd 5.0.12-dev - 20150407
**PHP version:**
7.0.15-0ubuntu0.16.04.4
**phpMyAdmin version:**
Version : 4.5.4.1deb2ubuntu2

### Client configuration
**Browser:**
Chromium Version 56.0.2924.76 Built on Ubuntu , running on Ubuntu 16.10 (64-bit)
**Operating system:**
Ubuntu 16.10 yakkety

nijel:
There are certainly some problems with exporting queries with aliases, see https://github.com/phpmyadmin/phpmyadmin/issues/13222. I'm not sure if the root cause is same...

dm79219:
Thanks @akashkumardce .  I was unable to use export due to this error. Was struck for a long time on this hampering my work. Your fix worked perfectly for me.

williamdes:
Duplicate of #12241

williamdes:
Will be fixed as #12241

PR #13296, opened 2017-05-24
Export Error while join on same table

akashkumardce:
Fix for bug #13187 reported-
Whenever we execute a query like this 
`select A.id,B.id from table A,table B where A.id=B.id;`

Query gets executed correctly, but on clicking EXPORT button(in Query Result Operation Widget) and then clicking on GO, gives us file with error """"1066 - Not unique table/alias: 'table'""""
**Actual Flow-**
1- Query get Parsed
2- Alias replaced by table name for every column.  
       _Example_- `A.id,B.id` =changes to=> `table.id,table.id` - _Cant identify Which column from which table_
3- Alias gets removed for table name. 
       _Example_ - `from table A,table B` =Changes to=> `from table, table` - _Cant differentiate between table_ 

This flow got failed for Queries have joins on same table, and working fine for rest cases.
**Conclusion and Changes Made-**
Step 2 and 3 were not required and hence removed those steps.

_**Now export is working fine in all the tested cases.**_

### DIFF
--- a/tbl_export.php
+++ b/tbl_export.php
@@ -57,8 +57,7 @@ if (! empty($sql_query)) {
         $aliases = array();
         foreach ($parser->statements[0]->from as $from) {
             if ((!empty($from->table)) && (!empty($from->alias))) {
-                $aliases[$from->alias] = $from->table;
-                
+                $aliases[$from->alias] = $from->table;         
                 $from->expr = null; // Force rebuild.
             }
         }
--- a/tbl_export.php
+++ b/tbl_export.php
@@ -58,9 +58,7 @@ if (! empty($sql_query)) {
         foreach ($parser->statements[0]->from as $from) {
             if ((!empty($from->table)) && (!empty($from->alias))) {
                 $aliases[$from->alias] = $from->table;
-                // We remove the alias of the table because they are going to
-                // be replaced anyway.
-                $from->alias = null;
+                
                 $from->expr = null; // Force rebuild.
             }
         }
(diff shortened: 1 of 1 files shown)


nijel:
@udan11 can you please review this? It's removing code you've written :-).

udan11:
I cannot remember exactly why I did this, but it was something about aliases conflicting.

Let's take into consideration the following schema and query:

```sql
CREATE TABLE tbl (
  `id` int(11)
  `name` varchar(30)
)
```

```sql
SELECT A.id, B.id
FROM tbl A, tbl B
WHERE A.id = B.id;
```

This will generate the following table:
```
| id | id |
|  1 |  1 |
|  2 |  2 |
|  3 |  3 |
| .. | .. |
```

Click on export button, will generate a file that will contain two queries like:

```sql
CREATE TABLE `A` (
  `id` int(11) DEFAULT NULL,
  `name` varchar(30) DEFAULT NULL

INSERT INTO `A` (`id`, `id`) VALUES
(1, 1),
(2, 2),
(3, 3),
(4, 4),
(5, 5),
(6, 6),
(7, 7),
(8, 8),
(9, 9),
(10, 10);
```
Both of those queries are invalid.

I think that we should rethink some aspects about regarding to the export mechanism.

madhuracj:
@udan11 We do not allow the SQL format for queries joining 2 or more tables. However, this query offers exporting to SQL format as it does a self-join. Maybe we should change the logic not to allow SQL export format to any query having a Join?

udan11:
In theory, it should be possible to export whatever result (even joins).

I guess that for now, we could restrict this.

nrv:
This patch is working fine on the 4.8.3 version to correct bug #1066

PR #13305, opened 2017-05-27
showing tooltip for user not in user table

### LINKED ISSUE #12718: Password No shown for user not in user table
When a user has access to a db (or table), but is not in the user table he's shown as not having a password. I don't think this is right, it'd be better if the "User accounts overview" page (how about just Users?) showed him not being in the user table.

manishbisht:
Fixes #12718 


nijel:
Thanks I've merged your change with minor edits (coding style and reusing message we use elsewhere for this) as 253c9bed0ed210d4569d7909291ed2e588a97cfb.

PR #13308, opened 2017-05-27
ServerConfigChecks.php, try to use phpseclib first

oittaa:
Logic here was backwards. We should use phpseclib, which uses random_bytes internally, if phpseclib is available. Refactored a bit, because the previous (and simpler) pull request #13307 failed on Travis with the following error:
"PHP Fatal error:  Call to undefined function phpseclib\Crypt\Random::string()"
https://travis-ci.org/phpmyadmin/phpmyadmin/builds/236686976?utm_source=github_status&utm_medium=notification

Additionally reduce the number of retries by discarding the highest bit. The byte's range changes from 0-255 to 0-127.

Signed-off-by: Eero Vuojolahti <eero@oittaa.net>

`testServerConfigChecksPerformConfigChecks()` in [test/libraries/PMA_SetupIndex_test.php](https://github.com/phpmyadmin/phpmyadmin/blob/d6b198e3fafcb6ce1a81db32d5b903c240c3f65a/test/libraries/PMA_SetupIndex_test.php#L167) should cover the changes.

### DIFF
--- a/libraries/config/ServerConfigChecks.php
+++ b/libraries/config/ServerConfigChecks.php
@@ -218,7 +218,7 @@ class ServerConfigChecks
         if ($cookieAuthServer && $blowfishSecret === null) {
             $blowfishSecret = '';
             while (strlen($blowfishSecret) < 32) {
-                if (class_exists('phpseclib\Crypt\Random')) {
+                if (is_callable(array('phpseclib\Crypt\Random', 'string'))) {
                     $byte = phpseclib\Crypt\Random::string(1);
                 } else {
                     $byte = openssl_random_pseudo_bytes(1);
--- a/libraries/config/ServerConfigChecks.php
+++ b/libraries/config/ServerConfigChecks.php
@@ -7,6 +7,7 @@
  */
 namespace PMA\libraries\config;
 
+use phpseclib\Crypt;
 use PMA\libraries\URL;
 
 /**
@@ -218,8 +219,8 @@ class ServerConfigChecks
         if ($cookieAuthServer && $blowfishSecret === null) {
             $blowfishSecret = '';
             while (strlen($blowfishSecret) < 32) {
-                if (is_callable('random_bytes')) {
-                    $byte = random_bytes(1);
+                if (is_callable('Crypt\Random::string')) {
+                    $byte = Crypt\Random::string(1);
                 } else {
                     $byte = openssl_random_pseudo_bytes(1);
                 }
--- a/libraries/config/ServerConfigChecks.php
+++ b/libraries/config/ServerConfigChecks.php
@@ -218,8 +218,8 @@ class ServerConfigChecks
         if ($cookieAuthServer && $blowfishSecret === null) {
             $blowfishSecret = '';
             while (strlen($blowfishSecret) < 32) {
-                if (is_callable(array('phpseclib\Crypt\Random', 'string'))) {
-                    $byte = phpseclib\Crypt\Random::string(1);
+                if (is_callable('random_bytes')) {
+                    $byte = random_bytes(1);
                 } else {
                     $byte = openssl_random_pseudo_bytes(1);
                 }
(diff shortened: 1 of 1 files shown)


libraries/config/ServerConfigChecks.php
```diff
@@ -217,16 +217,16 @@ protected function performConfigChecksServersSetBlowfishSecret(
...
         if ($cookieAuthServer && $blowfishSecret === null) {
             $blowfishSecret = '';
-            if (! function_exists('openssl_random_pseudo_bytes')) {
-                $random_func = 'phpseclib\\Crypt\\Random::string';
-            } else {
-                $random_func = 'openssl_random_pseudo_bytes';
-            }
             while (strlen($blowfishSecret) < 32) {
-                $byte = $random_func(1);
+                if (is_callable('random_bytes')) {
```
udan11: This does not work on PHP < 7.0 without OpenSSL.

`random_bytes` was added in PHP 7.0 and `openssl_random_pseudo_bytes` may be unavailable if PHP was compiled without OpenSSL support. 

The thing with `phpseclib\Crypt\Random::string()` is that it users `random_bytes` if available or uses a pure-PHP CSPRNG.

oittaa: Did it work before without OpenSSL? There was something weird with Travis and the tests failed if I used `phpseclib\Crypt\Random::string`.

phpseclib requires random_compat library so `random_bytes` will be available even on PHP < 7.0.

https://github.com/phpseclib/phpseclib/blob/master/composer.json#L55

oittaa:
Seems like the latest translations broke Travis, so I don't know if this is still broken or not.

libraries/config/ServerConfigChecks.php
```diff
@@ -217,16 +218,16 @@ protected function performConfigChecksServersSetBlowfishSecret(
...
-                $byte = $random_func(1);
+                if (is_callable('Crypt\Random::string')) {
+                    $byte = Crypt\Random::string(1);
+                } else {
+                    $byte = openssl_random_pseudo_bytes(1);
+                }
                 // We want only ASCII chars
-                if (ord($byte) > 32 && ord($byte) < 127) {
-                    $blowfishSecret .= $byte;
+                $byte = ord($byte) >> 1;
```
nijel: This doesn't really help to anything - you're still throwing away same amount of generated content as more bytes now fall into < 32 range.

What would actually help here is `$byte = ord($byte) & 0x7f;`, I've implemented such change in 6b18dfa588d0fce3598970ea4b0a339afe9db2f1.

libraries/config/ServerConfigChecks.php
nijel: We need either phpseclib or openssl support, so adding another code path is not required here.

libraries/config/ServerConfigChecks.php
```diff
@@ -217,16 +218,16 @@ protected function performConfigChecksServersSetBlowfishSecret(
...
         if ($cookieAuthServer && $blowfishSecret === null) {
             $blowfishSecret = '';
-            if (! function_exists('openssl_random_pseudo_bytes')) {
-                $random_func = 'phpseclib\\Crypt\\Random::string';
-            } else {
-                $random_func = 'openssl_random_pseudo_bytes';
-            }
             while (strlen($blowfishSecret) < 32) {
-                $byte = $random_func(1);
+                if (is_callable('Crypt\Random::string')) {
```
nijel: Calling this in the loop sounds like a bad idea performance wise.

nijel:
Thanks for the patch, however I've ended up implementing the same change separately. The main reason is that you mix too many changes into one pull requests and some of them are questionable.

libraries/config/ServerConfigChecks.php
oittaa: Thanks for implementing the change, but you're actually wrong about the bitshift. The range is 0-127 as I stated. https://3v4l.org/JgtF1

It really looks like that phpMyAdmin was broken without OpenSSL before and your commit 6b18dfa revealed it again. You need to add the following:
`use phpseclib\Crypt;`

nijel: You're right, both versions work pretty much same. The use is not really needed as we use full namespace there.

PR #13313, opened 2017-05-28
Double character encoding removed

### LINKED ISSUE #9419: Export with file option creates invalid commands
When using the export option with the Save as file, option enabled creates some commands that when doing the import of the very same generated file does not import the spanish characters correctly.
This in an example of an export file with the Save as file option enabled:

```
-- phpMyAdmin SQL Dump
-- version 3.3.4-rc1
-- http://www.phpmyadmin.net
--
-- Host: localhost
-- Generation Time: Jun 22, 2010 at 04:45 AM
-- Server version: 5.1.46
-- PHP Version: 5.3.2
SET SQL_MODE="NO_AUTO_VALUE_ON_ZERO";
/*!40101 SET @OLD_CHARACTER_SET_CLIENT=@@CHARACTER_SET_CLIENT */;
/*!40101 SET @OLD_CHARACTER_SET_RESULTS=@@CHARACTER_SET_RESULTS */;
/*!40101 SET @OLD_COLLATION_CONNECTION=@@COLLATION_CONNECTION */;
/*!40101 SET NAMES latin1 */;
--
-- Database: `carlos_photoalbums`
--
-- --------------------------------------------------------
--
-- Table structure for table `PhotoAlbums`
--

then the table and the data details and at the end:
/*!40101 SET CHARACTER_SET_CLIENT=@OLD_CHARACTER_SET_CLIENT */;
/*!40101 SET CHARACTER_SET_RESULTS=@OLD_CHARACTER_SET_RESULTS */;
/*!40101 SET COLLATION_CONNECTION=@OLD_COLLATION_CONNECTION */;


---------------------------------------------------------------------------------------------------------------------------------------------------------------------

This is the very same export but WITHOUT the Save as file option enabled
-- phpMyAdmin SQL Dump
-- version 3.3.4-rc1
-- http://www.phpmyadmin.net
--
-- Host: localhost
-- Generation Time: Jun 22, 2010 at 04:44 AM
-- Server version: 5.1.46
-- PHP Version: 5.3.2
SET SQL_MODE="NO_AUTO_VALUE_ON_ZERO";
--
-- Database: `carlos_photoalbums`
--
-- --------------------------------------------------------
--
-- Table structure for table `PhotoAlbums`
--
```

then the table and the data details and at the end there is nothing else.

This behavior is the same using PMA, 332, 333, and the recently released 334.rc1

I use Apache 2.2.11, Php 5.3.2 and MySQL client version: mysqlnd 5.0.7-dev - 091210 - $Revision: 294543 $

If I delete those 6 lines (3 at beginning and the 3 at end) the impor works correctly and the spanish characters are saved fine.

---
- **Original URL**: https://sourceforge.net/p/phpmyadmin/bugs/3049/
- **Original author**: yacar_suscri
- **Found in version**: 3.5.5

manishbisht:
Fixes: #9419

### DIFF
--- a/libraries/import.lib.php
+++ b/libraries/import.lib.php
@@ -318,7 +318,7 @@ function PMA_lookForUse($buffer, $db, $reload)
 function PMA_importGetNextChunk($size = 32768)
 {
     global $compression, $import_handle, $charset_conversion, $charset_of_file,
-        $read_multiply;
+        $read_multiply, $reset_charset;
 
     // Add some progression while reading large amount of data
     if ($read_multiply <= 8) {
@@ -357,6 +357,11 @@ function PMA_importGetNextChunk($size = 32768)
     $result = $import_handle->read($size);
     $GLOBALS['finished'] = $import_handle->eof();
     $GLOBALS['offset'] += $size;
+
+    if (strpos($result, 'SET NAMES') !== false) {
+        $charset_conversion = false;
+        $reset_charset = true;
+    }
 
     if ($charset_conversion) {
         return Encoding::convertString($charset_of_file, 'utf-8', $result);


libraries/import.lib.php
```diff
@@ -358,6 +358,11 @@ function PMA_importGetNextChunk($size = 32768)
     $GLOBALS['finished'] = $import_handle->eof();
     $GLOBALS['offset'] += $size;
 
+    if (strpos($result, 'SET NAMES') !== false) {
```
nijel: This looks quite fragile as it will happily use commented out strings or strings being actually values (eg. `INSERT INTO foo SET bar="SET NAMES"`). It should probably use SQL parser instead to actually understand the queries.

manishbisht: @nijel This is the sql parser that I have to use https://github.com/phpmyadmin/sql-parser ? Also is this already included in phpyadmin ?

nijel: Yes, that's it. However executing it on every query will slow down the import, so it probably should be used only on queries which match some simple check...

Another problem is that this code is used for all imports, while this makes sense only for SQL ones (so it should be rather placed in libraries/plugins/import/ImportSql.php).

nijel:
As mentioned before, this is wrong approach:

- it parses input for all imports, while it should affect on SQL one
- it matches by string, what can math in case the statement is used in comment or in field value

manishbisht:
@nijel Can we first check for `if (strpos($result, 'SET NAMES') !== false)` and then take use of SQL parser (https://github.com/phpmyadmin/sql-parser/blob/master/src/Statements/SetStatement.php). In this way we will be executing sql parser when "SET NAMES" is found.

nijel:
Let's move the discussion back to the issue (https://github.com/phpmyadmin/phpmyadmin/issues/9419)...

PR #13315, opened 2017-05-29
Update common.inc.php  #13316

LuciferJack:
fix You have enabled mbstring.func_overload in your PHP configuration. This option is incompatible with phpMyAdmin and might cause some data to be corrupted!  
because is **empyty string**

- [ ]  Any new functionality is covered by tests


LuciferJack:
php is unset this contition ,so is empyty string bug phpmyadmin is check regnize is set

nijel:
Thanks for your contribution, however it has few problems:

* It doesn't have Signed-Off-By, see https://github.com/phpmyadmin/phpmyadmin/blob/master/CONTRIBUTING.md
* It consists of several commits, one of them is minor change and another revering change done in previous, it's better to squash the commits to one.

Anyway I've fixed #13316 independently in c0bd19cfbb61b2986a1b034c4398b09d47647b6a, as I didn't notice the PR before (it was not linked to the issue as the commit or PR message didn't include the issue ID, for some reason having it in summary is not good enough for GitHub).

Issue #13270, opened 2017-05-12
Remove inline javascript events


emanuelb:
See more information in previous issue:
#12261 (Remove inline javascript)

Below remained places that use inline JS events:
```
./view_operations.php:        <td><input type="text" size="20" name="new_name" onfocus="this.select()"
./view_create.php:        . ' onfocus="this.select()" maxlength="64"'
./view_create.php:    . ' onfocus="this.select()"'
./view_create.php:    $htmlString .= ' onclick="selectContent(this, sql_box_locked, true)"';
./templates/columns_definitions/column_definitions_form.phtml:                               onfocus="this.select()" />
./templates/database/designer/having_query_panel.twig:                                    onclick="document.getElementById('query_having').style.display = 'none';" />
./templates/database/designer/aggregate_query_panel.twig:                                    onclick="document.getElementById('query_Aggregate').style.display = 'none';" />
./templates/database/designer/side_menu.phtml:           onclick="build_query('SQL Query on Database', 0)"
./templates/database/designer/rename_to_panel.twig:                                    onclick="document.getElementById('query_rename_to').style.display = 'none';" />
./templates/database/designer/where_query_panel.twig:                                    onclick="document.getElementById('query_where').style.display = 'none'" />
./templates/database/structure/check_all_tables.phtml:         / <a href="#" onclick="unMarkAllRows('tablesForm'); <?= $overhead_check; ?> return false;"><?= __('Check tables having overhead'); ?></a>
./templates/server/databases/table_row.phtml:        <a onclick="PMA_commonActions.setDb('<?= PMA\libraries\Sanitize::jsFormat($current['SCHEMA_NAME']) ?>');"
./templates/table/index_form.phtml:      onsubmit="if (typeof(this.elements['index[Key_name]'].disabled) != 'undefined') this.elements['index[Key_name]'].disabled = false;">
./templates/table/index_form.phtml:                       onfocus="this.select()" />
./templates/table/index_form.phtml:                                   onfocus="this.select()"name="index[columns][sub_parts][]"
./templates/table/index_form.phtml:                                   onfocus="this.select()"
./templates/table/structure/actions_in_table_structure.phtml:            <a href="#" onclick="$('input:checkbox').prop('checked', false);
./templates/table/structure/actions_in_table_structure.phtml:            <a href="#" onclick="$('input:checkbox').prop('checked', false);
[...]

PR #13330, opened 2017-05-30
Removed inline mouseover/mouseout to toggle sort icon.

karlvaniseghem:
Solving part of #13270

Signed-off-by: Karl Van Iseghem <karl.vaniseghem@icloud.com>


nijel:
Pretty much same logic is used in [libraries/DisplayResults.php](https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/DisplayResults.php#L2317), it would be great if those could share the javascript code.

nijel:

I've applied similar fix to DisplayResults in d639308f480a452b8b17b51329dc7ae94b78a392.

PR #13332, opened 2017-05-30
Removed inline javascript

karlvaniseghem:
Signed-off-by: Karl Van Iseghem <karl.vaniseghem@icloud.com>

solving part of #13270

### DIFF
--- a/js/normalization.js
+++ b/js/normalization.js
@@ -59,8 +59,14 @@ function goTo3NFStep1(newTables)
             });
             $("#mainContent").find("#newCols").html('');
             $('.tblFooters').html('');
+
             if (data.subText !== "") {
-                $('.tblFooters').html('<input type="button" onClick="processDependencies(\'\', true);" value="' + PMA_messages.strDone + '"/>');
+                $('<input/>')
+                    .attr({type: 'button', value: PMA_messages.strDone})
+                    .on('click', function() {
+                        processDependencies('', true);
+                    })
+                    .appendTo('.tblFooters');
             }
         }
     );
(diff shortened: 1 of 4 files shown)


js/normalization.js
```diff
@@ -81,8 +87,14 @@ function goTo2NFStep1() {
             $("#mainContent p").html(data.subText);
             $("#mainContent #extra").html(data.extra);
             $("#mainContent #newCols").html('');
+            
             if (data.subText !== '') {
-                $('.tblFooters').html('<input type="submit" value="' + PMA_messages.strDone + '" onclick="processDependencies(\'' + escapeJsString(escapeHtml(data.primary_key)) + '\');">');
+                var doneButton = $('<input />')
+                    .attr({type: 'submit', value: PMA_messages.strDone, })
+                    .on('click', function() {
+                        processDependencies('\'' + escapeJsString(escapeHtml(data.primary_key)) + '\'');
```
nijel: The escapeJsString should not be needed here (as you're not creating javascript code, but string literal).

js/normalization.js
```diff
@@ -637,8 +666,20 @@ AJAX.registerOnload('normalization.js', function() {
...
+                .attr({type: 'submit', value: PMA_messages.strCancel})
+                .on('click', function() {
+                    $('#newCols').html('');
+                    $('#extra input[type=checkbox]').prop('checked', false);
+                })
+                .appendTo('.tblFooters');
+            $('<input />')
+                .attr({type: 'submit', value: PMA_messages.strGo})
+                .on('click', function() {
+                    moveRepeatingGroup('\'' + escapeJsString(escapeHtml(repeatingCols)) + '\'');
```
nijel: ...and same here, escapeJsString is not needed here.

nijel (changes requested):
Thanks, can you please look at escapeJsString usage? IMHO it should not be used in those cases.

nijel:
Merged with following fixes:

* 58d966360e Avoid changing cursor on hover
* 024d53ab78 Remove not needed escaping

PR #13335, opened 2017-05-31
Fix setTimeout function on header_location.twig

### LINKED ISSUE #13269: setTimeout first argument is string instead of function (depend on unneeded unsafe-eval)

alvescleiton:
Fix #13269

Signed-off-by: Cleiton Alves <cleiton_jc@yahoo.com.br>

### DIFF
--- a/templates/header_location.twig
+++ b/templates/header_location.twig
@@ -8,7 +8,9 @@
     <meta http-equiv="Refresh" content="0;url={{ uri }}" />
     <script type="text/javascript">
         //<![CDATA[
-        setTimeout("window.location = decodeURI('{{ Sanitize_escapeJsString(uri) }}')", 2000);
+        setTimeout(function(){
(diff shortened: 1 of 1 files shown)


nijel:
Sorry, but there was another pull request implementing this: https://github.com/phpmyadmin/phpmyadmin/pull/13284

(Also the tests didn't pass here)

Issue #13145, opened 2017-04-02
Table Ops layout


OlafvdSpek:
This looks bad and I don't think it improves usability.
One column is probably best.

BTW, the rename table name field is longer then the containing table..

![pma](https://cloud.githubusercontent.com/assets/2218936/24585488/e3ad12bc-178b-11e7-8503-6dcf49caff1c.PNG)

Fenn-CS:
@OlafvdSpek If I am getting you right we have to eliminate the five column system and squash everything into one column.... I wondering if these wont bring a confusion amongst the "go" buttons or do we need to rename them too? I like to have more light on what you actually expect. thanks.

OlafvdSpek:
I'm expecting something that looks better. ;)

The number of columns currently depends on the window width, you can get a one-column layout already.
I don't think there's a problem with the Go buttons.

smileytechguy:
I am going to attempt this

smileytechguy:
![image](https://cloud.githubusercontent.com/assets/8005215/25213113/5368e832-255c-11e7-9ba8-51ace7162503.png)
I like it looking like this.  However, I am not sure how to modify the source to do so.  All I did was change the default widths to 100%

shucon:
I made some changes in css but seems like width : % changes automatically due to some reason.
Any help?

karlvaniseghem:
What about this?
Removing id="boxContainer" and setting min/max-width in css to 420px, margin: 20px auto;
Removing all "operations_half_width" classes in child divs.

Effect:
![schermafbeelding 2017-05-30 om 19 40 10](https://cloud.githubusercontent.com/assets/870825/26596679/d472127c-456f-11e7-8dfe-ed455f1518c9.png)

OlafvdSpek:
Much better, though I'm not entirely sure on the center alignment.

karlvaniseghem:
How about left aligned?
![schermafbeelding 2017-05-31 om 09 44 26](https://cloud.githubusercontent.com/assets/870825/26621098/c2b71c28-45e5-11e7-9904-3ab1e3b05de8.png)

OlafvdSpek:
Great!
There's a lot of whitespace around the Go buttons, perhaps we could reduce that too.

karlvaniseghem:
Overwrite default fieldset padding (1.5em).
Changed it to padding: 5px; for tblFooters only.

[...]

PR #13339, opened 2017-05-31
Enhance layout Table Operations

karlvaniseghem:
Remove boxContainer and set fixed width on operation tables
Removed extra whitespace around Go buttons
see #13145

### DIFF
--- a/tbl_operations.php
+++ b/tbl_operations.php
@@ -299,7 +299,7 @@ $columns = $GLOBALS['dbi']->getColumns($GLOBALS['db'], $GLOBALS['table']);
 /**
  * Displays the page
  */
-$response->addHTML('<div style="min-width: 450px; max-width: 450px; margin: 20px;">');
+$response->addHTML('<div class="operations_container">');
 
 /**
  * Order the table
--- a/tbl_operations.php
+++ b/tbl_operations.php
@@ -299,7 +299,7 @@ $columns = $GLOBALS['dbi']->getColumns($GLOBALS['db'], $GLOBALS['table']);
 /**
  * Displays the page
  */
-$response->addHTML('<div id="boxContainer" data-box-width="300">');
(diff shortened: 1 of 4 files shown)


tbl_operations.php
```diff
@@ -299,7 +299,7 @@
 /**
  * Displays the page
  */
-$response->addHTML('<div id="boxContainer" data-box-width="300">');
+$response->addHTML('<div style="min-width: 450px; max-width: 450px; margin: 20px;">');
```
mauriciofauth: I don't think it's a good idea to include inline style.

nijel:
Honestly I'm not sure if it's better to leave so much of empty space on the page. Though I agree that current looks is not good. Some feedback from @phpmyadmin/developers is welcome as well.

Current look:

screenshot-2017-6-6 localhost localhost bar2 cars phpmyadmin 4 8 0-dev-current

With this PR:

screenshot-2017-6-6 localhost localhost bar2 cars phpmyadmin 4 8 0-dev

nijel:
PS: The additional spaces around go buttons probably could be removed generally, not just in this specific case...

bakgat:
How about collapsible panels or tabs instead for viewing each fieldset at once ?

nijel:
I've changed default rendering of the fieldset buttons to be more compact in ae67d1c2a7, so at least this part is now better:

screenshot-2017-6-6 localhost localhost bar2 cars phpmyadmin 4 8 0-dev-master

As for layout, I'm not really sure what to choose here. How about using second level navigation to split the page?

nijel:
I've simplified the code in edad6fc2d9a336ed73aab52629479a2a8dfaed9a, any further improvements to the layout should be done on top of that.

ibennetch:
One thing that keeps bothering me is how the first and third items are
nearly the same - copy and move are similar functions, I wonder if we
should try to merge these two fieldsets.

Issue #13268, opened 2017-05-11
New UI for Multi-table query generator


Achilles-96:
Currently, we have two ways of building multi-table queries in order to reduce user effort while querying multiple tables - one is designer (db_designer.php) and the other is a form-based query builder (db_qbe.php) which is redundant. The new UI replaces these two interfaces. This UI helps in reducing the learning curve for users who are new to managing databases and also to make multi-table querying faster and easier.

This issue is part of GSoC project.
Mentor: @ibennetch

Achilles-96:
* Right now, after result of submitted query is fetched, a lot of space is occupied on top with various options pushing the actual builder down ([image](http://bit.ly/2nCHGQM)). This is unfriendly as it is common to want to modify the query soon. The new interface will ensure that builder itself will take much less amount of space on top leaving room for result to be shown below builder.
* I have made a [mockup](http://bit.ly/2nlONui) of what I planned to implement. Criteria will be entered as free text (we can provide interface for basic criteria like ‘AND’, ‘OR’ etc.)
* I also have a plan to add additional filters after results are displayed, as reflected in the mockup referred above.

Please share your opinions.

PR #13342, opened 2017-06-01
Interface design for Multi-table query

Achilles-96:
Not to merge.
I made layout for new UI. Please review this and let me know if I can proceed further implementing backend as well.
Issue - #13268 

### DIFF
--- a/db_multi_table_query.php
+++ b/db_multi_table_query.php
@@ -1,0 +1,58 @@
+<?php
+/* vim: set expandtab sw=4 ts=4 sts=4: */
+/**
+ * query by example the whole database
+ *
+ * @package PhpMyAdmin
+ */
+use PhpMyAdmin\Response;
+use PhpMyAdmin\DbMultiTableQuery;
+use PhpMyAdmin\Sql;
+
+require_once 'libraries/common.inc.php';
+
+if (isset($_REQUEST['sql_query'])) {
+    $sql_query = $_REQUEST['sql_query'];
+    $db = $_REQUEST['db'];
+    include_once 'libraries/parse_analyze.lib.php';
+    list(
+        $analyzed_sql_results,
+        $db,
+        $table_from_sql
+    ) = PMA_parseAnalyze($sql_query, $db);
+
+    extract($analyzed_sql_results);
+    $goto = 'db_multi_table_query.php';
+    $html_output = Sql::executeQueryAndSendQueryResponse(
+        null, // analyzed_sql_results
+        false, // is_gotofile
+        $db, // db
+        null, // table
+        null, // find_real_end
+        null, // sql_query_for_bookmark - see below
+        null, // extra_data
+        null, // message_to_show
+        null, // message
+        null, // sql_data
+        $goto, // goto
+        $pmaThemeImage, // pmaThemeImage
+        null, // disp_query
+        null, // disp_message
+        null, // query_type
+        $sql_query, // sql_query
+        null, // selectedTables
+        null // complete_query
+    );
+    exit;
+}
+
+$response = Response::getInstance();
+
+$header = $response->getHeader();
+$scripts = $header->getScripts();
+$scripts->addFile('vendor/jquery.md5.js');
+$scripts->addFile('db_multi_table_query.js');
+
+$QueryInstance = new DbMultiTableQuery($db);
+
+$response->addHTML($QueryInstance->getFormHTML());
--- a/js/db_multi_table_query.js
+++ b/js/db_multi_table_query.js
@@ -1,0 +1,256 @@
+/* vim: set expandtab sw=4 ts=4 sts=4: */
+/**
+ * @fileoverview    function used in QBE for DB
+ * @name            Database Operations
+ *
+ * @requires    jQuery
+ * @requires    jQueryUI
+ * @requires    js/functions.js
+ *
+ */
+
+/**
+ * js file for handling AJAX and other events in db_multi_table_query.php
+ */
+
+/**
+ * Unbind all event handlers before tearing down a page
+ */
+AJAX.registerTeardown('db_multi_table_query.js', function () {
+    $('.tableNameSelect').each(function(){
+        $(this).off('change');
+    });
+    $('#update_query_button').off('click');
+    $('#add_column_button').off('click');
+});
+
+AJAX.registerOnload('db_multi_table_query.js', function () {
+
+    var editor = PMA_getSQLEditor($('#MultiSqlquery'), {}, 'both');
+    $('.CodeMirror-line').css('text-align', 'left');
+    editor.setSize(-1, 50);
+
+    var column_count = 3;
+    PMA_init_slider();
+    addNewColumnCallbacks();
+
+    function escapeBacktick(s) {
+        return s.replace('`', '``');
+    }
+
+    function escapeSingleQuote(s) {
+        return s.replace('\\', '\\\\').replace('\'', '\\\'');
+    }
+
+    $('#update_query_button').on('click', function() {
+        var columns = [];
+        var table_aliases = {};
+        $('.tableNameSelect').each(function() {
+            $show = $(this).siblings('.show_col').first();
+            if ($(this).val() !== "" && $show.prop('checked')) {
+
+                var table_alias = $(this).siblings('.table_alias').first().val();
+                var column_alias = $(this).siblings('.col_alias').first().val();
+
+                if (table_alias !== "") {
+                    columns.push([table_alias, $(this).siblings('.columnNameSelect').first().val()]);
+                } else {
+                    columns.push([$(this).val(), $(this).siblings('.columnNameSelect').first().val()]);
+                }
+
+                columns[columns.length - 1].push(column_alias);
+
+                if ($(this).val() in table_aliases) {
+                    if (!(table_aliases[$(this).val()].includes(table_alias))) {
+                        table_aliases[$(this).val()].push(table_alias);
+                    }
+                } else {
+                    table_aliases[$(this).val()] = [table_alias];
+                }
+            }
+        });
+        if (Object.keys(table_aliases).length === 0) {
+            PMA_ajaxShowMessage('Nothing selected', false, 'error');
+            return;
+        }
+
+        query = 'SELECT ';
+        query += '`' + escapeBacktick(columns[0][0]) + '`.`' + escapeBacktick(columns[0][1]) + '`';
+        if (columns[0][2] !== "") {
+            query += ' AS ' + columns[0][2];
+        }
+        for (var i = 1; i < columns.length; i++) {
+            query += ', `' + escapeBacktick(columns[i][0]) + '`.`' + escapeBacktick(columns[i][1]) + '`';
+            if (columns[i][2] !== "") {
+                query += ' AS `' + escapeBacktick(columns[0][2]) + '`';
+            }
+        }
+        query += '\nFROM ';
+        var table_count = 0;
+        for (var table in table_aliases) {
+            for (var i = 0; i < table_aliases[table].length; i++) {
+                if (table_count > 0) {
+                    query += ', ';
+                }
+                query += '`' + escapeBacktick(table) + '`'
+                if (table_aliases[table][i] !== "") {
+                    query += ' AS `' + escapeBacktick(table_aliases[table][i]) + '`';
+                }
+                table_count++;
+            }
+        }
+
+        $criteria_col_count = $('.criteria_col:checked').length;
+        if ($criteria_col_count > 0) {
+            query += '\nWHERE ';
+
+            var logical_ops = [];
+
+            var count = 0;
+
+            $('.tableNameSelect').each(function() {
+                $criteria_div = $(this).siblings('.slide-wrapper').first();
+                $use_criteria = $(this).siblings('.criteria_col').first();
+                if ($(this).val() !== "" && $use_criteria.prop('checked')) {
+                    if (count > 0) {
+                        $criteria_div.find('input.logical_op').each(function(){
+                            if ($(this).prop('checked')) {
+                                query += ' ' + $(this).val() + ' ';
+                            }
+                        });
+                    }
+                    formats_text = {
+                        '=' : " = '%s'",
+                        '>' : " > '%s'",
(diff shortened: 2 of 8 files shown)


Achilles-96:
Access the page by going to http://<pma-address>/db_multi_table_query.php?db=<db_name>
eg: http://localhost/pma-dev/db_multi_table_query.php?db=hello

ibennetch:
This looks like a good start to me. I don't have much feedback right now on this, I thought your mockup looked good and this is beginning to look really good as well. I think you're clearly on the right path.

Any opinions from other @phpmyadmin/developers ?

Achilles-96:
This PR is ready for review.

ibennetch:
I'm currently getting an error:
> Warning in ./db_multi_table_query.php#12
require_once(libraries/sql.lib.php): failed to open stream: No such file or directory
> 
> Backtrace
> 
> ./db_multi_table_query.php#12: require_once()

Do you know what went wrong?

ibennetch:
I'm sure someone will correct me if I'm wrong, but I believe this is to replace the current Query tab and Visual Builder from db_designer. In that case, we probably should update the link from the Query tab to this new page and remove the old code (at least, db_qbe — I'm not certain offhand how much of the db_designer code is integrated with the rest of the Designer features).

Achilles-96:
I am not getting that error. sql.lib.php is present in libraries. I don't understand why you are getting that error. 
I am trying to change the link for query tab.

Achilles-96:
The link works now. Should I remove old code in this PR itself.

ibennetch:
I'll try again, if it's working for you then there must be a problem on my side. Thanks.

ibennetch:
I have a few thoughts about things that can be improved:

* After pressing Submit query, the results are shown under the query interface. I can scroll the results and down, but the query interface floats on top. I think ideally, it would collapse but still be present, so the user could go back and make changes; kind of like this:
image
* When used with `$cfg['InitialSlidersState'] = 'open';`, the Criteria area is expanded by default (which is fine), but then clicking the checkbox to enable the criteria results in the area being hidden. I think the best solution to this is simply to remove the collapse/hide event from the checkbox — if a user wants to hide it, they can click the text "- criteria". For those users in the default state, clicking the checkbox will still expand the area, which is great.
* When I combine two tables with criteria fields, they seem to get concatenated in an odd way, such as
  > SELECT \`actor\`.\`actor_id\`, \`film\`.\`film_id\`
  > FROM \`actor\`, \`film\`
  > WHERE \`actor\`.\`actor_id\` = 1\`film\`.\`film_id\` = 1;

  image

This looks nice, good work.

ibennetch (changes requested):
This doesn't quite work because libraries/Menu.php is still looking for the $tabs['qbe'] array but libraries/Util.php changes from qbe to 'multi_table_query'.

I have a slight preference toward renaming the array in Menu.php from qbe to multi_table_query.

Achilles-96:
Regarding the first two points: They make sense and I will be on them.
Regarding combining two tables, you have to select whether to use `AND` or `OR` by selecting the radio button. I guess one of them should be checked by default.
I will also do changes in Menu.php as you mentioned.

Thanks for the review.

Achilles-96:
Also, I want to suggest a different solution for slider thing. Can we add an additional parameter to slider to override default configuration. Please check my latest patch.
I feel the sliding the query window up after fetching result looks very good.

Achilles-96:
Also, I put `checked` for `AND` by default in criteria. I don't understand why it is not checking. Except that, I believe I addressed all the concerns raised by you.

db_multi_table_query.php
```diff
@@ -0,0 +1,59 @@
...
+use PMA\libraries\Response;
+use PMA\libraries\DbMultiTableQuery;
+
+require_once 'libraries/common.inc.php';
+require_once 'libraries/sql.lib.php';
+
+if (isset($_REQUEST['sql_query'])) {
+    $sql_query = $_REQUEST['sql_query'];
+    $db = $_REQUEST['db'];
+    //    $response = Response::getInstance();
```
ibennetch: Is this line old or debugging code that can be removed?

Achilles-96:
Hi @ibennetch , should I update the patch removing old code?

ibennetch:
Let's remove the old code as a separate pull request.

Achilles-96:
What is the conflict here? I am able to merge it on my local repo.

ibennetch (approved):

ibennetch:
It's reporting that `libraries/Menu.php` and `libraries/Util.php` have conflicts, but when I tried to merge it manually, git was able to do so with a three-way merge. I think this means that Github is unable to merge automatically while maintaining proper history.

I believe the correct resolution is to rebase while in your branch; that will bring in the changes from master and allow the merge to complete successfully. It wouldn't hurt to make a copy of your branch first (`git checkout Multi-Table-Query-UI` then `git branch Multi-Table-backup` should do).

From within Multi-Table-Query-UI, you should be able to simply run `git rebase origin/master` and it will give a message like

> First, rewinding head to replay your work on top of it...

Then it should think for a few minutes and give you a rebased version of your current branch (with no commits lost, but with new commit IDs). Then you can push that back to your github fork which will update the pull request here.

I think that's safer than me trying to merge and rebase it myself.

Achilles-96:
Moving files in `libraries/` caused some failures. I fixed them.

libraries/DbMultiTableQuery.php
```diff
@@ -0,0 +1,188 @@
+<?php
+/* vim: set expandtab sw=4 ts=4 sts=4: */
+/**
+ * Handles DB Multi-table query
+ *
+ * @package PhpMyAdmin
+ */
+namespace PMA\libraries;
+use PhpMyAdmin;
```
mauriciofauth: I think it's best to specify which classes you're using instead of including the entire namespace.

mauriciofauth:
You can move the `DbMultiTableQuery` class to the `libraries/classes` directory and use the `PhpMyAdmin` namespace. This would save the work of moving this file in the future.

Achilles-96:
Hi @mauriciofauth, thanks for the review. I updated the patch. Please verify.

mauriciofauth:
Hi, @Achilles-96. The only thing I see that is missing would be to move the `js/jquery.md5.js` file to `js/vendor/jquery.md5.js`, since it's a third-party library.

js/db_multi_table_query.js
```diff
@@ -0,0 +1,230 @@
...
+        $(this).off('change');
+    });
+    $('#update_query_button').off('click');
+    $('#add_column_button').off('click');
+});
+
+AJAX.registerOnload('db_multi_table_query.js', function () {
+
+    var editor = PMA_getSQLEditor($('#MultiSqlquery'), {}, 'both');
+    $('.CodeMirror-line').css('text-align','left');
```
ibennetch: Can you add a space after the comma here? It will make Codacy happier and is more readable.

js/db_multi_table_query.js
```diff
@@ -0,0 +1,230 @@
...
+    var column_count = 3;
+    PMA_init_slider();
+    addNewColumnCallbacks();
+
+    $('#update_query_button').on('click', function() {
+        var columns = [];
+        var table_aliases = {};
+        $('.tableNameSelect').each(function() {
+            $show = $(this).siblings('.show_col').first();
+            if ($(this).val() != "" && $show.prop('checked')) {
```
ibennetch: Can `!==` be used here instead of `!=`? The identity comparisons are generally considered safer than their equity counterparts (`===` instead of `==`).

js/db_multi_table_query.js
```diff
@@ -0,0 +1,230 @@
...
+
+                var table_alias = $(this).siblings('.table_alias').first().val();
+                var column_alias = $(this).siblings('.col_alias').first().val();
+
+                if (table_alias != "") {
```
ibennetch: Here as well.

js/db_multi_table_query.js
```diff
@@ -0,0 +1,230 @@
...
+                if ($(this).val() in table_aliases) {
+                    if (!(table_aliases[$(this).val()].includes(table_alias))) {
+                        table_aliases[$(this).val()].push(table_alias);
+                    }
+                } else {
+                    table_aliases[$(this).val()] = [table_alias];
+                }
+            }
+        });
+        if (Object.keys(table_aliases).length == 0) {
```
ibennetch: And here.

js/db_multi_table_query.js
```diff
@@ -0,0 +1,230 @@
...
+            PMA_ajaxShowMessage('Nothing selected', false, 'error');
+            return;
+        }
+
+        query = 'SELECT ';
+        query += '`' + columns[0][0] + '`.`' + columns[0][1] + '`';
+        if (columns[0][2] != "") {
```
ibennetch: Another one.

js/db_multi_table_query.js
```diff
@@ -0,0 +1,230 @@
...
+            query += ' AS ' + columns[0][2];
+        }
+        for (var i = 1; i < columns.length; i++) {
+            query += ', `' + columns[i][0] + '`.`' + columns[i][1] + '`';
+            if (columns[i][2] != "") {
```
ibennetch: Here also.

js/db_multi_table_query.js
```diff
@@ -0,0 +1,230 @@
...
+                table_count++;
+            }
+        }
+
+        $criteria_col_count = $('.criteria_col:checked').length;
+        if ($criteria_col_count > 0) {
+            query += '\nWHERE ';
+
+            var tables = [];
+            var columns = [];
```
ibennetch: I think you initialize the variables `tables` and `columns` here but I don't see them used; is this old code?

ibennetch (changes requested):
Okay, I think we're ready for merge here except for a few issues that Codacy found. I've commented on them in-line here; if you could take a look at those when you have a chance and let me know what you think.

Achilles-96:
Any more issues with this?

libraries/classes/DbMultiTableQuery.php
```diff
@@ -0,0 +1,187 @@
...
+        $html_output .= '<tr style="background:none" class="rhs_text">';
+        $html_output .= '<td></td>';
+        $html_output .= '<td colspan="2"><input type="text" style="width:91%" class="rhs_text_val" placeholder="Enter criteria as free text"></td>';
+        $html_output .= '</tr>';
+
+        $html_output .= '</table>';
+
+        $html_output .= '</div>';
+        $html_output .= '</div>';
+        $html_output .= '<a href="#" title="Remove this column" style="float:right;color:red" class="removeColumn">X</a>';
```
nijel: Can you please avoid inline styles? See https://github.com/phpmyadmin/phpmyadmin/issues/12262

libraries/classes/DbMultiTableQuery.php
```diff
@@ -0,0 +1,187 @@
...
+        $html_output .= '</fieldset>';
+        return $html_output;
```
nijel: It would be much better to use templates than generating HTML this way.

nijel (changes requested):

js/db_multi_table_query.js
```diff
@@ -0,0 +1,228 @@
...
+                }
+            }
+        });
+        if (Object.keys(table_aliases).length === 0) {
+            PMA_ajaxShowMessage('Nothing selected', false, 'error');
+            return;
+        }
+
+        query = 'SELECT ';
+        query += '`' + columns[0][0] + '`.`' + columns[0][1] + '`';
```
nijel: What if column name contains backtick? Generally this code is missing escaping.

ibennetch:
Indeed, here's a comment I tried to post over the weekend:

Also, we need to add single quotes (`'`) surrounding the information and escape the text the user can enter:

image

I believe that's as easy as adding the single quotes in your query and using the DBI function  `escapeString`. I suppose one could debate about whether integer values should be surrounded by single quotes (they shouldn't, technically, but MySQL seems to automatically cast and do the right thing here).

libraries/classes/DbMultiTableQuery.php
```diff
@@ -0,0 +1,103 @@
...
+    private $_default_no_of_columns;
+
+    public function __construct ($db_name)
+    {
+        $this->_db = $db_name;
+        $this->_default_no_of_columns = 3;
+    }
+
+    private function getColumnsHTML()
+    {
```
Achilles-96: @nijel, do you think this function also better templated?

nijel: I would prefer to use templates for all HTML generating, but I won't enforce this ;-)

js/db_multi_table_query.js
Achilles-96: Do we have any in-built escape function?

nijel: Yes, use ``Util::backquote`` (it adds the backticks as well).

ibennetch:
Please correct me if I'm wrong, but I believe this still lacks proper escaping; for instance a query with a single quote (such as `d'Arnaud`) fails.

ibennetch:
Good work @Achilles-96!

lem9:
Hi all,
in https://github.com/phpmyadmin/phpmyadmin/wiki/GSoC_2017_ideas_list I mentionned "the blocks could be moved left/right with the mouse, because their left-right order represent the order of the generated query". Is there a way to move the blocks in this pull request?

Also, what the point of having the "Query window" link for slider? When this is closed, nothing is left.

nijel:
Also the Selenium tests seem to no longer work:

```
1) PmaSeleniumQueryByExampleTest::testQueryByExample

InvalidArgumentException: Element not found.

/home/travis/build/phpmyadmin/phpmyadmin/test/selenium/TestBase.php:390

/home/travis/build/phpmyadmin/phpmyadmin/test/selenium/PmaSeleniumQueryByExampleTest.php:69
```

devenbansod:
Hi, I encountered a few issues while using the new interface:
* The `SORT` part of the query don't get generated
* I couldn't figure out a way to specify the sort order between two columns (for example to generate a clause like `ORDER BY article_rating DESC, article_time ASC`) which was present in older interface
* Do we have plans to introduce a better version of the visual builder or was it to be completely replaced by this?

PR #13354, opened 2017-06-04
Fix double escaping

CommanderRoot:

Fixes double escaping which for example breaks the "Change" and "Drop" buttons on the table structure view.

### DIFF
--- a/templates/prefs_autoload.twig
+++ b/templates/prefs_autoload.twig
@@ -3,7 +3,7 @@
         {{ hidden_inputs|raw }}
         <input type="hidden" name="json" value="" />
         <input type="hidden" name="submit_import" value="1" />
-        <input type="hidden" name="return_url" value="{{ return_url }}" />
+        <input type="hidden" name="return_url" value="{{ return_url|raw }}" />
         {% trans %}
             Your browser has phpMyAdmin configuration for this domain. Would you like to import it for current session?
         {% endtrans %}
--- a/templates/database/structure/tracking_icon.twig
+++ b/templates/database/structure/tracking_icon.twig
@@ -1,4 +1,4 @@
-<a href="tbl_tracking.php{{ url_query }}&amp;table={{ truename }}">
+<a href="tbl_tracking.php{{ url_query|raw }}&amp;table={{ truename }}">
     {% if is_tracked -%}
         {{ Util_getImage('eye.png', 'Tracking is active.'|trans) }}
     {%- else -%}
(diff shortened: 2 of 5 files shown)


nijel:
Maybe it would be better to pass url_query not escaped in the beginning? I'd like to hear opinion from @mauriciofauth (who did most of the templating work so far).

mauriciofauth:
@nijel, the Twig documentation says that the [escape](https://twig.sensiolabs.org/doc/1.x/filters/escape.html) filter internally uses the [htmlspecialchars](http://php.net/htmlspecialchars) function.
So, if `url_query` goes through this type of escape, then it's better to leave that job to Twig.

CommanderRoot:
I have tried to do it the other way around as suggested. The issue here is that `url_query` is used in non .twig templates too, at least for now. This would mean we would have to add "manual" escaping in the .phtml templates for it. To make things "worse" it's also used in HTML creating functions like [https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/rte/rte_footer.lib.php#L34](this) or [https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/rte/rte_list.lib.php#L227-L234](this) to give 2 examples.
While my solution is maybe not the best or nicest in the longrun it's the easiest and imo best for the current codebase.

nijel:
At least in some situations it's really easy to fix, I've fixed one of the cases in faf47b676ca37aee5c5a066cb6c7598fd01160c5.

The problem I see with your patch is that it introduces insecure handling of variables to templates and it will be harder to fix later. IMHO raw filter should be used only in rare cases and certainly not that widespread. Generally the template should be responsible for escaping to avoid deciding case by case whether the variable should be escaped or not.

templates/prefs_autoload.twig
```diff
@@ -3,7 +3,7 @@
         {{ hidden_inputs|raw }}
         <input type="hidden" name="json" value="" />
         <input type="hidden" name="submit_import" value="1" />
-        <input type="hidden" name="return_url" value="{{ return_url }}" />
+        <input type="hidden" name="return_url" value="{{ return_url|raw }}" />
```
nijel: Already fixed in faf47b676ca37aee5c5a066cb6c7598fd01160c5.

nijel:
I've also fixed the table structure in f887b65.

nijel:
...and the tracking is fixed in 8ea97c72a4.

PR #13355, opened 2017-06-05
Fix PMA_token generation

CommanderRoot:
The PMA_token is used in Javascript and HTML without escaping in most places.
If the token has ', ", < or > in it it can break the Javascript or the HTML on the page.


nijel:
Thanks, however I think it's better to fix cases where we handle token wrongly. Especially as there shouldn't be many of them:

* e01319a742 Avoid manual construction of hidden inputs with token
* 9ab34dbfd0 Remove unused pma_token js variable

PR #13359, opened 2017-06-06
Improve themes gitignore

MauricioFauth:
Improve `themes/.gitignore`.

In this case it is better to ignore everything and mark what should not be ignored than to mark everything that should be ignored.


nijel:
Thanks! I've merged this as e10c932133521908ff334e3246a83b0442619501 into QA_4_7 branch.

PR #13360, opened 2017-06-06
Refactor zip functions into static methods

MauricioFauth:


nijel:

Still I don't like mixing of ZipArchive and zip_* functions in this class, I'd really prefer if we can stick to only one of them (these were introduced quite some time ago).

mauriciofauth:
I did not pay much attention to the content of the functions while refactoring, but surely this class needs to be refactored again. I think the ideal would be for `ZipExtension` to be extended from `ZipArchive`.
Can you open an issue for this task, @nijel?

nijel:
Created https://github.com/phpmyadmin/phpmyadmin/issues/13365, it also covers ZipFile class, which is for some reason separate, but there is probably no good reason for that.

PR #13363, opened 2017-06-06
Not for merge: Add PHP 5.4 and 5.3 to test matrix to see what is broken

nijel:
This is mostly to see how much we depend on PHP 5.5 features, see #12790, #13242 and #13228.


nijel:
Some reasons why we don't support PHP 5.4:

* empty() now supports expressions, rather than only variables, there are at least 120 occurrences of this in the code, see http://php.net/manual/en/function.empty.php
* finally construct is used in few places, see http://php.net/manual/en/language.exceptions.php

There might be more, I just didn't look too deep...

PR #13367, opened 2017-06-08
Fix double escaping in table structure view

CommanderRoot:

`displayed_type` is either just escaped text or html + escaped text [depending on the length of the text](https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/Util.php#L2983-L2993). To fix the display issue which currently exists we can either use the already escaped content and display it raw as suggested with this commit or we can extract the length depending logic, put it into the template and use the unescaped text (`print_type`) and let the template do the escaping.


nijel:

PR #13373, opened 2017-06-11
Fix tbl_url_query escaping for tables containing +

CommanderRoot:

Fixes an issue where the links to browse, view the structure or modifiy a table are not working because for example `+` chars in a table name are not properly escaped, at least not to work in an `<a>` html tag. This has been broken by / since https://github.com/phpmyadmin/phpmyadmin/commit/c36d0001f93e2883faecbeaf07c25c2140f4500a . 


nijel:
Thanks for pull request, however I think it's better to use URL query construction helpers than doing that manually. I've done this in fc9140c0c5.

PR #13375, opened 2017-06-12
Update jQuery to 3.2.1

MauricioFauth:
Instead of including the compressed version and the source code, would not be better to just include the uncompressed version?


nijel:
The source code is there for licensing reasons...

nijel:

PR #13380, opened 2017-06-13
RFC: Add configuration for Probot Stale

nijel:
See https://github.com/probot/stale/

We have quite some pull requests which did not receive attention for quite some time and are not ready to be merged. This enables Probot Stale to label those as wontfix after 60 days and close them after additional 7 days without activity.

I think it's better to having a bot do the work because it, to some degree, is less personal and it can avoid bad feelings from the contributors as it's bureaucracy that closed your pull request.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,23 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale Issue or Pull Request is closed
+daysUntilClose: 7
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+exemptLabels:
+  - pinned
+  - security
+# Label to use when marking as stale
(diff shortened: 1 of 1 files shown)


mauriciofauth:
I think it's a good option. If there is no activity with these deadlines, it is better to close the PR.

ibennetch:
I somewhat rather disagree; it's easy for us to accidentally ignore a pull request or assume someone else is looking at it; closing such requests automatically may cause us to inadvertently ignore someone's work. However, I do understand that it can be seen as better to be rejected by a bot than a human. I'm not sure I have a better idea, but I'm concerned that automating rejections may cause problems.

nijel:
It first adds comment and if nothing happens for additional week, the issue is closed. The thresholds can be configured.

You can see example interaction here: https://github.com/WeblateOrg/weblate/pull/1128#issuecomment-308029451

udan11:
I am ok with this.

nijel:
Merged, I think it is really good to have something pushing on stale pull requests as otherwise they will stay not handled for really long.

PR #13402, opened 2017-06-21
Use non-breaking space entity instead of newline in the links on the browse page

ibennetch:
This prevents awkward breaking of the link which would otherwise wrap on small screens like:

[ Create PHP code
]

which looks particularly bad on mobile devices.


ibennetch:
I have this targeting master at the moment because this only really becomes noticeable on mobile screens and Manish's work towards responsive design won't go to QA_4_7. The problem is present in QA_4_7, so it could easily be merged there, but since it's not really breaking anything it seemed less important.

Before:
screen shot 2017-06-19 at 4 28 12 pm

After:
screen shot 2017-06-19 at 4 28 42 pm

nijel:
Merged to QA_4_7 as I think it's better to avoid diverging the code in this case.

PR #13407, opened 2017-06-24
Make grid editor NULL checkbox text clickable

mynetx:
So you can click the text "NULL" in addition to the checkbox itself.


nijel:

PR #13411, opened 2017-06-27
Move classes to PhpMyAdmin namespace

MauricioFauth:
I moved some classes to the `libraries/classes` directory and changed the namespace to `PhpMyAdmin` as discussed in the mailing list.

- Move Advisor to PhpMyAdmin namespace
- Move Bookmark to PhpMyAdmin namespace
- Move Charsets to PhpMyAdmin namespace
- Move Config class to PhpMyAdmin namespace
- Move Console class to PhpMyAdmin namespace
- Move Core to PhpMyAdmin namespace
- Move DatabaseInterface to PhpMyAdmin namespace
- Move DbList to PhpMyAdmin namespace
- Move DbQbe to PhpMyAdmin namespace
- Move DbSearch to PhpMyAdmin namespace
- Move DisplayResults to PhpMyAdmin namespace
- Move Encoding to PhpMyAdmin namespace
- Move Error to PhpMyAdmin namespace
- Move ErrorHandler to PhpMyAdmin namespace
- Move File to PhpMyAdmin namespace
- Move Font to PhpMyAdmin namespace
- Move Footer to PhpMyAdmin namespace
- Move Header to PhpMyAdmin namespace
- Move Index to PhpMyAdmin namespace
- Move IndexColumn to PhpMyAdmin namespace
- Move LanguageManager to PhpMyAdmin namespace
- Move Language to PhpMyAdmin namespace
- Move Linter to PhpMyAdmin namespace
- Move ListAbstract to PhpMyAdmin namespace
- Move ListDatabase to PhpMyAdmin namespace


nijel:
As mentioned on the list, I'd prefer to have separate namespace for the classes shipped with phpMyAdmin and external ones. Having ``PhpMyAdmin\SqlParser`` as external library and ``PhpMyAdmin\Advisor`` as internal sounds confusing. See also https://github.com/phpmyadmin/sql-parser/issues/126

mauriciofauth:
When you commented on the list, I didn't understand that this was your concern.

I understand the second level of the namespace as a kind of module. So I agree that external libraries have two levels. And for that reason too, I think it's okay not to have a sub namespace in the main library.

I think adding sub namespace like `PhpMyAdmin\Core` would end up losing the meaning a bit if used for all internal classes such as `PhpMyAdmin\Core\Plugins` and using `PhpMyAdmin\PhpMyAdmin` would get a bit repetitive. But I still prefer `PhpMyAdmin\Core` instead of `PhpMyAdmin\PhpMyAdmin`.

nijel:
I'm not really sure what's best approach, but I think it would be confusing when looking into the code that some of PhpMyAdmin\* load from libraries/classes, while other from vendor libs...

mauriciofauth:
@nijel, In fact this is very common in projects that modularize their components and it is common to use the same namespace.
When someone looks for `PhpMyAdmin\Advisor`, he will find it in `libraries/classes/Advisor` and when someone looks for `PhpMyAdmin\SqlParser`, as he will not find it in `libraries/classes/SqlParser`, he will find it in `vendor/phpmyadmin/sql-parser`.
This will become clearer when all classes are in a single location.

nijel:
As I've said before - I don't really have strong opinion on this, so having input from other @phpmyadmin/developers is certainly welcome...

ibennetch:
I admit I got a bit lost with the last round of comments.

> When someone looks for `PhpMyAdmin\Advisor`, he will find it in `libraries/classes/Advisor` and when someone looks for `PhpMyAdmin\SqlParser`, as he will not find it in `libraries/classes/SqlParser`, he will find it in `vendor/phpmyadmin/sql-parser`.

I have a slight preference to avoid this, but I'm not clear on what your position is. I also think `PhpMyAdmin\PhpMyAdmin` is repetitive and unnecessarily verbose, and sort of agree that using `PhpMyAdmin\Core` for every internal class is a bit much. So I don't have a good solution right now.

I suppose any good IDE would be able to tell a developer whether the library is in `./libraries/classes/` or `./vendor/phpmyadmin/` so perhaps this is really a non-issue.

mauriciofauth:
Sometimes I get a little bit lost in English, but basically what I meant was what you said, @ibennetch.

My preference is to use `PhpMyAdmin`, but I accept using `PhpMyAdmin\Core`.

nijel:
Okay, let's use just `PhpMyAdmin`...

nijel:
Thanks, merged. And sorry for requesting changes there and back...

PR #13413, opened 2017-06-28
fix brackets

remicollet:
Found testing PHP 7.2
```
Warning in ./libraries/sql.lib.php#615
count(): Parameter must be an array or an object that implements Countable
```

Brackets are obvious wrong (== in the count argument)


nijel:

PR #13420, opened 2017-06-29
Remove double escaping

CommanderRoot:
The href part of the url is already escaped, double escaping like it's done atm breaks the links on for example these buttons:
grafik


mauriciofauth:

PR #13422, opened 2017-06-30
Implemented responsive design

### LINKED ISSUE #6241: Responsive/mobile
Hi,

the world gets more mobile everyday, so I think it would be a good idea to make phpMyAdmin responsive, so it's optimized for smartphones and tablets too - instead of desktop only.

Sometimes I have to wait to edit something in te database, only because i'm not in reach of a desktop PC.

---
- **Original URL**: https://sourceforge.net/p/phpmyadmin/feature-requests/1612/
- **Original author**: roboroad

manishbisht:
Fixes #6241 


nijel:
Sorry for the conflicts, they are caused by merging https://github.com/phpmyadmin/phpmyadmin/pull/13411...

I've just tested it a bit and I've found that resizing window doesn't work properly, I quickly end up with messed up layout such as:

screenshot-2017-6-30 localhost localhost hokus pokus pma__savedsearches phpmyadmin 4 8 0-dev

manishbisht:
@nijel The page has to be refreshed again because the scrolling icon and more icon are loaded once and it is based upon the screen size. and if you move from screen size < 768 to > 768 or vice versa. This problem occurs.

manishbisht:
@nijel So there are two solutions now.
1. To reload the browser using JavaScript when required that is moving from < 768px to > 768px or vice versa.
2. To reduce 768 to a lower number to consider the screen as mobile device. Because I have taken screen < 768 px as a mobile device.

manishbisht:
@nijel I have fixed the issue. Can you have a look into it.

ibennetch:
I'm wondering if we should add a configuration directive to allow users to control the resolution at which the mobile version becomes active.

ibennetch:
Using the standard browser interface with your branch, the "More" dropdown no longer works correctly for me when the menu is larger than the screen.

manishbisht:
Using the configuration directive seems to be useless to me because when the user will going to view on mobile they will never going to check their screen size and put on the config file. And it become more useless when they are using multiple mobile devices.

manishbisht:
Fixed the more drop down not working issue.

nijel:
Can the loading be done a bit less flickery? Recalculating the layout seems to be quite slow:

pma

manishbisht:
@nijel Tried to make it little bit faster. Please have a look into it.

ibennetch:
My configuration directive idea came specifically because I was trying to use multiple devices (a phone and a tablet) and the tablet wasn't triggering the mobile interface. I thought some users would want to see the mobile interface on a tablet while others wouldn't. I agree that the configuration directive might be a bad way of handling it.

I'm still considering whether we could allow users to change the interface themselves in such a case and how that might work (and if it's worthwhile).

manishbisht:
If I am thinking right the configuration directive is a normal variable that that the user set ? If this is correct then when we put it in the configuration directive then we need that variable at two places in JavaScript and CSS files. We can easily use it in CSS as they are with PHP extension but in JavaScript we have to create a Ajax request to get the data of that variable (I am not sure if something better is already implemented to do this kind of work). Then when we have the data of the variable in both files we can just simply replace hard coded number 768 with the received data.

manishbisht:
The tablet may be not triggering the mobile interface if the screen width is less than 768 px.

ibennetch:
I'm pretty sure I responded to this, but don't see my response posted, so that's weird.

Anyway, yes, the configuration directive I was thinking of would be part of `config.inc.php` so the user could define what size they'd want the mobile display to begin. For instance, some users might want their tablet to  use the mobile display and others might want the tablet to have the desktop display.

I think I'm over-thinking it, though.

Is there an easy way to allow the user to toggle between interfaces? Like when viewing Github on mobile, there's a link at the very bottom to switch to desktop view.

js/menu-resizer.js
```diff
@@ -24,6 +24,30 @@
...
         self.widthCalculator = widthCalculator;
+        var windowWidth = $(window).width();
+
+        if (windowWidth < 768) {
+            $('#pma_navigation_resizer').css({'width': '0px'});
+        }
+        // Sets the image for the left and right scroll indicator
+        $('.scrollindicator--left').html($(PMA_getImage('b_left.png').toString()));
+        $('.scrollindicator--right').html($(PMA_getImage('b_right.png').toString()));
+        //$(PMA_getImage('b_left.png').toString()).prependTo($('.scrollindicator--left'));
```
ibennetch: Is this line and the next leftover debugging code?

js/menu-resizer.js
```diff
@@ -38,17 +62,17 @@
             .mouseenter(function() {
                 if ($(this).find('ul .tabactive').length === 0) {
                     $(this)
-                    .addClass('submenuhover')
-                    .find('> a')
-                    .addClass('tabactive');
+                        .addClass('submenuhover')
```
ibennetch: I believe these three lines are indented one level too far

js/menu-resizer.js
```diff
@@ -38,17 +62,17 @@
...
+                        .addClass('tabactive');
                 }
             })
             .mouseleave(function() {
                 if ($(this).find('ul .tabactive').length === 0) {
                     $(this)
-                    .removeClass('submenuhover')
-                    .find('> a')
-                    .removeClass('tabactive');
+                        .removeClass('submenuhover')
```
ibennetch: I believe the indentation is too much here as well.

js/navigation.js
```diff
@@ -1271,8 +1284,11 @@ var ResizeHandler = function () {
         // Set content bottom space beacuse of console
         $('body').css('margin-bottom', $('#pma_console').height() + 'px');
     };
-    /* Initialisation section begins here */
-    if (Cookies.get('pma_navi_width')) {
+    // Hide the pma_navigation initially when loaded on mobile
```
ibennetch: This works well; I like the page loading with navigation hidden.

libraries/Util.php
```diff
@@ -1722,15 +1722,19 @@ public static function getHtmlTabs(
 
         $tab_navigation = '<div id="' . htmlentities($menu_id)
             . 'container" class="menucontainer">'
-            . '<ul id="' . htmlentities($menu_id) . '" ' . $class . '>';
+            . '<i class="scrollindicator scrollindicator--left"><a href="#" class="tab"></a></i>'
+            . '<div class="navigationbar"><ul id="' . htmlentities($menu_id) . '" ' . $class . '>';
 
+        //$tab_navigation .= '<span class="scrollindicator--left" style="padding: 0;position: absolute;"><a href="#" class="tab"><img src="themes/pmahomme/img/b_left.png" title="Databases" alt="Databases"></a></span>';
```
ibennetch: Is this commented out line needed any longer?

themes/pmahomme/css/common.css.php
```diff
@@ -45,6 +45,18 @@
     margin: 0 .5em;
 }
 
+.desktop50 {
```
ibennetch: Did you also make all these CSS changes to 'original' or just 'pmahomme'?

ibennetch (changes requested):
Overall this is great work. It's a lot of changes to many of the files in the application and you've done nice work on it.

ibennetch:
@nijel I'm preparing to merge this, but wonder if you think I should squash all the commits in to one? It's 50 files +601 -328 lines of code, so I'm thinking that it's a lot of code for one commit, but the work is all dependent on each other and related so is probably easier to read in the history as one commit rather than 43. I'm leaning towards squashing it because then it's all together in one place in the history.

manishbisht:
Yes, The other better option I think is to set the cookie instead of configuration directive. And using that cookie we can check what display to show the mobile one or the desktop one. What you think of this option ?

js/menu-resizer.js
manishbisht: Yes, It was used for debugging. I removed it in new commit.

js/menu-resizer.js
manishbisht: Yes, Fixed the issue in new commit.

js/menu-resizer.js
manishbisht: Fixed the issue in new commit.

js/navigation.js
manishbisht: Thanks :)

themes/pmahomme/css/common.css.php
manishbisht: No, Right now it works only on `pmahomme`. Should I make the same changes to 'original' also ?
 There is one theme repository so the themes in the phpmyadmin are auto updated as soon as the themes repository is updated or we have to manually update the phpmyadmin themes. If they are auto updated then the good option will be make the changes in the theme repository.

ibennetch:
I like the cookie idea better than making a configuration directive.

themes/pmahomme/css/common.css.php
manishbisht: Made the css changes in original theme also.

nijel:
I think that having configuration directive which can be toggled in user settings as well (and maybe even directly on every page) is best approach. You can reuse existing framework for persistence (database stored / browser local storage / session) this way. Also see https://github.com/phpmyadmin/phpmyadmin/issues/11688 for problems with too many cookies.

@ibennetch as for merging, I'm fine with merging this as is or squashed. Both have it's goods (either showing how the changes have evolved or having all responsive changes in one changeset) and I really don't see either of approaches being substantively better.

manishbisht:
Localstorage also seems to be a good option instead of cookies.

nijel:
As for the configuration storage, I think it's time to provide one way to do that and avoid implementing it for every setting we have, see https://github.com/phpmyadmin/phpmyadmin/issues/13466

manishbisht:
The best option I can think of is to store it in JSON format and it will contain 2 objects. One will be mode that can be either desktop or mobile (this can further be used to move from desktop to mobile and mobile to desktop) and the other part will be width on which the view is triggered.
It seems like we have to think more during implementation as it may happen that when user see on mobile the mobile view is loaded but what if here user wants the desktop view on his mobile.

The other option is to do either based on width or display mode and trigger the view based on their values.

ibennetch (approved):

ibennetch:
My preference is to not store any resolution/size number; instead we can simply automatically present the mobile version if the size is under a threshold, like you've got now. I'm not against making a variable for this, but I don't see much of a point.

We can present a button/link at the very bottom of the page to "Switch to desktop view" which I think should be per device, so a cookie or local browser storage, rather than per user (a single user may visit from many devices, and having this setting follow them from phone to desktop as a default is probably not what we want).

On the desktop view we might even add a link in the "Appearance settings" area to "Switch to mobile view"

That's my two cents, what do you all think of that?

ibennetch:
I've merged this, we can continue improvements as a separate pull request as needed. Great work!

PR #13423, opened 2017-06-30
Properly fix #13420

CommanderRoot:
I edited the wrong row in GitHub after fixing #13420 in my local installation and testing that it worked. My bad. This is the correct fix for the double escaping of the href part of the url.


mauriciofauth:
When I checked locally if using raw filter was the best option, I checked the `href` key. So the mistake was also mine.

mauriciofauth:
Thanks for your contribution!

PR #13446, opened 2017-07-05
removed inline onclick javascript

### LINKED ISSUE #12261: Remove inline javascript

manishbisht:
Fixes #12261

### DIFF
--- a/js/pmd/init.js
+++ b/js/pmd/init.js
@@ -37,4 +37,24 @@ AJAX.registerOnload('pmd/init.js', function () {
             }
         });
     }
+
+    $('#query_Aggregate_Button').click(function(){
+        document.getElementById('query_Aggregate').style.display = 'none';
+    });
+
+    $('#query_having_button').click(function(){
+        document.getElementById('query_having').style.display = 'none';
+    });
+
+    $('#query_rename_to_button').click(function(){
+        document.getElementById('query_rename_to').style.display = 'none';
+    });
+
+    $('#build_query_button').click(function(){
+        build_query('SQL Query on Database', 0)
+    });
+
+    $('#query_where_button').click(function(){
+        document.getElementById('query_where').style.display = 'none';
+    });
 });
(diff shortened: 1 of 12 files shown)


manishbisht:
There is the small mistake in removing the inline css in `templates/database/structure/check_all_tables.phtml`

What does this `unMarkAllRows('tablesForm'); <?= $overhead_check; ?> return false;` line actually do ? means what is the value `$overhead_check;` stores and what action is performed based on the value of `$overhead_check;`

ibennetch:
So in libraries/controllers/database/DatabaseStructureController.php on line 442, when there is overhead present on a table, `$overhead_check` is set to some JavaScript query (`markAllRows('row_tbl_1');` for the first row as an example). The function is set and the row_tbl_ number is used for any row having the overhead; so on one of my test databases where the 8th and 16th tables have overhead, $overhead_check is set to `markAllRows('row_tbl_8');markAllRows('row_tbl_16');`.

`markAllRows()` is a function that simply selects the row indicated by the argument and corresponds to the id name of that `tr` (`tr id="tbl_row_10`, for instance).

So this bit of code you asked about first unselects all of the rows in the table with the unMarkAllRows() JavaScript function, then calls markAllRows() for each table having overhead (specifically, the JavaScript call for markAllRows() is added inline in DatabaseStructureController.php.

My gut feeling is that you won't be able to remove the JavaScript from being inline here specifically, but I haven't really thought it through.

manishbisht:
Yes, we will not be able to remove inline onclick from here because the variable directly contains the JavaScript functions. I have fixed the PR accordingly.

The other option that I can see here is to put the table id separated by , in the `$overhead_check` and then in JavaScript we can split the values by , and execute the `markAllRows()` on the table id's.

ibennetch:
I think the test error here is just from problems that were in master and have since been resolved; could you rebase your branch and push again (`git rebase master` from within the branch)?

manishbisht:
Rebased the branch

ibennetch:
Great, nice work.

PR #13458, opened 2017-07-08
Refactored view_create to use Twig templates.

udan11:

### DIFF
--- a/templates/view_create.twig
+++ b/templates/view_create.twig
@@ -1,0 +1,123 @@
+<!-- CREATE VIEW options -->
+<div id="div_view_options">
+    <form method="post" action="view_create.php">
+    {{ URL_getHiddenInputs(url_params) }}
+    <fieldset>
+        <legend>
+            {% if ajax_dialog %}
+                {% trans 'Details' %}
+            {% else %}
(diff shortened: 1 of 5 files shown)


templates/view_create.twig
```diff
@@ -0,0 +1,122 @@
...
+                    </td>
+                </tr>
+            {% endif %}
+
+            <tr>
+                <td class="nowrap"><label for="algorithm">ALGORITHM</label></td>
+                <td>
+                    <select name="view[algorithm]" id="algorithm">
+                        {% for option in view_algorithm_options %}
+                            <option value="{{ option|e }}"
```
mauriciofauth: You don't need to use the escape filter because it's enabled by default.

udan11: Is it the case for every use of the filter?

I could not find any documentation about this.

mauriciofauth: The `autoescape` [environment option](https://twig.sensiolabs.org/doc/1.x/api.html#environment-options) is enabled by default. So when you do not want to escape, you need to use the [raw filter](https://twig.sensiolabs.org/doc/1.x/filters/raw.html).

nijel:

PR #13461, opened 2017-07-09
Remember status of PMA's navigation bar.

### LINKED ISSUE #13397: hide side menu in standard
### Expected behaviour
when login, hide side menu in standard

### Actual behaviour
1. when login, view side menu.

2. and, I click on this button. 

### Server configuration
**Operating system**:
- windows 10 pro(1703, 64bit)
- xampp 7.1.2-0

**Web server:**
- Apache/2.4.25 (Win32) 

**Database:**
- libmysql - mysqlnd 5.0.12-dev - 20150407 - $Id: b396954eeb2d1d9ed7902b8bae237b287f21ad9e

**PHP version:**
- 7.1.2

**phpMyAdmin version:**
- 4.7.1

### Client configuration
**Browser:**
- chrome(59.0.3071.104, 64bit)

**Operating system:**
- windows 10 pro(1703, 64bit)

udan11:
Fixes #13397.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -24,6 +24,7 @@ phpMyAdmin - ChangeLog
 - issue #13095 Fixed alignmnet of foreign keys editing
 - issue #12944 Improved inline editor for JSON
 - issue #13145 Improved layout of operations pages
+- issue #13397 Remember navigation menu's state
 
 4.7.3 (not yet released)
 - issue #13447 Large multi-line query removes Export operation and blanks query box options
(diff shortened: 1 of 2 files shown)


nijel:
I really think we should not use one cookie per setting. This really doesn't scale. What I'd really prefer is to have some generic solution to store such configuration as there will be probably more things where such persistence makes sense. Preferably having this in the user configuration and having API to change that in javascript, I've created https://github.com/phpmyadmin/phpmyadmin/issues/13466 for this.

See also https://github.com/phpmyadmin/phpmyadmin/pull/13422#issuecomment-312433334 and https://github.com/phpmyadmin/phpmyadmin/issues/11688.

PR #13462, opened 2017-07-09
Fix Database-related selenium tests

devenbansod:
Fix Database-related (and CreateRemoveUser) selenium tests.

Selenium Tests from the following files should be fixed:
* CreateDropDatabase
* CreateRemoveUser
* DbEvents
* DbOperations
* DbProcedures
* DbStructure
* DbTriggers
 
### DIFF
--- a/test/selenium/PmaSeleniumCreateRemoveUserTest.php
+++ b/test/selenium/PmaSeleniumCreateRemoveUserTest.php
@@ -69,9 +69,17 @@ class PMA_SeleniumCreateRemoveUserTest extends PMA_SeleniumBase
 
         $this->scrollIntoView('add_user_anchor');
         $this->waitForElement('byId', 'usersForm');
-        usleep(1000000);
-        $this->waitForElement("byId", "add_user_anchor")->click();
-        usleep(1000000);
+        $ele = $this->waitForElement("byId", "add_user_anchor");
+        $this->moveto($ele);
+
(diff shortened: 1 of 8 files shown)


test/selenium/PmaSeleniumCreateRemoveUserTest.php
```diff
@@ -56,16 +62,18 @@ public function setUp()
...
         $this->waitForElement('byPartialLinkText', "User accounts")->click();
 
         // Let the User Accounts page load
         $this->waitForElementNotPresent('byId', 'ajax_message_num_1');
 
         $this->scrollIntoView('add_user_anchor');
-        $link = $this->waitForElement("byId", "add_user_anchor");
-        $link->click();
+        $this->waitForElement('byId', 'usersForm');
+        usleep(1000000);
```
udan11: Why is this `usleep` required?

devenbansod: The tests were actually running perfectly fine on my local setting without it. But while on Browserstack, they were failing without this. I have just added a second commit to try and do this in a different way.

nijel:
Merged, though I tend to agree with @udan11 that using sleep is not really best approach in most cases (on the other side, I've added bunch of them as well to solve some timing issues, so I know that they can be tricky to get rid of).

PR #13495, opened 2017-07-16
Add 2-factor authentication support for phpMyAdmin.

Achilles-96:
Signed-off-by: Raghuram <raghuram.4350@gmail.com>

### DIFF
--- a/libraries/session.inc.php
+++ b/libraries/session.inc.php
@@ -1,0 +1,186 @@
+<?php
+/* vim: set expandtab sw=4 ts=4 sts=4: */
+/**
+ * session handling
+ *
+ * @todo    add an option to use mm-module for session handler
+ *
+ * @package PhpMyAdmin
+ * @see     https://secure.php.net/session
+ */
+
+use PhpMyAdmin\Core;
+
+if (! defined('PHPMYADMIN')) {
+    exit;
+}
+
+require_once 'libraries/session.lib.php';
+
+// verify if PHP supports session, die if it does not
+
+if (!@function_exists('session_name')) {
+    Core::warnMissingExtension('session', true);
+} elseif (! empty(ini_get('session.auto_start')) && session_name() != 'phpMyAdmin' && !empty(session_id())) {
+    // Do not delete the existing non empty session, it might be used by other
+    // applications; instead just close it.
+    if (empty($_SESSION)) {
+        /* Ignore errors as this might have been destroyed in other request meanwhile */
+        @session_destroy();
+    } elseif (function_exists('session_abort')) {
+        /* PHP 5.6 and newer */
+        session_abort();
+    } else {
+        session_write_close();
+    }
+}
+
+// disable starting of sessions before all settings are done
+// does not work, besides how it is written in php manual
+//ini_set('session.auto_start', '0');
+
+// session cookie settings
+session_set_cookie_params(
+    0, $GLOBALS['PMA_Config']->getRootPath(),
+    '', $GLOBALS['PMA_Config']->isHttps(), true
+);
+
+// cookies are safer (use @ini_set() in case this function is disabled)
+@ini_set('session.use_cookies', 'true');
+
+// optionally set session_save_path
+$path = $GLOBALS['PMA_Config']->get('SessionSavePath');
+if (!empty($path)) {
+    session_save_path($path);
+}
+
+// use cookies only
+@ini_set('session.use_only_cookies', '1');
+// strict session mode (do not accept random string as session ID)
+@ini_set('session.use_strict_mode', '1');
+// make the session cookie HttpOnly
+@ini_set('session.cookie_httponly', '1');
+// do not force transparent session ids
+@ini_set('session.use_trans_sid', '0');
+
+// delete session/cookies when browser is closed
+@ini_set('session.cookie_lifetime', '0');
+
+// warn but don't work with bug
+@ini_set('session.bug_compat_42', 'false');
+@ini_set('session.bug_compat_warn', 'true');
+
+// use more secure session ids
+@ini_set('session.hash_function', '1');
+
+// some pages (e.g. stylesheet) may be cached on clients, but not in shared
+// proxy servers
+session_cache_limiter('private');
+
+// start the session
+// on some servers (for example, sourceforge.net), we get a permission error
+// on the session data directory, so I add some "@"
+
+/**
+ * Session failed function
+ *
+ * @param array $errors PhpMyAdmin\ErrorHandler array
+ *
+ * @return void
+ */
+function PMA_sessionFailed($errors)
+{
+    $messages = array();
+    foreach ($errors as $error) {
+        /*
+         * Remove path from open() in error message to avoid path disclossure
+         *
+         * This can happen with PHP 5 when nonexisting session ID is provided,
+         * since PHP 7, session existence is checked first.
+         *
+         * This error can also happen in case of session backed error (eg.
+         * read only filesystem) on any PHP version.
+         *
+         * The message string is currently hardcoded in PHP, so hopefully it
+         * will not change in future.
+         */
+        $messages[] = preg_replace(
+            '/open\(.*, O_RDWR\)/',
+            'open(SESSION_FILE, O_RDWR)',
+            htmlspecialchars($error->getMessage())
+        );
+    }
+
+    /*
+     * Session initialization is done before selecting language, so we
+     * can not use translations here.
+     */
+    Core::fatalError(
+        'Error during session start; please check your PHP and/or '
+        . 'webserver log file and configure your PHP '
+        . 'installation properly. Also ensure that cookies are enabled '
+        . 'in your browser.'
+        . '<br /><br />'
+        . implode('<br /><br />', $messages)
+    );
+}
+
+// See bug #1538132. This would block normal behavior on a cluster
+//ini_set('session.save_handler', 'files');
+
+$session_name = 'phpMyAdmin';
+@session_name($session_name);
+
+// Restore correct sesion ID (it might have been reset by auto started session
+if (isset($_COOKIE['phpMyAdmin'])) {
+    session_id($_COOKIE['phpMyAdmin']);
+}
+
+// on first start of session we check for errors
+// f.e. session dir cannot be accessed - session file not created
+if (!isset($_SESSION)){
+    $orig_error_count = $GLOBALS['error_handler']->countErrors(false);
+
+    $session_result = session_start();
+
+    if ($session_result !== true
+    ) {
+        setcookie($session_name, '', 1);
+        $errors = $GLOBALS['error_handler']->sliceErrors($orig_error_count);
+        PMA_sessionFailed($errors);
+    }
+    unset($orig_error_count, $session_result);
+}
+
+/**
+ * Disable setting of session cookies for further session_start() calls.
+ */
+@ini_set('session.use_cookies', 'true');
+
+/**
+ * Token which is used for authenticating access queries.
+ * (we use "space PMA_token space" to prevent overwriting)
+ */
+if (empty($_SESSION[' PMA_token '])) {
+    PMA_generateToken();
+
+    /**
+     * Check for disk space on session storage by trying to write it.
+     *
+     * This seems to be most reliable approach to test if sessions are working,
+     * otherwise the check would fail with custom session backends.
+     */
+    $orig_error_count = $GLOBALS['error_handler']->countErrors();
+    session_write_close();
+    if ($GLOBALS['error_handler']->countErrors() > $orig_error_count) {
+        $errors = $GLOBALS['error_handler']->sliceErrors($orig_error_count);
+        PMA_sessionFailed($errors);
+    }
+    session_start();
+    if (empty($_SESSION[' PMA_token '])) {
+        Core::fatalError(
+            'Failed to store CSRF token in session! ' .
+            'Probably sessions are not working properly.'
+        );
+    }
+}
--- a/setup2FA.php
+++ b/setup2FA.php
@@ -1,0 +1,164 @@
+<?php
+/* vim: set expandtab sw=4 ts=4 sts=4: */
+/**
+ * query by example the whole database
+ *
+ * @package PhpMyAdmin
+ */
+
+$session_name = 'phpMyAdmin';
+@session_name($session_name);
+session_start();
+
+require_once 'vendor/autoload.php';
+
+use RobThree\Auth\Providers\Qr\IQRCodeProvider;
+use PHPQRCode;
+
+class PMAQRProvider implements IQRCodeProvider {
+  public function getMimeType() {
+    return 'image/png';                             // This provider only returns PNG's
+  }
+
+  public function getQRCodeImage($qrtext, $size) {
+    ob_start();                                     // 'Catch' QRCode's output
+    PHPQRCode\QRcode::png($qrtext, null, PHPQRCode\Constants::QR_ECLEVEL_L, 3, 4);
+                                                    // We ignore $size and set it to 3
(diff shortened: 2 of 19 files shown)


Achilles-96:
I described the procedure of setting up in my [blog post](https://raghuramvadapalli.wordpress.com/2017/07/14/2-factor-authentication-for-phpmyadmin/).

As I mentioned there, I need advice on where to place link to access `setup2FA.php`.

js/get_scripts.js.php
```diff
@@ -7,6 +7,8 @@
  * @package PhpMyAdmin
  */
 
+/* var_dump($_GET['scripts']); */
+/* exit; */
```
nijel: I guess this is not intentional...

libraries/phpqrcode.php
```diff
@@ -0,0 +1,3625 @@
+<?php
+
+/*
+ * PHP QR Code encoder
+ *
+ * This file contains MERGED version of PHP QR Code library.
+ * It was auto-generated from full version for your convenience.
```
nijel: Please do not include third party libraries, can't this be loaded using composer? Also including generated code without source doesn't really play well with GPL, so we would have to distribute the source code as well.

libraries/session.inc.php
```diff
@@ -138,18 +138,19 @@ function PMA_sessionFailed($errors)
 
 // on first start of session we check for errors
 // f.e. session dir cannot be accessed - session file not created
-$orig_error_count = $GLOBALS['error_handler']->countErrors(false);
+if (!isset($_SESSION)){
```
nijel: What is purpose of this change? If really needed, please do that in separate commit explaining why it is needed.

libraries/plugins/auth/AuthenticationCookie.php
```diff
@@ -70,6 +71,94 @@ public function setUseOpenSSL($use)
...
+        }
+        $controllink = $GLOBALS['dbi']->connect(
+            DatabaseInterface::CONNECT_CONTROL
+        );
+        if (!$GLOBALS['dbi']->selectDb('phpmyadmin', $controllink)) {
+            $conn_error = $GLOBALS['dbi']->getError($controllink);
+            PMA_secureSession();
+            $this->auth();
+        }
+        $check_2FA_db_query = 'SELECT `pma_user`, `secret` FROM `pma__2fa_secrets` WHERE `pma_user` LIKE \'' . $GLOBALS['PHP_AUTH_USER'] . '\';';
```
nijel: * Please do not hardcode table name, see how others are handled.
* Why using LIKE when doing exact match?

libraries/plugins/auth/AuthenticationCookie.php
```diff
@@ -70,6 +71,94 @@ public function setUseOpenSSL($use)
...
+        $user = $GLOBALS['PHP_AUTH_USER'];
+
+        global $conn_error;
+        if (! empty($conn_error)) {
+            PMA_secureSession();
+            $this->auth();
+        }
+        $controllink = $GLOBALS['dbi']->connect(
+            DatabaseInterface::CONNECT_CONTROL
+        );
```
nijel: I think this can fail, so you should handle false return value here.

libraries/plugins/auth/AuthenticationCookie.php
```diff
@@ -449,6 +538,48 @@ public function authCheck()
...
+    {
+        global $conn_error;
+        $controllink = $GLOBALS['dbi']->connect(
+            DatabaseInterface::CONNECT_CONTROL
+        );
+        if (!$GLOBALS['dbi']->selectDb('phpmyadmin', $controllink)) {
+            $conn_error = $GLOBALS['dbi']->getError($controllink);
+            return false;
+        }
+        $check_2FA_db_query = 'SELECT `pma_user` FROM `pma__2fa_secrets` WHERE `pma_user` LIKE \'' . $GLOBALS['PHP_AUTH_USER'] . '\';';
```
nijel: Same as with auth2FA. Also these seem to share quite some amount of code, so you might want to factor that out.

setup2FA.php
```diff
@@ -0,0 +1,149 @@
...
+            $add_secret_query = 'INSERT INTO pma__2fa_secrets VALUES (\'' . $GLOBALS['PHP_AUTH_USER'] . '\' ,\'' . $_SESSION['secret'] . '\');';
+            if (!$GLOBALS['dbi']->tryQuery($add_secret_query, $GLOBALS['controllink'])) {
+                echo $GLOBALS['dbi']->getError($GLOBALS['controllink']);
+                exit;
+            }
+        }
+        if (isset($_REQUEST['logincheck']) && $_REQUEST['logincheck'] == 'true') {
+            $_SESSION['2FAVerified'] = true;
+        }
+        echo 'true';
```
nijel: Really?

setup2FA.php
```diff
@@ -0,0 +1,149 @@
...
+    } else {
+        echo 'Sorry. The key does not match. Please ensure that server time and your app\'s time are in sync.' . $_SESSION['secret'];
```
nijel: Should be localized, also there is probably no point in showing the secret here. Especially not unescaped as it seems to be user input.

setup2FA.php
```diff
@@ -0,0 +1,149 @@
...
+    }
+    exit;
+}
+
+if (isset($_REQUEST['delete_key'])) {
+    require_once 'libraries/common.inc.php';
+    $response = PhpMyAdmin\Response::getInstance();
+    if (!$GLOBALS['dbi']->selectDb('phpmyadmin', $GLOBALS['controllink'])) {
+        $response->disable();
+        echo $GLOBALS['dbi']->getError($GLOBALS['controllink']);
```
nijel: In case you want to pass the error details to client, please use JSON. Definitely do not output unescaped error messages.

nijel (changes requested):
I've quickly gone through your code, some general remarks:

* Please include documentation, having it in the blog is great, but it really should be part of our documentation.
* Any reason for not integrating 2FA into existing Settings tab?

Achilles-96:
@nijel, I addressed most of the issues you raised. I left comments for the rest. Please verify.

setup2FA.php
```diff
@@ -0,0 +1,154 @@
...
+            if (!$GLOBALS['dbi']->tryQuery($add_secret_query, $GLOBALS['controllink'])) {
+                echo htmlspecialchars($GLOBALS['dbi']->getError($GLOBALS['controllink']));
+                exit;
+            }
+        }
+        $_SESSION['2FAEnabled'] = true;
+        $_SESSION['2FAVerified'] = true;
+        echo 'true';
+    } else {
+        echo htmlspecialchars('Sorry. The key does not match. Please ensure that server time and your app\'s time are in sync.');
```
Achilles-96: By this time, translations may not be possible.

ibennetch: Could you please add a comment to the code as well stating that translations may not be possible?

setup2FA.php
```diff
@@ -0,0 +1,154 @@
...
+                DEFAULT CHARACTER SET utf8 COLLATE utf8_bin;';
+            if (!$GLOBALS['dbi']->selectDb('phpmyadmin', $GLOBALS['controllink'])) {
+                echo htmlspecialchars($GLOBALS['dbi']->getError($GLOBALS['controllink']));
+                exit;
+            }
+            if (!$GLOBALS['dbi']->tryQuery($create_table_query, $GLOBALS['controllink'])) {
+                echo htmlspecialchars($GLOBALS['dbi']->getError($GLOBALS['controllink']));
+                exit;
+            }
+            $add_secret_query = 'INSERT INTO ' . $cfg_2fa_table . ' VALUES (\'' . $GLOBALS['PHP_AUTH_USER'] . '\' ,\'' . $_SESSION['secret'] . '\');';
```
ibennetch: Forgive me if I missed you doing it earlier, but I think you need to escape PHP_AUTH_USER and secret here; you could use the Dbi function `escapeString` for that.

ibennetch:
I tried opening my main phpMyAdmin after pulling your branch (without configuring any user for 2FA or setting up config.inc.php) and got a number of errors. We should be more graceful about failing if the user hasn't configured their system for this.

> Error
> SQL query: DocumentationEdit Edit
> 
> 
> SELECT `pma_user` FROM `` WHERE `pma_user` = 'root';
> MySQL said: Documentation
> 
> #1103 - Incorrect table name ''
> Open new phpMyAdmin window
> Notice in ./libraries/plugins/auth/AuthenticationCookie.php#570
>  Undefined index: 2fa_secrets
> 
> Backtrace
> 
> ./libraries/plugins/auth/AuthenticationCookie.php#547: PMA\libraries\plugins\auth\AuthenticationCookie->check2FA()
> ./libraries/common.inc.php#656: PMA\libraries\plugins\auth\AuthenticationCookie->authCheck()
> ./index.php#22: require_once(./libraries/common.inc.php)
> Warning in ./libraries/dbi/DBIMysqli.php#556
>  mysqli_real_escape_string() expects parameter 1 to be mysqli, boolean given
> 
> Backtrace
> 
> ./libraries/dbi/DBIMysqli.php#556: mysqli_real_escape_string(
> boolean false,
> string '_',
> )
> ./libraries/classes/DatabaseInterface.php#2743: PMA\libraries\dbi\DBIMysqli->escapeString(
> boolean false,
> string '_',
> )
> ./libraries/navigation/nodes/Node.php#437: PhpMyAdmin\DatabaseInterface->escapeString(string '_')
> ./libraries/navigation/NavigationTree.php#289: PMA\libraries\navigation\nodes\Node->getData(
> string 'databases',
> integer 0,
> string '',
> )
> ./libraries/navigation/NavigationTree.php#847: PMA\libraries\navigation\NavigationTree->_buildPath()
> ./libraries/navigation/Navigation.php#45: PMA\libraries\navigation\NavigationTree->renderState()
> ./libraries/classes/Header.php#433: PMA\libraries\navigation\Navigation->getDisplay()
> ./libraries/classes/Response.php#266: PhpMyAdmin\Header->getDisplay()
> ./libraries/classes/Response.php#279: PhpMyAdmin\Response->_getDisplay()
> ./libraries/classes/Response.php#438: PhpMyAdmin\Response->_htmlResponse()
> PhpMyAdmin\Response->response()
> Warning in ./libraries/dbi/DBIMysqli.php#556
>  mysqli_real_escape_string() expects parameter 1 to be mysqli, boolean given
> 
> Backtrace
> 
> ./libraries/dbi/DBIMysqli.php#556: mysqli_real_escape_string(
> boolean false,
> string '_',
> )
> ./libraries/classes/DatabaseInterface.php#2743: PMA\libraries\dbi\DBIMysqli->escapeString(
> boolean false,
> string '_',
> )
> ./libraries/navigation/nodes/Node.php#447: PhpMyAdmin\DatabaseInterface->escapeString(string '_')
> ./libraries/navigation/NavigationTree.php#289: PMA\libraries\navigation\nodes\Node->getData(
> string 'databases',
> integer 0,
> string '',
> )
> ./libraries/navigation/NavigationTree.php#847: PMA\libraries\navigation\NavigationTree->_buildPath()
> ./libraries/navigation/Navigation.php#45: PMA\libraries\navigation\NavigationTree->renderState()
> ./libraries/classes/Header.php#433: PMA\libraries\navigation\Navigation->getDisplay()
> ./libraries/classes/Response.php#266: PhpMyAdmin\Header->getDisplay()
> ./libraries/classes/Response.php#279: PhpMyAdmin\Response->_getDisplay()
> ./libraries/classes/Response.php#438: PhpMyAdmin\Response->_htmlResponse()
> PhpMyAdmin\Response->response()
> Warning in ./libraries/dbi/DBIMysqli.php#556
>  mysqli_real_escape_string() expects parameter 1 to be mysqli, boolean given
> 
> Backtrace
> 
> ./libraries/dbi/DBIMysqli.php#556: mysqli_real_escape_string(
> boolean false,
> string '_',
> )
> ./libraries/classes/DatabaseInterface.php#2743: PMA\libraries\dbi\DBIMysqli->escapeString(
> boolean false,
> string '_',
> )
> ./libraries/navigation/nodes/Node.php#449: PhpMyAdmin\DatabaseInterface->escapeString(string '_')
> ./libraries/navigation/NavigationTree.php#289: PMA\libraries\navigation\nodes\Node->getData(
> string 'databases',
> integer 0,
> string '',
> )
> ./libraries/navigation/NavigationTree.php#847: PMA\libraries\navigation\NavigationTree->_buildPath()
> ./libraries/navigation/Navigation.php#45: PMA\libraries\navigation\NavigationTree->renderState()
> ./libraries/classes/Header.php#433: PMA\libraries\navigation\Navigation->getDisplay()
> ./libraries/classes/Response.php#266: PhpMyAdmin\Header->getDisplay()
> ./libraries/classes/Response.php#279: PhpMyAdmin\Response->_getDisplay()
> ./libraries/classes/Response.php#438: PhpMyAdmin\Response->_htmlResponse()
> PhpMyAdmin\Response->response()

ibennetch:
We'll have to deal gracefully with situations where the user either hasn't configured it or the table doesn't exist.

You can also see for instance libraries/classes/Bookmark.php:114 for an example of how we might deal with this. We have to test the array to make sure $cfgRelation['2fa_secret'] (or whatever configuration directive you use) isn't empty first, before trying to use it. We initialize it as part of the default configuration to an empty value, so before using we'll test if the user has configured it.

Also, you should add the new directive to libraries/config.default.php with a blank assignment such as $cfg['Servers'][$i]['2fa_secret'] = '';. You can add it right after the export_template area around line 503.

We should also add this new directive to libraries/relation.lib.php around line 353; this file is used by chk_rel.php which warns the user if their table structure is outdated.

Can you add the structure for the 2fa_secrets table to `sql/create_tables.sql`?

Instead of manually running the SQL, it might be more efficient to use PMA_queryAsControlUser such as
```
$sql_query
    = " SELECT `tables` FROM " . $this->_getPmaTable() .
    " WHERE `username` = '" . $GLOBALS['dbi']->escapeString($GLOBALS['cfg']['Server']['user']) . "'";
$result = PMA_queryAsControlUser($sql_query, false);
```

setup2FA.php
nijel: Hmm, maybe this should be part of the setup script in the end (which we indeed do not localize)?

setup2FA.php
```diff
@@ -0,0 +1,154 @@
...
+$session_name = 'phpMyAdmin';
+@session_name($session_name);
+session_start();
+
+require_once 'vendor/autoload.php';
+
+use RobThree\Auth\Providers\Qr\IQRCodeProvider;
+use PHPQRCode;
+
+class PMAQRProvider implements IQRCodeProvider {
```
nijel: Can you please place the class to separate file and include tests for it?

setup2FA.php
Achilles-96: By this time, user would not have completely authenticated i.e., it is in the middle of authentication. Including common.inc.php is not possible because it will circularly check for 2-factor authentication again.

This part is accessed only by ajax request and the error message is displayed in standard PMA error dialog once received.

nijel: Maybe it would be better to closer integrate this to the authentication so that it can use the authentication API to show the form

Achilles-96: Authentication form is shown by `auth2FA` function in `AuthenticationCookie.php`. This is for verification. Do you mean I should move verification to a different file.

libraries/session.inc.php
ibennetch: @nijel are you satisfied with this change?

nijel: I'd prefer to have the documentation in the code as well, not only in the commit message, but thanks for documenting this.

ibennetch:
Could you rebase this against current master to reflect the changes Maurício has done?

Issue #12401, opened 2016-07-23
Import progress bar


PatchRanger:
### Description
Similar and related to #6311 : could we have a progress bar for import task?
It would be great to split the entire job into several steps and track their progress separately:
- Upload a file
- Unzip it
- Import into a database (including how many tables and rows have been already processed and which is in progress right now)

It could also help to workaround server limitations related to maximum execution time - as we could handle it with continuous AJAX polling until finish.

nijel:
The progress bar is there: https://docs.phpmyadmin.net/en/latest/faq.html#seeing-an-upload-progress-bar

PatchRanger:
Thank you for response. It is upload progress bar while I mean import progress bar. It includes upload progress bar but also should address other parts of import process: upload, unzip and import. Right now it is done without any progress tracking, just similar to "Please wait" stub. It would be great to have a progress bar for import task: how many tables (and even rows) are already imported and how many still pending.

manishbisht:
Anyone have the larger database that takes some time to import so that I can do testing on that database.

ibennetch:
I suggest the 'employee data' database from https://dev.mysql.com/doc/index-other.html. It becomes rather large (mine claims to be 194 MiB). 'sakila' is another good choice but only around 6 MiB.

manishbisht:
This is how I am going to trying to approach for this.
1. During export I will set some data in session to track the upload progress.
2. Create a new php file that return the upload data from the session.
3. Using setTimeout to regularly get data from php file and update the progress bar.

Problem:
When I tried to access session using
```
<?php
session_start();
print_r($_SESSION);
?>
```
It always gives me the empty array. I think this might be due to some security. How can I access the session data in my new file.

ibennetch:
Is something like

```
use PhpMyAdmin\Response;
require_once 'libraries/common.inc.php';
$response = Response::getInstance();
$response->getHeader()->disableMenuAndConsole();
$response->getFooter()->setMinimal();
$response->addHTML(var_dump($_SESSION));
```

more like what you're after?

manishbisht:
Seems like it sends lot of HTML. Which is not needed in this case. Any better approach for use session variables.

manishbisht:
```
<?php
define('PHPMYADMIN', true);
require_once 'libraries/session.inc.php';
print_r($_SESSION);
?>
```
When I am using it shows me 500 error.
[...]

PR #13502, opened 2017-07-18
show export and import progress

manishbisht:
Fixes: #6311 and #12401 

### DIFF
--- a/export.php
+++ b/export.php
@@ -13,6 +13,10 @@ use PhpMyAdmin\Url;
 use PhpMyAdmin\Util;
 use PhpMyAdmin\Sanitize;
 
+// Export progress variables
+$data = array();
+$count = 0;
+$total = 0;
 /**
  * Get the variables sent or posted to this script and a core script
  */
@@ -162,7 +166,8 @@ $post_params = array(
         'latex_data_continued_caption',
         'latex_data_label',
         'latex_null',
-        'aliases'
+        'aliases',
+        'progressuuid'
 );
 
 foreach ($post_params as $one_post_param) {
@@ -369,6 +374,7 @@ if ($save_on_server) {
         $filename = Sanitize::sanitizeFilename($filename);
 
         Core::downloadHeader($filename, $mime_type);
+
     } else {
         // HTML
         if ($export_type == 'database') {
@@ -426,9 +432,10 @@ do {
         PMA_exportServer(
             $db_select, $whatStrucOrData, $export_plugin, $crlf, $err_url,
             $export_type, $do_relation, $do_comments, $do_mime, $do_dates,
-            $aliases, $separate_files
+            $aliases, $separate_files, $GLOBALS['progressuuid']
         );
     } elseif ($export_type == 'database') {
+        $total = 1;
         if (!isset($table_structure) || !is_array($table_structure)) {
             $table_structure = array();
         }
@@ -446,7 +453,7 @@ do {
                     $db, $tables, $whatStrucOrData, $table_structure,
                     $table_data, $export_plugin, $crlf, $err_url, $export_type,
                     $do_relation, $do_comments, $do_mime, $do_dates, $aliases,
-                    $separate_files
+                    $separate_files, $GLOBALS['progressuuid']
                 );
             } finally {
                 PMA_unlockTables();
@@ -455,7 +462,8 @@ do {
             PMA_exportDatabase(
                 $db, $tables, $whatStrucOrData, $table_structure, $table_data,
                 $export_plugin, $crlf, $err_url, $export_type, $do_relation,
-                $do_comments, $do_mime, $do_dates, $aliases, $separate_files
+                $do_comments, $do_mime, $do_dates, $aliases, $separate_files,
+                $GLOBALS['progressuuid']
             );
         }
     } else {
--- a/js/export.js
+++ b/js/export.js
@@ -7,8 +7,7 @@
 /**
  * Disables the "Dump some row(s)" sub-options
  */
-function disable_dump_some_rows_sub_options()
-{
+function disable_dump_some_rows_sub_options() {
     $("label[for='limit_to']").fadeTo('fast', 0.4);
     $("label[for='limit_from']").fadeTo('fast', 0.4);
     $("input[type='text'][name='limit_to']").prop('disabled', 'disabled');
@@ -18,8 +17,7 @@ function disable_dump_some_rows_sub_options()
 /**
  * Enables the "Dump some row(s)" sub-options
  */
-function enable_dump_some_rows_sub_options()
-{
+function enable_dump_some_rows_sub_options() {
     $("label[for='limit_to']").fadeTo('fast', 1);
     $("label[for='limit_from']").fadeTo('fast', 1);
     $("input[type='text'][name='limit_to']").prop('disabled', '');
@@ -31,8 +29,7 @@ function enable_dump_some_rows_sub_options()
  *
  * @returns template data
  */
-function getTemplateData()
-{
+function getTemplateData() {
     var $form = $('form[name="dump"]');
     var blacklist = ['token', 'server', 'db', 'table', 'single_table',
         'export_type', 'export_method', 'sql_query', 'template_id'];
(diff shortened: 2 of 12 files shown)


export.php
```diff
@@ -538,3 +538,7 @@
 } else {
     echo $dump_buffer;
 }
+
+$cfgRelation = PMA_getRelationsParam();
+$sql_query = 'DELETE FROM '.$cfgRelation['db'].'.' . $cfgRelation['progress'] . ' WHERE type = "export"';
```
nijel: You should properly escape database and table name here.

libraries/export.lib.php
```diff
@@ -606,7 +606,18 @@ function PMA_exportDatabase(
...
+    $cfgRelation = PMA_getRelationsParam();
+    $GLOBALS['dbi']->selectDb($cfgRelation['db']);
+    $data = array();
+    $count = 0;
+    $sql_query = 'INSERT INTO '.$cfgRelation['db'].'.' . $cfgRelation['progress'] . ' VALUES ("export", "' . implode(",", $data) . '", ' . $count . ', ' . sizeof($tables) . ')';
+    PMA_queryAsControlUser($sql_query);
     foreach ($tables as $table) {
+        $count++;
+        array_push($data, $table);
+        $sql_query = 'UPDATE '.$cfgRelation['db'].'.' . $cfgRelation['progress'] . ' SET data = "' . implode(",", $data) . '", value = ' . $count . ' WHERE type = "export"';
```
nijel: * What happens if more users will be exporting at same time? You need to identify session as well or generate UUID for every upload. 
* Missing escaping around data leading to SQL injection vulnerability for control user.

js/export.js
```diff
@@ -978,4 +978,6 @@ AJAX.registerOnload('export.js', function () {
         );
         $('#column_alias_name').val('');
     });
-});
+
+    setInterval(function(){ showProgress("export"); }, 1000);
```
nijel: This probably should happen after submitting export form, not on loading the export page....

export.php
manishbisht: It can be done using this right `PhpMyAdmin\Util::backquote($cfgRelation['db'])` ?

libraries/export.lib.php
manishbisht: escaping data can be done using `$GLOBALS['dbi']->escapeString($count)` right ?
do we have any function available that generate uuid ?

js/export.js
manishbisht: Yes, I was looking for this option only. This was not coming in my mind. Thanks :)

export.php
ibennetch: Yes, `Util::backquote()` is what I would use here.

libraries/export.lib.php
ibennetch: As far as I know, we don't have anything to make a UUID. There's a [bit of code](http://php.net/manual/en/function.uniqid.php#94959) that everyone seems to reference as being the way to do it in PHP, but without the author posting any license information it's not clear to me if we're able to use it. PHP doesn't yet have a UUID function itself, meaning I think the best solution is to use MySQL's [`UUID()`](https://dev.mysql.com/doc/refman/5.7/en/miscellaneous-functions.html#function_uuid) function. Perhaps Michal has a different suggestion, though.

ibennetch: And yes, I'd use `escapeString` here.

ibennetch:
When I open the main page, I get this error/warning message:

> Notice in ./libraries/relation.lib.php#562
>  Undefined index: progress
> 
> Backtrace
> 
> ./libraries/relation.lib.php#69: PMA_checkRelationsParam()
> ./libraries/user_preferences.lib.php#50: PMA_getRelationsParam()
> ./libraries/classes/Config.php#912: PMA_loadUserprefs()
> ./libraries/common.inc.php#862: PhpMyAdmin\Config->loadUserPreferences()
> ./index.php#22: require_once(./libraries/common.inc.php)

manishbisht:
This is due to some cookie issue. Try running in incognito mode or clear cookies before running.
Also have to re-imported the `.sql` file in configuration database ?

export.php
ibennetch: Before running the query (here and at other points in the code), we should test if $cfgRelation['progress'] exists, perhaps with
```
if (!empty($cfgRelation['progress'])) {
    $sql_query = 'DELETE FROM '.$cfgRelation['db'].'.' . $cfgRelation['progress'] . ' WHERE type = "export"';
    PMA_queryAsControlUser($sql_query); 
}
```

ibennetch: Before running the query (here and at other points in the code), we should test if $cfgRelation['progress'] exists, perhaps with
```
if (!empty($cfgRelation['progress'])) {
    $sql_query = 'DELETE FROM '.$cfgRelation['db'].'.' . $cfgRelation['progress'] . ' WHERE type = "export"';
    PMA_queryAsControlUser($sql_query); 
}
```

ibennetch:
So the reason for my error was because I hadn't yet configured the table in `config.inc.php`. We'll have to deal gracefully with situations where the user either hasn't configured it or the table doesn't exist.

I made a comment in-line with some code, but you can also see for instance libraries/classes/Bookmark.php:114 for an example of how we might deal with this. We have to test the array to make sure $cfgRelation['progress'] isn't empty first, before trying to use it. We initialize it as part of the default configuration to an empty value, so here we'll test if the user has configured it.

Also, you should add the new directive to `libraries/config.default.php` with a blank assignment such as `$cfg['Servers'][$i]['progress'] = '';`. You can add it right after the export_template area  around line 503.

We should also add this new directive to `libraries/relation.lib.php` around line 353; this file is used by chk_rel.php which warns the user if their table structure is outdated.

export.php
manishbisht: To optimize this can we perform the sql query using ajax request on configuration database.
So that instead of running setInterval() always we can do something like this.
1. Start setInterval when the button is clicked.
2. Track regularly the progress percentage and once we reach 100 stop setInterval.
3. Perform the ajax request to delete the data.

manishbisht:
Updated the PR making the suggested changes.

libraries/export.lib.php
nijel: Using MySQL's function would certainly work. Another option is to use Session ID as identifier, that would limit to one upload per session, what is probably okay.

That reminds me that the table should also contain timestamp to allow cleanup of no longer needed entries...

manishbisht:
It now support multiple user export using uuid. There is the change in the template of the table so delete the pma_progress from configuration database and re import the `.sql` file.

libraries/export.lib.php
manishbisht: I have added a custom JavaScript function to generate uuid because to update the progress on page it is needed on JavaScript part also. Updated the PR. Please take a look into it.

manishbisht:
The export progress is almost done. I am now waiting for the review/feedback if it need any changes.

manishbisht:
progress is now showing for both import and export. Here is the output for both.
screenshot from 2017-08-06 12-54-17
screenshot from 2017-08-06 12-54-46

js/export.js
```diff
@@ -7,8 +7,7 @@
 /**
  * Disables the "Dump some row(s)" sub-options
  */
-function disable_dump_some_rows_sub_options()
-{
+function disable_dump_some_rows_sub_options() {
```
ibennetch: It's better to keep such style fixes as a separate commit so it doesn't look like this was part of your code for the progress bar. Unstaging it from your commits might be too much hassle now, though. They're good fixes to perform when working with these older library files, but from the aspect of maintaining efficient history, it's best as another commit.

ibennetch:
It works pretty well for me, except on export the whole status box only fills with information after the download completes.

image

This is shown after the download finishes; while download is in progress it just shows a small bar like this:
screen shot 2017-08-07 at 5 07 18 pm

Any idea about why I'm seeing that?

ibennetch:
Also, I'm having trouble getting it to work on Firefox. Chrome works okay, but Firefox gives me an error code 0:

ff

Does anyone else see that error?

manishbisht:
I have added a little improvement for export progress and removed the progress bar because the export file is downloaded and in parallel it blocks the set Interval and Set Timeout till the download completed.

I am seeing the Mozilla error on the master branch also.

PR #13524, opened 2017-07-26
set the default configuration database and tables when controluser + controlpass options are set

### LINKED ISSUE #12887: Configuration storage - fallback to default table names when pmadb config is set
### Steps to reproduce
1. Use mysql root account to create phpmyadmin mysql database (ZeroConf)
2. Create mysql user pma/pma which only has access for Configuration storage database (phpmyadmin)
3. Create mysql user user/user which DOESN'T have access to phpmyadmin database 
4. Create config.inc.php with ONLY 3 variables set (controluser=pma, controlpass=pma, pmadb=phpmyadmin)
5. Login to pma using user/user
6. Go to chk_rel.php

### Expected behaviour
I think that PMA should fallback to default table names if controluser + pmadb options are set while other like 'relation', 'column_info' are NOT.

### Actual behaviour
All features are listed as Disabled even though tables were created with default names.

### Server configuration
**Operating system**: Debian 8.6

**Web server:** nginx

**Database:** mariadb 10.1

**PHP version:** 7.0.12

**phpMyAdmin version:** newest from STABLE

manishbisht:
Fixes: #12887

### DIFF
--- a/libraries/relation.lib.php
+++ b/libraries/relation.lib.php
@@ -497,6 +497,74 @@ function PMA_checkRelationsParam()
         return $cfgRelation;
     }
 
+    // Set the default configuration database and table names when not configured
+    // in config.inc.php file but the control user and control pass is set
+    if ($GLOBALS['cfg']['Server']['controluser'] && $GLOBALS['cfg']['Server']['controlpass']) {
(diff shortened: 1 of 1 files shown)


manishbisht:
Should I set default control user to `phpmyadmin` leaving only control pass to be set ?

PR #13532, opened 2017-07-28
Fix unnecessary escaping of html content

CommanderRoot:

The `privileges` string can contain html content which when escaped looks like this:
grafik


ibennetch:
Good catch.

nijel:

PR #13534, opened 2017-07-29
Add Session class

MauricioFauth:
`$_SESSION` handling methods (`get`, `set`, `has`, `delete`) does not handle multidimensional arrays very well.


mauriciofauth:
I decided to remove the methods for handling the `$_SESSION` variable. I think we can put that aside for now.

nijel (commented):
I'm not sure about crating new session object several times, this certainly brings some overhead. Maybe using singleton as we do with Response would be better?

mauriciofauth:
I don't think singleton is a good fit for this class. Switching to static methods will solve this problem, I think.

nijel:
That's option as well. I just found it strange to construct class for every use.

PR #13567, opened 2017-08-05
Fixed tracking report: cannot delete a statement

tejisav:
Signed-off-by: Tejisav Singh Brar <tejisav@gmail.com>


nijel:
Thanks I've merged this, however there won't be any further release for the 4.0 series, so this patch will stay in git only.

Issue #13538, opened 2017-07-30
"No user selected" warning should come, selecting "Export" button without selecting any checkbox again user in "User Accounts" tab


anantjain6:
### Steps to reproduce
1. Login with username password
2. From top navigation select "User Accounts"
3. Without selecting any checkbox against any user, click on "Export"
![improvement](https://user-images.githubusercontent.com/22341750/28752096-0d812e1e-7535-11e7-9d76-051d1a917f98.png)

### Expected behaviour
Warning/Message should come that no user is selected

### Actual behaviour
No action

**Web server:** Wamp 3.0.6 (32 bit)
**Database:** 5.7.14 - MySQL Community Server (GPL)
**PHP version:** 5.6.25
**phpMyAdmin version:** 4.7.3
**Browser:** Google Chrome 60.0.3112.78 (Official Build) (32-bit)

anantjain6:
@lukewhitehouse nice idea, but everywhere the pattern is for giving warning message such as "Drop" button on database page

PR #13571, opened 2017-08-06
fix issue #13538

codehimanshu:
Signed-off-by: Himanshu Agrawal <himanshuagrawal1998@gmail.com>

### DIFF
--- a/js/server_privileges.js
+++ b/js/server_privileges.js
@@ -292,6 +292,7 @@ AJAX.registerOnload('server_privileges.js', function () {
         event.preventDefault();
         // can't export if no users checked
         if ($(this.form).find("input:checked").length === 0) {
+            PMA_ajaxShowMessage(PMA_messages.strNoAccountSelected, 2000, 'success');
             return;
         }
         var $msgbox = PMA_ajaxShowMessage();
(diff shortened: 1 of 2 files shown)


js/server_privileges.js
```diff
@@ -292,6 +292,7 @@ AJAX.registerOnload('server_privileges.js', function () {
         event.preventDefault();
         // can't export if no users checked
         if ($(this.form).find("input:checked").length === 0) {
+            PMA_ajaxShowMessage('No User Account Selected', 3000, 'success');
```
nijel: Can you please make the message translatable? This can be done by adding it to the js_messages in js/messages.php and then using it as PMA_messages.strNoAccountSelected.

Also there is probably no reason to use uppercase letters...

codehimanshu: I have added the message in js/messages.php. Also changed the timeout to 2000, as used previously in Exporting Databases.

mauriciofauth:
Hello, @codehimanshu. Could you use your real name in the Signed-off-by message? See https://github.com/phpmyadmin/phpmyadmin/blob/master/DCO

codehimanshu:
@mauriciofauth I have edited the comment since the commit messages can not be changed now. Will use Full Name from now on. Thanks for the reference.

mauriciofauth:
@codehimanshu You can squash the commits and push with the force option.

nijel:
I've squashed it during merge, thanks!

PR #13581, opened 2017-08-10
Fix #13578: Added custom no of rows

### LINKED ISSUE #13578: Allow custom number of rows to be inserted
### Steps to reproduce
1. Login with username password
2. Select any database
3. Select any table
4. From top navigation select "Insert"

### Expected behaviour
We can add custom number of rows

### Actual behaviour
Bound to insert specific number of rows
enhs

codehimanshu:
Changed select typoe input to number type input.

Signed-off-by: Himanshu Agrawal <himanshuagrawal1998@gmail.com>

### DIFF
--- a/libraries/insert_edit.lib.php
+++ b/libraries/insert_edit.lib.php
@@ -1453,18 +1453,9 @@ function PMA_getContinueInsertionForm($table, $db, $where_clause_array, $err_url
                 . ' value="' . htmlspecialchars(trim($where_clause)) . '" />' . "\n";
         }
     }
-    $tmp = '<select name="insert_rows" id="insert_rows">' . "\n";
-    $option_values = array(1, 2, 5, 10, 15, 20, 30, 40);
 
-    foreach ($option_values as $value) {
-        $tmp .= '<option value="' . $value . '"';
-        if ($value == $GLOBALS['cfg']['InsertRows']) {
-            $tmp .= ' selected="selected"';
-        }
-        $tmp .= '>' . $value . '</option>' . "\n";
-    }
+    $tmp = '<input type="number" name="insert_rows" id="insert_rows" value="'.$GLOBALS['cfg']['InsertRows'].'" min="1" />';
 
-    $tmp .= '</select>' . "\n";
     $html_output .= "\n" . sprintf(__('Continue insertion with %s rows'), $tmp);
     unset($tmp);
     $html_output .= '</form>' . "\n";
--- a/js/tbl_change.js
+++ b/js/tbl_change.js
@@ -479,7 +479,7 @@ AJAX.registerOnload('tbl_change.js', function () {
     /**
      * Continue Insertion form
      */
-    $(document).on('change', "#insert_rows", function (event) {
+    $(document).on('keyup mouseup', "#insert_rows", function (event) {
         event.preventDefault();
         /**
          * @var columnCount   Number of number of columns table has.
--- a/test/libraries/PMA_insert_edit_test.php
+++ b/test/libraries/PMA_insert_edit_test.php
@@ -1479,7 +1479,7 @@ class PMA_InsertEditTest extends PHPUnit_Framework_TestCase
         );
 
         $this->assertContains(
-            '<option value="1" selected="selected">',
+            '<input type="number" name="insert_rows" id="insert_rows" value="1" min="1" />',
             $result
         );
     }


ibennetch:
It looks like this was covered under the test suite, so now there is a failure about the changed HTML. Could you also fix the test?

```
1) PMA_InsertEditTest::testGetContinueInsertionForm

Failed asserting that '<form id="continueForm" method="post" action="tbl_replace.php" name="continueForm"><input type="hidden" name="db" value="db" /><input type="hidden" name="table" value="tbl" /><input type="hidden" name="lang" value="en" /><input type="hidden" name="token" value="token" /><input type="hidden" name="goto" value="index.php" /><input type="hidden" name="err_url" value="localhost" /><input type="hidden" name="sql_query" value="SELECT 1" /><input type="hidden" name="where_clause[0]" value="a&lt;b" />

Continue insertion with <input type="number" name="insert_rows" id="insert_rows" value="2" min="1" /> rows</form>

' contains "<option value="1" selected="selected">".

/home/travis/build/phpmyadmin/phpmyadmin/test/libraries/PMA_insert_edit_test.php:1483
```

codehimanshu:
Ya sure, will fix this.
Any link or guidelines to setup travis on my local to run tests before making Commit?

nijel:
You do not need Travis to run the tests, see https://github.com/phpmyadmin/phpmyadmin/wiki/Unit_Testing. However to setup Travis you can just go to https://travis-ci.org/ and enable builds on your fork.

libraries/insert_edit.lib.php
```diff
@@ -1453,18 +1453,9 @@ function PMA_getContinueInsertionForm($table, $db, $where_clause_array, $err_url
...
-    $option_values = array(1, 2, 5, 10, 15, 20, 30, 40);
 
-    foreach ($option_values as $value) {
-        $tmp .= '<option value="' . $value . '"';
-        if ($value == $GLOBALS['cfg']['InsertRows']) {
-            $tmp .= ' selected="selected"';
-        }
-        $tmp .= '>' . $value . '</option>' . "\n";
-    }
+    $tmp = '<input type="number" name="insert_rows" id="insert_rows" value="2" min="1" />';
```
ibennetch: The value here should reflect the default setting from the [InsertRows](https://docs.phpmyadmin.net/en/latest/config.html#cfg_InsertRows) configuration directive.

ibennetch:
Thanks for your submission!

I like this change, but I have a concern about whether users will know they need to change focus out of the input field for the additional forms to be added to the page. I wonder if it would be possible to update it immediately after the user enters the number rather than requiring them to change focus.

With the current system, if I change the dropdown the number of fields immediately changes; I don't first need to click away (or perhaps the browser changes focus after selecting something from the dropdown, I don't recall offhand how the specifications handle this case). I think users will be looking for an "update" button or expect the change to be immediate.

codehimanshu:
@ibennetch  Changing **on change** to on **keyup/mouseup** seems to do the job. But if I have some data in 2 rows filled, and 2 other rows are empty. And I have to just go and change the number of rows to 2 from 4, I can simply press the down arrow key, but if I clear its content and type 2, all my data of the 2 rows in form is lost. 
Its better to have the change take place **on change** only.
` $(document).on('keyup mouseup', "#insert_rows", function (event) { `

codehimanshu:
@ibennetch any update on this?

libraries/insert_edit.lib.php
```diff
@@ -1453,18 +1453,9 @@ function PMA_getContinueInsertionForm($table, $db, $where_clause_array, $err_url
...
-    $option_values = array(1, 2, 5, 10, 15, 20, 30, 40);
 
-    foreach ($option_values as $value) {
-        $tmp .= '<option value="' . $value . '"';
-        if ($value == $GLOBALS['cfg']['InsertRows']) {
-            $tmp .= ' selected="selected"';
-        }
-        $tmp .= '>' . $value . '</option>' . "\n";
-    }
+    $tmp = '<input type="number" name="insert_rows" id="insert_rows" value="'.$GLOBALS['cfg']['InsertRows'].'" min="1" />';
```
nijel: AFAIK we're using `<div class='slider'></div>` is most  places, so this should be consistent with rest of the code.

codehimanshu: I didn't get where the class slider is being used! If someone could help out.

PR #13585, opened 2017-08-14
show error message on export page when database is empty

### LINKED ISSUE #13572: Nothing on Export page when database is empty
### Steps to reproduce
1. Create a new database
2. Open export page
3. Nothing on page

### Expected behaviour
Should show some message like "No tables in the database"

### Actual behaviour
Nothing on page.
screenshot from 2017-08-06 17-59-41

manishbisht:
Fixes: #13572

screenshot from 2017-08-14 06-57-35

### DIFF
--- a/db_export.php
+++ b/db_export.php
@@ -48,7 +48,8 @@ $export_page_title = __('View dump (schema) of database');
 
 // exit if no tables in db found
 if ($num_tables < 1) {
-    PhpMyAdmin\Message::error(__('No tables found in database.'))->display();
+    $message = PhpMyAdmin\Message::error(__('No tables found in database.'));
+    $message->getDisplay();
     exit;
 } // end if
 


db_export.php
```diff
@@ -48,7 +48,8 @@
 
 // exit if no tables in db found
 if ($num_tables < 1) {
-    PhpMyAdmin\Message::error(__('No tables found in database.'))->display();
+    $message = PhpMyAdmin\Message::error(__('No tables found in database.'));
+    $message->getDisplay();
```
nijel: The getDisplay method just returns the message to display, so this can not work. What was wrong on the previous code?

manishbisht: Earlier the error was not displaying on the page.

nijel: You're right. It seems that Reponse class is showing global $message if set, so that's why your change works, the getDisplay method is not really doing anything useful here.

nijel:
I've merged your changes, but in the end I've rewritten the code again in ec9fe20e48882e76197804e96ac6ae543665397e

Issue #13512, opened 2017-07-21
Add inde issue


OlafvdSpek:
"Add 1 column(s) to inde" is missing an X.

Chrome on Windows

![pma](https://user-images.githubusercontent.com/2218936/28457725-8e1e255c-6e07-11e7-8726-5fa9769dffbe.PNG)

PR #13590, opened 2017-08-16
#13512 Patch

anantjain6:
#13512 Patch

Signed-off-by: Anant Jain <anantjain60@gmail.com>

### DIFF
--- a/themes/pmahomme/css/common.css.php
+++ b/themes/pmahomme/css/common.css.php
@@ -2167,9 +2167,9 @@ input#input_import_file {
 /**
  * Indexes
  */
-#index_frm .index_info input,
+#index_frm .index_info input[type="text"],
 #index_frm .index_info select {
-    width: 17em;
+    width: 14em;
     margin: 0;
     box-sizing: border-box;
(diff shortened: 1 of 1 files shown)


themes/pmahomme/css/common.css.php
```diff
@@ -2169,7 +2169,7 @@
  */
 #index_frm .index_info input,
 #index_frm .index_info select {
-    width: 14em;
+    width: 17em;
```
nijel: This really doesn't fix the problem, but rather hide it. I think that we rather should not force any width on the button as the text is being translated and can really have different widths.

nijel (changes requested):

themes/pmahomme/css/common.css.php
anantjain6: @nijel Oh Ya, I forgot about different languages
Let me make its width flexible
Thanks for the guidance

nijel:
Thanks, I've merged your patch as 28143053df730e2fc93c95d666a4a3dfe5a71eb1.

Issue #13617, opened 2017-08-26
Unnecessary data should not come in clipboard copied table


anantjain6:
### Steps to reproduce
1. Export phpMyAdmin configuration to your browser
2. Open any table
3. Select "Copy to clipboard"

### Actual behaviour
Unnecessary data  in clipboard copied table
![bug](https://user-images.githubusercontent.com/22341750/29739479-73301e64-8a5c-11e7-916f-26dd040cc3fc.png)

**phpMyAdmin version:** 4.7.4

akritianand:
Hi, I would like to take up this issue. I think I know how to fix it. Can somebody guide me on how to submit my fix?

PR #13619, opened 2017-08-27
Solved issue #13617

akritianand:
Unnecessary data should not come in clipboard copied table - SOLVED


nijel:
But this will remove the "Showing rows..." part, which probably makes sense to include...

akritianand:
Okay, I can rectify that.

nijel:

PR #13630, opened 2017-09-01
Fix error on signon login, need create a cookie with sessionID

ricardojlrufino:


examples/signon.php
```diff
@@ -30,6 +30,7 @@
     /* Update another field of server configuration */
     $_SESSION['PMA_single_signon_cfgupdate'] = array('verbose' => 'Signon test');
     $id = session_id();
+    setcookie($session_name, $id , 0, "/");
```
nijel: The `session_start` is supposed to do this...

ricardojlrufino: Strange, it only worked after I did this. 
After about 3 hours trying
``
Versão do cliente de banco de dados: libmysql - 5.5.44-MariaDB
Extensão do PHP: mysqli Documentação
Versão do PHP: 5.4.45
```

Comment (nijel) [reply examples/signon.php]: What is your setting for `session.use_cookies`? Does calling following prior to session_start fix your problem?

```php
@ini_set('session.use_cookies', 'true');
```

nijel:
Closing as this should not be really needed. There might be other issue in the script, but this is certainly not the correct solution.

PR #13632, opened 2017-09-02
gitgnore ".github" folder

solracsf:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,5 @@ cache.properties
 # Composer
 composer.lock
 /vendor/
+# GitHub specific
+.github


nijel:
We use the `.github` folder, so it certainly should not be ignored: https://github.com/phpmyadmin/phpmyadmin/tree/master/.github

acs-ferreira:
What's the local use of that folder? Shouldn't it remain in github only?

nijel:
The local use is to be able to edit those files - it's configuration for GitHub issue tracking.

acs-ferreira:
OK, so no local (production) use.
No problem, as many other projects just `.gitignore` it, i've submitted this PR, but if it is non-sense, just keep it closed :)

nijel:
Well the files in git serve development purposes as well :-). The releases do not contain those. I've added .gitattributes to avoid shipping them using composer in f407601d5988f97b47ff53e5bace9e17463b8f50.

PR #13659, opened 2017-09-06
Attempt to optimize Selenium tests

nijel:
Trying to make Selenium tests faster.


nijel:
Okay, the speedup in the end is not that significat, but I've managed to get rid of all arbitrary sleeps and properly wait for desired events.

Issue #13662, opened 2017-09-06
Change password shown behind navigation


nijel:
### Steps to reproduce

This is scenario tested by PMA_SeleniumChangePasswordTest::testChangePassword

1. Make window not too widde
2. Login
3. Click on change password

### Expected behaviour
The change password dialog should be on top

### Actual behaviour
The change password dialog is below navigation panel

### Server configuration

**phpMyAdmin version:** master

![screenshot-91c51915b5](https://user-images.githubusercontent.com/212189/30133356-cfec0c4e-9353-11e7-953e-499c8441a203.jpeg)

shucon:
Where can I find the z-index properties change password dialog ?

nijel:
I *think* that this is going to be broken for all dialogs in general, not only this one.

PR #13664, opened 2017-09-07
Change password shown behind navigation #13662

shucon:

### DIFF
--- a/themes/original/jquery/jquery-ui.css
+++ b/themes/original/jquery/jquery-ui.css
@@ -53,7 +53,7 @@
 }
 
 .ui-front {
-	z-index: 100;
+	z-index: 800;
 }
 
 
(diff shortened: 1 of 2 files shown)


themes/original/jquery/jquery-ui.css
```diff
@@ -53,7 +53,7 @@
 }
 
 .ui-front {
-	z-index: 100;
+	z-index: 800;
```
nijel: AFAIK this is original file coming with jQuery UI, so better to override it in our CSS to avoid problems on jQuery UI upgrades...

shucon: Where is the CSS being overrided ? I think the 1 is added to the z-index value of ui-front but can't find it.

nijel:
Closing in favor of #13665.

themes/original/jquery/jquery-ui.css
nijel: There is probably no override for this element yet, has to be added to common.css.php.

PR #13670, opened 2017-09-10
Fix edit index field size bug

jakubstanecki:
Signed-off-by: Jakub Stanecki <staneckijakub@gmail.com>

### DIFF
--- a/themes/pmahomme/css/common.css.php
+++ b/themes/pmahomme/css/common.css.php
@@ -3606,3 +3606,7 @@ body .ui-dialog .ui-dialog-buttonpane .ui-dialog-buttonset {
         width: 95% !important;
     }
 }
+#index_frm > #index_edit_fields > .index_info > #index_columns > tbody > tr > td > select[name="index[columns][names][]"],
+#index_frm > #index_edit_fields > .index_info > div > input[name="index[Key_name]"] {
+    width: 100%;
+}
--- a/themes/original/css/common.css.php
+++ b/themes/original/css/common.css.php
@@ -3349,3 +3349,7 @@ body .ui-dialog .ui-button-text-only .ui-button-text {
         margin: 4px;
     }
 }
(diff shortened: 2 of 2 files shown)


themes/pmahomme/jquery/jquery-ui.css
```diff
@@ -1310,3 +1310,7 @@ a.ui-button:active,
 	-webkit-box-shadow: -8px -8px 8px #aaaaaa;
 	box-shadow: -8px -8px 8px #aaaaaa;
 }
+#index_frm > #index_edit_fields > .index_info > #index_columns > tbody > tr > td > select[name="index[columns][names][]"],
+#index_frm > #index_edit_fields > .index_info > div > input[name="index[Key_name]"] {
+	width: 100%;
```
mauriciofauth: I think it would be better to move this to another file, as this file is replaced in the jQuery UI updates if I am not mistaken.

themes/pmahomme/css/common.css.php
```diff
@@ -3606,3 +3606,7 @@
         width: 95% !important;
     }
 }
+#index_frm > #index_edit_fields > .index_info > #index_columns > tbody > tr > td > select[name="index[columns][names][]"],
+#index_frm > #index_edit_fields > .index_info > div > input[name="index[Key_name]"] {
+	width: 100%;
```
mauriciofauth: You are using tab character for indentation instead of spaces. See https://github.com/phpmyadmin/phpmyadmin/wiki/Developer_guidelines#indentation.

mauriciofauth: I don't know if this is the best way to solve it. Apparently this issue only occurs in the pmahomme theme. So there must be something setting a fixed width to the dialog box.

nijel:
As @mauriciofauth has mentioned, this problem only exists in the pmahomme theme, not in the original one. I've fixed in this theme by 1338294134234b1f4e8411fd9e7a45039b2d458f.

Issue #13517, opened 2017-07-24
Select only visible / filtered results


YoussefE123:
### Steps to reproduce
1. Go to the "Browse" tab of any table
2. Type some text in the "Filter rows" text box
3. Click the "Check all" link at the bottom

### Expected behaviour
After results are filtered, only visible / filtered results should be selected, which affects any operation from the "With selected" links, like "Edit" and "Delete" (it would be nice to have these links appear at the top as well).

### Actual behaviour
All results get selected instead, even those that are hidden !
The following code seems to fix the issue : `$('input[name^=rows_to_delete]:visible').prop("checked",true)`

### Server configuration
**Operating system**: Windows 7 x64

**Web server:** Apache 2.4.26

**Database:** MariaDB version 10.2.6

**PHP version:** 7.1.7

**phpMyAdmin version:** 4.7.2

nijel:
Indeed current situation is not intuitive, but when touching this, the behavior should be probably consistently applied in all cases - the very same applies to database structure and tables filtering.

Also it should probably behave consistently in other direction as well: Once you select all and apply filter, only filtered rows should be probably affected.

YoussefE123:
Hi,

I'm primarily using the "Delete" link, after filtering the results, so yes, you are right about the other cases and the "other direction" as well, but to make it simple, this only affects how JavaScript selects the checkboxes.

In fact, it has nothing to do with the logic behind actions like updating or deleting rows, as those are handled separately _after_ the action is requested (based on the selected checkboxes).
I think the only needed change is when you click "Check All" or when the text filter changes (for the other direction), which both can be fixed in JavaScript alone like the code above (no need to touch php files, or any other database related files).

the other direction is a bit tricky, because it depends on whether the "Check All" is active or not :
1- active : apply default behavior (select only visible rows)
2- inactive : I guess de-selecting all checkboxes is enough whenever the text changes, and the user can "Check All" again if needed.

spyrule:
This BUG actually also affects the database table view as well.  If you filter by table view, and use "check all", it selects the hidden tables as well. If you select drop, it will now delete all of the filtered AND unfiltered tables as well (found this bug by accident... wiped out 30+ tables instead of the 15 I though I had filtered/selected).

It should be noted that if you manually check each item in the view, it will only affect those items actually marked (so manually selecting the entire visible list is not affected by the same logic).

drtothferenc19:
[...]

PR #13690, opened 2017-09-17
Added filtering tables by name in export

jakubstanecki:
Signed-off-by: Jakub Stanecki <staneckijakub@gmail.com>

### DIFF
--- a/db_export.php
+++ b/db_export.php
@@ -61,6 +61,13 @@ if (isset($_GET['structure_or_data_forced'])) {
 } else {
     $force_val = 0;
 }
+$multi_values .= '<fieldset id="tableFilter">
+    <legend>Filters</legend>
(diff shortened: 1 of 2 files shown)


mauriciofauth:
Apparently you're only hiding the tables without unchecking the hidden tables. All checked tables are being exported, even if they are hidden.

mauriciofauth:
But I don't know if automatically unchecking the hidden tables would be a good idea.

jakubstanecki:
@mauriciofauth I can add button "Uncheck hidden".

mauriciofauth:
@jakubstanecki I think a button might be a good idea.

nijel:
See also https://github.com/phpmyadmin/phpmyadmin/issues/13517 for discussion on this topic...

PR #13696, opened 2017-09-20
Edit index Column alignment 

KeerthanaRatna:
Both edit index and Add index have a alignment issue.
Signed-off-by: Keerthana Ratnasingam <keerthyratnam@gmail.com>

### DIFF
--- a/themes/pmahomme/css/common.css.php
+++ b/themes/pmahomme/css/common.css.php
@@ -2232,7 +2232,7 @@ table#index_columns {
 
 table#index_columns select {
     width: 85%;
-    float: right;
+    float: left;
 }
 
 #move_columns_dialog div {
--- a/themes/pmahomme/css/common.css.php
+++ b/themes/pmahomme/css/common.css.php
@@ -2187,7 +2187,7 @@ input#input_import_file {
  */
 #index_frm .index_info input,
 #index_frm .index_info select {
-    width: 14em;
+    width: 17em;
     margin: 0;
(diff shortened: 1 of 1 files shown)


themes/pmahomme/css/common.css.php
```diff
@@ -2187,7 +2187,7 @@
  */
 #index_frm .index_info input,
 #index_frm .index_info select {
-    width: 14em;
+    width: 17em;
```
mauriciofauth: The issue #13512 as already been fixed by 2814305 commit. Your branch is outdated.

themes/pmahomme/css/common.css.php
```diff
@@ -2232,7 +2232,7 @@
 
 table#index_columns select {
     width: 85%;
-    float: right;
+    float: left;
```
mauriciofauth: This solves the problem, but I think it would be better to use `float: <?php echo $left; ?>;` instead of `float: left;`.

mauriciofauth:
Hi, @KeerthanaRatna. Thank you for your contribution.

You have opened the pull request from the master branch, which is outdated. Ideally, you should create a new branch for each issue you want to resolve, to avoid conflicts.

To solve the Signed-Off-By line issue, you must pass the `-s` option to git, for example `git commit -s`.

mauriciofauth:
Thanks for your contribution, merged as 73e6a37 to 4.7 maintenance branch.

Issue #13490, opened 2017-07-12
Importing table without selecting database shows both success and error message


manishbisht:
### Steps to reproduce
1. Import table without selecting any database.

### Expected behaviour
Should only shows an error

### Actual behaviour
Shows error + success message

![screenshot from 2017-07-11 21-16-06](https://user-images.githubusercontent.com/14255990/28128058-0fb61c36-674c-11e7-9c54-2a1a48c42154.png)

ibennetch:
So what we've decided is happening here is that phpMyAdmin is counting that it successfully passed on those 13 queries to MySQL, but of course MySQL wasn't successful in executing them all so the phpMyAdmin success message is a bit misleading.

Nilucshan:
Hi @manishbisht 

What do you mean by importing a table? Is it like importing a .csv or similar files to the phpMyadmin?

manishbisht:
Yes, I tried importing '.sql' and it has shown me the above error.

anantjain6:
Drag and drop .sql file without "database select" should show warning/error message

nijel:
Duplicate of #12856

PR #13698, opened 2017-09-21
Fixing issue #13490, removing success message when no db is selected

ryanml:
Signed-off-by: Ryan Lanese <rlanese@asu.edu>

### DIFF
--- a/import.php
+++ b/import.php
@@ -573,7 +573,7 @@ if (! empty($id_bookmark) && $_REQUEST['action_bookmark'] == 2) {
 } elseif ($finished && ! $error) {
     // Do not display the query with message, we do it separately
     $display_query = ';';
-    if ($import_type != 'query') {
+    if ($import_type != 'query' && $db != '') {
         $message = PhpMyAdmin\Message::success(
             '<em>'
             . _ngettext(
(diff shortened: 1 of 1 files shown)


ryanml:
Per issue #13490, when a user imports a file with a .sql extension without selecting a database, both a success and error message show, like pictured here:

screen shot 2017-09-20 at 8 39 18 pm

This small fix, when reaching the point that such a success message would be generated, will check to see if $db is empty string, and if so, will not show the message. This mimics the behavior of compressed file imports when no database is selected, where it shows only the error message.

Rather than modifying the number of successful queries message, it seems to make sense to mimic the current standing behavior mentioned above.

Result of fix:

screen shot 2017-09-20 at 8 36 32 pm

import.php
```diff
@@ -573,7 +573,7 @@
 } elseif ($finished && ! $error) {
     // Do not display the query with message, we do it separately
     $display_query = ';';
-    if ($import_type != 'query') {
+    if ($import_type != 'query' && $db != '') {
```
mauriciofauth: Wouldn't it be better to use the [empty](https://secure.php.net/empty) function?

Issue #12856, opened 2016-12-27
Error handling broken on import


Konstantinos-Sk:
### Steps to reproduce
1. Go to "Local Databases" page (without selecting any database).
2. Click on "Import" and browse through your files to select a .sql file.
3. Hit "Go" button.

### Expected behaviour
It should give **only** the _Error_ box output with error:
```
#1046 - No database selected
```

### Actual behaviour
It gives the _Error_ box correctly, but it also gives that the import has been successfully finished even though it failed.
![report](https://cloud.githubusercontent.com/assets/6353082/21508323/f279f060-cc88-11e6-9e70-8adee988cae2.png)

### Server configuration
**Operating system**:
Windows 7 Ultimate SP1
**Web server:**
Apache 2.4.25
**Database:**
MySQL 5.7.17
**PHP version:**
7.1.0
**phpMyAdmin version:**
4.6.5.2
### Client configuration
**Browser:**
Google Chrome 55.0.2883.87
**Operating system:**
Windows 7 Ultimate SP1

nijel:
This is not really related to database selection, the error handling in import is overall missing - it always shows success message regardless whether there are some failed queries.

For example try importing file with single line:

```
CREATE BAZ;
```

Importing on database actually shows two success messages, even though the query has failed:

![screenshot-2017-10-3 localhost localhost gammu phpmyadmin 4 8 0-dev](https://user-images.githubusercontent.com/212189/31132939-62820242-a85e-11e7-9290-7a70f37a6958.png)

williamdes:
@Konstantinos-Sk Closing, could not reproduce
Comment if still an issue :)

PR #13699, opened 2017-09-21
Using empty() to check for a set db

ryanml:
Signed-off-by: Ryan Lanese <rlanese@asu.edu>

### DIFF
--- a/import.php
+++ b/import.php
@@ -573,7 +573,7 @@ if (! empty($id_bookmark) && $_REQUEST['action_bookmark'] == 2) {
 } elseif ($finished && ! $error) {
     // Do not display the query with message, we do it separately
     $display_query = ';';
-    if ($import_type != 'query' && !empty($db)) {
+    if ($import_type != 'query' && $db != '') {
         $message = PhpMyAdmin\Message::success(
             '<em>'
             . _ngettext(
--- a/import.php
+++ b/import.php
@@ -573,7 +573,7 @@ if (! empty($id_bookmark) && $_REQUEST['action_bookmark'] == 2) {
 } elseif ($finished && ! $error) {
     // Do not display the query with message, we do it separately
     $display_query = ';';
-    if ($import_type != 'query') {
+    if ($import_type != 'query' && !empty($db)) {
         $message = PhpMyAdmin\Message::success(
             '<em>'
             . _ngettext(
--- a/test/selenium/PmaSeleniumImportTest.php
+++ b/test/selenium/PmaSeleniumImportTest.php
@@ -109,9 +109,16 @@ class PMA_SeleniumImportTest extends PMA_SeleniumBase
 
         $this->scrollToBottom();
         $this->byId("buttonGo")->click();
-        $this->waitForElement(
-            "byXPath",
-            "//div[@class='success' and contains(., 'Import has been successfully')]"
-        );
+        if (
+            $this->isElementPresent(
+                "byXPath",
+                "//a[@class='item' and contains(., 'Database:')]"
+            )
+        ) {
+            $this->waitForElement(
+                "byXPath",
+                "//div[@class='success' and contains(., 'Import has been successfully')]"
+            );
+        }
     }
 }


ryanml:
@mauriciofauth I opened a new PR using empty, as I made I pushed a few incorrect commits to the old one by mistake (I figured this was easier) [ref PR #13698]

I typically haven't used empty due to differences in behavior across PHP versions, but you're right in that it's better for readability, and will likely be fine for the purposes of this fix.

ryanml:
Also to note, this will fix issue #12856 as well as #13490

mauriciofauth:
This change broke one of the [Selenium tests](https://travis-ci.org/phpmyadmin/phpmyadmin/jobs/278190572). See:
```
1) PMA_SeleniumImportTest::testServerImport
PHPUnit_Extensions_Selenium2TestCase_WebDriverException: no such element: Unable to locate element: {"method":"xpath","selector":"//div[@class='success' and contains(., 'Import has been successfully')]"}
```

ryanml:
@mauriciofauth I imagine the test case would have to change a bit to accommodate this behavior

I think a check like this would have to happen in PmaSeleniumImportTest.php on line 112:

`if ($this->isElementPresent("byXPath", "//a[@class='item' and contains('Database: ')]"))`

From the dom perspective, this looks at the top nav for the presence of a selected database, and ideally the test would only expect the success message if that exists

Does this seem like a reasonable adjustment to you? I appreciate your patience :)

mauriciofauth:
What do you think about this, @devenbansod?

mauriciofauth:
@ryanml Feel free to make changes to the tests. :)

ryanml:
@mauriciofauth will do :) I'll be able to get to it this evening, thanks!

mauriciofauth:
@ryanml about the new PR. If you push some incorrect commits or just minor fixes to previous commits, you can squash them with `git rebase`.

ryanml:
@mauriciofauth thanks for the tips :)

It looks like everything is checking out with the selenium tests through Travis, are those code coverage checks anything I need to be concerned about?

devenbansod:
[bs-video-logs-use.mp4.zip](https://github.com/phpmyadmin/phpmyadmin/files/1323468/bs-video-logs-use.mp4.zip)
I could not find time to go through the code completely, but this condition of database being empty might actually be true for a server import.

Also, as you can see the attached video, it actually gives an error in the test. So adding the if condition in the test does not seem like a good idea.

ryanml:
@devenbansod the adjustment I made in the test will work fine when errors are shown, the issue is that the coverage didn't account for my code change, that a success message should only be show when a database is selected

On that note, if it is true for a server import, I'll be happy to withdraw this PR. However, any time I attempt any import without a db selected, MySQL throws a 1046, that and previous issues leads me to believe it's a constant. So just let me know with this one, thanks :)

cc: @mauriciofauth

import.php
```diff
@@ -573,7 +573,7 @@
 } elseif ($finished && ! $error) {
     // Do not display the query with message, we do it separately
     $display_query = ';';
-    if ($import_type != 'query') {
+    if ($import_type != 'query' && !empty($db)) {
```
nijel: Using empty() is usually not a good idea, imagine database called 0 (zero).

ryanml: @nijel Originally I had done ...&& $db != '', after @mauriciofauth suggested using empty().

Would the former be better, or may I have to do some more stringent checking to be sure that we in fact have an empty database? Eg maybe something like strlen($db) > 0, or perhaps use isset() and another validator

I'm glad to continue working on this, thank you both mauricio and nigel for your help

mauriciofauth: I don't know why, but I was sure that 0 was an invalid name. Anyway, we have to check the problem pointed out by Deven.

ryanml: I would have thought the same thing too, and cool. I'll check it out more as well

nijel: Actually 0 is valid database and table name (and quite lot of things in phpMyAdmin break with it exactly due to using empty as check in many places).

ryanml:
@nijel @mauriciofauth I have pushed another commit again making the check ($db != '')

In any situation a database is not chosen, $db always contains an empty string.

This probably the best solution for what this PR is trying to accomplish, let me know what you guys think and if you'd like to move forward with this :)

ryanml:
@nijel have you taken a look at this since my last commit or have any final thoughts on the issue in general?

nijel:
Honestly I'm not really sure what are you trying to achieve here. Why the server import should behave differently here?

mauriciofauth:
It's related to #12856 and #13490.

nijel:
I think we should not show the success message on error, regardless it is on server, database or table. Not showing success message at all sounds confusing.

ryanml:
@nijel Sorry, I'm a little confused by your response, do you think we should, or should not show the success message? The standing issue was that the success message implies that a _successful_ action has taken place

nijel:
We should not show the success message if there was error during the import. The fact whether $db is set or not is really not relevant to that.

nijel:
To make it more explicit: The error handling is broken in all import situations, not only server wide.

screenshot-2017-10-3 localhost localhost gammu phpmyadmin 4 8 0-dev

nijel:
Sorry, but this is wrong approach to the issue (which was wrongly described as well). See https://github.com/phpmyadmin/phpmyadmin/issues/12856#issuecomment-333875265

PR #13706, opened 2017-09-28
Remove jQuery and jQuery UI sources from repository

MauricioFauth:
The jquery and jquery-ui packages includes their source codes, which makes it unnecessary to include the source codes in the repository.


mauriciofauth:
@nijel We can do this, right?

nijel:
The source code is there for licensing reasons (in the end we release under GPLv2, so we have to distribute source code). I'm not really sure how using package manager fits into this...

mauriciofauth:
Currently only the jquery.min.js file is being included in the release archive. Source files are not included. We could add the node_modules directory in the phpmyadmin-sources archive to anyone who wants all the files.

mauriciofauth:
The changes I made now were just to make that PR more specific.

nijel:
In the end I've decide to merge this even though I'm not 100% sure this is okay from legal view. On the other side I have not heard opposite as well...

PR #13717, opened 2017-10-02
Simplify Template API

nijel:
Remove some features we do not seem to use:

- helper functions (this never worked with Twig)
- per class context, it now has to be passed to render
- this avoids need of array_merge call on every render


MauricioFauth (approved):
The helper functions are not being used in phtml templates if I'm not mistaken and in Twig we use the extensions, so it's safe to remove this functionality.
And passing the data directly to the render makes more sense.

PR #13720, opened 2017-10-04
Fix typo in Greek translation

grrnikos:
Signed-off-by: Nikos Grigoriadis <grrnikos@yahoo.gr>

There is a typo when searching the database in the Greek language, so it's frequently seen by many users.


mauriciofauth:

PR #13721, opened 2017-10-05
Attach syntax highlighter on single Routine/Trigger/Event edit page

Nekaravaev:
Hello, guys. 
I've noticed that on routine/trigger/event edit page CodeMirror didn't initialize.

### DIFF
--- a/js/rte.js
+++ b/js/rte.js
@@ -1080,4 +1080,29 @@ $(function () {
         // the input fields lose the correct order and need to be reordered.
         RTE.ROUTINE.reindexParameters();
     }); // end $(document).on()
+
+    /**
+     * Attach SyntaxHighlighter for the
+     * textarea on single Routine/Trigger/Event edit page
+     */
+    $(document).ready(function () {
+        var $singlePageForm = $('.rte_form');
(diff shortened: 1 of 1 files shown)


nijel:
It should be already done here:

https://github.com/phpmyadmin/phpmyadmin/blob/bd7ff2fe68a8b78ccc3bcc126f4e9269b5af3d5c/js/rte.js#L376

So IMHO it's better to investigate why current code doesn't work instead of adding yet another code to handle this....

Nekaravaev:
@nijel, yeah, I've seen it. 

https://github.com/phpmyadmin/phpmyadmin/blob/bd7ff2fe68a8b78ccc3bcc126f4e9269b5af3d5c/js/rte.js#L190

But this is a part of ajax functionality, which fire by click on link. 
https://github.com/phpmyadmin/phpmyadmin/blob/bd7ff2fe68a8b78ccc3bcc126f4e9269b5af3d5c/js/rte.js#L943

When we open in new tab and going to single page, this definitely will not work. 
And here I didn't see any opportunities to reuse method.

PR #13735, opened 2017-10-12
Fix for password change on Galera cluster

Seitanas:
Hello,
this patch fixes issue with changing passwords on Galera cluster.
The problem is, that user tables are stored in MyISAM storage engine, so changing password with
ordinary "UPDATE" queries will not be replicated. MariaDB uses "SET PASSWORD" command to update user credentials, which is a DDL command. DDL commands are replicated, thus phpMyAdmin should use "SET PASSWORD" on MariaDB versions 10 and above.

Signed-off-by: Tadas Ustinavičius <tadas@ring.lt>

### DIFF
--- a/libraries/server_privileges.lib.php
+++ b/libraries/server_privileges.lib.php
@@ -1905,7 +1905,19 @@ function PMA_updatePassword($err_url, $username, $hostname)
 
             $local_query = $query_prefix
                 . $GLOBALS['dbi']->escapeString($_POST['pma_pw']) . "'";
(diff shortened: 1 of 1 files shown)


Seitanas:
If this patch will be accepted, I will gladly make one for "main" branch

mauriciofauth:
If this is affecting version 4.7.x, please use branch [QA_4_7](https://github.com/phpmyadmin/phpmyadmin/tree/QA_4_7).

PR #13737, opened 2017-10-12
Fix for password change issues on Galera cluster

Seitanas:
Hello,
this patch fixes issue with changing passwords on Galera cluster.
The problem is, that user tables are stored in MyISAM storage engine, so changing password with
ordinary `UPDATE` queries will not be replicated. MariaDB uses `SET PASSWORD` command to update user credentials, which is a DDL command. DDL commands are replicated, thus phpMyAdmin should use `SET PASSWORD` on MariaDB versions 10 and above.

Signed-off-by: Tadas Ustinavičius <tadas@ring.lt>

### DIFF
--- a/libraries/server_privileges.lib.php
+++ b/libraries/server_privileges.lib.php
@@ -1905,7 +1905,19 @@ function PMA_updatePassword($err_url, $username, $hostname)
 
             $local_query = $query_prefix
                 . $GLOBALS['dbi']->escapeString($_POST['pma_pw']) . "'";
-        } else if ($serverType == 'MariaDB'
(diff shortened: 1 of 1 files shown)


Seitanas:
I assume checks failed because I've checked "Any new functionality is covered by tests", but there are no tests produced?

Seitanas:
Hello,
any progress on this issue?

nijel:
Thanks for your contribution, merged with following changes:

* 713fcd4195 Changelog entry for #13737
* 9ebd88061c Set $sql_query as well when changing password
* b8428ea70f Fix coding style for #13737

PR #13740, opened 2017-10-13
Fix input value escaping

CommanderRoot:

The value string is HTML escaped at this point so to get the desired effect we need to replace the HTML value for `'` which is `&#039;` instead of `'` directly.


mauriciofauth:

PR #13741, opened 2017-10-14
Fix favorite icon escaping

CommanderRoot:

Both `titles['Favorite']` and `titles['NoFavorite']` are HTML img tags defined here: https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/classes/Util.php#L3449-L3450 . In order to not return them as text but rather render them as an image we need to output them `raw`. 

This fixes this:
grafik


mauriciofauth:

PR #13749, opened 2017-10-16
Fix for table names being obscured when changing table structure

### LINKED ISSUE #13748: Table name hides column descriptions when creating table
### Steps to reproduce
1. Open database
2. Enter name of table to create and go to table creation

### Expected behaviour
The column names should be shown

screenshot-2017-10-16 localhost localhost baz3 hokus phpmyadmin 4 7 5-dev

### Actual behaviour
Column names are hidden behind table name input.

screenshot-2017-10-16 localhost localhost baz3 hokus phpmyadmin 4 8 0-dev

### Server configuration
**phpMyAdmin version:** master at 35e15e90c1, this is regression from QA_4_7

### Client configuration
**Browser:** Firefox

ankitjain28may:
Signed-off-by: Ankit Jain <ankitjain28may77@gmail.com>

### DIFF
--- a/js/config.js
+++ b/js/config.js
@@ -612,7 +612,7 @@ function adjustPrefsNotification () {
     var $prefsAutoShowing = ($prefsAutoLoad.css('display') !== 'none');
 
     if ($prefsAutoShowing && $tableNameControl.length) {
-        $tableNameControl.css('top', '120px');
+        $tableNameControl.css('top', '55px');
     }
 }
 


js/config.js
```diff
@@ -612,7 +612,7 @@ function adjustPrefsNotification () {
     var $prefsAutoShowing = ($prefsAutoLoad.css('display') !== 'none');
 
     if ($prefsAutoShowing && $tableNameControl.length) {
-        $tableNameControl.css('top', '120px');
+        $tableNameControl.css('top', '55px');
```
nijel: IMHO this code is not active in my case - the preferences loading box is not present....

ankitjain28may: This function is triggred and it adds the inline styling i.e top = 120px to the div which creates this bug. I don't know through which its triggering but this is the cause for the bug.

ankitjain28may:
@OlafvdSpek Check out the css.

js/config.js
ankitjain28may: @nijel Any other suggestions ?

nijel: Apparently I've misunderstood the code, the fix works fine.

nijel:

PR #13751, opened 2017-10-16
More descriptive text for ASC/DESC sort order

akritianand:
Replaced 1 with [Asc/Desc]


nijel:
Sorry, but this is wrong for several reasons:

- there can be different numbers than just one, when there are multiple sort ordering defined
- the text output should be localized
- I'm not sure if it makes sense to include this in the output at all, therefore https://github.com/phpmyadmin/phpmyadmin/pull/13823 was merged

Issue #13747, opened 2017-10-16
Drop table confirmation


OlafvdSpek:
The table name is burried in the fifth line of the dialog.
Could the title be like `Drop z_cms_m62_bike_equipment_brands` instead?

```
Confirm
You are about to DESTROY a complete table! Do you really want to execute "DROP TABLE `z_cms_m62_bike_equipment_brands`"?
```

Bhola-B2C:
Could you please explain the issue?? Do you want just the alert message to be changed ?

akritianand:
I have submitted a PR for this. Please review.

OlafvdSpek:
The diff / commit isn't right, it's way too large. I think it contains lots of other changes.

akritianand:
Yes. My first commit had different files. But I changed it, so now it only has the required changes. I'm sorry for my mistake. 
![capture](https://user-images.githubusercontent.com/23422253/31653262-977c470a-b33f-11e7-97a2-50b1408edcc2.PNG)

OlafvdSpek:
I'm not sure doing text matches is the right way to go..

akritianand:
@OlafvdSpek Since the input going into PMA_confirm was 'question' which contained the confirmation text, I thought to use that only to look for what kind of confirmation it was asking. Can you tell me what could be the more appropriate way to go about it?

OlafvdSpek:
I think title should be passed in from higher up instead of always using PMA_messages.strConfirm

PR #13752, opened 2017-10-16
Change phrasing of drop database dialog

akritianand:

### DIFF
--- a/js/functions.js
+++ b/js/functions.js
@@ -79,11 +79,11 @@ var spatial_indexes = [];
  * by appending a random variable to their parameters
  */
 $.ajaxPrefilter(function (options, originalOptions, jqXHR) {
-    var nocache = new Date().getTime() + "" + Math.floor(Math.random() * 1000000);
-    if (typeof options.data == "string") {
-        options.data += "&_nocache=" + nocache;
-    } else if (typeof options.data == "object") {
-        options.data = $.extend(originalOptions.data, {'_nocache' : nocache});
+    var nocache = new Date().getTime() + '' + Math.floor(Math.random() * 1000000);
+    if (typeof options.data === 'string') {
+        options.data += '&_nocache=' + nocache + '&token=' + encodeURIComponent(PMA_commonParams.get('token'));
+    } else if (typeof options.data === 'object') {
+        options.data = $.extend(originalOptions.data, { '_nocache' : nocache, 'token': PMA_commonParams.get('token') });
     }
 });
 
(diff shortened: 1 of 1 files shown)


ibennetch:
Hi @akritianand, thanks for your submission. In the future, it would be helpful if you can use a more descriptive title for your pull request and for the commits. "Update functions.js" or "Issue #13747" aren't very helpful when looking through the git log. Something like "Improve drop table confirmation message" makes a better pull request title.

The bot is upset that you haven't set the Signed-Off-By line which indicates that you agree with the terms of the [Developer Certificate of Origin (DCO)](https://github.com/phpmyadmin/phpmyadmin/blob/master/DCO). You can read a bit more about adding this in the [CONTRIBUTING document](https://github.com/phpmyadmin/phpmyadmin/blob/master/CONTRIBUTING.md#patches-submission), but it's most easily accomplished by passing the `--signoff` argument to git commit.

js/functions.js
```diff
@@ -2776,16 +2776,28 @@ jQuery.fn.PMA_confirm = function (question, url, callbackFn, openCallback) {
...
-            },
-            open: openCallback,
-            modal: true
-        });
+    if (question.search('DROP TABLE') != -1) {
+        var heading = question.split('execute "')[1].split('"')[0];
+    }
+
+    else if (question.search('DROP DATABASE') != -1) {
+        var heading = question.split('execute "')[1].split('"')[0];
```
nijel: Both these will break with localized messages.

js/functions.js
```diff
@@ -2776,16 +2776,28 @@ jQuery.fn.PMA_confirm = function (question, url, callbackFn, openCallback) {
...
+    }
+
+    else {
+        var heading = PMA_messages.strConfirm;
+    }
+
+    $('<div/>', {'id': 'confirm_dialog', 'title': heading})
+    .prepend(question)
```
nijel: Any reason for changing indentation on existing code?

nijel (changes requested):
Besides things @ibennetch has already mentioned, I think the patch itself is also wrong. Maybe it would be best to include the title as optional parameter for PMA_confirm, so that every caller could specify it instead of relying on parsing of passed text what will break with localized strings.

js/functions.js
akritianand: Can you tell me how will these break localised messages? I'm new to this, so sorry for my ignorance.

nijel: They won't contain the "execute" string (and can contain different quote as well). For example in Czech the string will be "Opravdu si přejete provést „%s“?". So split will return just one string and the code will fail.

PR #13755, opened 2017-10-18
RFC: Create CODE_OF_CONDUCT.md

nijel:
I think it's useful to have one. This one comes from the ones suggested by GitHub and I think is good match for us.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,46 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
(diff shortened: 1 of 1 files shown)


madhuracj:
+1
I wonder whether developers@phpmyadmin.net is the best place to report. Maybe, an email address to reach team or just the admin would be better? Because reporting to developers@phpmyadmin.net would not protect the confidentiality with regard to the reporter.

ibennetch:
I agree with Madhura that we should change the email address to one that is not public. Aside from that, this is fine by me.

nijel:
Landed as b916d5da6e with info@phpmyadmin.net as contact point, thanks for feedback.

PR #13763, opened 2017-10-20
Add JavaScript dependencies to package.json

MauricioFauth:
Added dependencies on specific versions because we do not use these files
directly which makes updating non-automatic.


ibennetch:
@mauriciofauth Should we also add yarn.lock to .gitignore?

mauriciofauth:
@ibennetch Yarn documentation suggests that the `yarn.lock` file should be checked into source control.
See https://yarnpkg.com/en/docs/yarn-lock#toc-check-into-source-control

Also, Travis CI detects the use of Yarn by the presence of yarn.lock in the repository root.
See https://blog.travis-ci.com/2016-11-21-travis-ci-now-supports-yarn

PR #13765, opened 2017-10-20
fix #13727 Copied table data add "1" after column name which had order by parameter

### LINKED ISSUE #13727: Copied table data add "1" after column name which had order by parameter
### Steps to reproduce
1. Select any table 
2. order by any column
3. from menu/option in bottom on table select "copy to clipboard"

### Expected behaviour
"1" do not depict the why it is used for
Maybe we can add [Asc] [Desc] or anything else or  nothing

### Actual behaviour
"1" is added after column name which had order by parameter
copy

### Server configuration
Online demo

**phpMyAdmin version:** 4.7.4

### Client configuration
**Browser:** 61.0.3163.100 (Official Build) (64-bit)

**Operating system:** Ubuntu 16.0.4 LTS

xypXiong:
Manipulate the title string so that the ordering text is not shown in the copied text. Web page still keeps the order numbering.


nijel:
I think this code is getting more and more complex and thus fragile. Maybe it would be better to make it work other way - indicate by class what should be copied and make this function copy only elements having this class?

xypXiong:
Hi. Thanks for suggestion. I will take a look in this direction.By the way, when I checked in code, I always get  complain about "trailing space" .Do you have any suggestions to get rid of that? I am very new to git right now.
thanks,Yanping 

    Michal Čihař <notifications@github.com> 于 2017年10月23日, 星期一, 上午 5:18 写道：
 

 I think this code is getting more and more complex and thus fragile. Maybe it would be better to make it work other way - indicate by class what should be copied and make this function copy only elements having this class?—

PR #13777, opened 2017-10-25
Removing inline CSS

ADFF2F:

### DIFF
--- a/themes/pmahomme/css/common.css.php
+++ b/themes/pmahomme/css/common.css.php
@@ -13,7 +13,6 @@ if (! defined('PMA_MINIMUM_COMMON') && ! defined('TESTSUITE')) {
 }
 ?>
 /******************************************************************************/
-
 /* general tags */
 html {
     font-size: <?php echo $theme->getFontSize(); ?>
@@ -382,8 +381,7 @@ fieldset legend {
     margin: 1.5em;
 }
 
-/* buttons in some browsers (eg. Konqueror) are block elements,
-   this breaks design */
+/* buttons in some browsers (eg. Konqueror) are block elements, this breaks design */
 button {
     display: inline;
 }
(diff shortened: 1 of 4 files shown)


themes/pmahomme/css/common.css.php
```diff
@@ -2503,6 +2501,13 @@
     padding: 2px;
 }
 
+.clearright {
+    clear:right;
+}
```
nijel: This is most likely wrong for RTL languages (probably should match floatright).

themes/pmahomme/css/common.css.php
```diff
@@ -2503,6 +2501,13 @@
...
+.displaynone {
+    display:none;
+}
```
nijel: There is already hide class doing this, no need to duplicate it.

nijel (changes requested):
Thanks for your patch. Besides things I've mentioned inline, the CSS changes should be also implemented in second theme we ship - in themes/original/css/common.css.php

themes/pmahomme/css/common.css.php
skokeers27: This is original line:
echo '<div class="choice floatright" style="clear:right;">';
Shoudl I left 2 classes, choice and floatright and remove "style="clear:right;" ?

nijel: Yes, the original code was wrong as well. Hmm, looking at it, this element seems to be hidden all the time anyway, so it's probably better remove anything else from it:

https://github.com/phpmyadmin/phpmyadmin/blob/701a8855fdb4c992e3882b76c40a92c2610891ba/themes/pmahomme/css/gis.css.php#L46-L48

skokeers27: Ok, I left only choice class.

ADFF2F (commented):
changes have been made

nijel:
Thanks, I've merged this PR with following fixes:

* 8109763b75 Remove not needed CSS for span elements
* 43cbcf6bd0 Remove unused class
* 636b22cb06 Remove clearright class

PR #13785, opened 2017-10-29
Port templates to Twig and remove PHP templates support

### LINKED ISSUE #13297: Port templates to Twig and remove PHP templates support
As you know, we started using the [Twig](https://twig.sensiolabs.org/doc/1.x/) template engine and some templates have already been ported.
But there are still several templates to be ported to Twig.

When all templates have been ported, we must then remove the support for the PHP templates (`.phtml` extension).

MauricioFauth:
Fix #13297


nijel:
Seems to break creating tables, see https://github.com/phpmyadmin/phpmyadmin/issues/13796

PR #13787, opened 2017-10-30
Second authentication factor

nijel:
Generic framework for supporting second authentication factor.

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -10,6 +10,7 @@ build:
     dependencies:
         before:
             - composer install
+            - composer require tecnickcom/tcpdf pragmarx/google2fa bacon/bacon-qr-code samyoul/u2f-php-server
             - ./vendor/bin/phpcs --config-set installed_paths `pwd`/vendor/phpmyadmin/coding-standard
     tests:
         override:
--- a/composer.json
+++ b/composer.json
@@ -57,7 +57,9 @@
         "symfony/polyfill-mbstring": "^1.3"
     },
     "conflict": {
-        "tecnickcom/tcpdf": "<6.2"
+        "tecnickcom/tcpdf": "<6.2",
+        "pragmarx/google2fa": "<2.0",
+        "bacon/bacon-qr-code": "<1.0"
     },
     "suggest": {
         "ext-openssl": "Cookie encryption",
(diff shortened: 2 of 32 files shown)


ibennetch:
In my experience, this is mostly called "two factor authentication" rather than "second authentication factor" — is there any objection to me adjusting the documentation?

nijel:
I was unsure about naming - the problem is that there is 2FA (Two factor authentication) and U2F (Universal second factor), which we now support both. Wikipedia now calls this [Multi-factor authentication](https://en.wikipedia.org/wiki/Multi-factor_authentication).

ibennetch:
This is very nice, thank you.

pwallner:
Not working, at settings I see
`Two-factor authentication is not available, please install optional dependencies to enable authentication backends.`

Did a git clone and then `composer update`

ibennetch:
@pwallner Since the dependencies are categorized as suggested packages in Composer, a `composer update` doesn't install them automatically.

The only way I know to do this is to add them to the Composer requirements (through `composer require pragmarx/google2fa bacon/bacon-qr-code`, and also adding samyoul/u2f-php-server if you need FIDO U2F) . That tells Composer to add those as required dependencies, then it installs them for you. That should fix your problem.

pwallner:
Thx for help. 

2 Problems:
1. If I activate Google auth, everything goes well during configuration, but when I logout and login, no Code request appears and I can login without google auth?!
2. If I go to "Settings" with Chrome, I cannot select anything, because there is only one line which shows all options but without ratio buttons and without Button to send the request. If I hack the template `<option name="2fa_configure">` then everything is shown, but I cannot use U2F, because if I connect the USB U2F device, the device isn't recognized (LED isn't flashing). On other sites (e.g. gmail) the usb U2F device is working...

nijel:
Apparently this needs more testing in other browsers. Anyway can you please report these issues separately (one problem per issue) instead of using merged pull request for this? This would be way easier to track...

1 sounds strange, let's deal with it in #13828.

2 is fixed in c9d290a116e551bacaca3052df42e3fc5bceb1d2

ibennetch:
Also, we've updated the documentation about this at https://docs.phpmyadmin.net/en/latest/two_factor.html

pwallner:
1 see #13828 . It doesn't ask for google auth during login
2 is fixed, led flash and device is registered, but I can login with user and pass only, no need of U2F, see #13829

pwallner:
Back from vacation. Did a `git pull` and `composer update` and now my FIDO U2F and google auth is working.

PR #13793, opened 2017-11-01
focus table search box on Ctrl+shift+f

deepaspl:
focus table search box on Ctrl+shift+f

Signed-off-by: deeppanara <deep@aspl.in>


nijel:
Any reason for not using existing key handler?

https://github.com/phpmyadmin/phpmyadmin/blob/master/js/shortcuts_handler.js#L25

Also we should consider that overriding default browser behavior might be not desired, see https://github.com/phpmyadmin/phpmyadmin/issues/13798

PR #13797, opened 2017-11-02
Remove sprites from the themes

nijel:
I'd like to get feedback on possibility of removing sprites feature from themes. These days most serves will probably support http/2, so benefit of sprites is pretty much zero, however the code overhead it has is quite significant.

This patch removes the sprites code, however it does not yet fix all testsuite errors, this will be done if we agree this is reasonable thing to do.


mauriciofauth:
I agree.

ibennetch:
I don't have a strong feeling about this; it's simply not an area where I've followed industry trends. It seems you're right that it isn't a large benefit anymore, and it does cause us maintenance hassle with the code, so I'm fine with removing the sprites.

Related, I wonder if we should incorporate png optimization as part of the release process (or perhaps through some other automated means), so that we don't have to remember to do it manually when adding a new icon. Perhaps the gains from doing so aren't worth the effort we'd put in to it, I doubt we'd really gain that much from running pngcrush on a small icon anyway.

nijel:
Okay, I will work on this next week and merge once tests pass.

On the PNG side, I'm not really sure. The pngcrush tool is sometimes not really generating smallest possible image and usually it is best to tweak the things manually depending on number of colors really used in the image. Eg. in Weblate I've ended up with using pngnq + zopflipng for that (but thats for pngs generated from svg): https://github.com/WeblateOrg/weblate/blob/2e4264ebda080547e0ee357e29ff2aeace36bd99/graphics/svg2png.sh#L10-L11

PR #13812, opened 2017-11-14
min php version 5.5.9

Gemorroj:
`symfony/expression-language` 3.2 requires php 5.5.9
see https://github.com/symfony/expression-language/blob/3.2/composer.json#L19

Signed-off-by: Gemorroj <wapinet@mail.ru>

### DIFF
--- a/js/functions.js
+++ b/js/functions.js
@@ -528,7 +528,7 @@ function PMA_current_version (data) {
         var version_information_message_link = document.createElement('a');
         version_information_message_link.href = url;
         version_information_message_link.className = 'disableAjax';
-        version_information_message_link_text = document.createTextNode(data.version);
+        var version_information_message_link_text = document.createTextNode(data.version);
         version_information_message_link.appendChild(version_information_message_link_text);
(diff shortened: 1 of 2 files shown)


js/functions.js
```diff
@@ -528,7 +528,7 @@ function PMA_current_version (data) {
         var version_information_message_link = document.createElement('a');
         version_information_message_link.href = url;
         version_information_message_link.className = 'disableAjax';
-        version_information_message_link_text = document.createTextNode(data.version);
+        var version_information_message_link_text = document.createTextNode(data.version);
```
ibennetch: This change seems to be unrelated to the PHP version so should probably be a separate commit.

However, since these changes are so small, it may be easier to just redo the commit in your name than to resubmit; so don't do anything yet until some other team member can comment on how they want to handle the changes.

Gemorroj: I did it because I thought it would work "Signed-Off-By".

nijel:
I think it's better to keep the dependency consistent with motranslator to allow older PHP versions as well: https://github.com/phpmyadmin/motranslator/blob/86e6032aa89f5d139b34d03dce0099055c079fa9/composer.json#L20

PR #13838, opened 2017-11-27
Deprecation of "referral" directive

bofh16:
The "referral" directive in the "Content-Security-Policy" header has been deprecated in favor of a separate "Referrer-Policy" one.  Using the legacy setting makes FF/Chrome to display an error in the console.

**FF57.0**
selection_039

**Chromium 62.0.3202.94**
selection_040

### DIFF
--- a/libraries/Header.class.php
+++ b/libraries/Header.class.php
@@ -525,6 +525,9 @@ class PMA_Header
             );
         }
         header(
+            "Referrer-Policy: no-referrer"
+        );
+        header(
             "Content-Security-Policy: default-src 'self' "
             . $captcha_url
             . $GLOBALS['cfg']['CSPAllow'] . ';'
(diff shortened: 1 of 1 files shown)


nijel:
The MAINT_4_4_15 branch is no longer maintained, so it makes no sense to merge changes there.

The supported versions already include the Referrer-Policy header:

https://github.com/phpmyadmin/phpmyadmin/blob/5a31fffbe20958f448c6925f74038e8c1b86bcac/libraries/classes/Header.php#L543

And the referer policy is only included in X-Content-Security-Policy, which should be parsed by browser not compliant with current specification:

https://github.com/phpmyadmin/phpmyadmin/blob/5a31fffbe20958f448c6925f74038e8c1b86bcac/libraries/classes/Header.php#L562-L574

The compliant browsers should use Content-Security-Policy header:

https://github.com/phpmyadmin/phpmyadmin/blob/5a31fffbe20958f448c6925f74038e8c1b86bcac/libraries/classes/Header.php#L544-L561

PR #13850, opened 2017-11-30
PHP 7.2 Countable Warning

zbmowrey:
7.2 throws Warnings if trying to count a non-array, non-object, or object w/o the Countable interface. This adjustment prevents that warning when viewing any table's contents.
Signed-off-by: Zach Mowrey


zbmowrey:
Signed-off-by: Zachary Mowrey

PR #13863, opened 2017-12-12
Refactoring tests

carusogabriel:

I've refactored some tests, using:
- `assertCount` instead of `count` function;
- `assertSame` instead of strict comparison `===`;
- `assertNull` instead of comparison with `null` keyword;
- `assertFalse` and `assertNotFalse` instead of comparison with `false` keyword;
- `assertArrayHasKey` and `assertArrayNotHasKey` instead of `isset` function.


mauriciofauth:

PR #13867, opened 2017-12-14
Try to use more recent phpunit-selenium

nijel:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -80,7 +80,7 @@
     "require-dev": {
         "phpunit/phpunit": "^4.8.36 || ^5.7",
         "codacy/coverage": "^1.3.0",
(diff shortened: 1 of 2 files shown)


carusogabriel:
Some [Type Hinting problem](https://travis-ci.org/phpmyadmin/phpmyadmin/jobs/316404704#L710) going on at [`PMA_SeleniumBase`](https://github.com/phpmyadmin/phpmyadmin/blob/master/test/selenium/TestBase.php#L732)

nijel:
Yes that's what I've noticed as well, but in the end this was not the root cause - the root cause was stuck BrowserStack and bad error handling in the on phpunit-selenium which got pulled in after this PR.

PR #13871, opened 2017-12-17
Fix Row coloring issue@Indexes

### LINKED ISSUE #13090: Row coloring issue @ Indexes
pma

shucon:
Fixes https://github.com/phpmyadmin/phpmyadmin/issues/13090

### DIFF
--- a/templates/privileges/privileges_summary.twig
+++ b/templates/privileges/privileges_summary.twig
@@ -31,9 +31,9 @@
                     </tr>
                 {% else %}
                     {% for privilege in privileges %}
-                        {% include 'privileges/privileges_summary_row.twig' with [
+                        {% include 'privileges/privileges_summary_row.twig' with
                             privilege|merge({'type': type})
-                        ] only %}
+                        only %}
                     {% endfor %}
                 {% endif %}
             </tbody>


mauriciofauth:
This fixes the issue only in the sub rows, see the screenshot.
screenshot from 2017-12-17 10-31-47

This issue also happens with the Original theme, not only with pmahomme theme.

shucon:
@mauriciofauth I fixed the issue in Original theme too.

libraries/classes/Index.php
```diff
@@ -796,7 +798,12 @@ static public function getHtmlForIndexes($table, $schema, $print_mode = false)
 
             foreach ($index->getColumns() as $column) {
                 if ($column->getSeqInIndex() > 1) {
-                    $r .= '<tr class="noclick" >';
+                    if($iter%2 == 0){
+                        $r .= '<tr class="noclick" id="odd_row" >';
```
mauriciofauth: Using `id` here does not seem like a good idea, because ids need to be unique. See: [id in MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id)

mauriciofauth:
As I said earlier, your solution doesn't fully fix the issue.
For example, in the image below, the two lines have white background and this should not happen.
screenshot from 2017-12-17 10-31-47

shucon:
@mauriciofauth I couldn't reproduce the issue.

libraries/classes/Index.php
shucon: I tried using `odd_row` in class but it was somehow coming below in priority order , so it was not working as it is supposed to.

Issue #13854, opened 2017-12-06
Column options are ignored for GENERATED / VIRTUAL / STORED columns


mvorisek:
### Steps to reproduce
1. create or update any GENERATED / VIRTUAL / STORED column with "NOT NULL", specifict charset or type options like "UNSINGED"
2. save

### Example
Create table with one STORED column with type VARCHAR(10), collation ascii_general_ci and NOT NULL. This is currently not possible via GUI. The GUI currently also ignores these options on column def. change, resulting that all options are removed.

### Expected behaviour
Configured column options should be saved as normal (not generated) columns. Columns options are fully supported by MySQL and honored / validated as for normal columns.

### Actual behaviour
Column options are ignored in columns create or update via GUI.

### Server configuration
any

mvorisek:
Can someone fix this by setting the column options for GENERATED columns as usual for standard columns?

williamdes:
Can you try with version `4.8.1` ?
I am unable to reproduce this issue.

mvorisek:
Issue evaluated in 4.8.1:

Create table:
CREATE TABLE test (
  test int(11) NOT NULL,
  testvirt varchar(11) COLLATE utf8_bin GENERATED ALWAYS AS (test) VIRTUAL
) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_bin;

Go to the "Structure" tab and try to change the collation of the generated column to "utf8_general_ci".

Currently the PhpMyAdmin generates:
ALTER TABLE `test` CHANGE `testvirt` `testvirt` VARCHAR(11) AS (`test`) VIRTUAL; 

But it should generate something like this:
ALTER TABLE `test` CHANGE `testvirt` `testvirt` VARCHAR(11) CHARACTER SET utf8 COLLATE utf8_general_ci AS (`test`) VIRTUAL NULL;

williamdes:
MariaDB _does not support_ `COLLATE utf8_bin` in CREATE statement.
And the query the PMA _should_ generate does not work in MariaDB.

Maybe this explains why this issue.
I will try to investigate.

mvorisek:
Yes, tested on MySQL. Maria added the virtual columns support sooner and the syntax and features differs.

Here is my proposal for MySQL:
https://github.com/phpmyadmin/phpmyadmin/pull/13875/commits/d254e8accd3baa2dde322b60b17e096c84454486

[...]

PR #13875, opened 2017-12-18
Fix column options for virtual generated columns

mvorisek:
Fix issue #13854, based on official syntax:
column_definition: https://dev.mysql.com/doc/refman/5.7/en/create-table.html

Signed-Off-By: Michael Voříšek <mvorisek@mvorisek.cz>

### DIFF
--- a/libraries/classes/Table.php
+++ b/libraries/classes/Table.php
@@ -504,26 +504,27 @@ class Table
             }
         }
 
+        $matches = preg_match(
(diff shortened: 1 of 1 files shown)


mvorisek:
Can someone review and merge this pull?

mvorisek:
So is it not a bug?

williamdes:
@mvorisek Can you re-create this PR ?
Sorry that nobody could review the one you posted.

PR #13877, opened 2017-12-18
Fix Row coloring issue@Indexes 

### LINKED ISSUE #13090: Row coloring issue @ Indexes

shucon:
Signed-off-by: Saksham Gupta [shucon01@gmail.com](shucon01@gmail.com)

Fixes: https://github.com/phpmyadmin/phpmyadmin/issues/13090
Accidentally deleted PR : https://github.com/phpmyadmin/phpmyadmin/pull/13871


libraries/classes/Index.php
```diff
@@ -796,7 +798,12 @@ static public function getHtmlForIndexes($table, $schema, $print_mode = false)
 
             foreach ($index->getColumns() as $column) {
                 if ($column->getSeqInIndex() > 1) {
-                    $r .= '<tr class="noclick" >';
+                    if($iter%2 == 0){
+                        $r .= '<tr class="noclick" id="odd_row" >';
+                    }
+                    else{
+                        $r .= '<tr class="noclick" id="even_row" >';
```
mauriciofauth: I don't think this is the proper solution to this issue, as there is the CSS's [`:nth-child()`](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child) pseudo-class that serves that purpose.
I think a more appropriate solution would be to wrap the entire row and apply the pseudo-class to the entire row instead of just applying to the `<tr>` tag.
Perhaps this happens with other tables that also make use of rowspan. It would be good to check this out.

MauricioFauth (changes requested):

libraries/classes/Index.php
shucon: @mauriciofauth the CSS's :nth-child() pseudo-class is creating the problems , that's why I created another class to override it . I'll see what else could be done.

shucon:
@mauriciofauth please review the changes.

libraries/classes/Index.php
```diff
@@ -796,7 +798,12 @@ static public function getHtmlForIndexes($table, $schema, $print_mode = false)
 
             foreach ($index->getColumns() as $column) {
                 if ($column->getSeqInIndex() > 1) {
-                    $r .= '<tr class="noclick" >';
+                    if($iter%2 == 0){
```
mauriciofauth: I think that adding an iterator isn't the best option, besides not solving the problem completely, since `:nth-child()` pseudo-class already does this function.

A possible solution to this problem would be to add a `div`, or another tag, that wraps each index row and apply the pseudo-class to that `div` instead of applying it to the `tr`.

MauricioFauth (changes requested):

shucon:
@mauriciofauth I've done the required changes.

themes/pmahomme/css/common.css.php
```diff
@@ -554,13 +554,15 @@
...
+table tbody:first-of-type tr:nth-child(odd) th,
+#table_index tbody:nth-child(odd) tr {
     background: #fff;
 }
 
 /* even items 2,4,6,8,... */
 table tbody:first-of-type tr:nth-child(even),
-table tbody:first-of-type tr:nth-child(even) th {
+table tbody:first-of-type tr:nth-child(even) th,
+#table_index tbody:nth-child(even) tr {
```
mauriciofauth: This solution is much closer to the expected result, but it still has some problems.
See:
screenshot from 2018-01-25 19-50-31

The first problem is that the Keyname column is not being colored. But this can be solved by adding the `th` selector to the rule set. For example:
```css
#table_index tbody:nth-child(even) tr,
#table_index tbody:nth-child(even) th
```

Another problem is that the colors of the rows are inverted, because the `:nth-child()` pseudo-class is counting the `<thead>` as the first element. This can be solved by replacing `:nth-child()` with the [`:nth-of-type()`](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-of-type) pseudo-class.

MauricioFauth (changes requested):

MauricioFauth (approved):

mauriciofauth:
Merged, thanks for your contribution and patience with my feedback!

PR #13878, opened 2017-12-20
Update DatabaseInterface.php

jasgripen:
ignore errors if variable lc_messages is not supported by the server

Signed-off-by: Andreas Grip <andreas@wk.se>

### DIFF
--- a/libraries/classes/DatabaseInterface.php
+++ b/libraries/classes/DatabaseInterface.php
@@ -1446,7 +1446,7 @@ class DatabaseInterface
         /* Locale for messages */
         $locale = LanguageManager::getInstance()->getCurrentLanguage()->getMySQLLocale();
         if (! empty($locale)) {
-            $this->query(
+            $this->tryQuery(
                 "SET lc_messages = '" . $locale . "';",
                 DatabaseInterface::CONNECT_USER,
                 self::QUERY_STORE


nijel:
Do you get error here in other case than when running on unsupported MySQL server? Current phpMyAdmin releases require MySQL 5.5 or newer....

jasgripen:
I realize I run an unsupported version of MySQL, but so far this is the only thing I found that is not working.

And I must argue that not catching errors so the software crashes are always bad and since this is a non important function it's better to ignore any error and let the software continue. This function is not even run for all languages because MySQL do not support all languages that phpMyAdmin supports.

nijel:
Sorry, but silently ignoring errors is usually not a good idea. You will never know that this code is broken once you start to ignore errors. I really prefer to let user know that something is wrong than ignoring failure silently.

kvtb:
`lc_messages` is read-only in TiDB since recently (https://github.com/pingcap/tidb/pull/33708)
So it is the issue again

ibennetch:
As presented, this pull request wasn't found to be a good solution, so it's better to open an issue than to try to get this old pull request merged (you could reference it for some developer to use as a starting point, though). 

However, TiDB claims to be MySQL compatible, so it seems like the problem is on their end, not ours. We don't claim to be compatible with TiDB since we don't have the resources to test against their systems, but the phpMyAdmin behavior with `lc_messages` is compatible with MySQL.

kvtb:
Read only `lc_messages` is the only incompatibility

PR #13905, opened 2017-12-31
Use CSS to hide table structure action buttons

### LINKED ISSUE #13694: MenuResizer is VERY slow
It would be a great thing if loding of "MenuResizer" is optional:

e.g. having a table with 150 columns, most of them sets/enums, loading of the datastructure takes ~ 1 minute. After killing the above script loading was done in 1s!

This is really annoying to wait for the reformating, just to see the table-layout. Maybe there is some other solution, but this one is really slow.

MauricioFauth:
Fix #13694

### DIFF
--- a/js/tbl_structure.js
+++ b/js/tbl_structure.js
@@ -19,38 +19,6 @@
  */
 
 /**
- * This function returns the horizontal space available for the menu in pixels.
- * To calculate this value we start we the width of the main panel, then we
- * substract the margin of the page content, then we substract any cellspacing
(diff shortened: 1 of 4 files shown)


mauriciofauth:
A con to this solution is that the buttons will always be hidden regardless of screen size, but the page loads much faster.

nijel:
Sorry for not doing review, I don't feel well this week, so I'm a bit slower in doing things...

While I don't like that it's hiding the actions even when they would fit on the screen, I agree that this is better situation than the slow resizer.

PR #13917, opened 2018-01-08
Force PHP platform in composer to oldest supported version

### LINKED ISSUE #13831: PHP 5 / 7 compatibility
As more and more PHP libraries are starting to be PHP 7 only, we should address this somehow to ensure our releases will not break for some users. This came up in https://github.com/phpmyadmin/phpmyadmin/issues/13826, but can affect other parts of our code as well. In this particular case one library is using PHP 7 features, what fails on PHP 5. This seems to be correctly expressed in deps in this case, but this doesn't have to be true.

Anyway on the release time, we bundle libraries into tarball. This should work for all supported versions, not only for the one building release. For example right now the snapshots are built using PHP 7, so at least 2FA will fail in them on PHP 5.

Few ways to address this:

* **Always run release creation with oldest PHP we support.**

  This is probably achievable, however that means that we're relying on EOLed version to do the releases. That's probably not problem as it's used on command line only, but still feels strange.

  Same version should be used when @ibennetch is doing releases and on the server to build snapshots to have the releases and snapshots consistent.

* **Use composer configuration to enforce PHP 5.5 deps**

  Composer can be used to override PHP version for deps using `composer config platform.php 5.5`. Having this in our `composer.json` would make *all* installations behave like having PHP 5.5. This can be configured just on the releasing system by `composer config -g platform.php 5.5`.

* **Define PHP 5 dependencies in composer.json.**

  We need to keep eye on problematic libs so that composer always chooses older versions compatible with both PHP 5 and 7.

* **Drop PHP 5 support.**

  Sounds unrealistic as even dropping support for PHP 5.5 was not yet agreed, see https://github.com/phpmyadmin/phpmyadmin/issues/13228. However that would allow us to use modern PHP features such as type hinting...

Note that all but last options apparently lead to using old and thus possibly insecure or unsupported libraries in our releases.

Do you see other options how to deal with this?

nijel:
This way the packages should be compatible with all versions we support
regardless PHP version used to build the release.

Fixes #13831

### DIFF
--- a/scripts/create-release.sh
+++ b/scripts/create-release.sh
@@ -220,12 +220,18 @@ rm -f .travis.yml .coveralls.yml .scrutinizer.yml .jshintrc .weblate codecov.yml
 rm -f README.rst
 
 if [ ! -d libraries/tcpdf ] ; then
-    echo "* Running composer"
-    composer update --no-dev
+    PHP_REQ=`sed -n '/"php"/ s/.*">=\([0-9]\.[0-9]\).*/\1/p' composer.json`
+    if [ -z "$PHP_REQ" ] ; then
+        echo "Failed to figure out required PHP version from composer.json"
+        exit 2
+    fi
     # Okay, there is no way to tell composer to install
     # suggested package. Let's require it and then revert
     # composer.json to original state.
     cp composer.json composer.json.backup
+    echo "* Running composer"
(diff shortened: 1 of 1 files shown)


nijel:
Yes, it can be set in composer.json, but we don't want to force this in all cases, just for the generated releases to be sure they are compatible with oldest supported PHP version.

Issue #13124, opened 2017-03-27
Table SQL: Align Query and Columns fields


OlafvdSpek:
Could the Columns field be extended downward to consume the empty space?

![pma](https://cloud.githubusercontent.com/assets/2218936/24347004/1d74c4c2-12d7-11e7-9233-1e3351e3b2d1.PNG)

nijel:
Certainly they could be sized same as the textarea.

Achilles-96:
This is how I have it.
![screenshot_20170327_164011](https://cloud.githubusercontent.com/assets/10742413/24353659/2e7cef08-130c-11e7-96dd-67efbe7d164a.png)

OlafvdSpek:
Could the difference be due to $cfg['CodemirrorEnable'] = false; ?

nijel:
Yes, the default dimensions are different. Also as the editing area is resizeable, the height of columns selector should match it even when resizing.

OlafvdSpek:
Should the default size be different?

nijel:
That's probably just a side effect of different fonts, the number of rows should be same and should match [ $cfg\['TextareaRows'\]](https://docs.phpmyadmin.net/en/latest/config.html#cfg_TextareaRows).

Fenn-CS:
Thats the reason why the pages have to be responsive to avoid such errors caused by user preferences and browsers like in this case fonts.

JSn1nj4:
Hi, is there still room to jump in on this? I'd like to start contributing.

KasperFranz:
Is this not already implemented - when I just tried to open this it looked like it was correct

williamdes:
@JSn1nj4 Are you still interested in contributing ?

willwalsh96:
I want to try this issue. How do I see how my changes affect the running application? I do not want to host my own server to do testing.

williamdes:
@willwalsh96 you could do the changes online and submit a patch afterwards
For sure having a local php web server would be a better idea
Demo servers (`root`, no password)
- [`5.1.0-dev`](https://demo.phpmyadmin.net/master/)
- [`5.0.x-dev`](https://demo.phpmyadmin.net/QA_5_0/)
- [`4.9.x`](https://demo.phpmyadmin.net/QA_4_9/)

JSn1nj4:
@williamdes I'm sorry, I was, but I'm too short on time to add a project into my schedule. Sorry for the late reply.

[...]

PR #13946, opened 2018-01-18
change #tablefieldscontainer select height setting to syn better with .CodeMirror height for #13124

xypXiong:


xypXiong:
after adjusting, firefox and chrome look nicer, however, IE looks more off. Seems can't make all three  good altogether

xypXiong:
If we use pixel instead of em, we could get all three browsers behave same . Pixel is independent with the font size. But they will be not something related with $cfg['TextareaRows'] settings. Any thought about it?

mauriciofauth:
Hi, @xypXiong. Thanks for your contribution!
You included the `.metadata` directory, but I don't think that was your intention. Could you remove it?

ibennetch:
There's certainly a place for using pixels instead of em or percentages, but in general I try to avoid defining sizes in pixels. I know we do it a lot in the theme code, and it may be the best way here, but just my two cents is that using pixels isn't always the best solution. Others may disagree, of course.

xypXiong:
@mauriciofauth I have removed the .meta dir, please take a look. @ibennetch the current version still uses em which makes firefox and chrome looks better. Hope this something at least better than before. thanks

themes/pmahomme/css/common.css.php
```diff
@@ -1622,7 +1622,8 @@
 div#tablefieldscontainer select {
     width: 100%;
     background: #fff;
-    /* height: 12em; */
+   /* height: <?php echo ceil($GLOBALS['cfg']['TextareaRows'] * 1.2)-3.35; ?>em; */
+     height: 192px;
```
mauriciofauth: Using a fixed value in pixels is not a good choice because height is influenced by [`$cfg['TextareaRows']`](https://docs.phpmyadmin.net/en/latest/config.html#cfg_TextareaRows) and [`$cfg['CodemirrorEnable']`](https://docs.phpmyadmin.net/en/latest/config.html#cfg_CodemirrorEnable).

MauricioFauth (changes requested):

xypXiong:
Thanks, currently I don't use fixed pixels. 

    Maurício Meneghini Fauth <notifications@github.com> 于 2018年3月23日, 星期五, 下午 11:54 写道：
 

 @mauriciofauth requested changes on this pull request.In themes/pmahomme/css/common.css.php:> @@ -1622,7 +1622,8 @@
 div#tablefieldscontainer select {
     width: 100%;
     background: #fff;
-    /* height: 12em; */
+   /* height: <?php echo ceil($GLOBALS['cfg']['TextareaRows'] * 1.2)-3.35; ?>em; */
+     height: 192px;
Using a fixed value in pixels is not a good choice because height is influenced by $cfg['TextareaRows'] and $cfg['CodemirrorEnable'].—

PR #13970, opened 2018-02-06
Adjustments in syntax

Edumatsu:


browse_foreigners.php
```diff
@@ -39,23 +39,23 @@
...
 );
 
 $foreignData = Relation::getForeignData(
     $foreigners, $_REQUEST['field'], true,
-    isset($_REQUEST['foreign_filter'])
-    ? $_REQUEST['foreign_filter']
-    : '',
-    isset($foreign_limit) ? $foreign_limit : null,
+    $_REQUEST['foreign_filter'] ?: '',
+    $foreign_limit ?: null,
```
mauriciofauth: There is the possibility that these variables are undefined. If so, this can generate a PHP notice.

Issue #13969, opened 2018-02-06
Profiling breaks user session


mauriciofauth:
### Steps to reproduce
1. Browse some table
2. Click the Profiling checkbox

### Expected behaviour
Do the profiling

### Actual behaviour
Breaks the user session and the only way to resolve is to start a new session.

![screenshot-profiling](https://user-images.githubusercontent.com/120970/35834880-d24818d8-0abe-11e8-8677-a388dfe80924.png)

```
PHP Fatal error:  Uncaught Error: Call to a member function getScripts() on null in phpmyadmin/libraries/common.inc.php:428
Stack trace:
#0 phpmyadmin/ajax.php(15): require_once()
#1 {main}
  thrown in phpmyadmin/libraries/common.inc.php on line 428
```
```
Notice in ./libraries/common.inc.php#428
 Undefined property: PhpMyAdmin\Header::$header
Backtrace
./index.php#26: require_once(./libraries/common.inc.php)
```

### Server configuration
Demo server
29d92a1 from master branch

nulll-pointer:
I have made a few changes and now it works for me. 
![screenshot_20180207_163737](https://user-images.githubusercontent.com/20344314/35913345-3763ec40-0bf7-11e8-9829-8fa4efe5ce9d.png)

Should i make a PR.

mauriciofauth:
@nulll-pointer Absolutely!

nulll-pointer:
I have a pr pending on the master branch.
https://github.com/phpmyadmin/phpmyadmin/pull/13971

mauriciofauth:
@nulll-pointer Create a new branch for each pull request, this way each PR becomes independent. It's not recommended to commit directly to the master branch.

PR #13982, opened 2018-02-09
Fix for issue #13969

aroralakshya:
Changed file ./libraries/common.inc.php
Removed member header from line '$response->getHeader()->header->getScripts()', there is no such member.

### DIFF
--- a/libraries/common.inc.php
+++ b/libraries/common.inc.php
@@ -425,7 +425,7 @@ if (! defined('PMA_MINIMUM_COMMON')) {
      */
     $response = Response::getInstance();
     if (isset($_SESSION['profiling'])) {
-        $scripts  = $response->getHeader()->header->getScripts();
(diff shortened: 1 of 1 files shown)


libraries/common.inc.php
```diff
@@ -425,7 +425,7 @@
      */
     $response = Response::getInstance();
     if (isset($_SESSION['profiling'])) {
-        $scripts  = $response->getHeader()->header->getScripts();
+        $scripts  = $response->getHeader()->getScripts(); // Fix for issue 13969
```
mauriciofauth: I don't think this comment is relevant. This type of information already exists in the commit message.

MauricioFauth (changes requested):

libraries/common.inc.php
nulll-pointer: I'll make the changes.

PR #13987, opened 2018-02-10
Prompt before reducing number of rows

### LINKED ISSUE #13952: Insert page should prompt before reducing number of rows
On the Insert page, when we add rows to the page with the "Continue insertion with __ rows" function, there is a risk of data loss if the user adds data to these rows then accidentally reduces the number of rows. When changing the number of rows, we should check if we're reducing the number, and if we are then we should check if there's been data input and prompt the users.

shucon:
Fixes: https://github.com/phpmyadmin/phpmyadmin/issues/13952

### DIFF
--- a/js/ajax.js
+++ b/js/ajax.js
@@ -512,23 +512,21 @@ var AJAX = {
      * Displays alert if data loss possible on decrease
      * of rows.
      */
-    rowCheck: function (row,row_count) {
-        var new_count = Number(row.value);
-        if (new_count < row_count && !jQuery.isEmptyObject(AJAX.lockedTargets)
+    rowCheck: function (row, rowCount) {
+        var newCount = Number(row.value);
+        if (newCount < rowCount && !jQuery.isEmptyObject(AJAX.lockedTargets)
             && confirm(PMA_messages.strConfirmRowChange) === false
         ) {
-            row.value = row_count;
-            return row_count;
+            row.value = rowCount;
+            return rowCount;
         }
 
-        if (new_count < row_count && !jQuery.isEmptyObject(AJAX.lockedTargets)
+        if (newCount < rowCount && !jQuery.isEmptyObject(AJAX.lockedTargets)
             && confirm(PMA_messages.strConfirmRowChange) === true
         ) {
-            return new_count;
-        }
-
-        else {
-            return new_count;
+            return newCount;
+        } else {
+            return newCount;
         }
     },
     /**
--- a/js/ajax.js
+++ b/js/ajax.js
@@ -750,17 +750,6 @@ AJAX.registerOnload('functions.js', function () {
     $('form.lock-page').on('reset', function (event) {
         AJAX.resetLock();
     });
-    /**
-     * Displays warning if number of rows are reduced
-     * and chances of data loss.
-     */
-    var row = document.getElementById('insert_rows');
-    if (row !== null) {
-        var row_count = Number(row.value);
-        row.addEventListener("change", function(){
-            row_count = AJAX.rowCheck(row,row_count);
-        });
-    }
 });
 
 /**
--- a/js/ajax.js
+++ b/js/ajax.js
@@ -509,21 +509,6 @@ var AJAX = {
         }
     },
     /**
-     * Displays alert if data loss possible on decrease
-     * of rows.
-     */
-    rowCheck: function (row, rowCount) {
-        var newCount = Number(row.value);
-        if (newCount < rowCount && !jQuery.isEmptyObject(AJAX.lockedTargets)
-            && confirm(PMA_messages.strConfirmRowChange) === false
-        ) {
-            row.value = rowCount;
-            return rowCount;
-        } else {
-            return newCount;
-        }
-    },
-    /**
      * This object is in charge of downloading scripts,
      * keeping track of what's downloaded and firing
      * the onload event for them when the page is ready.
--- a/js/ajax.js
+++ b/js/ajax.js
@@ -509,6 +509,29 @@ var AJAX = {
         }
     },
     /**
+     * Displays alert if data loss possible on decrease
+     * of rows.
+     */
+    rowCheck: function (row,row_count) {
+        var new_count = Number(row.value);
+        if (new_count < row_count && !jQuery.isEmptyObject(AJAX.lockedTargets)
+            && confirm(PMA_messages.strConfirmRowChange) === false
+        ) {
+            row.value = row_count;
+            return row_count;
+        }
+
+        if (new_count < row_count && !jQuery.isEmptyObject(AJAX.lockedTargets)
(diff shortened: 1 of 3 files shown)


js/messages.php
```diff
@@ -69,6 +69,8 @@ function () {
     = __('Do you really want to delete the search "%s"?');
 $js_messages['strConfirmNavigation']
     = __('You have unsaved changes; are you sure you want to leave this page?');
+$js_messages['strConfirmRowChange']
+    = __('You have unsaved changes; are you sure you want to reduce number of rows?');
```
ibennetch: Your message is okay, but I've been trying to make error messages and warnings more user-friendly. How does "You are trying to reduce the number of rows, but have already entered data in those rows which will be lost. Do you wish to continue?" sound to everyone?

ibennetch:
I haven't tested it but this looks quite good.

js/ajax.js
```diff
@@ -727,6 +750,17 @@ AJAX.registerOnload('functions.js', function () {
...
         AJAX.resetLock();
     });
+    /**
+     * Displays warning if number of rows are reduced
+     * and chances of data loss.
+     */
+    var row = document.getElementById('insert_rows');
+    if (row !== null) {
+        var row_count = Number(row.value);
+        row.addEventListener("change", function(){
```
mauriciofauth: An event listener already exists for this selector. Why not use it?

https://github.com/phpmyadmin/phpmyadmin/blob/ef99fef8d6b760a2f3552c89afe50ba6d5562add/js/tbl_change.js#L463

js/ajax.js
```diff
@@ -508,6 +508,27 @@ var AJAX = {
...
+
+        if (newCount < rowCount && !jQuery.isEmptyObject(AJAX.lockedTargets)
+            && confirm(PMA_messages.strConfirmRowChange) === true
+        ) {
+            return newCount;
+        } else {
+            return newCount;
+        }
+    },
     /**
```
devenbansod: @shucon I'm not sure if this file would be an ideal place to add this function. It's not related to AJAX as such, more related to a specific page.

shucon: @devenbansod I did this for two reasons.
1. A similar function for the same page is written here (line 249).
2. I had to use AJAX.lockedTargets . Although I could use it anywhere but writing it here seemed right because of the above reason.

js/ajax.js
```diff
@@ -508,6 +508,21 @@ var AJAX = {
             }
         }
     },
+    /**
+     * Displays alert if data loss possible on decrease
+     * of rows.
+     */
+    rowCheck: function (row, rowCount) {
+        var newCount = Number(row.value);
+        if (newCount < rowCount && !jQuery.isEmptyObject(AJAX.lockedTargets)
```
mauriciofauth: You don't need to check if the number of rows the user entered is less than or greater than the number of rows on the page. This check already exists, you only need to add the confirmation message.

js/tbl_change.js
```diff
@@ -460,7 +460,17 @@ AJAX.registerOnload('tbl_change.js', function () {
...
+    if (row !== null) {
+        var rowCount = Number(row.value);
+    }
     $(document).on('change', '#insert_rows', function (event) {
+
+        /**
+         * Displays warning if number of rows are reduced
+         * and chances of data loss.
+         */
+        rowCount = AJAX.rowCheck(row, rowCount);
```
mauriciofauth: You are returning a value, but you are not using it for anything.

MauricioFauth (changes requested):

js/ajax.js
mauriciofauth: I have to agree with Deven.

js/ajax.js
shucon: Where does this check exists? Also , I would have to provide some condition to show the confirmation message.

js/tbl_change.js
shucon: I'm using the returned value to update the old row count and compare it with the new count that the user wants to check the rowCheck condition.

js/ajax.js
shucon: Should I transfer the code to tbl_change.js ?

js/ajax.js
mauriciofauth: This check is already done when adding or removing lines from the page.

https://github.com/phpmyadmin/phpmyadmin/blob/ef99fef8d6b760a2f3552c89afe50ba6d5562add/js/tbl_change.js#L483

https://github.com/phpmyadmin/phpmyadmin/blob/ef99fef8d6b760a2f3552c89afe50ba6d5562add/js/tbl_change.js#L669

shucon: Thanks @mauriciofauth I've done the required changes.

js/tbl_change.js
```diff
@@ -667,12 +672,22 @@ AJAX.registerOnload('tbl_change.js', function () {
...
-                    .addBack()
-                    .remove();
-                curr_rows--;
+            /**
+             * Displays alert if data loss possible on decrease
+             * of rows.
+             */
+            if (jQuery.isEmptyObject(AJAX.lockedTargets) || (!jQuery.isEmptyObject(AJAX.lockedTargets)
+                && confirm(PMA_messages.strConfirmRowChange) === true)
+            ) {
```
mauriciofauth: Don't perform complex actions in if statements. Put this jQuery call into a variable.

This second jQuery call is redundant and can be removed. The right side of the `||` will only run if the left side is false.

js/tbl_change.js
```diff
@@ -667,12 +672,22 @@ AJAX.registerOnload('tbl_change.js', function () {
...
+                while (curr_rows > target_rows) {
+                    $('input[id^=insert_ignore]:last')
+                        .nextUntil('fieldset')
+                        .addBack()
+                        .remove();
+                    curr_rows--;
+                }
+            } else {
+                    document.getElementById('insert_rows').value = curr_rows;
```
mauriciofauth: Fix the indentation of this block.

MauricioFauth (changes requested):

js/tbl_change.js
```diff
@@ -461,6 +461,11 @@ AJAX.registerOnload('tbl_change.js', function () {
      * Continue Insertion form
      */
     $(document).on('change', '#insert_rows', function (event) {
+
+        /**
+         * Displays warning if number of rows are reduced
+         * and chances of data loss.
+         */
```
mauriciofauth: Is this comment really necessary?

shucon: I left it by mistake , thanks.

js/tbl_change.js
shucon: @mauriciofauth this jQuery call is used in ajax.js too without variable

MauricioFauth (approved):

mauriciofauth:
Squashed and merged.
Thanks for your contribution and patience with the feedback.

PR #14001, opened 2018-02-12
Fix table horizontal scrolling

aroralakshya:
Changed files:
	themes/original/css/common.css.php
	themes/pmahomme/css/common.css.php
Added style definition for .responsivetable element to prevent horizontal overflow

### DIFF
--- a/themes/original/css/common.css.php
+++ b/themes/original/css/common.css.php
@@ -3324,7 +3324,7 @@ body .ui-dialog .ui-button-text-only .ui-button-text {
 }
 
 .responsivetable {
-    overflow-x: auto !important;
+    overflow-x: auto;
 }
 
 @media only screen and (max-width: 768px) {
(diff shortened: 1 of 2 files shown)


themes/original/css/common.css.php
```diff
@@ -3323,6 +3323,10 @@
     display: none;
 }
 
+.responsivetable {
+    overflow-x: auto !important;
```
mauriciofauth: Using `!important` is generally not recommended. Could you solve it differently?

MauricioFauth (changes requested):

themes/original/css/common.css.php
nulll-pointer: It works without important, i just added it for safety reasons.

nulll-pointer: i could remove it if you suggest.

mauriciofauth:

Issue #13941, opened 2018-01-16
Argument 5 passed to PhpMyAdmin\Table::_getSQLToCreateForeignKey() must be of the type array, string given


xuefer:
### Steps to reproduce
1. submit Foreign key constraints

### Expected behaviour
Foreign key constraints added

### Actual behaviour
```
Uncaught TypeError: Argument 5 passed to PhpMyAdmin\Table::_getSQLToCreateForeignKey() must be of the type array, string given, called in /var/www/htdocs/phpMyAdmin/libraries/classes/Table.php on line 2384 and defined in /var/www/htdocs/phpMyAdmin/libraries/classes/Table.php:2464
Stack trace:
#0 /var/www/htdocs/phpMyAdmin/libraries/classes/Table.php(2384): PhpMyAdmin\Table-&gt;_getSQLToCreateForeignKey('users', Array, 'tracker', 'lkp', 'type', '', 'RESTRICT', 'RESTRICT')
#1 /var/www/htdocs/phpMyAdmin/libraries/classes/Controllers/Table/TableRelationController.php(253): PhpMyAdmin\Table-&gt;updateForeignKeys(Array, Array, Array, Array, Array, 'users', Array)
#2 /var/www/htdocs/phpMyAdmin/libraries/classes/Controllers/Table/TableRelationController.php(127): PhpMyAdmin\Controllers\Table\TableRelationController-&gt;updateForForeignKeysAction()
#3 /var/www/htdocs/phpMyAdmin/tbl_relation.php(70): PhpMyAdmin\Controllers\Table\TableRelationController-&gt;indexAction()
#4 {main}
 thrown in /var/www/htdocs/phpMyAdmin/libraries/classes/Table.php on line 2464
```

### Server configuration
**Operating system**: Gentoo

**Web server:** lighttpd

**Database:** mariadb

**PHP version:** 7.1

**phpMyAdmin version:** git master rev 2823d27cd3

### Client configuration
**Browser:** Chrome

**Operating system:** Windows

PR #14031, opened 2018-02-20
Fixes error in foreign key sql generation + drop foreign key contraint

aroralakshya:
Changed:
	templates/table/relation/foreign_key_row.twig
The destination_column_names were getting passed as a 1d array, the names when extracted gave a string. Passing it as a 2d array (see templates/table/relation/foreign_key_row.twig lines 123,133) fixes the issue.


Gemorroj:
I think it is also necessary to correct the error display.
The response "success: true", which is wrong.

nulll-pointer:
I thing making success:false also expects an error parameter in the json object returned, the error is generated during parsing so i'm still figuring out how to display the message.

mauriciofauth:
Related to #13941

mauriciofauth:

Issue #14009, opened 2018-02-14
Problem with saving the relation view


BloomCake:
### Steps to reproduce
1. Open the Relation View of a table
2. Save your settings

### Expected behaviour
The Foreign key constraints should be saved without error.

### Actual behaviour
I get a error message in phpmyadmin 
`Error in processing request
Error code: 500
Error text: error (rejected)
It seems that the connection to server has been lost. Please check your network connectivity and server status.`

and an error in the nginx log
`2018/02/14 11:26:34 [error] 1050#1050: *3617 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Uncaught Error: Cannot unset string offsets in /opt/phpmyadmin/libraries/classes/Table.php:2307
Stack trace:
#0 /opt/phpmyadmin/libraries/classes/Controllers/Table/TableRelationController.php(252): PhpMyAdmin\Table->updateForeignKeys(Array, Array, Array, Array, Array, 'duplicity', Array)
#1 /opt/phpmyadmin/libraries/classes/Controllers/Table/TableRelationController.php(126): PhpMyAdmin\Controllers\Table\TableRelationController->updateForForeignKeysAction()
#2 /opt/phpmyadmin/tbl_relation.php(70): PhpMyAdmin\Controllers\Table\TableRelationController->indexAction()
#3 {main}
  thrown in /opt/phpmyadmin/libraries/classes/Table.php on line 2307" while reading response header from upstream, client: 10.8.0.42, server: db1.server.com, request: "POST /tbl_relation.php HTTP/2.0", upstream: "fastcgi://unix:/run/php/php7.1-fpm.sock:", host: "db1.server.com"`

On another server with the same setup and the phpmyadmin version 4.2.12deb2+deb8u2 it works fine without errors.

### Server configuration
**Operating system**: Ubuntu 16.04.3 LTS

**Web server:** NGINX (1.10.3 (Ubuntu))

**Database:** MariaDB-10.2.12

**PHP version:** 7.1.14-1

**phpMyAdmin version:** 4.8.0-dev (Git revision: 095c3a8)

### Client configuration
**Browser:** Google Chrome (62.0.3202.62)

**Operating system:** Ubuntu 17.10

shucon:
I'm starting to work on this.

shucon:
Related to https://github.com/phpmyadmin/phpmyadmin/issues/13941

PR #14010, opened 2018-02-14
Fix broken Relation View

shucon:
Signed-off-by: Saksham Gupta <shucon01@gmail.com>

Before submitting pull request, please check that every commit:
Fixes: https://github.com/phpmyadmin/phpmyadmin/issues/14009
- [x] Has proper Signed-Off-By
- [x] Has commit message which describes it
- [x] Is needed on it's own, if you have just minor fixes to previous commits, you can squash them
- [x] Any new functionality is covered by tests

PR #14037, opened 2018-02-22
fix: wrong argument type passed when creating foreign key

philly-php:
we were passing a string into the function to create sql to create a foreign key constraint. This function expects an array, since foreign keys can reference multiple columns.

Fixes Issue #13941

Signed-off-by: Leonid Sklyut <lsklyut@kbra.com>

- [x ] Has proper Signed-Off-By

### DIFF
--- a/libraries/classes/Table.php
+++ b/libraries/classes/Table.php
@@ -2286,7 +2286,7 @@ class Table
             $master_field = $multi_edit_columns_name[$master_field_md5];
 
             $foreign_table = $destination_foreign_table[$master_field_md5];
-            $foreign_field = $destination_foreign_column[$master_field_md5];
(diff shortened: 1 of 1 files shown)


nijel (changes requested):
I think this should be addressed elsewhere, similarly as https://github.com/phpmyadmin/phpmyadmin/pull/14010

lsklyut:
@nijel , I've pulled the array casting up to where we grab the $foreign_field from $destination_foreign_column. As it stands, it does not look like this input would ever be anything but a string, even though the code further down and in _getSQLToCreateForeignKey relies on it being an array. I do think the array logic makes sense, since mysql does allow multi-column foreign keys (although it doesn't look like the UI allows you to do that yet)

lsklyut:
this would also fix #14009

PR #14040, opened 2018-02-22
Add Yarn install to create-release script

### LINKED ISSUE #13918: Add Yarn to create-release script
We need to run `yarn install --production` to include the `node_modules` directory in the `phpMyAdmin-4.8.0-source.tar.xz` file.

This way jQuery source code will be included, since it was removed in #13706.

MauricioFauth:
Fix #13918

### DIFF
--- a/scripts/create-release.sh
+++ b/scripts/create-release.sh
@@ -273,6 +273,11 @@ if [ ! -d libraries/tcpdf ] ; then
     fi
 fi
 
+if [ -f package.json ] ; then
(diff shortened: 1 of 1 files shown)


nijel:
This needs to be approved by @ibennetch as he is doing the releases.

ibennetch:
I don't have much of an opinion about this. I'm fine with it in terms of managing the release; if this is the solution you think is best then go for it.

ibennetch (approved):

PR #14046, opened 2018-02-28
Restore SQL query after session expire

### LINKED ISSUE #13654: Add modal to login after session expired
Hi, Many times i want continue my work on phpmyadmin when click on something this will go to login page and i lost my data on page.
I think its good idea to create a login modal and open when detect session expired, and login on modal immediately and do not redirect to any page. Just like Wordpress.
Thanks all.

shucon:
Partially Fixes: https://github.com/phpmyadmin/phpmyadmin/issues/13654

### DIFF
--- a/js/sql.js
+++ b/js/sql.js
@@ -865,6 +865,11 @@ AJAX.registerOnload('sql.js', function () {
      * create resizable table
      */
     $('.sqlqueryresults').trigger('makegrid').trigger('stickycolumns');
+
+    /**
+     * Restores SQL query after timeout login
+     */
+    codemirror_editor.setValue(window.localStorage.auto_saved_sql);    
 });
 
 /*
--- a/js/sql.js
+++ b/js/sql.js
@@ -41,12 +41,10 @@ function PMA_urlencode (str) {
  * @return void
  */
 function PMA_autosaveSQL (query) {
-    if (query) {
-        if (isStorageSupported('localStorage')) {
-            window.localStorage.auto_saved_sql = query;
-        } else {
-            Cookies.set('auto_saved_sql', query);
-        }
+    if (isStorageSupported('localStorage')) {
+        window.localStorage.auto_saved_sql = query;
+    } else {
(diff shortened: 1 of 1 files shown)


shucon:
Once this is approved I'll try to do something similar for Insert page.

js/sql.js
```diff
@@ -865,6 +865,11 @@ AJAX.registerOnload('sql.js', function () {
      * create resizable table
      */
     $('.sqlqueryresults').trigger('makegrid').trigger('stickycolumns');
+
+    /**
+     * Restores SQL query after timeout login
+     */
+    codemirror_editor.setValue(window.localStorage.auto_saved_sql);
```
nijel: There is already code to handle this:

https://github.com/phpmyadmin/phpmyadmin/blob/733a3f1c0b16ca0fe486fecf5940c3143fdda651/js/functions.js#L1180-L1187

So it should load it from cookies as well (`insertQuery('saved')` would do it, but it would also error on no saved query what is not desired behavior here).

nijel (changes requested):
It should handle values stored in cookies as well, preferably reusing existing code.

shucon:
@nijel I think it's ok now, can you please review it?

js/sql.js
```diff
@@ -41,12 +41,10 @@ function PMA_urlencode (str) {
...
-    if (query) {
-        if (isStorageSupported('localStorage')) {
-            window.localStorage.auto_saved_sql = query;
-        } else {
-            Cookies.set('auto_saved_sql', query);
-        }
+    if (isStorageSupported('localStorage')) {
+        window.localStorage.auto_saved_sql = query;
+    } else {
+        Cookies.set('auto_saved_sql', query);
```
nijel: Should it really save the empty query here?

shucon: Yes, because if the query is empty then on restore the last character was saved and got restored, that wasn't looking good.

nijel:

Issue #14059, opened 2018-03-06
Indexes are not created on Structure tab after update to 4.7.9


yura3d:
### Steps to reproduce
1. Open a **Structure** tab.
2. Click an **Primary**/**Unique**/**Index** link at **Action** column on any field row.

### Expected behaviour
An index should be created.

### Actual behaviour
An index is not created on **4.7.9**. On **4.7.8** and below everything was fine.

### Server configuration
**Operating system**: Debian 9.3 (4.9.0-6-amd64)

**Web server:** nginx/1.12.2

**Database:** 5.7.21-20 - Percona Server (GPL), Release '20', Revision 'ed217b06ca3'

**PHP version:** 7.2.3

**phpMyAdmin version:** 4.7.9

### Client configuration
**Browser:** Google Chrome 64.0.3282.186 x64, Mozilla Firefox 58.0.2 x64

**Operating system:** Windows 10 Pro x64 (1709, build 16299.248)

yura3d:
Also, I see the same behaviour on the demo server (https://demo.phpmyadmin.net/master-config/), which has **4.8.0-dev** installed version.

kampalex:
I also ran in this bug - trying to resolve using differences between 4.7.8 and 4.7.9.

The way requests are fired to tbl_structure.php is different. In 4.7.9, all GET values are merged into a POST request. Method getJSConfirmCommonParam (/js/functions.js) collects the arguments.

=============== 4.7.8 ==================
**GET**
db: (database)
table: (table)
token: (token)
goto: tbl_structure.php
back: tbl_structure.php
add_key: 1
sql_query: ALTER TABLE `(table)` ADD INDEX(`(field)`);
message_to_show: (Message on success)
**POST**
ajax_request: true
ajax_page_request: true
token: (token)
_nocache: (timestamp)

=============== 4.7.9 ==================
**POST**
db: (database)
table: (table)
token: (token)
goto: tbl_structure.php
back: tbl_structure.php
add_key: 1
sql_query: ALTER TABLE `(table)` ADD INDEX(`(field)`);
message_to_show: (Message on success)
is_js_confirmed: 1
ajax_request: true
_fk_checks0:_ 
ajax_page_request: 1
_nocache: (timestamp)

I've spotted a migration bug here:
/js/functions.js - r1856
`params += 'is_js_confirmed=1&ajax_request=true&fk_checks' + ($elem.find('#fk_checks').is(':checked') ? 1 : 0);`
[...]

PR #14075, opened 2018-03-12
Fixes for issue 14059

kampalex:
Fixes for issue #14059
Missing '=' after fk_checks (value was appended onto field name)
sql_query variable not set due to GET-POST method change in tbl_structure.php 'add_key' situation

Signed-off-by: Alexander Kamp git@akamp.nl


mauriciofauth:

PR #14095, opened 2018-03-18
added a keyboard shortcut n for directly jumping for creating a new database

akshilshah:
added a keyboard shortcut n for directly jumping for creating a new database

Signed-off-by: Akshil shah <akshilshah4444@gmail.com>

### DIFF
--- a/js/shortcuts_handler.js
+++ b/js/shortcuts_handler.js
@@ -22,6 +22,7 @@ $(document).ready(function () {
     var keyH = 72;
     var keyC = 67;
     var keyBackSpace = 8;
+    var keyN = 78;
     $(document).on('keyup', function (e) {
         if (e.target.nodeName === 'INPUT' || e.target.nodeName === 'TEXTAREA' || e.target.nodeName === 'SELECT') {
             return;
@@ -96,6 +97,8 @@ $(document).ready(function () {
             window.history.back();
         } else if (e.keyCode === keyH) {
             $('.ic_b_home').first().trigger('click');
+        } else if (e.keyCode === keyN) {
+            window.open('http://localhost/phpmyadmin/server_databases.php?server=1', '_self');
         }
     });
 });


js/vendor/jquery/jquery-migrate.js
```diff
@@ -1,626 +1,215 @@
...
-
-	return deferred;
-};
-
-// Preserve handler of uncaught exceptions in promise chains
-jQuery.Deferred.exceptionHook = oldDeferred.exceptionHook;
-
-return jQuery;
-} );
+/*! jQuery Migrate v3.0.1 | (c) jQuery Foundation and other contributors | jquery.org/license */
```
mauriciofauth: Hi, @akshil4444. Updating jQuery Migrate does not seem to be related to this PR. Could you remove this change? Thank you.

js/shortcuts_handler.js
```diff
@@ -96,6 +97,8 @@ $(document).ready(function () {
             window.history.back();
         } else if (e.keyCode === keyH) {
             $('.ic_b_home').first().trigger('click');
+        } else if (e.keyCode === keyN) {
+            window.open('http://localhost/phpmyadmin/server_databases.php?server=1', '_self');
```
ibennetch: This takes me to an error page that says "Unable to connect. Firefox can’t establish a connection to the server at localhost."

ibennetch (changes requested):

ibennetch:
I'm not sure how we want to handle shortcut keys; this may not be the best way (for instance, the last time this came up we discussed how "N" for "new" makes sense in English, but not for all other languages where the word "new" translates to something that doesn't start with an 'n'.

js/shortcuts_handler.js
akshil4444: confirmed, working on this.

akshil4444: @ibennetch The firefox error gives only for the first time and it's actually not an error as it is working.  After that, if you try to do the same there won't be any error. I actually don't know for what reason is the error occurred for the initial time.

saulotoledo: @akshil4444 I suggest to change the port of your local HTTP server and test again.

Issue #14128, opened 2018-03-27
[4.8 RC1] Show 2FA Secret next to QR


My1:
### Steps to reproduce
1. go to settings -> 2FA
2. try to configure an Authentication Application (2FA)

### Expected behaviour
get the QR code and the secret as text in case you cannot scan the QR code.

### Actual behaviour
only QR Code gets shown

### Server configuration
**Operating system**: Win 8.1 Pro x64

**Web server:** Caddy 0.10.11

**Database:** MySQL 5.6.21

**PHP version:** 7.2.0

**phpMyAdmin version:** 4.8 RC1

### Client configuration
**Browser:** Opera 52

**Operating system:** Win 8.1 Pro x64

PR #14154, opened 2018-04-03
Display the secret for 2FA - #14128

ramantehlan:
After this commit, users will be able to see secret/key for 2FA, in case they can't scan QR code. #14128 

- [X] Is needed on its own, if you have just minor fixes to previous commits, you can squash them


MauricioFauth (changes requested):
Hi, @ramantehlan. Thanks for your contribution.
I just reviewed your pull request.

PR #14206, opened 2018-04-18
patch fixed for the datepickerui broken #14199

### LINKED ISSUE #14199: UI Issue - date picker hides in scroll frame
### Steps to reproduce
Regression introduced in 4.8.0 release.
1. Go to [Demo table on PMA demo server](https://demo.phpmyadmin.net/master-config/sql.php?server=2&db=sakila&table=staff&pos=0)
2. Double click last update 

### Expected behaviour
Show date picker on top of everything.

### Actual behaviour
Shows the date picker in a scroll frame.

#### 4.8.0

selection_015

#### 4.7.9

selection_014

akshilshah:
Fixes #14199 
It is not a permanent solution but as of now, I could think of this option to make it go.

Implementation:
Removed the scroll frame when clicked on the date picker and add it back as soon as date picker is destroyed.

Signed-off-by: Akshil Shah <akshilshah4444@gmail.com>

### DIFF
--- a/js/functions.js
+++ b/js/functions.js
@@ -115,6 +115,8 @@ function PMA_addDatepicker ($this_element, type, options) {
         altFieldTimeOnly: false,
         showAnim: '',
         beforeShow: function (input, inst) {
+            //temporary adding the class for visible datepicker
+            $( ".responsivetable" ).removeClass('responsivetable').addClass('temp');
             // Remember that we came from the datepicker; this is used
             // in tbl_change.js by verificationsAfterFieldChange()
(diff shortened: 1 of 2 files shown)


nulll-pointer:
I  was thinking about changing how the positioning of this element works. or, 'overflow-x' could be removed as suggested by users.

akshil4444:
User was facing problem at the time of datepicker only. so I don't really think there is need to remove overflow as I personally like the responsive table. 
this patch will remove overflow particular when the user wants to use datepicker and will add back once it has been done.

PR #14218, opened 2018-04-20
Fix for "No tables found" problem on deleting tables in the last tab, #14205

### LINKED ISSUE #14205: "No tables found in database" when you delete all tables from last page
### Steps to reproduce
1. Create enough tables, so you have 2 pages
2. Delete all tables from 2nd page
3. "You get message: No tables found in database"

### Expected behaviour
You should see tables from 1st page

aroralakshya:
Fix #14205

Signed-Off-By: Lakshya arora <arora.lakshya123@gmail.com>

### DIFF
--- a/js/navigation.js
+++ b/js/navigation.js
@@ -259,6 +259,9 @@ function traverseNavigationForPaths () {
                     .find('span.pos2_value:last')
                     .text();
             }
+            if ($('#pma_navigation_tree').find('li.database.selected').children('div.list_container').find('li.table').length === 1) {
+                pos2_value = 0;
+            }
 
             params['n' + count + '_pos2_name'] = pos2_name;
             params['n' + count + '_pos2_value'] = pos2_value;


ibennetch:
I apparently have better luck than Maurício did; it worked for me. After deleting the table, the navigation pane did not update; I had to manually use the `<` button to go to the previous page, but I did not get an error.

I wonder if a better solution would be to automatically refresh/move the user to the new last page.

This isn't a bad or wrong solution as-is, but I wonder if we can do better for presenting a smooth interface to the user.

nulll-pointer:
@mauriciofauth  @ibennetch , try hard reload. Probably, it's loading the script from the cache.

MauricioFauth (commented):
I tried again, checked that the script had been loaded correctly, but the issue keeps happening.

nulll-pointer:
I'm looking into it.

ibennetch:
I know this one wasn't part of the GSoC plan, but did you figure out what was wrong?

js/navigation.js
```diff
@@ -259,6 +259,9 @@ function traverseNavigationForPaths () {
                     .find('span.pos2_value:last')
                     .text();
             }
+            if( $('#pma_navigation_tree').find('li.database.selected').children('div.list_container').find('li.table').length === 1) {
```
mauriciofauth: ESLint failed. See: https://travis-ci.org/phpmyadmin/phpmyadmin/jobs/414006024#L744

MauricioFauth (changes requested):

js/navigation.js
nulll-pointer: I've fixed it, kindly have a look.

nulll-pointer:
@mauriciofauth, fixed

MauricioFauth (commented):
@nulll-pointer Now I realize that we are not seeing the same thing.
You are looking at the list of tables in the navigation menu. However, the issue is about the list of tables in the database Structure tab.
You can see the pagination if you change the maximum number of tables, for example: `$cfg['MaxTableList'] = 2;`
Now go to the last page and drop all the tables. You will see the message _No tables found in database._
The only ways I found to stop seeing this message was by changing the database or manually adding the `pos` parameter, for example: `db_structure.php?db=sakila&pos=0`

nulll-pointer:
oh, yes. But, this was also a bug :smile:

mauriciofauth:
@nulll-pointer Yes, you've found a related bug. :)

williamdes:
@sudo-bot Deploy PR

williamdes (changes requested):
This PR did not solve the issue from my tests

williamdes:
Closing since this PR does not solve the issue, sorry.

Issue #14248, opened 2018-05-02
Date selection in search menu missing higher Z-index value


DharmarajPS:
### Steps to reproduce
1.Please find the attached files where red marked image is bug. Not able to select/no left right arrow to navigate to next month or previous month.
2. Green marked image shows how to fix that bug. z-index: 99 works and shows navigation to next and prvious months
2. If we give give z-index 99, it will work.
3..div#ui-datepicker-div{z-index:99}

### Expected behaviour
We should give z-index:99 to the #ui-datepicker so that we can select next and previous months.

### Actual behaviour
 If number of columns in table is less, then if we select date field, we are not able to select month because of Z-index value is 1(css part).

### Server configuration
**Operating system:** windows 10

**Web server:** Apache

**Database:** MySQL

**PHP version:** PHP 7.1

**phpMyAdmin version:** phpMyAdmin 4.7.4

### Client configuration
**Browser:** Google Chrome

**Operating system:** Windows 10
![phpmyadminbug](https://user-images.githubusercontent.com/38908133/39515362-7cf5811c-4e17-11e8-9cb9-9ff625dac4a0.PNG)
![phpmyadminbugsolved](https://user-images.githubusercontent.com/38908133/39515541-0ce2b858-4e18-11e8-8890-319df9d1fd42.PNG)

rijaspk:
@DharmarajPS, phpMyAdmin 4.9 have the same issue.I will try to fix this issue

DharmarajPS:
It can be solved by giving higher index to the datepicker UI.

rijaspk:
hi, @DharmarajPS.Inline Styles are used and I couldn't find the file where it is written.can you help me find that file.Input field of time also has a style issue.

DharmarajPS:
You can do like this if you not found respective file.

z-index:99 !important in stylesheet.

rijaspk:
That works fine.

BO41:
@rijaspk @ibennetch will this PR get merged or should someone open another one with a maybe better solution?

PR #14251, opened 2018-05-03
Date selection in search menu missing higher Z-index value

rijaspk:
Date selection in search menu missing a higher Z-index value
Issue #14248

Signed-off-by: Rijas Ahammed PK <rijaspk5195@gmail.com>

### DIFF
--- a/js/server_databases.js
+++ b/js/server_databases.js
@@ -101,7 +101,7 @@ AJAX.registerOnload('server_databases.js', function () {
 
         // TODO Remove this section when all browsers support HTML5 "required" property
         var newDbNameInput = $form.find('input[name=new_db]');
-        if (newDbNameInput.val() === '') {
+        if (newDbNameInput.val() == '') {
             newDbNameInput.focus();
             alert(PMA_messages.strFormEmpty);
             return;
--- a/themes/original/jquery/jquery-ui.css
+++ b/themes/original/jquery/jquery-ui.css
@@ -170,7 +170,7 @@
 }
 .ui-selectable-helper {
 	position: absolute;
-	z-index: 100;
+       z-index: 100;
 	border: 1px dotted black;
 }
 .ui-sortable-handle {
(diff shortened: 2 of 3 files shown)


themes/original/jquery/jquery-ui.css
```diff
@@ -403,6 +403,7 @@ button.ui-button::-moz-focus-inner {
 	width: 17em;
 	padding: .2em .2em 0;
 	display: none;
+       z-index: 99 !important;
```
mauriciofauth: Do you have any other way to solve it without using `!important`?

MauricioFauth (commented):
@rijaspk Thanks for your contribution.

themes/original/jquery/jquery-ui.css
rijaspk: Hi @mauriciofauth, Date picker UI used Inline style and unfortunately I couldn't find the file where it is written. Can you help me find that file

themes/original/jquery/jquery-ui.css
```diff
@@ -170,7 +170,7 @@
 }
 .ui-selectable-helper {
 	position: absolute;
-	z-index: 100;
+       z-index: 100;
```
saulotoledo: This file uses tab as indent, please fix it.

themes/original/jquery/jquery-ui.css
saulotoledo: Hi everyone. I am new here, but I would like to head the opinion of an old repository reviewer about. This is a JQuery UI file, so I think we should do these changes in another place, not here. Are we really allowed do change these files?

saulotoledo (changes requested):

js/server_databases.js
```diff
@@ -101,7 +101,7 @@ AJAX.registerOnload('server_databases.js', function () {
 
         // TODO Remove this section when all browsers support HTML5 "required" property
         var newDbNameInput = $form.find('input[name=new_db]');
-        if (newDbNameInput.val() === '') {
+        if (newDbNameInput.val() == '') {
```
mauriciofauth: This change does not appear to be related to this pull request.
If this is fixing another issue, it is better to open a new pull request with this change.

themes/original/jquery/jquery-ui.css
mauriciofauth: Hi, @saulotoledo. You're right, this is a jQuery UI file. Although this file already contains some changes, I agree that we should not edit this file, and that those changes should go elsewhere.

Issue #14260, opened 2018-05-06
Language does not change in 4.8.0.1


AlBundyLoves69:
### Steps to reproduce
Installed 4.8.0.1

### Expected behaviour
The language should be as set in `$cfg['Lang'] = 'en';`

### Actual behaviour
All settings according [PhpMyAdmin Language Documentation](https://docs.phpmyadmin.net/en/latest/config.html#languages) are ignored. My Windows has German localization and PMA uses the German texts.

### Server configuration
**Operating system**:
SLES 12.3

**Web server:**
Apache 2

**Database:**
MySQL 5.7.21

**PHP version:**
7.0.7

**phpMyAdmin version:**
4.8.0.1

### Client configuration
**Browser:**
IE, Chrome, FF

**Operating system:**
Windows Server 2012 R2

williamdes:
Confirmed on branch master at revision : https://github.com/phpmyadmin/phpmyadmin/commit/bb9fc5d6e970c7a72d8888175074121816cd7792
```php
//not working
$cfg['DefaultLang']
//also not working
$cfg['Lang']
```

AlBundyLoves69:
Yep, I implemented it locally on my phpMyAdmin copy and it works!

PR #14261, opened 2018-05-06
#14260 Fixed configuration for DefaultLang and Lang

williamdes:
`$GLOBALS['cfg']` is not set before :
https://github.com/phpmyadmin/phpmyadmin/blob/bb9fc5d6e970c7a72d8888175074121816cd7792/libraries/common.inc.php#L287
So `$GLOBALS['cfg']` cannot be used in
https://github.com/phpmyadmin/phpmyadmin/blob/bb9fc5d6e970c7a72d8888175074121816cd7792/libraries/common.inc.php#L255

https://github.com/phpmyadmin/phpmyadmin/blob/bb9fc5d6e970c7a72d8888175074121816cd7792/libraries/classes/LanguageManager.php#L822-L831

And

https://github.com/phpmyadmin/phpmyadmin/blob/bb9fc5d6e970c7a72d8888175074121816cd7792/libraries/classes/LanguageManager.php#L886-L892

# Solution
Migrate to `$GLOBALS['PMA_Config']->get('nameofkey')` instead of `$GLOBALS['cfg']['nameofkey']`

# More info

Unit tests never failed because the globals where setup for the test.
I updated the test and added a test case that was not covered (cookie test case)

Closes: #14260

### DIFF
--- a/libraries/classes/LanguageManager.php
+++ b/libraries/classes/LanguageManager.php
@@ -822,8 +822,8 @@ class LanguageManager
     public function selectLanguage()
     {
         // check forced language
-        if (! empty($GLOBALS['cfg']['Lang'])) {
-            $lang = $this->getLanguage($GLOBALS['cfg']['Lang']);
+        if (! empty($GLOBALS['PMA_Config']->get('Lang'))) {
+            $lang = $this->getLanguage($GLOBALS['PMA_Config']->get('Lang'));
             if ($lang !== false) {
                 return $lang;
             }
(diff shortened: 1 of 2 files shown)


williamdes:
@mauriciofauth CodeCov seems broken ...
https://codecov.io/gh/phpmyadmin/phpmyadmin/commit/2f84b4034e4c423512029534e7a6132e6f8dc4d3

Should we add `php-recode` extension ?
PhpMyAdmin\Tests\EncodingTest::testRecode
recode extension missing

mauriciofauth:

mauriciofauth:
The test should not fail if the recode extension is missing.
Apparently it was a network error.
`WARNING Unable to get report from archive, will try again.`
`ERROR Max retries exceeded`

PR #14265, opened 2018-05-08
Issue #14019 fix: Allow parameter in AllowThirdPartyFraming

### LINKED ISSUE #14019: Allow parameter in 'AllowThirdPartyFraming'
`$cfg['AllowThirdPartyFraming'] ` is set to TRUE/FALSE only. I think it's a kinda limitation, because we should be able to set the intermediate value 'SAMEORIGIN'  (as mentioned [here](https://stackoverflow.com/a/2896705/2377343) )

at this moment, PMA uses this code:

```
        if (! $GLOBALS['cfg']['AllowThirdPartyFraming']) {
            header(
                'X-Frame-Options: DENY'
            );
        }
```

so, you should allow to use also 'SAMEORIGIN' value

aswaniprakash:
Fixes #14019

Signed-off-by: Aswani Prakash <aswani15prakash@gmail.com>

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -212,12 +212,14 @@ Basic settings
 
 .. config:option:: $cfg['AllowThirdPartyFraming']
 
-    :type: boolean
+    :type: boolean|string
     :default: false
 
     Setting this to ``true`` allows phpMyAdmin to be included inside a frame,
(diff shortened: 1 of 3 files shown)


mauriciofauth:

PR #14267, opened 2018-05-08
Require a polyfill instead of the cytpe extension

BackEndTea:

Ctype is an indirect dependency due to twig,  and it isn't resolved there until a later version than we require.
By requiring the symfony polyfill for cytpe everything will still run, regardless of the ctype availability. 

Twig will require this polyfill itself in a later version, but this makes sure we still allow earlier versions of twig to be used.

Signed-off-by: Gert de Pagter <BackEndTea@gmail.com>

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,6 @@
         "ext-xml": "*",
         "ext-pcre": "*",
         "ext-json": "*",
-        "ext-ctype": "*",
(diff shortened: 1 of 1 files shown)


mauriciofauth:

BackEndTea:
Thank you @mauriciofauth

Issue #6356, opened 2015-05-13
Theme creation tool


pma-import:
Linked to PR https://github.com/phpmyadmin/phpmyadmin/pull/1669, we thought about having a complete tool to create/manage a pMA theme, instead of defining some specific needs.

This tool would help to create a full theme:
- colors
- icons / logo
- navigation tree style
- …

We also could imagine to be able to load an existing theme to duplicate / update it.

---
- **Original URL**: https://sourceforge.net/p/phpmyadmin/feature-requests/1670/
- **Original author**: @Tithugues

shucon:
Can be closed in favour of https://github.com/phpmyadmin/phpmyadmin/issues/14933

williamdes:
Good catch, thank you so much !
Initial author: @tithugues

PR #14293, opened 2018-05-15
Automated Theme generator

shucon:
For easy access to my work progress and any suggestions.
All of the work regarding GSoC has been included in this PR.


js/theme_generator/color_picker.js
```diff
@@ -0,0 +1,2080 @@
+var UIColorPicker = (function UIColorPicker() {
```
Piyush3079: Is this a library of javascript?

shucon: yes

theme_generator.php
```diff
@@ -0,0 +1,28 @@
...
+use PhpMyAdmin\Response;
+
+require_once 'libraries/common.inc.php';
+
+$response = Response::getInstance();
+$header   = $response->getHeader();
+$scripts  = $header->getScripts();
+$scripts->addFile('theme_generator/color_picker.js');
+
+$response->addHTML('<div id="container">');
```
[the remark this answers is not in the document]
shucon: I may go with creating a twig template, as I'll be making preview panels later.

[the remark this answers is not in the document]
shucon: I have  created a mockup for the same https://researchweb.iiit.ac.in/~saksham.gupta/mockup2.jpg

js/theme_generator/color_picker.js
mauriciofauth: Did you write this library or is it a third-party library?

libraries/classes/ThemeGenerator.php
```diff
@@ -0,0 +1,589 @@
...
+        $output = '<div id="container">';
+        $output .= '<div id="palette" class="block">';
+        $output .= '<div id="color-palette"></div>';
+        $output .= '</div>';
+        $output .= '<div id="picker" class="block">';
+        $output .= '<div class="ui-color-picker" data-topic="picker" data-mode="HSB"></div>';
+        $output .= '<div id="picker-samples" sample-id="master"></div>';
+        $output .= '</div>';
+        $output .= '</div>';
+        return $output;
```
mauriciofauth: It would be better to use Twig templates instead of adding HTML directly in the code.

js/theme_generator/color_picker.js
```diff
@@ -0,0 +1,1942 @@
+/* vim: set expandtab sw=4 ts=4 sts=4: */
+/**
+ * Library for the Color Picker Tool
```
mauriciofauth: Did you write this library or is it a third-party library?

libraries/classes/ThemeGenerator.php
shucon: As I have to read about twig, I'll fix it in the end.

js/theme_generator/color_picker.js
shucon: Part of it is written and edited by me and part of it is a third-party library from https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Colors/Color_picker_tool

mauriciofauth: I was only concerned with the license to use the code, but it seems that this code is under public domain, so no problems.

libraries/classes/ThemeGenerator.php
mauriciofauth: No problem.

shucon: Is it fine now?

test/classes/Theme/LayoutTest.php
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * Test for Layout::createLayoutFile
+     *
+     * @param string $name   name of new theme
+     *
+     * @param string $output layout file data
+     *
+     * @return void
+     */
+    public function createLayoutFile($post , $output)
```
ibennetch: I'm not sure about this test. It looks to me like it's just testing whether the layout.inc.php file exists and contains the NaviBackground directive. I'm a little weak on our unit tests, to be honest, but I don't think this is really testing any of your code; it's instead testing something on the file system which is sort of beyond the scope of what we should be testing, unless we use the function that directly writes the file and test against the output, but even then it's better to just test the function output directly as you do in testcolorPicker().

I _think_ this is unnecessary, but would love for another developer to chime in if I'm wrong.

test/classes/ThemeGeneratorTest.php
```diff
@@ -0,0 +1,113 @@
...
+    /**
+     * Test for ThemeGenerator::createJsonFile
+     *
+     * @param string $name   name of new theme
+     *
+     * @param string $output JSON file data
+     *
+     * @return void
+     */
+    public function createJsonFile($name , $output)
```
ibennetch: This seems fine to me because you're creating the file and testing the output all at once.

test/classes/ThemeGeneratorTest.php
```diff
@@ -0,0 +1,113 @@
...
+        $this->assertContains('<input type="text" name="theme_name"></input>' , $output);
+        $this->assertContains('<input type="submit">' , $output);
+    }
+
+    /**
+     * Test for ThemeGenerator::form
+     *
+     * @return void
+     */
+    public function testcreateFileStructure()
```
ibennetch: I think testCreateFileStructure() is a better name here as well.

test/classes/ThemeGeneratorTest.php
```diff
@@ -0,0 +1,113 @@
...
+        $this->assertContains('<div class="ui-color-picker" data-topic="picker" data-mode="HSB"></div>' , $output);
+        $this->assertContains('<div id="picker-samples" sample-id="master">' , $output);
+    }
+
+    /**
+     * Test for ThemeGenerator::form
+     *
+     * @return void
+     */
+    public function testform()
```
ibennetch: I think testForm() would be a better name.

ibennetch (commented):
I think you're on a good path here. I've added some comments and suggestions about the tests.

test/classes/Theme/LayoutTest.php
shucon: I haven't completed all my tests for this function, but I feel it is an important test as any failed tests in this function would mean that the whole theme would be screwed. The layout file would be used as a function for all other files. Any failed test would mean a lot of errors.

libraries/classes/Theme/Common.php
```diff
@@ -0,0 +1,3637 @@
...
+        $txt .= '    border-bottom-right-radius: 0;';
+        $txt .= '    padding-bottom: 5px;';
+        $txt .= '}';
+
+        $txt .= '#PMA_slidingMessage code.sql,';
+        $txt .= 'div.sqlvalidate {';
+        $txt .= '    background: <?php echo $GLOBALS[\'cfg\'][\'BgOne\']; ?>;';
+        $txt .= '}';
+
+        $txt .= '#main_p
```
ibennetch: I doubt all this was fun or glamorous to copy/paste and format! :-o

libraries/classes/ThemeGenerator.php
```diff
@@ -0,0 +1,95 @@
...
+     * @param string $name name of new theme
+     *
+     * @return string $txt JSON file data
+     */
+    public function createJsonFile($name)
+    {
+        $file = fopen("themes/" . $name . "/theme.json", "w");
+        $txt = '{';
+        $txt .= '"name": "' . $name . '",';
+        $txt .= '"version": "5.0",';
```
ibennetch: Do you plan to allow the user to set the version number from the theme generator interface?

shucon: I don't know what's the use of version number, so right now it's what the other themes are using.

libraries/classes/Theme/Common.php
shucon: No it wasn't :stuck_out_tongue: . Although I learned a bit of regex and some efficient ways while formatting them.

libraries/classes/Theme/Common.php
```diff
@@ -0,0 +1,3637 @@
...
+    /**
+     * Creates navigation.css.php
+     *
+     * @param string $name name of new theme
+     *
+     * @return null
+     */
+    public function createCommonFile($name)
+    {
+        $file = fopen("themes/" . $name . "/css/common.css.php", "w");
```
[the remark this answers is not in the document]
ibennetch: Just posting this here for completeness after our Hangouts chat earlier this weekend.

On further review, I think I was mistaken about fopen(). I believe I was incorrectly remembering the configuration directive `allow_url_fopen` which of course won't affect us.

Of course we should still fail gracefully if fopen() doesn't work, but we don't need to test for it as I first thought.

libraries/classes/ThemeGenerator.php
ibennetch: I can help a bit with this: version numbers are set by whoever makes the themes and they're completely arbitrary. From the themes repository, we have version numbers like 2.8, 1.0b, 0.2, and more...they don't really correspond to anything. Sometimes a pull request for a theme will bump the version number by a full number (from 2.5 to 3.0), other times it's an incremental change (from 2.7 to 2.8). I don't even know what the 1.0b is supposed to be.

There's nothing tying theme version numbers to phpMyAdmin version numbers, either.

I'd say we should let the theme designer add their own theme version number, description, author name, and URL.

shucon: What if the user leaves the fields? What should be the default value?

libraries/classes/Theme/Common.php
```diff
@@ -0,0 +1,3637 @@
+<?php
+/* vim: set expandtab sw=4 ts=4 sts=4: */
+/**
+ * Creates common.css.php file
+ *
+ * @package PhpMyAdmin
+ */
+declare(strict_types=1);
+
+namespace PhpMyAdmin\Theme;
```
mauriciofauth: I'm not very sure about this namespace. What if it was `PhpMyAdmin\Theme\Generator` or `PhpMyAdmin\ThemeGenerator`? What do you think?

libraries/classes/Theme/Common.php
mauriciofauth: I think you would make good use of the [heredoc syntax](http://php.net/manual/en/language.types.string.php#language.types.string.syntax.heredoc) or [nowdoc syntax](http://php.net/manual/en/language.types.string.php#language.types.string.syntax.nowdoc) here.

mauriciofauth: You can remove this return. This way, the return of the method will be considered as `void`.

shucon: Is it necessary?? I have already changed the format once, so if it's not necessary can we leave it like this?

libraries/classes/Theme/Common.php
shucon: Yes! I'll correct it.

libraries/classes/ThemeGenerator.php
```diff
@@ -0,0 +1,143 @@
...
+     * Creates theme.json
+     *
+     * @param array $post POST data
+     *
+     * @return string $txt JSON file data
+     */
+    public function createJsonFile($post)
+    {
+        $name = $post['theme_name'];
+        $file = fopen("themes/" . $name . "/theme.json", "w");
```
ibennetch: We should test if $file is false (if the fopen failed) and handle the failure gracefully here.

libraries/classes/ThemeGenerator/Layout.php
```diff
@@ -0,0 +1,107 @@
...
+     * Creates layout.inc.php
+     *
+     * @param array $post POST form data
+     *
+     * @return string $txt layout.inc.php file data
+     */
+    public function createLayoutFile(array $post)
+    {
+        $name = $post['theme_name'];
+        $file = fopen("themes/" . $name . "/layout.inc.php", "w");
```
ibennetch: And here as well

libraries/classes/ThemeGenerator.php
```diff
@@ -93,10 +93,14 @@ public function createFileStructure(array $post)
...
-        $common->createCommonFile($name);
-        $out['layout'] = $layout->createLayoutFile($post);
-        $nav->createNavigationFile($name);
+        if (is_dir("themes/" . $name)) {
+            $out['json'] = $this->createJsonFile($post);
+            $common->createCommonFile($name);
+            $out['layout'] = $layout->createLayoutFile($post);
+            $nav->createNavigationFile($name);
+        } else {
+            trigger_error("The 'themes' directory is not writable by the webserver process. You must change permissions for the theme generator to be able to write the generated theme.", E_USER_ERROR);
```
ibennetch: This is a good solution, but the user will have already gone through and made their theme before realizing there's a problem. What about something like this as a slight tweak? https://gist.github.com/4788cd15d56c911b25be28d596f2d48c

ibennetch: Regardless, this error message should be translatable: use the `__("translatable text here") wrapper around the message like
```
trigger_error(__("The 'themes' directory is not writable by the webserver process. You must change permissions for the theme generator to be able to write the generated theme."), E_USER_ERROR);
```

libraries/classes/ThemeGenerator.php
```diff
@@ -0,0 +1,152 @@
...
+        }
+        if ($post['description'] != "") {
+            $txt .= '"description": "' . $post['description'] . '",';
+        } else {
+            $txt .= '"description": "Theme Generated by Automated Theme Generator",';
+        }
+        if ($post['author'] != "") {
+            $txt .= '"author": "' . $post['author'] . '",';
+        } else {
+            $txt .= '"author": "phpMyAdmin developers",';
```
ibennetch: In my opinion, since the phpMyAdmin developers didn't really make the theme, we can just leave this as an empty string.

libraries/classes/ThemeGenerator.php
```diff
@@ -0,0 +1,152 @@
...
+        }
+        if ($post['url'] != "") {
+            $txt .= '"url": "' . $post['url'] . '",';
+        } else {
+            $txt .= '"url": "https://www.phpmyadmin.net/",';
+        }
+        $txt .= '"supports": ["5.0"]';
```
ibennetch: Is this meant to be changed by the phpMyAdmin development team as new phpMyAdmin versions are released? Perhaps it would be better as a form field on the page, automatically filled in with the current version of phpMyAdmin (which can be obtained from the `PMA_VERSION` constant from libraries/classes/Config.php).

libraries/classes/ThemeGenerator.php
```diff
@@ -0,0 +1,152 @@
...
+        $txt .= '}';
+        // Check if the file is writable as this condition would only occur if files are overwritten.
+        if ($file) {
+            fwrite($file, $txt);
+            fclose($file);
+        } else {
+            trigger_error("The theme.json file is not writable by the webserver process. You must change permissions for the theme generator to be able to write the generated theme.", E_USER_ERROR);
```
ibennetch: This needs to be translatable as well.

templates/theme_generator/form.twig
```diff
@@ -0,0 +1,40 @@
+<form action="theme_generator.php" method="post" id="save">
+    <select name="type" id="theme">
+        <option value="0">Triadic</option>
```
ibennetch: These color types probably should be translatable as well.

templates/theme_generator/form.twig
```diff
@@ -0,0 +1,40 @@
...
+        <option value="1">Complementary</option>
+        <option value="2">Adjacent</option>
+        <option value="3">Monochrome</option>
+    </select>
+    <br>
+    &nbsp; Font :<br>
+    <select name="font" id="font">
```
ibennetch: I'm not sure about whether we would make these font options translatable, since they really directly reference the font name. @nijel do you have an opinion about that?

templates/theme_generator/form.twig
```diff
@@ -0,0 +1,40 @@
...
+        <option value="Palatino Linotype">Palatino Linotype</option>
+        <option value="Book Antiqua">Book Antiqua</option>
+        <option value="Arial">Arial</option>
+        <option value="Helvetica">Helvetica</option>
+        <option value="Arial Black">Arial Black</option>
+        <option value="Courier New">Courier New</option>
+        <option value="Verdana">Verdana</option>
+    </select>
+    <br>
+    &nbsp; Theme Name :<br>
```
ibennetch: This should be translatable as well.

templates/theme_generator/form.twig
```diff
@@ -0,0 +1,40 @@
...
+    <input type="text" name="theme_name" required></input> <br>
+    <br>
+    &nbsp; Version :<br>
```
ibennetch: And this

templates/theme_generator/form.twig
```diff
@@ -0,0 +1,40 @@
...
+    <input type="text" name="version"></input> <br>
+    <br>
+    &nbsp; Description :<br>
```
ibennetch: And this :)

templates/theme_generator/form.twig
```diff
@@ -0,0 +1,40 @@
...
+    <input type="text" name="description"></input> <br>
+    <br>
+    &nbsp; Author :<br>
```
ibennetch: And this :)

templates/theme_generator/form.twig
```diff
@@ -0,0 +1,40 @@
...
+    <input type="text" name="author"></input> <br>
+    <br>
+    &nbsp; URL :<br>
+    <input type="text" name="url"></input> <br>
+    <input type="checkbox" name="text_contrast">Automatically set text colours for best visibility<br>
```
ibennetch: And this :)

templates/theme_generator/form.twig
```diff
@@ -0,0 +1,40 @@
...
+    &nbsp; Version :<br>
+    <input type="text" name="version"></input> <br>
+    <br>
+    &nbsp; Description :<br>
+    <input type="text" name="description"></input> <br>
+    <br>
+    &nbsp; Author :<br>
+    <input type="text" name="author"></input> <br>
+    <br>
+    &nbsp; URL :<br>
```
ibennetch: And this :)

templates/theme_generator/form.twig
```diff
@@ -0,0 +1,40 @@
...
+    <input type="text" name="url"></input> <br>
+    <input type="checkbox" name="text_contrast">Automatically set text colours for best visibility<br>
+    <br><br>
+    <input type="button" value="Preview" id="preview">
```
ibennetch: And this text

templates/theme_generator/form.twig
```diff
@@ -0,0 +1,40 @@
...
+    <input type="submit">
```
ibennetch: Let's give this a less generic text value as well, perhaps `value="Save theme"`?

ibennetch (changes requested):
This is excellent, and I'm very pleased with your work here. There are a handful of suggestions and improvements I've made comments about in line.

There's one other thing I'd like to see if we can improve; after saving the file, the page just seems to reload for me, so I don't know if it's been successful. I think a popup message could be best saying "Theme saved, go to the main page to try it" or something along those lines. What do you think?

libraries/classes/ThemeGenerator.php
shucon: Yes! thanks for the idea :smile:

libraries/classes/ThemeGenerator.php
shucon: Or could we write 'custom generated theme' ??

libraries/classes/ThemeGenerator.php
shucon: PMA_VERSION is returning 5.0.0-dev for my case, so it's not valid. Should I trim it??

templates/theme_generator/form.twig
shucon: Does translation work the same way for twig files??

libraries/classes/ThemeGenerator.php
ibennetch: Sorry, I thought I responded to this but don't see it posted here. I think you should take the first two digits only, so 5.0, or 4.8, or 9.2, but nothing after that.

ibennetch (approved):

templates/theme_generator/form.twig
```diff
@@ -0,0 +1,40 @@
...
+    <br>
+    &nbsp; {% trans 'Author' %} :<br>
+    <input type="text" name="author"></input> <br>
+    <br>
+    &nbsp; {% trans 'URL' %} :<br>
+    <input type="text" name="url"></input> <br>
+    <input type="checkbox" name="text_contrast">{% trans 'Automatically set text colours for best visibility' %}<br>
+    <br><br>
+    <input type="button" value="{% trans 'Preview' %}" id="preview">
+    <input type="submit" value="{% trans 'Save theme' %}">
```
ibennetch: This breaks the test at /home/travis/build/phpmyadmin/phpmyadmin/test/classes/ThemeGeneratorTest.php:83

libraries/classes/ThemeGenerator.php
ibennetch: Reminder about testing if the fopen failed.

ibennetch (changes requested):
This is looking very good. There are three comments remaining, if you have time to check for a failed fopen() in two places and fix the ThemeGeneratorTest for `<input type="submit">` then I think everything else is set.

The documentation build warning/error is telling us that the theme_generator.rst document wasn't included in any table of contents. That can easily be fixed with this patch:
```
diff --git a/doc/user.rst b/doc/user.rst
index a98e2cb894..06843709b3 100644
--- a/doc/user.rst
+++ b/doc/user.rst
@@ -13,4 +13,5 @@ User Guide
     charts
     import_export
     themes
+    theme_generator
     other
```

libraries/classes/ThemeGenerator.php
shucon: It's being tested at line 155

shucon:
@ibennetch Can you please rerun Travis??

shucon:
@ibennetch can you please review this?? I think it's ready to be merged.

MauricioFauth (commented):
I think we should have a setting to enable/disable the Theme Generator. What do you think, @ibennetch?

williamdes:
Related to: #6356

shucon:
Commenting as the PR went stale. @ibennetch @mauriciofauth

mauriciofauth:
As I commented before, I think we should have a setting to enable/disable the Theme Generator, because some people will not want to have this feature enabled.

shucon:
@mauriciofauth where do you think it would be best to add this feature?

ibennetch:
Maybe a configuration directive and a simple test early in the theme generator code. Plus of course hiding the link on the main page. Would that be suitable?

shucon:
@ibennetch @mauriciofauth can you have a look at it now?
Also, there are some errors in Travis, it's something new to me.
Can you please help me how to fix it.

mauriciofauth:
@shucon PHPStan found 15 errors in your code.
See: https://travis-ci.org/phpmyadmin/phpmyadmin/jobs/464923677#L639

You can run PHPStan with: `composer phpstan`

sadaqtu (commented):
@phpmyadmin-bot

PR #14313, opened 2018-05-20
 Login modal after session expiration

aroralakshya:


libraries/common.inc.php
```diff
@@ -441,10 +441,11 @@
...
     }
+
 }
 
 // load user preferences
 $GLOBALS['PMA_Config']->loadUserPreferences();
 
 /* Tell tracker that it can actually work */
-Tracker::enable();
+Tracker::enable();
```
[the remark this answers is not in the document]
nulll-pointer: I might have accidently added it while trying some implementations, I'm removing all these right now.

shucon:
I think instead of displaying a modal , it should redirect to login page , and on logging in the previous work should be displayed. @ibennetch and I have discussed it previously and both of us think it would be better for security.

williamdes:
@shucon I agree for security, but for productivity going back to login page is just a waste of time.

nulll-pointer:
@shucon, I agree that security is a major concern for a login modal as the user can easily monitor the page using browser inspector.
I think the user can't do anything when the session expires, because all the requests are first check if the user is authenticated, if not they autmoatically redirect the user to the login page.
So, if the user tries to delete the modal or do anything suspicious, they will be redirected to the login page.

js/ajax.js
```diff
@@ -329,7 +329,118 @@ var AJAX = {
...
+            }
+        }
+    },
+    /**
+     * Response handler to handle login request from login modal after session expiration
+     *
+     * To refer to self use 'AJAX', instead of 'this' as this function
+     * is called in the jQuery context.
+     *
+     * @param object e Event data
```
mauriciofauth: Is that correct? I think it should be `@param type name description`, shouldn't it?

libraries/classes/Plugins/Auth/AuthenticationCookie.php
```diff
@@ -107,7 +126,12 @@ public function showLoginForm()
             $autocomplete   = ' autocomplete="off"';
         }
 
-        echo Template::get('login/header')->render(['theme' => $GLOBALS['PMA_Theme']]);
+        // wrap the login form in a div which overlays the whole page.
+        if($session_expired) {
+            echo Template::get('login/header')->render(['theme' => $GLOBALS['PMA_Theme'], 'add_class' => ' modal_form', 'session_expired' => 1]);
```
ibennetch: The get method has been removed, the render function can be used as in 244d4e8e81e40ebd621a6c285a3b108922cb5109 or 5edbdf2063789d9bacb56804634ed77be6a0bb08. Could you make the adjustment here and the other few places in this code?

ibennetch (changes requested):
Overall this works excellently for me, I think it's a great improvement for the users and well done. I've made a few comments for you to look at when you can.

ibennetch:
Also, just to be clear: some of your earlier commits don't have the sign off line, was your intention to agree to the DCO for every commit? I believe for this commit we can squash and rebase the merge through Github but need to be explicitly clear about your intentions :)

js/ajax.js
nulll-pointer: I was a little busy with the Add check constraint feature. I'm working on these now.

js/ajax.js
```diff
@@ -329,7 +329,118 @@ var AJAX = {
...
+        });
+        $('#pma_ignore_all_errors_popup').on('click', function () {
+            PMA_ignorePhpErrors(false);
+        });
+
+        if (typeof data.success !== 'undefined' && data.success) {
+            // reload page if user trying to login has changed
+            if(typeof data.user_changed !== 'undefined' && data.user_changed == 1) {
+                window.location = "index.php";
+                PMA_ajaxShowMessage("Loading...", false);
```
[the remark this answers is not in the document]
nulll-pointer: I was a little busy with the Add check constraint feature. I'm working on these now.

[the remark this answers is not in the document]
ibennetch: 👍

nulll-pointer:
Sorry I forgot to add Sign-off earlier.

nulll-pointer:
I've applied git rebase to-
Resolve the merge conflicts.
Remove trailing whitespace.
Add Signed-Off-By to older commits.

libraries/classes/Plugins/Auth/AuthenticationCookie.php
```diff
@@ -472,6 +512,9 @@ public function rememberCredentials()
     {
         // Name and password cookies need to be refreshed each time
         // Duration = one month for username
+        
+        // check if the user trying to login after session expiration is the same user as before
+        $user_changed = !hash_equals($_COOKIE['pmaUserHashed-' . $GLOBALS['server']], crypt($this->user, "SillYSAlTstRInG"));
```
ibennetch: We already encrypt username and password cookies in the cookie authentication (https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/classes/Plugins/Auth/AuthenticationCookie.php#L534), would it make sense to use that (or re-use that code)?

ibennetch: We already encrypt username and password cookies in the cookie authentication (https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/classes/Plugins/Auth/AuthenticationCookie.php#L534), would it make sense to use that (or re-use that code)?

nulll-pointer: I have an idea, we can use session storage to store the hash of the user, whenever user logs in again, we can validate if the new user is different from the old one their, this way, we don't need an extra cookie. But, this won't work if session storage is not supported.

ibennetch:
Hm, this was working for me, but now I get this JavaScript message after logging in

```
Notice in ./libraries/classes/Plugins/Auth/AuthenticationCookie.php#557
Undefined variable: user_changed

Backtrace

./libraries/common.inc.php#379: PhpMyAdmin\Plugins\Auth\AuthenticationCookie->rememberCredentials()
./index.php#29: require_once(./libraries/common.inc.php)
```

ibennetch:
I missed this before merging. It looks like you're still using the `user_changed` variable at https://github.com/phpmyadmin/phpmyadmin/pull/14313/files#diff-7b7c059f64cce443122a835f56a2fb07R557, but removed where you set it in https://github.com/phpmyadmin/phpmyadmin/pull/14313/commits/9d2ebe7b79ff2e8ba1e1193c1a859cb94ee83798#diff-7b7c059f64cce443122a835f56a2fb07L517.

Could you take a look?

nulll-pointer:
definitely

mauriciofauth:
@nulll-pointer @ibennetch I fixed in 12dfaafa3124ccc0826da8fd0a359bfbb3d921e0 the tests that were failing. I forgot to mention this PR in the commit message.

nulll-pointer:
@mauriciofauth thank you :smile:

PR #14316, opened 2018-05-22
Fix #14323, Parse type error when exporting

DvDMH:
This FIX an error exporting databases, bug #14323.

libraries/clases/Export.php::getFilenameAndMimetype():278

Before submitting pull request, please check that every commit:
Signed-off-by: Dave Macías Hurtado<TheDave86@hotmail.com>

- [x] Has proper Signed-Off-By
- [x] Has commit message which describes it
- [x] Is needed on it's own, if you have just minor fixes to previous commits, you can squash them
- [x] Any new functionality is covered by tests

PR #14325, opened 2018-05-25
Declare strict type issue in export.php fix

aswaniprakash:
While trying to export the following two errors were found in export.php

PHP Fatal error:  Uncaught TypeError: Argument 4 passed to PhpMyAdmin\\Export::getFilenameAndMimetype() must be of the type string, boolean given

PHP Fatal error:  strict_types declaration must be the very first statement in the script

Signed-off-by: Aswani Prakash <aswani15prakash@gmail.com>

### DIFF
--- a/export.php
+++ b/export.php
@@ -1,11 +1,12 @@
 <?php
+declare(strict_types=1);
 /* vim: set expandtab sw=4 ts=4 sts=4: */
 /**
  * Main export handling code
  *
  * @package PhpMyAdmin
  */
-declare(strict_types=1);
+
 
 use PhpMyAdmin\Core;
 use PhpMyAdmin\Encoding;
@@ -217,7 +218,7 @@ $compression_methods = array(
 /**
  * init and variable checking
  */
-$compression = false;
+$compression = '';
 $onserver = false;
 $save_on_server = false;
 $buffer_needed = false;


export.php
```diff
@@ -217,7 +218,7 @@
 /**
  * init and variable checking
  */
-$compression = false;
+$compression = '';
```
Piyush3079: This issue has already been solved in an earlier pr yet to be merged
https://github.com/phpmyadmin/phpmyadmin/pull/14316/commits/c3df4e4ef58b01b98203ba0e7fea27f026cf289a

export.php
```diff
@@ -1,11 +1,12 @@
 <?php
+declare(strict_types=1);
 /* vim: set expandtab sw=4 ts=4 sts=4: */
 /**
  * Main export handling code
  *
  * @package PhpMyAdmin
  */
-declare(strict_types=1);
```
Piyush3079: In all the other php files, `declare(strict_types=1);` is used at the same place not at the beginning of the the file and they are not producing any error too.
If you are getting the error, can you please mention the steps to reproduce it, because I am not able to reproduce the error.

aswaniprakash: PHP Fatal error: strict_types declaration must be the very first statement in the script

I got that error early but now its working fine :)

Piyush3079: So this pr does not contain any new changes. Correct me if I am wrong.

aswaniprakash: No, there are no other changes.

PR #14338, opened 2018-05-29
Composer install instead of update

meons:
I suggest to use `composer install` here. With `composer update`, last versions of dependencies will be installed and it could potentially break something into the code that is actually incompatible with latest versions of some of those dependencies.

https://getcomposer.org/doc/03-cli.md#install

Signed-off-by: Simeon Bobylev <simeon.bobylev@gmail.com>

### DIFF
--- a/doc/setup.rst
+++ b/doc/setup.rst
@@ -114,14 +114,14 @@ Additionally you need to install dependencies using the `Composer tool`_:
 
 .. code-block:: sh
 
-    composer update
+    composer install
 
 If you do not intend to develop, you can skip the installation of developer tools
 by invoking:
 
 .. code-block:: sh
 
-    composer update --no-dev
+    composer install --no-dev
 
 .. _composer:
 


meons:
Sorry I forgot Signed-Off-By line =/
I added it by editing my comment. Now what shall I do to run checks once again ?

williamdes:
Amend your commit and force push it, it will re trigger the tests
`git commit --amend -s`

doc/setup.rst
```diff
@@ -114,14 +114,14 @@ Additionally you need to install dependencies using the `Composer tool`_:
 
 .. code-block:: sh
 
-    composer update
+    composer install
 
 If you do not intend to develop, you can skip the installation of developer tools
 by invoking:
```
ibennetch: What about:
If you do not intend to develop for phpMyAdmin, you can restrict the minor version updates
of library files with the `install` directive and skip the installation of developer tools
by invoking the `--no-dev` option:

doc/setup.rst
```diff
@@ -114,14 +114,14 @@ Additionally you need to install dependencies using the `Composer tool`_:
 
 .. code-block:: sh
 
-    composer update
+    composer install
```
ibennetch: With my suggestion, this would remain as 'update'.

ibennetch (commented):
I'm a bit torn on this. I think when we first introduced Composer, this was deliberate; for contributors and testers I think it makes more sense to use the latest version of the dependencies specifically to notice if anything is broken. However, for people who are using the STABLE branch and prefer the Git version just to easily stay updated, it's going to make more sense to have the stability of 'composer install'. So perhaps we can do a bit better than just changing it, by mentioning both commands.

Perhaps something like what I've added below as comments?

meons:
I partially agree with you.
A user, which mostly wants a stable working version, should never (be able to) upgrade dependencies which have not been tested and could potentially break some features.
From the contributor's perspective it's a different approach, if for some reasons he wants to perform a `composer update` , no problem; composer.json file will also be updated and then commited/pushed to repository with the rest of code changes.
Furthermore, from my point of view, you don't have to mechanically perform a `composer update` everytime. You should use it when you really need benefits provided by dependencies updates such as features improvement or security fixes.

To summarize, I think the default usage should be `composer install`. Then if you want to specify that contributors could use `composer upgrade` to get latest versions of dependencies, I'm ok with that even if I guess that they are certainly very comfortable with composer commands =).

seyfer:
I agree with @meons on that.

Composer installation must be done with `composer install`, not `composer update`.
Because of `update` updates `composer.lock` file, which should be updated only in case of **intentional** dependency change (add, remove, version).
`composer install` just installs versions preserved in composer.lock.

Means if you use `install`, you always have latest working and tested state of dependencies.
It will not be broken accidentally, as it could happen when dependencies are updated.
For that reason, we commit `composer.lock`.

https://stackoverflow.com/questions/12896780/should-composer-lock-be-committed-to-version-control

williamdes:
@seyfer we commit the package lock for each version and delete it between versions https://github.com/phpmyadmin/phpmyadmin/commit/a00f23711e33a09dd8a472bfc6f938e5ab51ced4 https://github.com/phpmyadmin/phpmyadmin/commit/249c34d1effc9aa6f78c77861abcea38ff9dcfb5

https://github.com/phpmyadmin/phpmyadmin/commits/QA_4_9?after=b11d68386bac0795c76281c5aef83d754bf46260+34

mauriciofauth:
I agree we should start using `composer install` and add the `composer.lock` to the repository. Besides being much faster to install the dependencies, it will avoid breaking our test suite when a dependency releases a new version. Some recent cases are fc393eac393fadb858225374f799f87415645d92 and dd5d6d834dfae86297315cd07b7e21c972170495.

PR #14347, opened 2018-06-02
Add phpstan / finding errors in php code

williamdes:
# This PR is a proposal

Using `phpstan/phpstan` would help solving issues since strict mode is now enabled.

I intentionally put `--level=max` to show all errors, this mode goes very far in analysis.

Try levels from 0 to 7 or max.

Many errors found by `phpstan` can be fixed (IMO).

Other can be hidden :)

## Manual launch
`./vendor/bin/phpstan analyse ./ --level=max --memory-limit=1G --configuration=phpstan.neon`

`./` can be `./libraries ./test ./*.php` ...

`bootstrap.php` file is needed, otherwise `phpstan` will crash (no time to find why, maybe constants).

`PhpMyAdmin\MoTranslator\Loader::loadFunctions()` is required to avoid having errors for undefined function with code like: `__("blablabla");`

See : https://github.com/phpstan/phpstan

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,6 +66,8 @@ matrix:
       env: CI_MODE=docs
     - php: "7.1"
       env: CI_MODE=apidocs
+    - php: "7.2"
+      env: CI_MODE=phpstan
     - os: osx
       language: generic
       env: CI_MODE=test
@@ -80,6 +82,8 @@ matrix:
   allow_failures:
     - php: "nightly"
       env: CI_MODE=test
+    - php: "7.2"
+      env: CI_MODE=phpstan
 
 cache:
   pip: true
--- a/composer.json
+++ b/composer.json
@@ -87,7 +87,8 @@
         "pragmarx/google2fa": "^3.0",
         "bacon/bacon-qr-code": "^1.0",
         "samyoul/u2f-php-server": "^1.1",
-        "phpmyadmin/coding-standard": "^0.3"
+        "phpmyadmin/coding-standard": "^0.3",
+        "phpstan/phpstan": "^0.9.2"
     },
     "extra": {
         "branch-alias": {
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,0 +1,7 @@
+parameters:
+    bootstrap: %rootDir%/../../../test/bootstrap-phpstan.php
+    ignoreErrors:
+        - '#Unexpected token "&", expected TOKEN_VARIABLE at offset#'
+    excludes_analyse:
+        - %rootDir%/../../../tmp/*
+        - %rootDir%/../../../vendor/*
--- a/test/bootstrap-phpstan.php
+++ b/test/bootstrap-phpstan.php
@@ -1,0 +1,19 @@
+<?php
+/* vim: set expandtab sw=4 ts=4 sts=4: */
+/**
+ * Bootstrap file for phpstan
+ *
+ * @package PhpMyAdmin-test
+ */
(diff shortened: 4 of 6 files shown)


bootstrap.php
```diff
@@ -0,0 +1,11 @@
...
+/* vim: set expandtab sw=4 ts=4 sts=4: */
+declare(strict_types=1);
+
+// Setting constants for testing
+define('PHPMYADMIN', 1);
+define('TESTSUITE', 0);
+
+require_once 'libraries/vendor_config.php';
+require_once AUTOLOAD_FILE;
+PhpMyAdmin\MoTranslator\Loader::loadFunctions();
```
mauriciofauth: Just before you opened this pull request, I was testing phpstan and was using this bootstrap file:
```php
<?php

declare(strict_types=1);

define('PHPMYADMIN', true);
define('TESTSUITE', true);

require_once 'libraries/config.default.php';
$GLOBALS['cfg'] = $cfg;
$GLOBALS['server'] = 0;

\PhpMyAdmin\MoTranslator\Loader::loadFunctions();

\PhpMyAdmin\DatabaseInterface::load();
```
What do you think?

And I think the `bootstrap.php` name is generic, I think it would be better to use `phpstan-bootstrap.php`.

scripts/ci-phpstan.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+# from: github:wdes/account-manager scripts/ci
+cd $(dirname $0)/../
+echo "Running in : $(pwd)"
+# 1G is a litte high, should use < 512M
+./vendor/bin/phpstan analyse ./ --configuration=phpstan.neon --level=max --memory-limit=1G
```
mauriciofauth: I think it is more appropriate to start at the lowest level and once we clear all the errors, we can raise the level up.
Is it really necessary to change the `memory-limit` value?

williamdes: Lets start at level=0, `memory-limit` was an issue when I misconfigured phpstan and It scanned /vendor :laughing: 
I think we should set a value that is slightly overkill to avoid crashes of phpstan because of too much classes :)
`memory-limit` is not required.

bootstrap.php
williamdes: `phpstan-bootstrap.php` is a good idea.

I think that you know more than me what should be in that file, for phpMyAdmin: having nothing in the bootstrap file crashes phpstan.

`phpstan` will help avoiding bugs when we will have reached max level (IMO).

scripts/ci-phpstan.sh
mauriciofauth: I'm not very convinced about the location of this file. Would not it be better to put it in the test directory along with the other ci files? And would it be interesting to put it on Travis CI?

bootstrap.php
mauriciofauth: > `phpstan` will help avoiding bugs when we will have reached max level (IMO).

I agree with that.

scripts/ci-phpstan.sh
williamdes: I agree with you but It would make more sense to change the name to : `bootstrap-phpstan.php` to match `bootstrap-dist.php`.
What do you think about adding it to Travis CI with failure allowed, when max level will be reached, remove it from allow failure ?

mauriciofauth: I think it's a good idea.

mauriciofauth:

PR #14354, opened 2018-06-03
Fix for #14295

### LINKED ISSUE #14295: Issue in Changing theme
### Steps to reproduce
1. Login to PMA
2. Select theme Metro.
3. Now try to change theme to pmahomme

### Expected behaviour
Theme must change from metro to pmahomme

### Actual behaviour
Theme will not change and remains metro only.

### Client configuration
**Browser:**
chome 62

**Operating system:**
Ubuntu 17.10

williamdes:
Fixes: #14295

This is not the best way of fixing this issue.

Concurrent browsers keep overwriting the phpmyadmin config, the last to refresh has it's config written to phpmyadmin config database.

It is not clear what must be the last one to decide in theme selection. Cookie, PMA DB ?

Why the theme cookie is deleted if the theme is the default one, and why the theme it is not written into PMA DB if default one ?

**I need clarifications on theme selection process ...**

Why `original`, what if `original` is deleted in theme folder ?
https://github.com/phpmyadmin/phpmyadmin/blob/2a34c018633764ab0149d453dd2c8bfc8e79396f/libraries/classes/Config.php#L956-L970

### DIFF
--- a/index.php
+++ b/index.php
@@ -21,6 +21,7 @@ use PhpMyAdmin\Server\Select;
 use PhpMyAdmin\ThemeManager;
 use PhpMyAdmin\Url;
 use PhpMyAdmin\Util;
+use PhpMyAdmin\UserPreferences;
 
 /**
  * Gets some core libraries and displays a top message if required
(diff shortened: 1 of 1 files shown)


mauriciofauth:

mauriciofauth:
> This is not the best way of fixing this issue.
> 
> Concurrent browsers keep overwriting the phpmyadmin config, the last to refresh has it's config written to phpmyadmin config database.
> 
> It is not clear what must be the last one to decide in theme selection. Cookie, PMA DB ?
> 
> Why the theme cookie is deleted if the theme is the default one, and why the theme it is not written into PMA DB if default one ?
> 
> I need clarifications on theme selection process ...

@williamdes Could you open a new issue to discuss this?

> Why original, what if original is deleted in theme folder ?

Probably this line was forgotten when the default theme was changed from Original to pmahomme.

williamdes:
@mauriciofauth I would prefer that you create the issue because I do not know how to introduce the discussion.

Issue #14359, opened 2018-06-04
[5.0.0-dev] javascript report loop / flood server with report requests


williamdes:
### Steps to reproduce
1. Install phpMyAdmin 5.0.0-dev
2. block request to `js/messages.php` in chrome dev tools
2. **Or** break javascript :

In `js/messages.php` :
change on line 766 (remove simple coma) :

```    , " ' ;\n";``` to ```    , "  ;\n";```

# Warning
This issue will create an **infinite loop** of sending reports to server, be **prepared to close** this browser tab.
Open the browser developer console to see all the requests.

### Expected behaviour
Send one report or none

### Actual behaviour
Sending reports (thousands, forever)

### Error

```
error_report.js?v=5.0.0-dev:124 Uncaught ReferenceError: PMA_messages is not defined
```

PR #14360, opened 2018-06-05
Prevent reposting of javascript errors

williamdes:
Prevent reposting of javascript errors.

If an error is from the error handler then it will be caught by the error handler that will re-error and so on, forever.
#infiniteloop
This PR stops that behavior.

Closes: #14359


mauriciofauth:

PR #14379, opened 2018-06-10
Increase input field width for varchar

### LINKED ISSUE #13627: Increase input field width for varchar(255)
The input field width (on Insert Row page) is currently 40 chars, could this be increased?
Perhaps the value fields could use all remaining width in the table.

aroralakshya:
Fixes #13627

Signed-Off-By: Lakshay arora <arora.lakshya123@gmail.com>

### DIFF
--- a/libraries/classes/InsertEdit.php
+++ b/libraries/classes/InsertEdit.php
@@ -980,8 +980,8 @@ class InsertEdit
              * @todo clarify the meaning of the "textfield" class and explain
              *       why character columns have the "char" class instead
              */
-            $the_class = 'char';
-            $textAreaRows = $GLOBALS['cfg']['CharTextareaRows'];
+            $the_class = 'char charField';
+            $textAreaRows = max($GLOBALS['cfg']['CharTextareaRows'], 7);
             $textareaCols = $GLOBALS['cfg']['CharTextareaCols'];
             $extracted_columnspec = Util::extractColumnSpec(
                 $column['Type']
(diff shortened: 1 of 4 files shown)


MauricioFauth (changes requested):
@nulll-pointer Thanks for your contribution!
You need to fix the tests before we can merge your pull request.
See: https://travis-ci.org/phpmyadmin/phpmyadmin/jobs/390383283#L697

mauriciofauth:
I have the impression that [`$cfg['CharTextareaCols']`](https://docs.phpmyadmin.net/en/latest/config.html#cfg_CharTextareaCols) no longer works with this patch. Is this expected?

nulll-pointer:
"I have the impression that $cfg['CharTextareaCols'] no longer works with this patch. Is this expected?" 
Yes, this shouldn't work, I've set the width of the textarea to 90% of the page width(no matter what the configuration settings).
Should I set a minimum to the number of cols instead??

nulll-pointer:
I have made some changes in the tests for insertEdit, kindly have a look.

nulll-pointer:
I checked, this configuration isn't used anywhere else, I think using css to override it would be an overkill. Changing the default value sounds better.

MauricioFauth (approved):

mauriciofauth:

libraries/classes/InsertEdit.php
```diff
@@ -980,8 +980,8 @@ private function getTextarea(
              * @todo clarify the meaning of the "textfield" class and explain
              *       why character columns have the "char" class instead
              */
-            $the_class = 'char';
-            $textAreaRows = $GLOBALS['cfg']['CharTextareaRows'];
+            $the_class = 'char charField';
+            $textAreaRows = max($GLOBALS['cfg']['CharTextareaRows'], 7);
```
ibennetch: Is max() really a good choice here? What if (it's extremely unlikely, but still) a user deliberately wants more than 7 rows?

Issue #13621, opened 2017-08-28
QBE selenium tests broken since merge of #13342


nijel:
The Selenium tests for QBE got broken since merge of #13342:

```
1) PmaSeleniumQueryByExampleTest::testQueryByExample

InvalidArgumentException: Element not found.

/home/travis/build/phpmyadmin/phpmyadmin/test/selenium/TestBase.php:390

/home/travis/build/phpmyadmin/phpmyadmin/test/selenium/PmaSeleniumQueryByExampleTest.php:69
```

PS: This was hard to catch before, I'm working on bringing Selenium test to pull requests in https://github.com/phpmyadmin/phpmyadmin/issues/13603

nijel:
I've marked it as skipped for now (see eefc2cc47e) to avoid failures on every commit.

ibennetch:
Thanks @nijel, Raghuram and I have both been a bit too occupied elsewhere to fix this yet.

PR #14417, opened 2018-06-21
Automatically add index while editing an existing row and setting auto increment

aroralakshya:
This PR - 
1. Automates the addition of PRIMARY KEY index when user tries to edit an existing row and set AUTO INCREMENT.
2. Disables the popup on insert new column page when user sets AUTO INCREMENT and automatically sets the index to PRIMARY KEY.


ibennetch:
This is odd to me, I fixed one error with Selenium tests for you but there are two that don't seem to me to be related. Do you have any thought about the errors shown around https://travis-ci.org/phpmyadmin/phpmyadmin/jobs/398511173#L859 ?

nulll-pointer:
isn't this related to https://github.com/phpmyadmin/phpmyadmin/issues/13621

nulll-pointer:
I tried looking into the tests-
1. The first one tries to drop a table, and asserts if it is dropped, but the test shows -
>  Unable to locate element: {"method":"id","selector":"input_username"}
which is quite strange because #input_username textbox used to input username in login form. **So I feel this is unrelated to my changes.**
2. The second one tries to submit an empty sql query and asserts if it shows an error alert, the test shows-
> element not visible
Which is expected after [e5ff7f3](https://github.com/phpmyadmin/phpmyadmin/pull/14046/commits/e5ff7f34fe944975ce5c2d770d8a97b8d5534b7e), because after this commit, sql query field is never left blank, when the test tries to submit the query, it's never empty. **So I feel this on is also unrelated to my changes.**

nulll-pointer:
@ibennetch I've fixed tests. Only lint tests failed, which can be fixed by --fix.

MauricioFauth (approved):

mauriciofauth:

Issue #14184, opened 2018-04-11
Failed to set session cookie. Maybe you are using HTTP instead of HTTPS


nik6018:
### Steps to reproduce
1. Install PMA 4.8.0
2. Login via HTTP

### Expected behaviour
After entering the login information, I should be able to log in

### Actual behaviour
It doen't log me in and fails with this error : **Failed to set session cookie. Maybe you are using HTTP instead of HTTPS** 
In previous version of PMA 4.7.9 I didn't face this issue I was able to log in via HTTP and HTTPS without any issue.
Now I am only able to login via HTTPS.

How can I disable this as I cannot find anything in the docs, the closest thing I was able to find is this [Link](https://docs.phpmyadmin.net/en/latest/setup.html?#http-authentication-mode)

I am really confused about this.

I checked the file `common.inc.php` where this error is mentioned, but still wasn't able to figure out the issue.

### Web Server configuration

      server{

		listen	   2002;
		listen	   2004;
		server_name  localhost;
		
		# The Document Root
		root	/custom/path;
		
		rewrite "/sess([0-9a-zA-Z]{16})/(.*)" /$2;
		
		# Redirect server error pages to the static pages
		error_page  404					/nginx/404.html;
		error_page  497					/nginx/497.html;
		error_page  500 502 503 504		/nginx/50x.html;		
		location ~ (index.php|/)$ {
		
			fastcgi_pass	unix:/usr/local/php/var/fpm9178.sock;
			if ($uri ~ "^/phpmyadmin/") {
				fastcgi_pass	unix:/usr/local/php/var/fpm9179.sock;
			}
			fastcgi_read_timeout 3600;
			fastcgi_index	index.php;
			fastcgi_param	SCRIPT_FILENAME  $document_root$fastcgi_script_name;
			fastcgi_param	SCRIPT_NAME	  $fastcgi_script_name;
			include			fastcgi_params;
		}
	
		location ~ \.php$ {
			fastcgi_pass	unix:/usr/local/php/var/fpm9178.sock;
			if ($uri ~ "^/phpmyadmin/") {
				fastcgi_pass	unix:/usr/local/php/var/fpm9179.sock;
			}

			fastcgi_read_timeout 3600;
			fastcgi_index	index.php;
			fastcgi_param	SCRIPT_FILENAME	$document_root$fastcgi_script_name;
			fastcgi_param	SCRIPT_NAME		$fastcgi_script_name;
			include			fastcgi_params;
		}
[...]

PR #14420, opened 2018-06-23
Proposal to fix HTTPS/HTTP cookie separation issue

williamdes:
I am trying to fix the issue with login.
**This PR fixes 1/2 of the issue.**

@ibennetch @nijel @mauriciofauth I want to separate cookie names like in screenshot, please _approve_ or not the proposal code of this PR :)

Tests are failing, I will correct them if this PR seems OK for you.

It works !

See issue #14184
- REPO: https://github.com/williamdes/issue-14184-pma-login

### Follow instructions below
Do the following :
- Open a **private browser**
- Login on **https** version
- It works !
- Login on **http** version
- Does not work
- **Close the browser**
- Open a **private browser**
- Login on **http** version
- It works !
- Login on **https** version
- It works !

This is exactly what happens on phpMyAdmin (excluding the cookie token/expire bug).

### RTFM
```
4.1.2.5.  The Secure Attribute

   The Secure attribute limits the scope of the cookie to "secure"
   channels (where "secure" is defined by the user agent).  When a
   cookie has the Secure attribute, the user agent will include the
   cookie in an HTTP request only if the request is transmitted over a
   secure channel (typically HTTP over Transport Layer Security (TLS)
   [RFC2818]).

   Although seemingly useful for protecting cookies from active network
   attackers, the Secure attribute protects only the cookie's
   confidentiality.  An active network attacker can overwrite Secure
   cookies from an insecure channel, disrupting their integrity (see
   Section 8.6 for more details).
```
> The Secure attribute **limits the scope** of the cookie to "secure"
   channels

> the user agent will include the
   cookie in an HTTP request **only** if the request is transmitted over a
   **secure channel**

> An active network attacker can **overwrite Secure
   cookies from an insecure channel**

### DIFF
--- a/libraries/classes/Config.php
+++ b/libraries/classes/Config.php
@@ -1432,6 +1432,20 @@ class Config
     }
 
     /**
+     * removes cookie
+     *
+     * @param string $cookieName name of cookie to remove
+     *
+     * @return void
+     */
+    public function unsetCookie(string $cookieName): void
+    {
+        if ($this->issetCookie($cookieName)) {
+            unset($_COOKIE[$this->getCookieName($cookieName)]);
+        }
+    }
+
+    /**
      * sets cookie if value is different from current cookie value,
      * or removes if value is equal to default
      *
@@ -1482,7 +1496,7 @@ class Config
                 return true;
             }
             return setcookie(
-                $cookie,
+                $this->getCookieName($cookie),
                 $value,
                 $validity,
                 $this->getRootPath(),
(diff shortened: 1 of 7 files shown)


nijel:
I'm not sure this is good idea to introduce this as it makes it easier to run insecure setup. Once https is available it should be used and preferably redirected server side and served with HSTS headers. On the other side educating users is always hard...

williamdes:
@nijel I agree with you but I know that if a software/tool forces me to do something I will end up "hacking it" to get back my freedom, #sudo. Maybe adding a bypass is an option ?

And where would you store that https was available at one point ?

williamdes:
@nijel Good idea !
So no more http usage of phpMyAdmin for https enabled severs ?

ibennetch:
@williamdes Do you have another pull request in place to replace this one? Seems like you made good progress towards solving this and I'd love to be able to merge a fix in to master.

williamdes:
@ibennetch No, maybe if I have time to do so :)

PR #14466, opened 2018-07-08
Console and Error Report files modularized 

Piyush3079:
This pr contains the PMA Console and PMA Error Report code in the modular form. The old files were split into multiple files to separate out functions required for base functionality.
There is some issue with this pr.
1). Earlier all the global functions were wrapped in the context of ErrorReport but now due to js modules it will not be possible to wrap all the imports in the context of ErrorReport.
2). Cross-origin files are now not allowed in Tracekit so stacktrace cannot be generated in developer mode.
3). The stack trace generated will contain a single line with column number because of uglified js file as output. See Tracekit Issue: https://github.com/occ/TraceKit/issues/61

This pr is in continuations of Mod_Js_Server_Privileges. Before reviewing this pr, review Mod_Js_Server_Privileges and merge and then compare with this pr.

### DIFF
--- a/js/src/classes/ErrorReport.js
+++ b/js/src/classes/ErrorReport.js
@@ -1,0 +1,338 @@
+/**
+ * general function, usually for data manipulation pages
+ *
+ */
+import { AJAX } from '../ajax';
+import { PMA_Messages as PMA_messages } from '../variables/export_variables';
+import { PMA_commonParams } from '../variables/common_params';
+import { PMA_ajaxShowMessage } from '../utils/show_ajax_messages';
+import { PMA_getImage } from '../functions/get_image';
+import TraceKit from 'tracekit';
+import { jQuery as $ } from '../utils/extend_jquery';
+
+/**
+ * This Object uses the library TraceKit to generate the backtrace of the
+ * error but since we are using webpack and the development and production
+ * mode have different url leading to cross origin query, the error
+ * report cannot be generated in development mode with context.
+ * The context will be null in deveopment mode.
+ *
+ * @see TraceKit url checking implemented to avoid Cross Origin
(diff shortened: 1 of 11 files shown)


devenbansod (commented):
Looks good, though the files and functions might require renaming later.

I'll test locally once you've confirmed it works on your end.

js/src/classes/ErrorReport.js
```diff
@@ -0,0 +1,338 @@
...
+    },
+    /**
+     * Wraps all global functions that start with PMA_
+     *
+     * @return void
+     *
+     * This function needs to be changes as no objects are available
+     * in the global scope so window does not contain any function
+     */
+    wrap_global_functions: function () {
```
Piyush3079: Functions are now not available in global scope, therefore need to find some alternate way to wrap exported functions in ErrorReport object and call all the functions in the context of ErrorReport.

js/src/classes/ErrorReport.js
```diff
@@ -0,0 +1,338 @@
+/**
```
Piyush3079: This file is not in working state completely because of some issues with webpack output files.
1) The output is the uglified file, therefore the stack trace is not that clear as it was earlier.
2) Due to cross-origin file loading in developer mode, context will not be available in developer mode.

Piyush3079 (commented):
General comments to increase understandability of this pr

js/src/classes/ErrorReport.js
devenbansod: Do you foresee this working in near future? can something like sourcemaps or something be used to improve this?

devenbansod:
Everything except the selenium job is passing at https://travis-ci.org/phpmyadmin/phpmyadmin/builds/408533738.

Merging this to unblock #14467

PR #14499, opened 2018-07-21
removes outdated browser prefixes

ghost:
on many css property's with user-select as an exception, there is no need for the browser prefix.
remove them will save some filespace

### DIFF
--- a/themes/metro/jquery/jquery-ui.css
+++ b/themes/metro/jquery/jquery-ui.css
@@ -5,303 +5,267 @@
 * Copyright jQuery Foundation and other contributors; Licensed MIT */
 
 .ui-draggable-handle {
-    -ms-touch-action: none;
-    touch-action: none;
+	-ms-touch-action: none;
+	touch-action: none;
 }
-
 /* Layout helpers
 ----------------------------------*/
-
 .ui-helper-hidden {
-    display: none;
+	display: none;
 }
-
 .ui-helper-hidden-accessible {
-    border: 0;
-    clip: rect(0 0 0 0);
-    height: 1px;
-    margin: -1px;
-    overflow: hidden;
-    padding: 0;
-    position: absolute;
-    width: 1px;
+	border: 0;
+	clip: rect(0 0 0 0);
+	height: 1px;
+	margin: -1px;
+	overflow: hidden;
+	padding: 0;
+	position: absolute;
+	width: 1px;
 }
-
 .ui-helper-reset {
-    margin: 0;
-    padding: 0;
-    border: 0;
-    outline: 0;
-    line-height: 1.3;
-    text-decoration: none;
-    font-size: 100%;
-    list-style: none;
+	margin: 0;
+	padding: 0;
+	border: 0;
+	outline: 0;
+	line-height: 1.3;
+	text-decoration: none;
+	font-size: 100%;
+	list-style: none;
 }
-
 .ui-helper-clearfix:before,
 .ui-helper-clearfix:after {
-    content: "";
-    display: table;
-    border-collapse: collapse;
+	content: "";
+	display: table;
+	border-collapse: collapse;
 }
-
 .ui-helper-clearfix:after {
-    clear: both;
+	clear: both;
 }
-
 .ui-helper-zfix {
-    width: 100%;
-    height: 100%;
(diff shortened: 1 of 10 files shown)


williamdes:
All commits must have signoff :)

BO41:
now, finally!

please review my changes. These are my first PR's to this project.

Please tell me if such changes are welcome. Or where I can find easy issues to solve :)

Piyush3079:
The `-moz` and `webkit` prefixes are supported in latest browses only. Removing these may create some styling issues in some of the old `Firefox` and `Safari` browser.
Are you sure about removing these?

BO41:
As far as I know, and I might be mistaken here, the prefixes allow the use of some propertys in earlier browser version. Or in some examples like [`user-select`](https://www.w3schools.com/csSref/css3_pr_user-select.asp) only cross browser with the prefix.
`user-select | 54.0 6.0-webkit- | 10.0-ms- | 2.0-moz- | 3.1-webkit- | 41.0 15.0-webkit-
--`

but other propertys like [`border-radius`](https://www.w3schools.com/cssref/css3_pr_border-radius.asp) don't really need the prefixes for ANY relative new browser
`border-radius | 5.0 4.0-webkit- | 9.0 | 4.0 3.0-moz- | 5.0 3.1-webkit- | 10.5
--`

of course there is a better support with the prefixes, but does it really make a difference if you support chrome 4 instead of 5 while the current version is 58 or something?
I don't think `html5` is supported back then.


But, I don't want to force anything, or even break any style. I just want to propose those changes and let you decide what is right :)
(sry, wasn't able to upload screenshots :/)

williamdes:
TLDR;
https://caniuse.com

BO41:
btw, not everywhere are the same prefixes used.
so if you decide to not remove them, we should use everywhere the same prefixes

themes/metro/jquery/jquery-ui.css
```diff
@@ -5,945 +5,1109 @@
 * Copyright jQuery Foundation and other contributors; Licensed MIT */
```
Piyush3079: Just a small suggestion:
This file is a vendor file downloaded from jquery.com and making code styling changes in vendor files may not be a good approach as we have lots of javascript and php vendor files which are not to the coding standards of PMA.
What's your suggestion regarding this? @williamdes

themes/original/jquery/jquery-ui.css
```diff
@@ -5,945 +5,1109 @@
 * Copyright jQuery Foundation and other contributors; Licensed MIT */
```
Piyush3079: See above comment regarding this file

themes/pmahomme/jquery/jquery-ui.css
```diff
@@ -5,945 +5,1109 @@
 * Copyright jQuery Foundation and other contributors; Licensed MIT */
```
Piyush3079: See above comment regarding this file

themes/metro/jquery/jquery-ui.css
williamdes: @Piyush3079 I think vendor files must not be changed, since they contain a version code.
They must be left intact.
Do we have a script that checks CS in vendor files ?

themes/original/jquery/jquery-ui.css
BO41: ok, now I know. will remove those changes :)

themes/metro/jquery/jquery-ui.css
Piyush3079: I don't think so we have any such script.
Is it really required?

williamdes: @Piyush3079 No, just asking :)
It would have explained why the files where prettified

Piyush3079:
@BO41 Your signoff is by the name of `Lukas Walter` but DCO was expecting `BOB41`.
I think you should check you git configuration or rebase your commits with correct signoff :)

BO41:
@Piyush3079 yeah thanks. I am trying to fix it all.
don't know what the problem of DCO is. I was asked to sign with my real name instead of my account name.
But I will figure it out, eventually

BO41:
replaced with #14522 because this PR is broken.

PR #14501, opened 2018-07-21
removes not needed code

ghost:
"In HTML5, the type attribute is no longer required for CSS. The default value is "text/css"."

Signed-off-by: BO41 <lukasw41@gmail.com>

PR #14523, opened 2018-07-25
removes type="text/css"

ghost:
"In HTML5, the type attribute is no longer required for CSS. The default value is "text/css"."

replaces #14501 since it is broken

please review

### DIFF
--- a/libraries/classes/Header.php
+++ b/libraries/classes/Header.php
@@ -686,27 +686,27 @@ class Header
         $v          = self::getVersionParameter();
 
         if ($this->_isPrintView) {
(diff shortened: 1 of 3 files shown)


BO41:
do I need to change anything or just trigger the pipe again?

ibennetch (approved):

BO41:
@ibennetch ah okay. thanks for the explanation.
I had some experience with bots/pipes which got triggered on a commit. But they did fail sometimes without a reason, so by re-triggering them they would pass. I wasn't sure if this is the same case in this project.

BO41:
closing for now because of inactivity.
reopen if you want to merge

Issue #14516, opened 2018-07-23
Cannot INSERT into tables with SYSTEM VERSIONING


darkain:
Reference: https://mariadb.com/kb/en/library/system-versioned-tables/

### Steps to reproduce
1. Create a table with SYSTEM VERSIONING and manually specify the `ts` and `te` columns.
2. Go to the "Insert" tab in phpMyAdmin for the that table
3. Attempt to insert any valid data

### Expected behaviour
New row INSERTed into table

### Actual behaviour
SQL query error: 1906 - The value specified for generated column 'ts' in table 'tbl' ignored.

Generated columns should not show up on the Insert tab in phpMyAdmin. They also should not attempt to be populated by phpMyAdmin when generating the INSERT query.

### Server configuration
**Operating system**: Debian Stretch 9.5

**Web server:** Apache 2.4.25-3+deb9u5

**Database:** 10.3.8-MariaDB-1:10.3.8+maria~stretch-log - mariadb.org binary distribution

**PHP version:** 7.0.30-0+deb9u1

**phpMyAdmin version:** 4.8.2

### Client configuration
**Browser:** Opera 54

**Operating system:** Windows 10 1803

williamdes:
`SHOW FULL COLUMNS FROM tablenamehere` does not indicate that the columns are versioned

darkain:
It looks like as of today, this is finally being addressed upstream: https://jira.mariadb.org/browse/MDEV-16804

williamdes:
Closing as it seems to have been fixed in MariaDB 10.3.16, 10.4.6 and above

PR #14536, opened 2018-07-31
Fix for SYSTEM VERSIONED tables #14514, #14515, and #14516

### LINKED ISSUE #14514: Tables with SYSTEM VERSIONING show up as views instead of tables
Reference: https://mariadb.com/kb/en/library/system-versioned-tables/

### Steps to reproduce
1. Create a table with SYSTEM VERSIONING.
2. Refresh phpMyAdmin
3. Table will be listed under VIEWS instead of TABLES

### Expected behaviour
Table should be listed as a table.

### Actual behaviour
Table is instead incorrectly listed as a view.

### Server configuration
**Operating system**: Debian Stretch 9.5

**Web server:** Apache 2.4.25-3+deb9u5

**Database:** 10.3.8-MariaDB-1:10.3.8+maria~stretch-log - mariadb.org binary distribution

**PHP version:** 7.0.30-0+deb9u1

**phpMyAdmin version:** 4.8.2

### Client configuration
**Browser:** Opera 54

**Operating system:** Windows 10 1803

### LINKED ISSUE #14515: Values cannot be edited in SYSTEM VERSIONING tables with INVISIBLE timestamps
Reference: https://mariadb.com/kb/en/library/system-versioned-tables/
Reference: https://mariadb.com/kb/en/library/invisible-columns/

### Steps to reproduce
1a. Create a table with SYSTEM VERSIONING without manually specifying the TIMESTAMP columns.
1b. Create a table with SYSTEM VERSIONING, set the `ts` and `te` columns to INVISIBLE, and add a PRIMARY KEY to any other column
2. Browse the table in phpMyAdmin
3. phpMyAdmin says: "Current selection does not contain a unique column. Grid edit, checkbox, Edit, Copy and Delete features are not available."

### Expected behaviour
Normal browsing and editing behavior

### Actual behaviour
"Current selection does not contain a unique column. Grid edit, checkbox, Edit, Copy and Delete features are not available."

phpMyAdmin identifies that the primary key is adjusted to have two columns (the one specified, and the `te` GENERATED column). With `te` set to INVISIBLE, when doing a SELECT * on the table, the column does not show up in the results. With SYSTEM VERSIONING, `te` is a read-only GENERATED column and should not be used for UPDATE or DELETE. These queries are fully deterministic using just the initially specified PRIMARY KEY column without the generated `te` column.

### Server configuration
**Operating system**: Debian Stretch 9.5

**Web server:** Apache 2.4.25-3+deb9u5

**Database:** 10.3.8-MariaDB-1:10.3.8+maria~stretch-log - mariadb.org binary distribution

**PHP version:** 7.0.30-0+deb9u1

**phpMyAdmin version:** 4.8.2

### Client configuration
**Browser:** Opera 54

**Operating system:** Windows 10 1803

darkain:
Fixes https://github.com/phpmyadmin/phpmyadmin/issues/14514
Fixes https://github.com/phpmyadmin/phpmyadmin/issues/14515
Possible fix for https://github.com/phpmyadmin/phpmyadmin/issues/14516

#14514 Tables which enable the new MariaDB 10.3 versioning show up as "SYSTEM VERSIONED" instead of "BASE TABLE" inside of information schema. This patch now checks for both, so these new table types don't incorrectly display as views.

#14515 MariaDB 10.3 also supports INVISIBLE columns, which is optionally paired with SYSTEM VERSIONED tabled (or implemented separately). This patch helps handle this new column attribute.

#14516 This patch cannot be fully tested and verified with current MariaDB builds. Version 10.3.9 (currently unreleased) corrects an issue with column information for SYSTEM VERSIONED columns that this patch relies on. https://jira.mariadb.org/browse/MDEV-16804 - the reason for checking with preg_match instead of adding to the list is because there are now combined "extra" information for columns, such as INVISIBLE columns. Just matching entire strings also has a bug with these other "extra" parameters.


williamdes (approved):

MauricioFauth (approved):

MauricioFauth (changes requested):
Thanks for your contribution!
I don't know why Travis CI didn't run, but you also need to fix the failing tests.

mauriciofauth:

PR #14538, opened 2018-08-01
change of typo

johnlow95:
change of typo in the intro.rest where logs in is change to login

### DIFF
--- a/doc/intro.rst
+++ b/doc/intro.rst
@@ -63,7 +63,7 @@ A word about users
 ------------------
 
 Many people have difficulty understanding the concept of user
-management with regards to phpMyAdmin. When a user logs in to
+management with regards to phpMyAdmin. When a user login to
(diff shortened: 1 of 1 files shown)


ghost (changes requested):
This does not really sound right for me.
I don't think any change is needed

johnlow95:
because im kinda confusing with this part since it sound more like the password and username will be pass directly to mySQL when a user login, or am i misunderstood it?

lem9:
I confirm that the original sentence is fine, no change needed. Please close.

williamdes (changes requested):
I think no change is needed.

ibennetch:
Thanks for your submission. I agree with the others that this isn't a proper change. "Login" is a noun, which is why we use the verb "log in" or, in the context of this sentence, "logs in".

You're correct that phpMyAdmin passes the username and password directly to MySQL.

Issue #13843, opened 2017-11-28
Better integrate storing user preferences in browser local storage


nijel:
After https://github.com/phpmyadmin/phpmyadmin/issues/13466 will be fixed, all user configuration will be stored in the user preferences. When the phpMyAdmin configuration storage is not configured, these do not persist over session.

This should be addressed by using browser local storage in this case. There already is existing code for this in the user preferences, it just needs manual actions, but this should happen automatically without need for user interventions:

- Save any changes to local storage. As this is on the client and we maintain the database on the server, we need to push changes somehow. It can be probably included in the AJAX responses as an additional field listing which values should be changed (or complete configuration). We can keep track of changed values in the session when calling Config::setUserValue...
- Automatically load settings from local storage. This should happen once at login time.

nulll-pointer:
Hello @nijel, I'm working on this as a part of my GSOC project. I looked into your commits and found that change in local storage is not bound with change in values in the form. So I guess I'll have to work on this. Some other things I wanted to clarify-
1. I think we need to uniquely identify the configuration under username and server name.
2. There are tons of configuration directives, I think we can save the values for all the changed configurations.
3. At login time, we can send this data to the server which can change its $cfg[''] directives accordingly.
4. Whenever a user changes a configuration, it's saved for the session and in the local storage. When they relogin, it's loaded directly from the local storage.
@ibennetch

nijel:
Your approach looks reasonable to me.

ibennetch:
I agree that this seems reasonable.

PR #14543, opened 2018-08-02
Integrate user preferences with local storage

aroralakshya:
#13843 
Signed-Off-By: Lakshay arora <arora.lakshya123@gmail.com>

### DIFF
--- a/js/functions.js
+++ b/js/functions.js
@@ -5012,7 +5012,7 @@ function PMA_getImage (image, alternate, attributes) {
  * A configuration value may be set in both browser's local storage and
  * remotely in server's configuration table.
  *
- * If the `only_local` argument is `true`, the value is store is stored only in
+ * If the `only_local` argument is `true`, the value is stored only in
  * browser's local storage and may be lost if the user resets his browser's
  * settings.
  *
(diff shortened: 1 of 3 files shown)


nulll-pointer:
I've given preference to the configuration stored in local storage if the data between local storage and database is inconsistent.

js/functions.js
```diff
@@ -5106,3 +5106,47 @@ jQuery.fn.getPostData = function () {
...
+    } else if(isStorageSupported('localStorage') && $("#login_form").length === 0) {
+        var username = PMA_commonParams.get('user');
+        var server = PMA_commonParams.get('server');
+        var configurations = localStorage.getItem(username + "/" + server);
+        if(typeof configurations === 'undefined' || configurations === null) {
+            configurations = [];
+        } else {
+            configurations = JSON.parse(configurations);
+        }
+        for(i = 0; i < configurations.length; ++i) {
```
saulotoledo: Use `var` to declare the loop variable, or you are creating a global one in your script.

js/functions.js
```diff
@@ -5106,3 +5106,47 @@ jQuery.fn.getPostData = function () {
...
+                    .attr('name', 'configurations')
+                    .attr('value', configurations)
+                    .appendTo('#login_form');
+            }
+        });
+    } else if(isStorageSupported('localStorage') && $("#login_form").length === 0) {
+        var username = PMA_commonParams.get('user');
+        var server = PMA_commonParams.get('server');
+        var configurations = localStorage.getItem(username + "/" + server);
+        if(typeof configurations === 'undefined' || configurations === null) {
```
saulotoledo: Try `if (configurations) {` here.

js/functions.js
```diff
@@ -5106,3 +5106,47 @@ jQuery.fn.getPostData = function () {
...
+            configurations = [];
+        } else {
+            configurations = JSON.parse(configurations);
+        }
+        for(i = 0; i < configurations.length; ++i) {
+            currentConfig = configurations[i];
+            type = currentConfig["type"];
+            if(type === "") break;
```
saulotoledo: I particularly do not like this `break` here. If says that if you have an empty value in the configuration you should stop the full processing. Why that happens? It is not easy to understand just by reading the code. Is there no other way to write this more clearly?

saulotoledo (changes requested):

nulll-pointer:
@saulotoledo, thanks for the suggested changes.

js/functions.js
```diff
@@ -5106,3 +5106,47 @@ jQuery.fn.getPostData = function () {
...
+                    .attr('name', 'configurations')
+                    .attr('value', configurations)
+                    .appendTo('#login_form');
+            }
+        });
+    } else if (isStorageSupported('localStorage') && $("#login_form").length === 0) {
+        username = PMA_commonParams.get('user');
+        server = PMA_commonParams.get('server');
+        var configurations = localStorage.getItem(username + "/" + server);
+        if (typeof configurations === 'undefined' || configurations === null) {
```
saulotoledo: I still prefer just `if (configurations) {` here, but it is ok.

saulotoledo (changes requested):
Yw @nulll-pointer :smile:

I just found some other minor details. Please address them if possible.

In your next commit you can use a `git commit --amend`.

saulotoledo:
@nulll-pointer Ops, I forgot something: there is a codeclimate problem to be solved [here](https://codeclimate.com/github/phpmyadmin/phpmyadmin/pull/14543).

js/functions.js
nulll-pointer: Actually, getItem function explicitly returns null if the item is not found.
https://stackoverflow.com/questions/3262605/how-to-check-whether-a-storage-item-is-set

saulotoledo: You are right, but if `configurations` is `null` or `undefined`, the expression `if (configurations)` will return `false`:

```javascript
if (undefined) {
  console.log("returning true");
} else {
  console.log("returning false"); // <- this is the result
}

if (null) {
  console.log("returning true");
} else {
  console.log("returning false"); // <- this is the result
}
```

saulotoledo:
@nulll-pointer There are still some lint errors found by travis, ok? You can find them by clicking on "Details" in the travis link below with a red "x".

nulll-pointer:
Yes, there are lint errors in some other prs as well. Im fixing these one
by one.

saulotoledo:
I will give you my ok, but now you need a review from someone on the project team. @ibennetch, could you point to the right person to do it?

saulotoledo (approved):
@nulll-pointer I'm giving you my ok, but now you need a review from someone on the project team. @ibennetch, could you point to the right person to do it?

PR #14553, opened 2018-08-06
Fix bad callers to getError where null was being passed.

### LINKED ISSUE #14552: Bad callers to getError in four Rte classes cause blank errors to be reported when adding things like triggers
### Steps to reproduce
1. Create a dummy table if you don't already have one for testing.
2. Navigate to the table and attempt to add a trigger. Name it anything you want, but make sure that the 'Definition' field contains a syntax error instead of a valid definition so that it would cause an error.
3. Click 'Go'.

### Expected behaviour
You see an error like the following:
MySQL Said: #1064 - You have an error in your SQL syntax; yadda yadda yadda

### Actual behaviour
MySQL Said:           <--- The error is blank

### Notes
I have a PR incoming on this issue, as I know the cause. In four of the Rte classes it appears we are passing null to `$GLOBALS['dbh']->getError`, which causes the default set in `libraries/classes/DatabaseInterface.php` (the CONNECT_USER named constant) to not be used as the actual value for our database handle. As such, when you enter the getError function within `libraries/classes/Dbi/DbiMysqli.php`, it actually is passed in a null dbh (just as was specified by the caller), though this results in the conditional attempting to get the *connection* error instead of the query error. As such, the error returned is blank.

Apologies if this issue has already been filed elsewhere. I did attempt to search for this, but found no cases in the tracker that currently appeared to speak to this issue. However, considering the number of currently open issues, the possibility that I overlooked a case seems likely to me.

### Server configuration
**Operating system**:
CentOS 7 (LINUX 3.10.0-693.17.1.el7.x86_64)

**Web server:**
PHP executed via CGI handler (Apache 2.4, perl CGI passthrough, etc.) or over socket via PHP-FPM.

**Database:**
MySQL 5.7

**PHP version:**
7.2.7

**phpMyAdmin version:**
4.8.2, though current master/HEAD appears impacted as well.

### Client configuration
**Browser:**
Firefox, Chromium, though this shouldn't matter since it isn't a JS error (backend response error).

**Operating system:**
Ubuntu 16.04.5 (LINUX 4.4.0-127-generic)

troglodyne:
Fixes #14552: Considering how function signatures work in PHP, this should
be a reltatively 'safe' change, as the mysqli lib currently would still
do a check for cases where the passed in $link was either null or false.
Otherwise the bug's source appears to be a common mistake in PHP:
null != undefined

Signed-off-by: Andy Baugh <thomas.baugh@cpanel.net>

### DIFF
--- a/libraries/classes/Rte/Events.php
+++ b/libraries/classes/Rte/Events.php
@@ -179,7 +179,7 @@ class Events
                             htmlspecialchars($drop_item)
                         )
                         . '<br />'
-                        . __('MySQL said: ') . $this->dbi->getError(null);
+                        . __('MySQL said: ') . $this->dbi->getError();
                     } else {
                         $result = $this->dbi->tryQuery($item_query);
                         if (! $result) {
@@ -188,7 +188,7 @@ class Events
                                 htmlspecialchars($item_query)
                             )
                             . '<br />'
-                            . __('MySQL said: ') . $this->dbi->getError(null);
+                            . __('MySQL said: ') . $this->dbi->getError();
                             // We dropped the old item, but were unable to create
                             // the new one. Try to restore the backup query
                             $result = $this->dbi->tryQuery($create_item);
@@ -219,7 +219,7 @@ class Events
                             htmlspecialchars($item_query)
                         )
                         . '<br /><br />'
-                        . __('MySQL said: ') . $this->dbi->getError(null);
+                        . __('MySQL said: ') . $this->dbi->getError();
                     } else {
                         $message = Message::success(
                             __('Event %1$s has been created.')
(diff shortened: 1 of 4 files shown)


troglodyne:
Let me know if I need to look into adding some assertions around this. Considering the above result from codecov (if I'm reading that right, it seems to say there's 0% coverage of affected statements), that seems like it might be indicated here. May take me till later this week to get familiarized enough with the testsuite to provide the updated tests, however.

ibennetch:
Your commit 793b062d572936b86b7d7278a22a9f430f31c6b2 for master and c55755d0ca9cb58371f9d6f667aa3dc4be984400 for QA_4_8 have been merged. Thank you very much for your contribution. The fix will be part of the next release, 4.8.3.

ibennetch:
We love increasing our test coverage and would welcome a pull request to do that (especially if this part of the code is fresh in your mind!), but I've gone ahead with the merge anyway since this is a good improvement.

PR #14559, opened 2018-08-09
Fix displaying of indexes

CommanderRoot:
Fixes an issue where you can't see or edit indexes on tables which don't have foreign keys support but support "normal" ones. An example engine like this would be TokuDB.


mauriciofauth:

PR #14589, opened 2018-08-23
Fix several issues with table headers in Safari

btacordex:
The width += 1 does not apply to Safari, either.

Safari requires min-width and max-width to be set. “width” is not honored.

Add a line break to template so comments stay on separate line.

Moved browser checks to higher scope; the user agent only needs to be checked once, not once per cell.


ibennetch:
Thanks for researching this and contributing a fix!

Even though the Safari issue affects the current 4.8 branch, I think I'll merge this only in to master (which will become version 5.0) as it does change the look of the comments.

ibennetch:

This will be a part of version 5.0 when it comes out.

PR #14650, opened 2018-10-08
Issue #12158: auto expand the single database

### LINKED ISSUE #12158: Auto open only_db
It would be nice, if when only_db is set to only one Database or you just have access to one single database, if you would not have to select it first in the left panel but the tables are automatically displayed. I like to work with the only_db feature and it's a bit annoying to click on the database every time I access the server.

Lkopo:
### Description

Implemented a new feature as discussed in #12158 with the following scenario:
If the logged user has only one database in the navigation tree, that database will be automatically expanded. I've also added this as a configurable option with default value **true**, so people can disable this change whenever they want.

Closes #12158

### DIFF
--- a/js/navigation.js
+++ b/js/navigation.js
@@ -742,7 +742,7 @@ function PMA_showCurrentNavigation () {
     var table = PMA_commonParams.get('table');
 
     var autoexpand = $('#pma_navigation_tree').hasClass('autoexpand');
-    
+
     $('#pma_navigation_tree')
         .find('li.selected')
         .removeClass('selected');
@@ -769,17 +769,7 @@ function PMA_showCurrentNavigation () {
                 handleTableOrDb(table, $('#pma_navigation_tree_content'));
             }
         } else if ($dbItem) {
-            var $expander = $dbItem.children('div:first').children('a.expander');
-            // if not loaded or loaded but collapsed
-            if (! $expander.hasClass('loaded') ||
-                $expander.find('img').is('.ic_b_plus')
-            ) {
-                expandTreeNode($expander, function () {
-                    handleTableOrDb(table, $dbItem);
-                });
-            } else {
-                handleTableOrDb(table, $dbItem);
-            }
+            fullExpand(table, $dbItem);
         }
     } else if ($('#navi_db_select').length && $('#navi_db_select').val()) {
         $('#navi_db_select').val('').hide().trigger('change');
@@ -800,6 +790,11 @@ function PMA_showCurrentNavigation () {
             $('#pma_navigation_tree').find('> div'), dbItemName, 'database', !table
         );
 
+        fullExpand(table, $dbItem);
+    }
+    PMA_showFullName($('#pma_navigation_tree'));
+
+    function fullExpand(table, $dbItem) {
         var $expander = $dbItem.children('div:first').children('a.expander');
         // if not loaded or loaded but collapsed
         if (! $expander.hasClass('loaded') ||
@@ -812,7 +807,6 @@ function PMA_showCurrentNavigation () {
             handleTableOrDb(table, $dbItem);
         }
     }
-    PMA_showFullName($('#pma_navigation_tree'));
 
     function handleTableOrDb (table, $dbItem) {
         if (table) {
--- a/js/navigation.js
+++ b/js/navigation.js
@@ -740,6 +740,9 @@ function scrollToView ($element, $forceToTop) {
 function PMA_showCurrentNavigation () {
     var db = PMA_commonParams.get('db');
     var table = PMA_commonParams.get('table');
+
+    var autoexpand = $('#pma_navigation_tree').hasClass('autoexpand');
+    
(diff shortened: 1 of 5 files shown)


lkopo:
I see there is a problem with sign-off as I've signed-off my commits by my real name (as it is written in https://github.com/phpmyadmin/phpmyadmin/blob/master/DCO) which conflicts with my nickname. What should I do now?

williamdes:
@lkopo The DCO details show:
```
Commit sha: 7167a63, Author: Eduard Karpiel, Committer: Eduard Karpiel; Expected "Eduard Karpiel lkopo@lkopo.local", but got "Eduard Karpiel eduard@karpiel.sk".
Commit sha: 920be41, Author: lkopo, Committer: lkopo; Expected "lkopo eduard@karpiel.sk", but got "Eduard Karpiel eduard@karpiel.sk".
```

williamdes:
@lkopo I think that you can force ```--author``` and/or ```--commiter```
The best for me is to configure globally or for this repo you author settings:
```bash
git config --global user.name "John Doe"
git config --global user.email "john@doe.org"
```

williamdes:
@lkopo Only one left (double signoff) ;)

lkopo:
@williamdes all checks passed. Is it OK now? :)

williamdes (approved):

williamdes:
@lkopo Yes :)

lkopo:
Hi @williamdes. Is there something missing for merge or some reviews are still pending? :)

williamdes:
@lkopo For me everything is OK
@mauriciofauth Can you merge this PR if it seems good for you ?

js/navigation.js
```diff
@@ -766,23 +769,45 @@ function PMA_showCurrentNavigation () {
...
-                    handleTableOrDb(table, $dbItem);
-                });
-            } else {
-                handleTableOrDb(table, $dbItem);
-            }
+            fullExpand(table, $dbItem);
         }
     } else if ($('#navi_db_select').length && $('#navi_db_select').val()) {
         $('#navi_db_select').val('').hide().trigger('change');
+    } else if (autoexpand && $('#pma_navigation_tree_content > ul > li').length === 1) {
```
mauriciofauth: `$('#pma_navigation_tree_content > ul > li').length` isn't equal to 1 if the user has permission to create new databases.
db-list

MauricioFauth (changes requested):

lkopo:
@mauriciofauth Thank you, I didn't notice that! Already fixed. I've also fixed JS error when no database was in the tree, only link to create new database.

mauriciofauth:
Nice job. It's working now, however I noticed that it doesn't work the first time, when you just logged in, but when you refresh, it works. What do you think it could be?

lkopo:
@mauriciofauth I've found the issue, but I don't think it's not caused by my changes. The problem is here: in navigation.js:

Line 1814:
```
                if ($('#pma_navigation_tree').hasClass('synced')) {
                    PMA_selectCurrentDb();
                    PMA_showCurrentNavigation();
                }
```

If you open PMA in incognito window and log in, it will work as expected. But if you log out and log in again, it won't because of part of code mentioned above. I've checked and somehow after log out and log in, the `#pma_navigation_tree` element is not "synced" (the condition will fail).

I'm not sure what I can do here.

lkopo:
@mauriciofauth Ok, I'm sorry for previous deletions of my posts, but finally I've managed to find what is the real cause of this issue - Session storage. In navigation.js there is a function to reload navigation tree `PMA_reloadNavigation()`. It is called by multiple places, but after some debugging, the default behaviour when you log in from incognito window (so your session storage is empty), the condition on lines 640-641 will pass:

```
        if ($('#pma_navigation_tree_content').length &&
            typeof storage.navTreePaths === 'undefined'
        ) {
            PMA_reloadNavigation();
        }
```
Part `storage.navTreePaths` is undefined. But when you log out, session storage is not cleared so old data for navigation tree are here. After log in, `else` block is executed on line 649 which doesn't contain call for function `PMA_reloadNavigation()`, it calls instead `navTreeStateUpdate()` which will update three sessionStorage items: `navTreePaths`, `server` and `token`. Why it works on refresh is that `token` field was updated and now it evaluates the condition on lines 644-645:

```
        } else if (PMA_commonParams.get('server') === storage.server &&
            PMA_commonParams.get('token') === storage.token
        ) {
            // Reload the tree to the state before page refresh
            PMA_reloadNavigation(navFilterStateRestore, JSON.parse(storage.navTreePaths));
        }
```
as `true`. I've put the call for `PMA_reloadNavigation()` as a workaround for it, but I think it should be better for the future to somehow clear / remove sessionStorage fields after log out.

MauricioFauth (approved):

mauriciofauth:

> I've put the call for PMA_reloadNavigation() as a workaround for it, but I think it should be better for the future to somehow clear / remove sessionStorage fields after log out.

Can you open a new issue describing that?

PR #14712, opened 2018-11-04
Refactor .travis.yml

MauricioFauth:
### Description

* Moves the lint stage to before the test stage
* Moves most of the test/ci* scripts to .travis.yml file


williamdes (approved):
Looks good to me, I would like to still have the script to launch unit tests on my local setup
See: https://github.com/williamdes/mariadb-mysql-kbs/tree/master/scripts/ci

ibennetch (approved):
I'm not really familiar with the advantages of moving everything to Travis, but I don't object and don't need you to explain it for me to feel comfortable approving this. If you think this is better, go for it.

mauriciofauth:
I started making these changes just to move the linting stage to before the testing stage. Then I realized that I could make the file cleaner by moving the script files into the .travis.yml file. This made both .travis.yml file and Travis log cleaner.

> I'm not really familiar with the advantages of moving everything to Travis

I didn't move all the scripts to .travis.yml, just those which were one or two lines long. I agree that shell scripts doesn't belong to the .travis.yml file.

> I would like to still have the script to launch unit tests on my local setup

What do you think about using [Composer scripts](https://getcomposer.org/doc/articles/scripts.md#writing-custom-commands)? I think it would be a good option.

williamdes:
@mauriciofauth Good idea, composer scripts will be easier than manual scripts!

Issue #13773, opened 2017-10-24
Export page of Setting is not responsive


anantjain6:
### Steps to reproduce
1. From top navigation goto "Setting"
2.  Goto "Export"

### Expected behaviour
Page should be responsive

### Actual behaviour
Not responsive
I think that it is because of options with long text in drop down/list of "export method"

![screenshot_20171024-234038](https://user-images.githubusercontent.com/22341750/31961128-76c0dbb0-b917-11e7-9b82-6e9e9185a9bf.png)

**phpMyAdmin version:** Online Demo from git master

**Browser:** Latest Stable Google Chrome for Android

**Operating system:** Android v6.0.1

kartik1000:
Can anyone give some details on what exactly needs to be fixed in this issue? and if anyone is also working currently please let me know..I wish to work on it.

ibennetch:
It seems like no one is working on it.

You could look at https://github.com/phpmyadmin/phpmyadmin/pull/13422 for some inspiration from the project that made many of the other pages responsive. This feature request refers to a mobile-friendly design.

ZsgsDesign:
@kartik1000 Still working on it? I would like to have a try if nobody is doing this

kartik1000:
No I am currently not onto this issue, you can freely take it up, In case I find more info. Related to it which can help you i'll let you know ;)

shifatbuet:
Can i work on this??
Any confusion about this issue?
is it necessary to fix??

@williamdes

williamdes:
@shifatbuet Yes of course, just try on different demo servers before starting to fix

Demo servers (`root`, no password)
- [`5.1.0-dev`](https://demo.phpmyadmin.net/master/)
- [`5.0.0-dev`](https://demo.phpmyadmin.net/QA_5_0/)
- [`4.9.x`](https://demo.phpmyadmin.net/QA_4_9/)

Use the first lowest branch that is affected 4.9 > 5.0 > and finally master

shifatbuet:
@williamdes can't produce in demo. Maybe it is fixed. check.

AhmedEHaddad:
is this still open/not fixed ?

Issue #13077, opened 2017-03-14
support for large monitors


paveldvorak5:
### Steps to reproduce
1. try clicking button far on the rught
2. see the image
![screenshot 9](https://cloud.githubusercontent.com/assets/14863835/23905504/910946f8-08cb-11e7-9644-2cab34be0203.png)

3.

### Expected behaviour
move buttons to the right

### Actual behaviour
buttons are on the far left

### Server configuration
**Operating system**:

**Web server:**

**Database:**

**PHP version:**

**phpMyAdmin version:**

### Client configuration
**Browser:**

**Operating system:**

ibennetch:
Sorry, which buttons do you mean here?

paveldvorak5:
on the image profiling, [Edit inline], [Edit]..
or Go button in similar boxes

ibennetch:
That's what I thought you meant, but I got confused because I think you said the expected behavior is for the buttons to be on the right, and they already are.

I agree, this looks unfriendly.

paveldvorak5:
yes, i used wrong section

ibennetch:
No problem, I understand now.

OlafvdSpek:
I think left is the only proper option here.

paveldvorak5:
yes all other buttom are on the left

udan11:
What is the expected behaviour here?

I think the `Go` button looks a little weird if it is not on the right side of the screen. I may think that just because I got used to it.

![screencapture-192-168-168-10-tbl_sql-php-1491820702462](https://cloud.githubusercontent.com/assets/4147664/24858004/01741d46-1df3-11e7-8e21-9eeafddaf0e4.png)
![screencapture-192-168-168-10-tbl_sql-php-1491820670039](https://cloud.githubusercontent.com/assets/4147664/24858003/01720b28-1df3-11e7-96ed-0728188b3f8a.png)

paveldvorak5:
![myadmin](https://cloud.githubusercontent.com/assets/14863835/24859281/3088ab40-1df0-11e7-9fb3-d4e28c048322.jpg)
check this screenshots, button is totally weird on the rights
btw bookmark query should be on one line, not two

udan11:
I agree it looks weird on the right, but so it does in the middle. I think the `Go` button should stand out and if it were in the middle, it would not.

paveldvorak5:
maybe you can change button color or make a setting to make button float left

rajat315315:
[...]

Issue #13036, opened 2017-02-24
Disability tools


edwarddekker:
People who have a disability with seeing don't have the ability to experience phpMyAdmin.

Could we make a tool for them where these person's can view with a virtual keyboard and/or magnifying glass the letters with a built-in application. Also do we have the choice to make it with light and dark contrast to see it better for them all. And can we use braille for blind people.

nijel:
We certainly could do more for accessibility, but most of things you mention are IMHO rather in scope for the web browser rather than phpMyAdmin...

edwarddekker:
Give it a try would i say so.

edwarddekker:
I thinking about buttons on your screen.
Just a simple screen with only buttons and a command behind that button to execute a command and/or show a specific database.

This is of course theme related.

Soullivaneuh:
Why note taking profit to give a try to a CSS framework like Bootstrap? https://v4-alpha.getbootstrap.com/

A framework will provide a lot of built-in functionalities with very less css code.

nijel:
That's certainly an option, but for many things such massive change is not needed and including proper aria markup would help (and it would be needed with Bootstrap anyway).

Issue #13004, opened 2017-02-15
Suggestion - About the Whole phpmyadmin UI 


nimeshmora:
I hope that phpmyadmin web interface should be more attractive and most important features should be highlighted than it is at the moment.

ibennetch:
Hi @nimeshmora and thanks for your suggestions.

Which important features do you think aren't highlighted? Have you tried some other themes (these are available on the demo server at https://demo.phpmyadmin.net/QA_4_6; once logged in look for the Theme dropdown on the main page)? Do you think the interface looks old and outdated? Is it difficult to find the functionality you're looking for?

nimeshmora:
yah, several nice themes are available in that demo server. But looks like still the same layout. right hand side for DBs, top header for functionalities and all the stuff.. How about a new layout? I'm just thinking about something better than this layout.

jublo:
@nimeshmora Thank you for putting thought into this. Do you have some specifics in mind, or would you be able to draw a Paint scribble sketch to highlight some of these ideas? Just to visualise and help us get an idea about them :)

nimeshmora:
yeah, i will try to sketch something within next few weeks. thank you for replying back :)

edwarddekker:
Hi,

Maybe could you make a drag and drop theme. Where you can re-order different areas at your screen.

Greetings,
@edwarddekker

nimeshmora:
Yah i also thinking about something customizable as user would like , have to think about it more

jeeskdenny:
I think we don't need to display all table at once in sidebar. If we have to put a select dropdown option for the database, every time user get back into the phpmyadmin can see the working directory of their currant database. This is my small suggestion. 

 I am also a newbie like you. i am also interested to work with @nimeshmora  with the layout enhancement.

edwarddekker:
#13004

OlafvdSpek:
@jeeskdenny There's already a setting to only display tables of one DB (and have a dropdown to select the DB).

ervin210:
I think somonet take  from me

anantjain6:
Responsive theme/design should be implemented

ibennetch:
@anantjain6 a responsive design has recently been implemented by @manishbisht and is already a part of the development version. It will be part of the next major release, which probably will be called version 4.8.

rajat315315:
Can be closed I guess.
@williamdes ?

PR #14765, opened 2018-12-02
Add initial support for Bootstrap 4

MauricioFauth:
### Description

When implementing Sass, I noticed some inconsistencies in phpMyAdmin's CSS and UI, some of these inconsistencies were described in #14237.

Thinking about this, a good way to solve this would be to adopt a CSS framework, as this would be like a CSS guideline (#14237). Another advantage is that a theme developed for this framework would be easily adaptable to phpMyAdmin.

So, I think it would be interesting to use [Bootstrap](https://getbootstrap.com/) because, in addition to being one of the most popular, we already use it in the [Website](https://www.phpmyadmin.net/) and in the [Error Reporting Server](https://reports.phpmyadmin.net/).

Bootstrap 4.1.3 has approximately 141 kB in size when minified.

The idea is to refactor part by part, until we have the entire layout redone with Bootstrap.

Related to #11392, #13004, #13036, #13077, #13773 and #14237

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -24,6 +24,7 @@ phpMyAdmin - ChangeLog
 - issue #13023 Show errors at the bottom of the page and add copy query button for errors in processing sql queries
 - issue #14633 Use Sass instead of PHP to compile CSS
 - issue #14053 Fix missed padding on query results
+- issue #14765 Add initial support for Bootstrap 4
 
 4.8.5 (not yet released)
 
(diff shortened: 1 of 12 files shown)


williamdes (approved):

devenbansod (approved):
LGTM 💯 
Thanks @mauriciofauth for taking this up. This definitely seems like a step in right direction! 👍

PR #14788, opened 2018-12-16
Fix issue when adding user on MySQL 8.0.11

### LINKED ISSUE #14217: Issue adding new user with MySql 8.0.11
### Steps to reproduce
1. In phpmyadmin, open a database
2. Select tab Priviledges
3. Select Add user account, provide username and password + re-type (plugin = Native MySQL authentication), otherwise default options, select OK

### Expected behaviour
New user should be created without error

### Actual behaviour
`#1064 - You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'REQUIRE NONE WITH MAX_QUERIES_PER_HOUR 0 MAX_CONNECTIONS_PER_HOUR 0 MAX_UPDATES_' at line 1`

Additional information: 
1) The SQL command in the inline-box is : `CREATE USER 'aaa'@'%' IDENTIFIED WITH mysql_native_password AS '***';GRANT USAGE ON *.* TO 'aaa'@'%' REQUIRE NONE WITH MAX_QUERIES_PER_HOUR 0 MAX_CONNECTIONS_PER_HOUR 0 MAX_UPDATES_PER_HOUR 0 MAX_USER_CONNECTIONS 0;`
2) User is created, but with no password.
3) With same conf except MySQL=5.7, there is no error.

### Server configuration
Windows 10 Pro x64 1709

**Web server:**
Apache x64 2.4.33

**Database:**
MySQL 8.0.11

**PHP version:**
7.2.4

**phpMyAdmin version:**
4.8.0.1

### Client configuration
**Browser:**
Chrome

**Operating system:**
Windows 10 Home x64 1709

MauricioFauth:
It's no longer possible to use GRANT to modify account properties other than privilege assignments.

Fixes #14217

### DIFF
--- a/libraries/classes/Server/Privileges.php
+++ b/libraries/classes/Server/Privileges.php
@@ -2110,8 +2110,10 @@ class Privileges
     public static function getWithClauseForAddUserAndUpdatePrivs()
     {
         $sql_query = '';
-        if ((isset($_POST['Grant_priv']) && $_POST['Grant_priv'] == 'Y')
-            || (isset($GLOBALS['Grant_priv']) && $GLOBALS['Grant_priv'] == 'Y')
+        if (((isset($_POST['Grant_priv']) && $_POST['Grant_priv'] == 'Y')
+            || (isset($GLOBALS['Grant_priv']) && $GLOBALS['Grant_priv'] == 'Y'))
+            && ! ((Util::getServerType() == 'MySQL' || Util::getServerType() == 'Percona Server')
+                && $GLOBALS['dbi']->getVersion() >= 80011)
         ) {
             $sql_query .= ' GRANT OPTION';
         }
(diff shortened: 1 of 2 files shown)


libraries/classes/Server/Privileges.php
```diff
@@ -5236,6 +5254,8 @@ public static function getSqlQueriesForDisplayAndAddUser($username, $hostname, $
                 $create_user_stmt .= ' USING \'%s\'';
             } elseif ($serverType == 'MariaDB') {
                 $create_user_stmt .= ' IDENTIFIED BY \'%s\'';
+            } elseif (($serverType == 'MySQL' || $serverType == 'Percona Server') && $serverVersion >= 80011) {
+                $create_user_stmt .= ' BY \'%s\'';
```
mauriciofauth: I used clear text here, because the `PASSWORD()` function is no longer available, does anyone know if there is a problem with this?

williamdes (approved):

PR #14793, opened 2018-12-19
Remove unnecessary HTML closing tags

MauricioFauth:

### DIFF
--- a/changelog.php
+++ b/changelog.php
@@ -102,10 +102,10 @@ header('Content-type: text/html; charset=utf-8');
 <!DOCTYPE HTML>
 <html lang="en" dir="ltr">
 <head>
-    <link rel="icon" href="favicon.ico" type="image/x-icon" />
-    <link rel="shortcut icon" href="favicon.ico" type="image/x-icon" />
+    <link rel="icon" href="favicon.ico" type="image/x-icon">
+    <link rel="shortcut icon" href="favicon.ico" type="image/x-icon">
     <title>phpMyAdmin - ChangeLog</title>
-    <meta charset="utf-8" />
+    <meta charset="utf-8">
 </head>
 <body>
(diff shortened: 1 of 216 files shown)


williamdes:
Can you send links to why it is unnecessary ?
Any issues with browser support ?

mauriciofauth:
Closing all tags is only mandatory when using XML syntax (XHTML), but since we are using HTML5 with HTML syntax, it is not recommended to add an end tag on empty tags.

ibennetch:
My gut was to agree with William that this is probably not a positive change, but I didn't want to speak up until I looked in to it. Turns out, I was wrong.

https://stackoverflow.com/questions/3558119/are-non-void-self-closing-tags-valid-in-html5 and in particular the answer at https://stackoverflow.com/a/38068359/2385479 address using self-closing tags for void elements and the consensus there is that either they're not valid at all or are valid only because they're ignored. I think I'd like to test this by running some test pages through the W3C validator to verify that everything is being processed correctly, but it does indeed look like we should remove self-closing tags on void elements.

mauriciofauth:
Some W3C [HTML5 Specification](https://www.w3.org/TR/html50/) references:

> When a document is transmitted with an XML MIME type, such as `application/xhtml+xml`, then it is treated as an XML document by Web browsers, to be parsed by an XML processor. Authors are reminded that the processing for XML and HTML differs; (https://www.w3.org/TR/html50/introduction.html#html-vs-xhtml)

> Void elements only have a start tag; end tags must not be specified for void elements. (https://www.w3.org/TR/html50/syntax.html#void-elements)

> 6. Then, if the element is one of the void elements, or if the element is a foreign element, then there may be a single "/" (U+002F) character. This character has no effect on void elements, but on foreign elements it marks the start tag as self-closing. (https://www.w3.org/TR/html50/syntax.html#syntax-start-tag)

PR #14814, opened 2018-12-23
Move public files to the public directory

MauricioFauth:
By moving public files to a public directory, we improve the security of phpMyAdmin because users will not have access to files that are outside the public directory.

Since this is a major change, I think it makes sense to do this in version 5.0.0.

- [x] Move public files to the public directory
- [x] Update `ROOT_PATH` with new location
- [x] Move public Setup files to the public directory
- [x] Fix tests
- [x] Update documentation


williamdes (approved):
Very good idea !
:100:

mauriciofauth:
@ibennetch Could you review the documentation changes?

mauriciofauth:
I'll close this pull request and redo it in the future. I'm refactoring entry points so it's easier to make that change later.

PR #14821, opened 2018-12-28
Fix 14607 - Count rows only if needed

### LINKED ISSUE #14607: Count rows only if needed
Currently the PhpMyAdmin counts the total rows of a query result every time when LIMIT is not present.

If LIMIT is not presented in the user query, default LIMIT (like LIMIT 0, 50) is appended to the query.

But currently the total rows count is queried even when the total rows are below the default limit. This is not needed as has very large impact on long running queries.
<br>
Example:

User query:
```
SELECT count(*) tablewith10megrows
```

<br>PhpMyAdmin sends to DB:
```
SELECT count(*) tablewith10megrows LIMIT 0, 50
```
and gets exactly one row.
<br>

But the PhpMyAdmin sends additional unneeded query:
```
SELECT SQL_CALC_FOUND_ROWScount(*) tablewith10megrows
```
which is never needed if the rows count of the previous query was below the limit (exactly below max. count as LIMIT clause can have offset).

dhyey35:
Signed-off-by: Dhyey Thakore <dhyey35@gmail.com>

### Description

It does not send another count query in case the rows returned in a query with no limit clause are less than what we are supposed to display on that page eg: By default phpmyadmin sets limit to 25 by default and query returns 13 rows than there is no need to count total rows for that query as it will be 13 only.

Fixes #14607 

### DIFF
--- a/libraries/classes/Sql.php
+++ b/libraries/classes/Sql.php
@@ -1212,6 +1212,11 @@ EOT;
             // "Showing rows..." message
             // $_SESSION['tmpval']['max_rows'] = 'all';
             $unlim_num_rows = $num_rows;
+        } elseif ($this->isAppendLimitClause($analyzed_sql_results) && $_SESSION['tmpval']['max_rows'] > $num_rows) {
+            // When user has not defined a limit in query and total rows in
+            // result are less than max_rows to display, there is no need
+            // to count total rows for that query again
+            $unlim_num_rows = $_SESSION['tmpval']['pos'] + $num_rows;
         } elseif ($analyzed_sql_results['querytype'] == 'SELECT'
             || $analyzed_sql_results['is_subquery']
         ) {


dhyey35:
@mvorisek Done. @williamdes Can you please take a look ?

mvorisek:
I would put the if block as a 2nd if (more logical, also use `elseif` instead of `else if`).
Also remove the .gitignore modifications from your commit.

Otherwise the code seems to fully address the https://github.com/phpmyadmin/phpmyadmin/issues/14607 issue and I recommend merging it to the next release. Thanks.

dhyey35:
@mvorisek I have edited the source as per your suggestions. I was using XDebug with VSCode that's what the `launch.json` file was for in `.gitignore`. I have now removed it.

williamdes (approved):
LGTM
Thank you!

dhyey35:
@williamdes You are welcome :)

williamdes:
@mauriciofauth Can you please review this PR ?

MauricioFauth (approved):

mauriciofauth:

Issue #14800, opened 2018-12-20
Status -> Processes doesn't show full query


spacepille:
### Describe the bug

A click on the <-T-> on SQL query in Status -> Processes is without function. It doesn't show the full query.

### To Reproduce

1. Go to Status -> Processes
2. Click on <-T-> behind SQL query
3. And see that nothing happens

### Expected behavior

It should show the full SQL query.

### Screenshots

![grafik](https://user-images.githubusercontent.com/11977563/50302634-efff6180-048a-11e9-8a7c-a444700bfa66.png)

### Server configuration

- Operating system: Ubuntu 14.04.1 LTS
- Web server: nginx/1.4.6
- Database version: 10.0.27-MariaDB-1~xenial-wsrep - mariadb.org binary distribution, wsrep_25.16.rc3fc46e 
- PHP version: 5.6.34-1+ubuntu14.04.1+deb.sury.org+1 
- phpMyAdmin version: 4.8.4 (up to date)

### Client configuration

- Browser: Firefox Devloper Edition 65.0b4 (64-Bit)
- Operating system: Windows 10 Pro

williamdes:
Git bisect found that 0fe1a3bea88a553407930f83380b88d7591d2bdd is the first bad commit

Successfully reproduced on demo servers (root, no password)
- [`5.0.0-dev`](https://demo.phpmyadmin.net/master-config/)
- [`4.8.x`](https://demo.phpmyadmin.net/QA_4_8/)

ibennetch:
@williamdes @mauriciofauth Has there been any more progress on fixing this bug?

williamdes:
@ibennetch I did not have any time to fix this issue and #14833 is maybe related to this issue ;)

still-dreaming-1:
The problem in the stable version is it is using the POST value instead of the GET value. I peeped at the latest code and it seems that may have been fixed. However, I also feel the logic should be fully reversed to use FULL by default rather than after explicitly requesting this, otherwise it is basically setup to be annoyingly almost useful by default and only actually useful upon request. Or at the very least there should be a configuration option to make FULL the default. Occasionally people send me screenshots of partial long running queries from this page, which is almost helpful.

PR #14835, opened 2019-01-08
Issue #14833 -- fixed (Sort by Time not working in Processlist)

### LINKED ISSUE #14833: Sort by Time not working in Processlist
### Describe the bug

Under Status -> Processes, clicking on Time to sort, will make applied filters (e.g. Show only active) disappear. Also, it doesn't actually sort, and clicking 2 times doesn't change the order (DESC or ASC).

Version 4.8.4

### To Reproduce

1. Go to Status -> Processlist
2. Click on 'Show only active'
3. Click on the column title 'Time'

### Expected behavior

This should maintain the previous selection and sort by time. Instead, it removes the current filter and doesn't actually sort the processes by time. 

### Screenshots

[Screencast of the problem](https://youtu.be/FsNVSDQ_wYI)

### Server configuration

- Operating system: Amazon Linux AMI release 2018.03 and Mac under Homebrew 
- Web server: nginx/1.12.1
- Database version: 5.7.23 and 8.0.12 
- PHP version: 7.0.31 and  7.3.0
- phpMyAdmin version: 4.8.4

### Client configuration

- Browser: Chrome 71.0.3578.98 
- Operating system: Mac OS 10.14.2

saurass:
Signed-off-by: saurass <saurabhsrivastava312@gmail.com>

### Description

Fixes #14833

- In \libraries\classes\Server\Status\Processes.php -- getHtmlForServerProcesslist(), getHtmlForServerProcessItem()
It handled only POST requests.
- Where in the case above ajax call was made using GET
- So fixed points to handle both types of request.

### DIFF
--- a/libraries/classes/Server/Status/Processes.php
+++ b/libraries/classes/Server/Status/Processes.php
@@ -127,23 +127,21 @@ class Processes
             ],
         ];
         $sortableColCount = count($sortable_columns);
-
         $sql_query = $show_full_sql
             ? 'SHOW FULL PROCESSLIST'
             : 'SHOW PROCESSLIST';
-        if ((! empty($_POST['order_by_field'])
-            && ! empty($_POST['sort_order']))
-            || (! empty($_POST['showExecuting']))
+        if ( (!empty($_REQUEST['order_by_field']) && !empty($_REQUEST['sort_order']))
+                || (!empty($_REQUEST['showExecuting'])) 
         ) {
             $sql_query = 'SELECT * FROM `INFORMATION_SCHEMA`.`PROCESSLIST` ';
         }
-        if (! empty($_POST['showExecuting'])) {
+        if (! empty($_REQUEST['showExecuting'])) {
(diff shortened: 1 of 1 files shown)


MauricioFauth (changes requested):
Hi, @saurass. Thanks for your contribution.

Please do not use the `$_REQUEST` superglobal. This needs to be fixed in a different way.

saurass:
Yeah that is really a very bad method

I just figured that issue
#14800 and #14833 are closely related.

Both using GET for ajax whereas the code In Response.php is compatible only with POST request.

Working on that  :)

saurass:
@mauriciofauth should I make changes to server_status_processes.js - to make the AJAX call via POST method or we gotta make changes to backend code (Processes.php) ??

Please lemme know which one should I work on ?

mauriciofauth:
@saurass Yes, the request should use POST method.

saurass:
Thanks @mauriciofauth , I will try to fix the js file.

PR #14846, opened 2019-01-12
Replace if+return by return

Tithugues:
### Description

Replace if + return boolean by a simple return of the condition.


MauricioFauth (approved):

mauriciofauth:

PR #14860, opened 2019-01-12
Use real function names, not their aliases

Tithugues:
### Description

Don't use aliases, so everybody will use the same function names (and aliases may be still here for historical reasons, like `rand` -> `mt_rand`).


williamdes (approved):
Thank you !

MauricioFauth (approved):

PR #14861, opened 2019-01-12
Use Elvis operator

Tithugues:
### Description

Use Elvis operator when possible.


williamdes (approved):

ibennetch:
I personally find this less readable so I’m not sure about whether we should implement this change. If others are okay with it, I don’t object.

MauricioFauth (approved):

mauriciofauth:

PR #14865, opened 2019-01-12
Remove useless casts

Tithugues:
### Description

Some casts seem to be useless, let's try to remove them when confident about the type of the value.

### DIFF
--- a/libraries/classes/Import.php
+++ b/libraries/classes/Import.php
@@ -467,7 +467,7 @@ class Import
 
         if ($num > 26) {
             $div = (int) ($num / 26);
-            $remain = (int) ($num % 26);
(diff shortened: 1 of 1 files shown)


libraries/classes/Config.php
```diff
@@ -1258,7 +1258,7 @@ public function getSource(): string
      */
     public function getThemeUniqueValue(): int
     {
-        return (int) (
+        return (
```
mauriciofauth: This change will revert 2b4e71657b929984e4c7f4264c00ab62a5b1f953.

Tithugues: Sorry, I just reverted this one.

MauricioFauth (approved):

PR #15147, opened 2019-04-02
Fixes #15136 Hex converter giving invalid result

### LINKED ISSUE #15136: Hex converter giving invalid results
### Describe the bug
The hexadecimal converter 'HEX' in the insert field is giving invalid results. Suppose if I enter 70 as my int value it should give answer as '46' in HEX but it is giving answer as '3730'.
The 'OCT' converter for the same is working correct as I verified.
### To Reproduce
1.  Add int type column in any table.
2. Go to insert section.
3. Select the function 'HEX' in that 'int' column.
4. Enter some int value and observe the result in the added column
### Expected behaviour
The 'HEX' function should give out the corrected value.
### Screenshots
Screenshot 2019-03-31 at 10 47 55 PM
This value '3730' was received from type conversion of int '70' to hex
### Server configuration

Operating system: macOS Server(5.7.1 )
Web server: Apache/2.4.37 (Unix) OpenSSL/1.0.2q PHP/7.3.0 mod_perl/2.0.8-dev Perl/v5.16.3
Database version: Localhost via UNIX socket
PHP version: 7.3.0
phpMyAdmin version: 4.8.4

### Client configuration
Browser: Safari (12.0.3)
Operating system: macOS Mojave (version 10.14.2)

apoorvkhare07:
Signed-off-by: apoorv <apoorvkhare007@gmail.com>

### Description
The HEX() function in mysql behaves differently for string and int input. 
Changing the value to int which is string currently fixes the issue.

Fixes #15136

### DIFF
--- a/libraries/classes/InsertEdit.php
+++ b/libraries/classes/InsertEdit.php
@@ -2563,6 +2563,11 @@ class InsertEdit
                 return $multi_edit_funcs[$key] . '(' . $current_value . ",'"
                     . $this->dbi->escapeString($multi_edit_salt[$key]) . "')";
             }
+            elseif ($multi_edit_funcs[$key] == "HEX") {
+                $len = strlen($current_value);
+                $value = substr($current_value,1,$len-2);
+                return $multi_edit_funcs[$key] . '(' . (int)$value . ')';
+            }
 
             return $multi_edit_funcs[$key] . '(' . $current_value . ')';
         }


kartik1000:
LGTM👍. What do you think @williamdes?

AghaSaad04:
@kartik1000 , @apoorvkhare07  Actually, It does not work for some integers that are from 10 - 15, because these are Alphabets in Hex as the column is already in int, So you cannot save them in an int column and 0 will be stored for those numbers.

apoorvkhare07:
@AghaSaad04 This works fine, the mysql query will now return the correct hex value for all values. The issue is we cannot store hex values in int field. 
I think which is not the purpose of this PR. 
@williamdes @kartik1000 what do you'll think ?

AghaSaad04:
@apoorvkhare07 I am not saying your changes are wrong. They are actually perfect 👍. I am just saying that if we cannot save hex in int column. Then there's no need for that converter. Why don't we just remove that hex conversion for int? This is what I am thinking :)

apoorvkhare07:
@AghaSaad04 No we need the converter to return correct values as we cannot save HEX values in int but we can save them in strings or other types.

kartik1000:
@apoorvkhare07, Is it now working fine, for all the values or is it giving incorrect answers?

williamdes (changes requested):
Thank you for contributing, unfortunately I can not merge the PR
Please have a look at the discussion on #15136

Issue #13197, opened 2017-04-16
SignonScript do not perform successful login in


sritd:
### Steps to reproduce
1. I do a login to my general application and get successfully registered https://admin.DOMAIN
2. I have configured phpMyAdmin with following configuration options:
 2.a $cfg['Servers'][$i]['auth_type'] = 'signon';
 2.b $cfg['Servers'][$i]['SignonScript'] ='PMA-LOGIN.php';
 2.c $cfg['Servers'][$i]['SignonURL'] = 'https://admin.DOMAIN';
 2.d $cfg['Servers'][$i]['LogoutURL'] = 'https://admin.DOMAIN';
3. I call a script, which will create credentials for phpMyadmin and stores it in database and forward to https://phpmadmin.DOMAIN/index.php
4. I get the following displayed. Have a look at phpMyAdmin-AfterLogin.png
5. I can not use it. I can not click any link or do any other action than a logout

### Expected behaviour
I should get a working login to phpMyAdmin, that I can use your great tool.

### Actual behaviour
I get a login to phpMyAdmin which is not working. A session is created, but I am not able to use it.
I think, that there is a problem with the signon-script.php.

### Server configuration
**Operating system**: Debian 8 latest patchlevel

**Web server**: Apache 2.4.10

**Database**: MySQL 5.5.54-0+deb8u1 - (Debian)

**PHP version**: 5.6.30-0+deb8u1

**phpMyAdmin version**: 4.7.0

### Client configuration
**Browser**: Firefox 52.0.2 (64-bit)

**Operating system**: Ubuntu 16.04.2 LTS

Information after my automatic login:
![phpmyadmin-afterlogin](https://cloud.githubusercontent.com/assets/27511750/25071399/207f6f5c-22b6-11e7-8574-f4168a2a4269.png)

This file is used to create the SSO info for other applications:
[phpMyAdmin-createSSO.txt](https://github.com/phpmyadmin/phpmyadmin/files/924287/phpMyAdmin-createSSO.txt)

This file is included in phpMyAdmin and named PMA-LOGIN.php:
[phpMyAdmin-useSSO.txt](https://github.com/phpmyadmin/phpmyadmin/files/924288/phpMyAdmin-useSSO.txt)

udan11:
I did not work with SSO before, but It looks like the error code here is 302, which could be caused by this line:

```php
header('LOCATION: https://phpmyadmin.DOMAIN/index.php');
```

[...]

PR #15192, opened 2019-04-14
Modified single sign on and storing credentials

sven-ruess:
Signed-off-by: Sven Rueß <github@sritd.de>

### Description

I had a problem with single sign on in my environment. The used credentials are not stored globally for later usage in phpMyAdmin. For every interaction, the credentials are loaded from the source. This does not make sense in single sign on environments. Those credentials are used for the first login and will be deleted after it. If needed, they have to be stored locally in the application.

Fixes open issue #13197

### DIFF
--- a/libraries/classes/Plugins/AuthenticationPlugin.php
+++ b/libraries/classes/Plugins/AuthenticationPlugin.php
@@ -267,8 +267,7 @@ abstract class AuthenticationPlugin
      */
     public function isAuthenticated()
     {
-        if (isset($_SESSION['isAuthenticated']) && !empty($_SESSION['isAuthenticated']) && $_SESSION['isAuthenticated'] == true)
-        {
+        if (isset($_SESSION['isAuthenticated']) && ! empty($_SESSION['isAuthenticated']) && $_SESSION['isAuthenticated'] == true) {
             return true;
         } else {
             return false;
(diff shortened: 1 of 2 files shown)


lem9:
According to https://www.php.net/manual/en/features.session.security.management.php, "The session module can not guarantee that the information stored in a session is only viewed by the user who created the session. Additional measures are needed to protect the confidentiality of the session, depending on the value associated with it."  Therefore, storing the password in clear in the session is risky.

sritd:
Thank you for your comments. I know, that this is not the best solution. What I am missing is a general information about your information management. Do you have the possibility to store the dynamic username and password somewhere else? What I want is to read once the credentials and used it for session for all authentications. It is no need to read it every time for all new connections.
Otherwise I have to include another way, where I can store those credentials encrypted and secure.

Do you have a general wrapper class for accessing globals in PHP? I mean the variable $_SESSION for example. This is another recommended way for the implementation. I searched the source code but could not find something like that. $_SESSION is used directly in the code. If there is one wrapper class available all the checks can be merged together and security is increased.

Thank you very much for further information and have a great day.

Best regards
Sven

sritd:
Okay, another solution for solving this problem.

Waht about if we store the session data in a table of the database, if it is configured. Otherwise it will be stored as crypted values in the session data.

If database is configured:
I will create a new table for storing/reading session data. In the session and cookie only the ID will be stored. The other data is read from database. Data can be crypted stored in the database. This will be no problem for security, if it is implemented on a safe way. All data can be only accessed by database user for accessing phpmyadmin database.

If no database is configured:
All the data will be stored in the session directly. But private data will be encrypted. Those this will be protect against the access of getting access to session data files.

If you accept those modifications I will modify my pull request and implement the needed changes.

Feel free to give my a feedback.

Best regards
Sven

PR #15226, opened 2019-04-25
Fix #14986 problems with inserting data of DEFAULT values using PMA-GUI insert Tab

### LINKED ISSUE #14986: #1292 error for insert format fractional current_timestamp, support custom default functions
phpMyAdmin version 4.8.5
MariaDB 10.3.12
php 7.3.1

create table log:
```sql
CREATE TABLE IF NOT EXISTS log (
  logtime datetime(2) NOT NULL DEFAULT current_timestamp(2),
  logtext varchar(255) NOT NULL
) ENGINE=MyISAM DEFAULT CHARSET=latin1;
```

select the table and click insert

the logtime value is prompted as `current_timestamp(2).000000`

1) this result in a `#1292 - Incurrent datetime value: 'current_timestamp(2).00000' for column logtime`
2) removing the trailing .000000 also results in the same error

A working statement would be `insert into log values( default, 'test' )`
[ts-test.gz](https://github.com/phpmyadmin/phpmyadmin/files/2918629/ts-test.gz)

saurass:
## Description

Fixes #14986 problems with inserting data of DEFAULT values using PMA-GUI insert Tab

## Problem
* When inserting data using PMA GUI insert tab, the generated query also contains the default value to insert.
* If we want to insert default value in field, then that field can be ignored from the query.

## Solution
* So let's remove such fields from the query. And let MySQL/MariaDB server take care if we are trying to insert default value.

## Contribution Checklist

### DIFF
--- a/libraries/classes/Util.php
+++ b/libraries/classes/Util.php
@@ -4048,7 +4048,7 @@ class Util
         }
 
         if (mb_strpos($value, '.') === false) {
-            return $value . '.000000';
+            return $value;
         }
 
         $value .= '000000';
--- a/libraries/classes/InsertEdit.php
+++ b/libraries/classes/InsertEdit.php
@@ -2604,8 +2604,10 @@ class InsertEdit
         $key,
         $multi_edit_columns_null_prev
     ) {
+        $field = $multi_edit_columns_name[$key];
+        $isEqualToDefault = $this->checkIfFieldValueEqualToDefault($field, $current_value);
         //  i n s e r t
-        if ($is_insert) {
+        if ($is_insert && ! $isEqualToDefault) {
             // no need to add column into the valuelist
             if (strlen($current_value_as_an_array) > 0) {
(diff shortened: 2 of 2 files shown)


saurass:
@sudo-bot Deploy pr

libraries/classes/Util.php
```diff
@@ -4048,7 +4048,7 @@ public static function addMicroseconds($value)
         }
 
         if (mb_strpos($value, '.') === false) {
-            return $value . '.000000';
+            return $value;
```
williamdes: Thanks for removing this

On the screenshot below, the first row was updated without removing `.000000` and the second one after removing it.

image

saurass: Are you referring to a problem ?

williamdes: No, just posting a comment for others to understand the impact of this change and for the record.

williamdes:
@sudo-bot Deploy PR

PR #15234, opened 2019-04-29
Fix #14614 - Commands out of sync; you can't run this command now

### LINKED ISSUE #14614: "#2014 - Commands out of sync; you can't run this command now" returned when stored procedure does not return any result set
### Describe the bug

In **phpMyAdmin 4.8.3** if you call a stored procedure that contains a SELECT statement, when this SELECT does not return a result set, this error is displayed:
`#2014 - Commands out of sync; you can't run this command now`

When calling the same stored procedure from the mysql command line the expected output is returned instead:
```
Empty set (0.70 sec)
Query OK, 0 rows affected (0.70 sec)
```

### Server configuration

- Operating system:
- Web server:
- Database version: Server version: 5.5.5-10.1.19-MariaDB MariaDB Server
- PHP version:
- phpMyAdmin version: phpMyAdmin 4.8.3

### Client configuration

- Browser:
- Operating system:

williamdes:
### Description

Fixes `#2014 - Commands out of sync; you can't run this command now" returned when stored procedure does not return any result set`

Fixes #14614

### DIFF
--- a/libraries/classes/DatabaseInterface.php
+++ b/libraries/classes/DatabaseInterface.php
@@ -327,12 +327,17 @@ class DatabaseInterface
             $this->_dbgQuery($query, $link, $result, $time);
             if ($GLOBALS['cfg']['DBG']['sqllog']) {
                 if ($options & DatabaseInterface::QUERY_STORE == DatabaseInterface::QUERY_STORE) {
-                    $tmp = $this->_extension->realQuery('
-                        SHOW COUNT(*) WARNINGS', $this->_links[$link], DatabaseInterface::QUERY_STORE
+                    $tmp = $this->_extension->realQuery(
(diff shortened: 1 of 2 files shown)


ibennetch (approved):
Looks good to me and works fine, I approve.

libraries/classes/DatabaseInterface.php
```diff
@@ -327,12 +327,17 @@ public function tryQuery($query, $link = DatabaseInterface::CONNECT_USER, $optio
...
                     );
                     $warnings = $this->fetchRow($tmp);
+
+                    if ($this->_extension->moreResults($this->_links[$link])) {
+                        $this->_extension->nextResult($this->_links[$link]);
+                    }
+
                 } else {
-                    $warnings = 0;
+                    $warnings = [ '' ];
```
williamdes: Use `array()`

PR #15246, opened 2019-05-04
Fix #14412 - AuthenticationSignon.php's showFailure function does not consider SignonCookieParams

### LINKED ISSUE #14412: AuthenticationSignon.php's showFailure function does not consider SignonCookieParams
### Steps to reproduce
1. Use "signon" mode in a way that an error occurs on phpMyAdmin side (ex: set incorrect mysql server name) so `showFailure()` is called (In `libraries/classes/Plugins/Auth/AuthenticationSignon.php`)

### Expected behaviour
`showFailure` sets a session parameter called "`PMA_single_signon_error_message`" so the external login script can have access to the error message in case some error happens on phpMyAdmin side.

### Actual behaviour
`showFailure` forgets to consider `SignonCookieParams` and call `session_set_cookie_params` before calling `session_start`. In corner cases, this creates a new session and cookie, so the external script will not have access to the error message.

williamdes:
Fixes: #14412
Call `session_set_cookie_params` to set the cookie params

### DIFF
--- a/libraries/classes/Plugins/Auth/AuthenticationSignon.php
+++ b/libraries/classes/Plugins/Auth/AuthenticationSignon.php
@@ -41,6 +41,61 @@ class AuthenticationSignon extends AuthenticationPlugin
     }
 
     /**
+     * Set cookie params
+     *
+     * @param array $sessionCookieParams The cookie params
+     * @return void
+     */
+    public function setCookieParams(array $sessionCookieParams = null)
+    {
+        /* Session cookie params from config */
+        if ($sessionCookieParams === null) {
+            $sessionCookieParams = (array) $GLOBALS['cfg']['Server']['SignonCookieParams'];
+        }
+
+        /* Sanitize cookie params */
+        $defaultCookieParams = function ($key) {
+            switch ($key) {
+                case 'lifetime':
+                    return 0;
+                case 'path':
+                    return '/';
+                case 'domain':
+                    return '';
+                case 'secure':
+                    return false;
+                case 'httponly':
+                    return false;
+            }
+            return null;
+        };
+
+        foreach (array('lifetime', 'path', 'domain', 'secure', 'httponly') as $key) {
+            if (! isset($sessionCookieParams[$key])) {
+                $sessionCookieParams[$key] = $defaultCookieParams($key);
+            }
+        }
+
+        if (isset($sessionCookieParams['samesite'])
+            && ! in_array($sessionCookieParams['samesite'], array('Lax', 'Strict'))) {
+                // Not a valid value for samesite
+                unset($sessionCookieParams['samesite']);
+        }
+
+        if (version_compare(phpversion(), '7.3.0', '>=')) {
+            session_set_cookie_params($sessionCookieParams);
+        }
+
+        session_set_cookie_params(
+            $sessionCookieParams['lifetime'],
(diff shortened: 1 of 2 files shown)


williamdes:
Hi @nhatier can you please review this PR ?

williamdes:
@mauriciofauth @ibennetch can you please review this PR ?

libraries/classes/Plugins/Auth/AuthenticationSignon.php
```diff
@@ -220,6 +220,15 @@ public function showFailure($failure)
                 session_write_close();
 
                 /* Load single signon session */
+                /* Session cookie params */
+                $sessionCookieParams = (array) $GLOBALS['cfg']['Server']['SignonCookieParams'];
+                session_set_cookie_params(
```
mauriciofauth: You could check if PHP > 7.3 to use the PHP 7.3 signature of `session_set_cookie_params`.

libraries/classes/Plugins/Auth/AuthenticationSignon.php
```diff
@@ -220,6 +220,15 @@ public function showFailure($failure)
...
+                    $sessionCookieParams['lifetime'],
```
mauriciofauth: I think it would be interesting to check if the key exists and also add a type cast.

libraries/classes/Plugins/Auth/AuthenticationSignon.php
williamdes: I copied the line https://github.com/phpmyadmin/phpmyadmin/blob/9de453fa314863b8f728e80addfe10411e30324d/libraries/classes/Plugins/Auth/AuthenticationSignon.php#L118 

Do I fix the other occurrences (adding a if statement) ?

mauriciofauth: The PHP 7.3 signature adds support for the SameSite cookie attribute. I think it will be good to add that.

williamdes: Okay, will do that tomorrow :)

mauriciofauth: But that change is not related to the issue. Maybe open another PR for this changes?

williamdes: Yes, good idea

libraries/classes/Plugins/Auth/AuthenticationSignon.php
williamdes: Yes, in an other PR :)

MauricioFauth (approved):

nhatier (approved):
The changes work for me.

williamdes:
@mauriciofauth I made the changes you wanted, can you re-review please ?

libraries/classes/Plugins/Auth/AuthenticationSignon.php
williamdes: I move this to a function because checking that the key exists already existed in the code

libraries/classes/Plugins/Auth/AuthenticationSignon.php
williamdes: Done !

MauricioFauth (approved):
Very good!

Issue #15247, opened 2019-05-05
[Discussion] Remove defined('TESTSUITE') from codebase


williamdes:
### We can do better than that

- Remove all hacks `if (defined('TESTSUITE'))`
- Implement a session handling system for the testsuite

Please post your thoughts on why we could not remove `defined('TESTSUITE')`

mauriciofauth:
I agree with you. We can definitely find a solution that doesn't require this constant.

williamdes:
Needs to be keeped or detected for `ErrorHandler.php` & `Config.php` and `DatabaseInterface.php` for `DbiDummy`

williamdes:
- [ ] `libraries/classes/Core.php:        if (! defined('TESTSUITE')) {`
- [ ] `libraries/classes/Core.php:        if (defined('TESTSUITE')) {`
- [ ] `libraries/classes/Core.php:        if (defined('TESTSUITE')) {`
- [ ] `libraries/classes/Relation.php:        if (defined('TESTSUITE')`
- [ ] `libraries/classes/Response.php:        if (! defined('TESTSUITE')) {`
- [ ] `libraries/classes/Response.php:        if (! defined('TESTSUITE')) {`
- [ ] `libraries/classes/Html/MySQLDocumentation.php:        if (! defined('TESTSUITE') && @file_exists(ROOT_PATH . 'doc/html/index.html')) {`
- [ ] `libraries/classes/DatabaseInterface.php:        if (defined('TESTSUITE')) {`
- [ ] `libraries/classes/Session.php:        if (session_status() === PHP_SESSION_ACTIVE && ! defined('TESTSUITE')) {`
- [ ] `libraries/classes/OutputBuffering.php:        if (! defined('TESTSUITE')) {`
- [ ] `libraries/classes/InsertEdit.php:        if (! defined('TESTSUITE')) {`
- [ ] `libraries/classes/Util.php:        if (defined('TESTSUITE')) {`
- [ ] `libraries/classes/ErrorHandler.php:        if (! defined('TESTSUITE')) {`
- [ ] `libraries/classes/Plugins/AuthenticationPlugin.php:            if (! defined('TESTSUITE')) {`
- [ ] `libraries/classes/Plugins/AuthenticationPlugin.php:            if (defined('TESTSUITE')) {`
- [ ] `libraries/classes/Plugins/AuthenticationPlugin.php:        if (! defined('TESTSUITE')) {`
[...]

PR #15248, opened 2019-05-05
Remove exit from code

williamdes:
### Description

I have no idea is this is a good enhancement but I wanted to remove as much 'exit;' to be able to mock the response function in order to test code and be able to remove some if TESTSUITE from codebase

Ref #15247

### DIFF
--- a/libraries/classes/Sql.php
+++ b/libraries/classes/Sql.php
@@ -1361,7 +1361,7 @@ class Sql
      *
      * @return void
      */
-    private function sendResponseForGridEdit($result)
+    private function getResponseForGridEdit($result): void
     {
         $row = $GLOBALS['dbi']->fetchRow($result);
         $field_flags = $GLOBALS['dbi']->fieldFlags($result, 0);
@@ -1370,7 +1370,6 @@ class Sql
         }
         $response = Response::getInstance();
         $response->addJSON('value', $row[0]);
-        exit;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


williamdes:
Hi @mauriciofauth I hope this is a good idea, can you please review ?

MauricioFauth (approved):

mauriciofauth:
I want to improve how we handle request and response in the future. Using a library that implements [PSR-7](https://www.php-fig.org/psr/psr-7/) or other HTTP library.

mauriciofauth:
There is a lot of errors on Browserstack.

williamdes:
A very much smaller change but if it passes and you approve @mauriciofauth that will be a small but good step to remove the 80+ exits in our code base

libraries/classes/Controllers/Database/StructureController.php
```diff
@@ -425,7 +425,7 @@ public function multiSubmitAction(): void
                         /** @var ExportController $controller */
                         $controller = $containerBuilder->get(ExportController::class);
                         $controller->index();
-                        exit;
+                        return;
```
mauriciofauth, line 428: You also have to add a return after the `$this->multiSubmitAction();` call in the `index` method.

MauricioFauth (changes requested):

libraries/classes/Controllers/Database/StructureController.php
mauriciofauth: This is not correct. :disappointed: 
Removing these exits will change the code flow. The best way to remove them is to extract each action into a dedicated route.

williamdes: I see, that will take me more time :)
I will have a try when possible

mauriciofauth: I've been extracting these `mult_submit` actions recently. They are confusing and difficult to read.

The other changes are good changes. :+1:

PR #15254, opened 2019-05-07
Use magic variables when possible

Tithugues:
### Description

Use magic variables instead of functions when possible. This should be just a bit faster.


MauricioFauth (approved):

mauriciofauth:

ibennetch:
Not to mention, this improves readability for those values. Nice submission, @Tithugues.

Tithugues:
Thank you @ibennetch ! :)

PR #15258, opened 2019-05-07
Use "strpos" instead of "strstr" to improve performances

Tithugues:
### Description

Regarding documentation: `If you only want to determine if a particular needle occurs within haystack, use the faster and less memory intensive function strpos() instead.`


williamdes (approved):

MauricioFauth (approved):

mauriciofauth:

PR #1627, opened 2015-04-01
feature 844 structure in PDF export

whitepiegon:
Signed-off-by: Aayush Anand aayushism12@gmail.com

PR #15259, opened 2019-05-07
Remove use of unexisting / useless variables

Tithugues:
### Description

Remove use of unexisting variables
Remove useless variables

### DIFF
--- a/libraries/classes/Plugins/Export/Helpers/Pdf.php
+++ b/libraries/classes/Plugins/Export/Helpers/Pdf.php
@@ -164,10 +164,6 @@ class Pdf extends PdfLib
         $this->SetAutoPageBreak(false);
         // Check if header for this page already exists
         if (! isset($this->headerset[$this->page])) {
-            $fullwidth = 0;
-            foreach ($this->tablewidths as $width) {
-                $fullwidth += $width;
-            }
             $this->SetY($this->tMargin - ($this->FontSizePt / $this->k) * 5);
             $this->cellFontSize = $this->FontSizePt;
             $this->SetFont(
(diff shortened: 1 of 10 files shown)


libraries/classes/Plugins/Export/Helpers/Pdf.php
```diff
@@ -593,13 +589,13 @@ public function getTableDef(
          * Presently, this information is not used. We will have to find out
          * way of displaying it.
          */
-        $unique_keys = [];
+        /*$unique_keys = [];
```
mauriciofauth: Could you remove the commented code?

MauricioFauth (changes requested):

libraries/classes/Plugins/Export/Helpers/Pdf.php
Tithugues: As you can see in the comment above, it was expected to find a way to display the information.

Can you please confirm that you want me to remove the 2 comments?

Thank you.

mauriciofauth: Apparently, this piece of code was added in #1627 and was never used. I think is safe to remove it.

mauriciofauth: Maybe we can open an issue about it, I don't know. An issue is better than a comment on the code.

Tithugues: Comment removed.

MauricioFauth (approved):

mauriciofauth:

PR #15264, opened 2019-05-11
Fix property visibility

Tithugues:
### Description

`$template` was `public`. But it doesn't seem to be used externally and I don't see any reason to keep it `public`.


williamdes (approved):

MauricioFauth (approved):

mauriciofauth:

PR #15266, opened 2019-05-11
External dependency injection system

Tithugues:
### Description

What do you think about this kind of DI system please?

The current system seems to be restrictive: the definition of services should be done in PHP (which doesn't seem very readable for me), also the parameters can't be defined at the beginning (only when getting the object).

With this system, we defined the template class and it can be passed everywhere quite easily, and this may be done for other classes. Also in a second step, this would enforce the principle to use injection instead of building objects in other objects. (I would love to split the current `Response` class in `HtmlResponse` and `AjaxResponse` classes with a common `ResponseInterface`.)

Fixes #

### DIFF
--- a/ajax.php
+++ b/ajax.php
@@ -33,11 +33,8 @@ $dbi = $container->get(DatabaseInterface::class);
 
 $response->setAjax(true);
 
-$controller = new AjaxController(
-    $response,
-    $dbi,
-    $GLOBALS['PMA_Config']
-);
+/** @var AjaxController $controller */
+$controller = $containerBuilder->get(AjaxController::class);
 
 if (empty($_POST['type'])) {
     Core::fatalError(__('Bad type!'));
--- a/db_central_columns.php
+++ b/db_central_columns.php
@@ -18,6 +18,7 @@ use PhpMyAdmin\DatabaseInterface;
 use PhpMyAdmin\Di\Container;
 use PhpMyAdmin\Message;
 use PhpMyAdmin\Response;
+use Symfony\Component\DependencyInjection\Definition;
 
 global $db;
 
@@ -33,13 +34,24 @@ $response = $container->get(Response::class);
 $dbi = $container->get(DatabaseInterface::class);
 
 $centralColumns = new CentralColumns($dbi);
+/* Define dependencies for the concerned controller */
+$dependency_definitions = [
+    'db' => $container->get('db'),
+    'centralColumns' => $centralColumns,
+];
 
-$controller = new CentralColumnsController(
-    $response,
-    $dbi,
-    $db,
-    $centralColumns
+/** @var Definition $definition */
+$definition = $containerBuilder->getDefinition(CentralColumnsController::class);
+array_map(
+    static function (string $parameterName, $value) use ($definition) {
+        $definition->replaceArgument($parameterName, $value);
+    },
+    array_keys($dependency_definitions),
+    $dependency_definitions
 );
+
+/** @var CentralColumnsController $controller */
+$controller = $containerBuilder->get(CentralColumnsController::class);
 
 if (isset($_POST['edit_save'])) {
     echo $controller->editSave([
--- a/db_datadict.php
+++ b/db_datadict.php
@@ -14,6 +14,7 @@ use PhpMyAdmin\Relation;
 use PhpMyAdmin\Response;
 use PhpMyAdmin\Transformations;
 use PhpMyAdmin\Util;
+use Symfony\Component\DependencyInjection\Definition;
 
 if (! defined('ROOT_PATH')) {
     define('ROOT_PATH', __DIR__ . DIRECTORY_SEPARATOR);
@@ -34,13 +35,25 @@ $response = $container->get(Response::class);
 /** @var DatabaseInterface $dbi */
 $dbi = $container->get(DatabaseInterface::class);
 
-$controller = new DataDictionaryController(
-    $response,
-    $dbi,
-    $db,
-    new Relation($dbi),
-    new Transformations()
+/* Define dependencies for the concerned controller */
+$dependency_definitions = [
+    'db' => $container->get('db'),
+    'relation' => new Relation($dbi),
+    'transformations' => new Transformations(),
+];
+
+/** @var Definition $definition */
+$definition = $containerBuilder->getDefinition(DataDictionaryController::class);
+array_map(
+    static function (string $parameterName, $value) use ($definition) {
+        $definition->replaceArgument($parameterName, $value);
+    },
+    array_keys($dependency_definitions),
+    $dependency_definitions
 );
+
+/** @var DataDictionaryController $controller */
+$controller = $containerBuilder->get(DataDictionaryController::class);
 
 $header = $response->getHeader();
 $header->enablePrintView();
--- a/db_events.php
+++ b/db_events.php
@@ -13,6 +13,7 @@ use PhpMyAdmin\Di\Container;
 use PhpMyAdmin\Response;
 use PhpMyAdmin\Url;
 use PhpMyAdmin\Util;
+use Symfony\Component\DependencyInjection\Definition;
 
 if (! defined('ROOT_PATH')) {
     define('ROOT_PATH', __DIR__ . DIRECTORY_SEPARATOR);
@@ -33,11 +34,12 @@ $dbi = $container->get(DatabaseInterface::class);
 
 $_PMA_RTE = 'EVN';
 
-$controller = new EventsController(
-    $response,
-    $dbi,
-    $db
-);
+/** @var Definition $definition */
+$definition = $containerBuilder->getDefinition(EventsController::class);
+$definition->replaceArgument('db', $container->get('db'));
+
+/** @var EventsController $controller */
+$controller = $containerBuilder->get(EventsController::class);
 
 if (! $response->isAjax()) {
     /**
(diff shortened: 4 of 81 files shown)


MauricioFauth (approved):

mauriciofauth:
> I would love to split the current `Response` class in `HtmlResponse` and `AjaxResponse` classes with a common `ResponseInterface`.

Great idea. As I said in https://github.com/phpmyadmin/phpmyadmin/pull/15248#issuecomment-490535168, I also want to improve how we handle request and response with a HTTP library.

Tithugues:
If you find any bug to fix or any code to improve, please tell me, I'll do my best to do it.

Tithugues:
Hi,
Do you please know how to get rid of the errors thrown in https://travis-ci.org/phpmyadmin/phpmyadmin/jobs/531442853 ?

`Undefined variable: $containerBuilder`

This variable is defined in `common.inc.php` which is called at the beginning of all files which have the error. :confused: 

Thank you.

mauriciofauth:
> This variable is defined in common.inc.php which is called at the beginning of all files which have the error. :confused:

The static analyzer doesn't know about variables defined in another files. We have been ignoring these errors.

https://github.com/phpmyadmin/phpmyadmin/blob/720e302de5c1ad5db1fb1cbcad3eaaec948ca465/phpstan.neon.dist#L18

mauriciofauth:
Some ways to solve this would be to get the variable some other way, or if it is a global variable, declare it in the file.

Tithugues:
> Some ways to solve this would be to get the variable some other way, or if it is a global variable, declare it in the file.

Yes, I know this is done for other variables. But doing this would encourage use of the container through the global variable in many places, whereas the idea would rather be to provide the data as an input.
So if this is OK for you, I would prefer to fill the `phpstan.neon.dist` file.

mauriciofauth:
> So if this is OK for you, I would prefer to fill the `phpstan.neon.dist` file.

No problem.

Tithugues:
Please don't merge it for now, I've an error after the creation of a table.

Tithugues:
> Please don't merge it for now, I've an error after the creation of a table.

Please ignore my previous message, it was on the other PR I'm working on.

ajax.php
```diff
@@ -33,11 +33,8 @@
 
 $response->setAjax(true);
 
-$controller = new AjaxController(
-    $response,
-    $dbi,
-    $GLOBALS['PMA_Config']
-);
+/** @var AjaxController $controller */
+$controller = $containerBuilder->get('ajax_controller');
```
mauriciofauth: What do you think about using `AjaxController::class` instead of `'ajax_controller'`?

Tithugues: This is OK for me. I'll rename the controller services with the class name of the controllers.

db_events.php
```diff
@@ -33,11 +34,17 @@
...
-    $db
-);
+/* Define dependencies for the concerned controller */
+$dependency_definitions = [
+    'db' => $container->get('db'),
+];
+
+/** @var Definition $definition */
+$definition = $containerBuilder->getDefinition('PhpMyAdmin\Controllers\Database\EventsController');
+$definition->setArguments(array_merge($definition->getArguments(), $dependency_definitions));
```
williamdes: Is it possible to avoid having array merges?

db_multi_table_query.php
```diff
@@ -29,11 +30,17 @@
...
-    $dbi,
-    $db
-);
+/* Define dependencies for the concerned controller */
+$dependency_definitions = [
+    'db' => $container->get('db'),
+];
+
+/** @var Definition $definition */
+$definition = $containerBuilder->getDefinition('PhpMyAdmin\Controllers\Database\MultiTableQueryController');
```
williamdes: Can you use ::class ?

williamdes (changes requested):
Good work!

db_events.php
Tithugues: 2 things please:
1/ thank you because I wanted to replace it
2/ what would be your reason to avoid `array_merge` please?

db_multi_table_query.php
Tithugues: Oups… Missed this one…

db_events.php
williamdes: It feels strange, kind of broken
I would love to see a `->setArgument('name', $value);`
And also if you invert or do a mistake in the order of an array merge the output will not be what you expected.

Tithugues: OK, thanks for feedback. It should be OK now.

db_central_columns.php
```diff
@@ -33,13 +34,18 @@
...
 
-$controller = new CentralColumnsController(
-    $response,
-    $dbi,
-    $db,
-    $centralColumns
-);
+/** @var Definition $definition */
+$definition = $containerBuilder->getDefinition('database_central_columns_controller');
+$definition->setArguments(array_merge($definition->getArguments(), $dependency_definitions));
```
williamdes: @Tithugues Can you remove `array_merge` and all the other occurrences ?

db_central_columns.php
```diff
@@ -33,13 +34,18 @@
...
+];
 
-$controller = new CentralColumnsController(
-    $response,
-    $dbi,
-    $db,
-    $centralColumns
-);
+/** @var Definition $definition */
+$definition = $containerBuilder->getDefinition('database_central_columns_controller');
```
williamdes: missing `::class` :)

db_central_columns.php
```diff
@@ -33,13 +34,18 @@
...
+$definition->setArguments(array_merge($definition->getArguments(), $dependency_definitions));
+
+/** @var CentralColumnsController $controller */
+$controller = $containerBuilder->get('database_central_columns_controller');
```
williamdes: here also

db_central_columns.php
Tithugues: Yes, I'm going to try.

server_databases.php
```diff
@@ -12,26 +12,25 @@
...
 use PhpMyAdmin\Response;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 
 if (! defined('ROOT_PATH')) {
     define('ROOT_PATH', __DIR__ . DIRECTORY_SEPARATOR);
 }
 
 require_once ROOT_PATH . 'libraries/common.inc.php';
 
+$response = $containerBuilder->get('response', ContainerInterface::NULL_ON_INVALID_REFERENCE) ?? Response::getInstance();
```
williamdes: Can you use `::class` ?

williamdes (changes requested):

MauricioFauth (approved):
Excellent work!

mauriciofauth:
@Tithugues Do you think it's now possible to remove our custom DIC (`PhpMyAdmin\Di\Container`)?

server_databases.php
Tithugues: As I defined `response` as a service in the other PR, I would like to do it in the other one if this is OK for you.

db_central_columns.php
Tithugues: Done.

db_central_columns.php
Tithugues: Done.

Tithugues:
> @Tithugues Do you think it's now possible to remove our custom DIC (`PhpMyAdmin\Di\Container`)?

Not yet. But I should confess this is one on my aims…
I think we should go further before this. Your current `Container` has one advantage that I would like not to implement with the current one: `getInstance`. Thanks to this, you can retrieve every service everywhere. But if we achieve to inject dependencies instead of creating objects, we should be able to remove the current `container`, I think.
(This is a bit confusing because I think that I'm currently doing 2 things at the same time: defining the services externally thanks to the Symfony DI + injecting dependencies.)
The first things I'm currently trying to get rid of the current `container` are:
* registering every old services as a service with the new `container`;
* replacing as much as possible the `get` of services by a `get` through the new `container`.

When all the `get` will be replaced, it won't be fully finished, because there is also some magic in your `container` that loads the parameters regarding name and/or type (don't really remember, I had a quick look) of registered services.

db_central_columns.php
Tithugues: This is done!
As there is no `replaceArguments`, I had to do it for each parameter.

Later, I think that we'll have to find a way not to duplicate this closure.

williamdes: @Tithugues Seems like you did not push on the right branch :)

Tithugues: Seems you're right… :sob:

db_datadict.php
```diff
@@ -34,13 +35,19 @@
...
+/* Define dependencies for the concerned controller */
+$dependency_definitions = [
+    'db' => $container->get('db'),
+    'relation' => new Relation($dbi),
+    'transformations' => new Transformations(),
+];
+
+/** @var Definition $definition */
+$definition = $containerBuilder->getDefinition(DataDictionaryController::class);
+$definition->setArguments(array_merge($definition->getArguments(), $dependency_definitions));
```
williamdes: You missed this one :)

Tithugues: Thanks. It should be OK this time…

williamdes (approved):
:100:

Tithugues:
Do you want me to squash the commits before merging the branch?

mauriciofauth:
You can do it, if you want.

Tithugues:
Done.

MauricioFauth (approved):

mauriciofauth:
There are a few Selenium errors. You can open the link in the log to see the video of the test. 

https://travis-ci.org/phpmyadmin/phpmyadmin/jobs/534311833#L494

Tithugues:
> There are a few Selenium errors. You can open the link in the log to see the video of the test.
> 
> https://travis-ci.org/phpmyadmin/phpmyadmin/jobs/534311833#L494

Thanks, I did some fixes. I'll wait for Selenium feedback.

mauriciofauth:
Excellent work! Thank you for your contribution.

PR #15298, opened 2019-06-01
Convert Login and Home page layout to Bootstrap

murex971:
### Description
implementation of Bootstrap


shucon:
Please try to use better commit message and sign-off each commit.

templates/home/index.twig
```diff
@@ -184,6 +185,6 @@
       </ul>
     </div>
   </div>
-</div>
+</div></div></div>
```
shucon: Proper indentation of closing tags, and each tag in new line is required.

templates/login/header.twig
```diff
@@ -2,6 +2,8 @@
     <div id="modalOverlay">
 {% endif %}
 <div class="container{{ add_class }}">
+<div class="row">
+<div class="col-12">
```
shucon: Closing tags for them are not present.

shucon (changes requested):

MauricioFauth (commented):
Nice work!

You should try to keep the style of the themes (pmahomme, Metro and Original) as same as possible.
For example, the login page is looking different:
Screenshot from 2019-06-01 14-50-38

MauricioFauth (changes requested):
Also, don't forget to add your signature tag in the commit messages with `git commit -s`.

murex971:
> Nice work!
> 
> You should try to keep the style of the themes (pmahomme, Metro and Original) as same as possible.
> For example, the login page is looking different:
> Screenshot from 2019-06-01 14-50-38

I will be working on forms particularly once I am done with the grids part.

mauriciofauth:
Closed in favor of #15302

Issue #14363, opened 2018-06-05
Broken relational links in tables


Jay04653:
### Steps to reproduce
1. Open any table with related objects (FK)
2. Clink on link - working fine :heavy_check_mark:
3. Open link in new window/tab - not working (all links are **/sql.php** only) :x:

### Expected behaviour
Open related table with filter by related record

### Actual behaviour
Open page **/sql.php**

### Server configuration
**Operating system**: GNU/Linux Ubuntu 16.04 and 18.04

**Web server:** Apache 2.4

**Database:** MariaDB 10.1.31 / MySQL 5.7.21

**PHP version:** 7.1 / 7.2

**phpMyAdmin version:** 4.8.1

### Client configuration
**Browser:** Google Chrome

**Operating system:** GNU/Linux Ubuntu 18.04

nulll-pointer:
does tbl_change.php work? for me it doesn't

williamdes:
@nulll-pointer Blank page for `search` and `insert` on `5.0.0-dev` ?

nulll-pointer:
Yes

williamdes:
Maybe this is: #14388

pietersnld:
Not the same! Since sql.php changed from GET to POST there is no way you can open a link in a new tab/window. Which is really annoying since opening a link in a new tab keeps your selection from the other table.

alekc:
This one is really affecting our workflow, since you can't just right click on foreign key and see the relative child row without loosing your previous results.

Drakar2007:
For reference this issue seems to be the same as #14596 and #14270.  This issue just hit our environments a few months ago and it's starting to become a real pain.

Issue #14270, opened 2018-05-09
Middle-click on foreign key link broken


OlafvdSpek:
https://demo.phpmyadmin.net/master-config/sql.php?server=1&db=sakila&table=film_actor&pos=0

Middle-clicking on any of the actor_id rows opens a kinda 'blank' page.

fliespl:
Confirming it's still this way in 4.8.3

fliespl:
Confirming that this bug is still valid for 4.9.0.1.

williamdes:
@fliespl Yes..
Needs to be fixed soon

mauriciofauth:
This will not be fixed in 5.0 either. This requires a lot of work to be fixed, because we have to change how we handle requests. But we are working on it.

OlafvdSpek:
> This will not be fixed in 5.0 either. 

Why not? IMO it should be a showstopper.

> This requires a lot of work to be fixed, because we have to change how we handle requests. But we are working on it.

I'd use GET for simple requests and POST for the more complex ones.

williamdes:
You know what, I fixed it !!
Opening a PR in less than one hour !

cc @OlafvdSpek @mauriciofauth

PR #15325, opened 2019-06-12
Fix broken foreign key links

williamdes:
### Description

Fixes: #14270 - Middle-click on foreign key link broken
Fixes: #14363 - Broken relational links in tables

Fixes the broken foreign key links

### DIFF
--- a/libraries/classes/Core.php
+++ b/libraries/classes/Core.php
@@ -1298,7 +1298,9 @@ class Core
      */
     public static function signSqlQuery($sqlQuery)
     {
-        return hash_hmac('sha256', $sqlQuery, $_SESSION[' HMAC_secret ']);
+        /** @var array $cfg */
+        global $cfg;
+        return hash_hmac('sha256', $sqlQuery, $_SESSION[' HMAC_secret '] . $cfg['blowfish_secret']);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


williamdes:
@sudo-bot Deploy pull-request

williamdes:
@ftrotter @Drakar2007 @OlafvdSpek @Jay04653 @fliespl

Jay04653:
Hi,

Tested here: https://phpmyadmin-pr-15325.wdes.eu.org/db_structure.php?db=public&server=10
Tables: t1 and t2

Seems all working fine

MauricioFauth (changes requested):
This pull request fixes the issue but it is not safe because the token is session based and not request based, so it is possible to forge the query signature using a token from a previous request.

williamdes:
> This pull request fixes the issue but it is not safe because the token is session based and not request based, so it is possible to forge the query signature using a token from a previous request.

Nope, I think that you got confused between the the phpmyadmin session and how I use it

Currently, you can not change the contents or it will void the signature

The signature is query and session based

I will added a Unit test that proves my point of view

williamdes:
@mauriciofauth I think I understand your comment, maybe I can store a random secret that is never sent to the user into the session ?

OlafvdSpek:
> This pull request fixes the issue but it is not safe because the token is session based and not request based, so it is possible to forge the query signature using a token from a previous request.

If one has the token it's game over already isn't it?

williamdes:
> If one has the token it's game over already isn't it?

Yes absolutely, if the phpmyadmin session token is stolen then game over.

williamdes:
@sudo-bot /deploy

williamdes:
@mauriciofauth I made the secret a secret for the user (80a7f0a75c72ec2b92216647ac66890ff58002f8)
Can you please re-review ?

MauricioFauth (approved):

devenbansod (approved):
💯

williamdes:
@ibennetch @madhuracj I would like to have some feedback ;)

@madhuracj Do you have some security advice ?

I will merge soon :)

PR #14464, opened 2018-07-06
Basic initial webpack configuration for Modular js code

Piyush3079:
This pr contains the basic webpack configuration for modular js code with some basic changes in Header and Script class and the javascript AJAX handler for handling the new js code files.

Before submitting pull request, please check that every commit:

- [x] Has proper Signed-Off-By
- [x] Has commit message which describes it
- [x] Is needed on it's own, if you have just minor fixes to previous commits, you can squash them
- [ ] Any new functionality is covered by tests

PR #15331, opened 2019-06-14
Add working environment configuration directive

MauricioFauth:
### Description

Adds a configuration directive named `environment` that sets the working environment mode to `production` (default) or to `development`.

For now, it only enables the debug mode for Twig templates, however the idea is to use it in more places.

Related to the `environment` configuration introduced in #14464.

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -3433,6 +3433,18 @@ Developer
 
     These settings might have huge effect on performance or security.
 
+.. config:option:: $cfg['environment']
+
+    :type: string
+    :default: ``'production'``
+
+    Sets the working environment.
+
+    This only needs to be changed when you are developing phpMyAdmin itself.
+    The ``development`` mode may display debug information in some places.
+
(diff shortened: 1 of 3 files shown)


williamdes (approved):

doc/config.rst
```diff
@@ -3433,6 +3433,15 @@ Developer
 
     These settings might have huge effect on performance or security.
 
+.. config:option:: $cfg['environment']
+
+    :type: string
+    :default: ``'production'``
+
+    Sets the working environment.
```
ibennetch: Can we add another sentence or two about what exactly this means?

mauriciofauth: I agree. What about that:

> Sets the working environment. This only needs to be changed when you are developing phpMyAdmin itself. The _development_ mode may display some debug information in some places.

williamdes: @mauriciofauth I think your proposal is good and you can merge the PR

PR #15333, opened 2019-06-15
Add ignored files and version check in create-release.sh

MauricioFauth:
### Description

Add ignored files and version check in create-release.sh

- Remove some test suite files from release 
- Add version check for package.json in create-release.sh

### DIFF
--- a/scripts/create-release.sh
+++ b/scripts/create-release.sh
@@ -316,6 +316,8 @@ if [ $do_test -eq 1 ] ; then
     if [ $test_ret -ne 0 ] ; then
         exit $test_ret
     fi
+    # Remove PHPUnit cache file
+    rm -f .phpunit.result.cache
     # Remove libs installed for testing
     rm -rf build
     composer update --no-dev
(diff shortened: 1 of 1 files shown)


scripts/create-release.sh
```diff
@@ -134,10 +134,12 @@ if [ $do_ci -eq 0 -a -$do_daily -eq 0 ] ; then
 
 Please ensure you have incremented rc count or version in the repository :
      - in $CONFIG_LIB Config::__constructor() the line
-          " \$this->set( 'PMA_VERSION', '$version' ); "
+          " \$this->set('PMA_VERSION', '$version'); "
      - in doc/conf.py the line
           " version = '$version' "
      - in README
+     - in package.json the line
+          " "version": "$version", "
```
williamdes: also in the -lock if one exists

williamdes (approved):

scripts/create-release.sh
mauriciofauth: We don't use npm to install JavaScript dependencies, so a `package-lock.json` file doesn't exist, but even if this was not the case, the version in the `package-lock.json` file is automatically updated when running `npm install`.

ibennetch:
Merged, thanks!

PR #15338, opened 2019-06-17
Refactoring init.js of Designer

Bournvita1998:
Signed-off-by: Bournvita1998 <mohit.kuri@research.iiit.ac.in>

### Description

Refactoring in the init.js of the designer section

Fixes #

### DIFF
--- a/js/designer/init.js
+++ b/js/designer/init.js
@@ -54,15 +54,15 @@ AJAX.registerOnload('designer/init.js', function () {
     }
 
     $('#query_Aggregate_Button').on('click', function () {
-        document.getElementById('query_Aggregate').style.display = 'none';
+        $('#query_Aggregate').style.display = 'none';
     });
 
     $('#query_having_button').on('click', function () {
-        document.getElementById('query_having').style.display = 'none';
+        $('#query_having').style.display = 'none';
     });
 
     $('#query_rename_to_button').on('click', function () {
-        document.getElementById('query_rename_to').style.display = 'none';
+        $('#query_rename_to').style.display = 'none';
     });
 
     $('#build_query_button').on('click', function () {
@@ -70,6 +70,6 @@ AJAX.registerOnload('designer/init.js', function () {
     });
 
     $('#query_where_button').on('click', function () {
-        document.getElementById('query_where').style.display = 'none';
+        $('#query_where').style.display = 'none';
     });
 });


shucon (approved):

williamdes (approved):

ibennetch:
Good job.

js/designer/init.js
```diff
@@ -54,22 +54,22 @@ AJAX.registerOnload('designer/init.js', function () {
     }
 
     $('#query_Aggregate_Button').on('click', function () {
-        document.getElementById('query_Aggregate').style.display = 'none';
+        $('#query_Aggregate').style.display = 'none';
```
devenbansod: @Bournvita1998 shouldn't we be using `$("#id").css("display", "none");` or `$("#id").hide();` here?

Bournvita1998: @devenbansod I didn't find a good article comparing the difference between these 3. I know they all does the same job but no idea what's the difference. Can you help me knowing the difference between them?

Bournvita1998: @devenbansod, for now, I am thinking of moving ahead making the changes in a similar way as I have done here. If required, we will change them again.

devenbansod: @Bournvita1998 have you tested this in the browser?

AFAIK, `$('selector')` returns an array-like JQuery-native object and doesn't return a native DOM element. So if I'm not mistaken, this code change would throw an error: `Uncaught TypeError: Cannot set property 'display' of undefined` in the console.

Ref for the `css` API: https://api.jquery.com/css/

Bournvita1998: @devenbansod Yeah you are right here. $('selector') will return an array of all possible matches but document.getElementById('ID') returns just the first match it finds. To resolve this error we can either do the things as you suggested in your first comment or we can change $('selector') to $('selector')[0]. I found [this](https://stackoverflow.com/questions/6974582/jquery-object-and-dom-element) on stackoverflow explaining the difference between jQuery object and DOM elements.

Also, I found the same issue in my previously created PRs too, which need to be changed soon. I am thinking of replacing them using the css format like:  $("#id").css("display", "none");  is that fine to move ahead?

devenbansod: Sure, please go ahead 👍

williamdes: This was indeed invalid: https://github.com/phpmyadmin/phpmyadmin/issues/16960

williamdes:
Gosh, why did I approve this invalid code.. 🤦🏻 
Ref: #16960

PR #15344, opened 2019-06-19
Implemented password_hash function

### LINKED ISSUE #14772: php_password_hash input function
When inserting data, it'd be handy if PHP's password_hash function was available. 
Unlike the other functions this one has to be run by PHP though.

http://php.net/password_hash

kartik1000:

### Description

Password_hash function has been implemented through this PR by php. To use it you must go to Insert field of any column and then select 'PASSWORD_HASH' as a function in some column which supports text like when we use 'PASSWORD' function. Then Enter the desired password and click on 'Go' and you will see that the row has been inserted with password in the form of hash from password_hash function.
 
Fixes #14772 

### DIFF
--- a/js/table/change.js
+++ b/js/table/change.js
@@ -184,6 +184,15 @@ function verificationsAfterFieldChange (urlField, multiEdit, theType) {
         $('#salt_' + target.id).remove();
     }
 
+    // Alert Message for PASSWORD_HASH to alert it as a PHP function and not SQL function
+    var new_alert_box = '<br><p name=salt[multi_edit][' + multi_edit + '][' + urlField + ']' +
+        ' id=salt_' + target.id + ' placeholder=\'' + PMA_messages.strEncryptionKey + '\'+ style="background-color:#FEF3CD;color:#8E6F29;padding:20px;">' + 'ALERT: PASSWORD_HASH is a PHP function and not a SQL function' + '<p>';
+    if (target.value === 'PASSWORD_HASH'){
+        if (!($('#salt_' + target.id).length)) {
+            $this_input.after(new_alert_box);
+        }
+    }
+
     if (target.value === 'AES_DECRYPT'
             || target.value === 'AES_ENCRYPT'
             || target.value === 'MD5') {
--- a/libraries/classes/InsertEdit.php
+++ b/libraries/classes/InsertEdit.php
@@ -2613,6 +2613,12 @@ class InsertEdit
                     . $this->dbi->escapeString($multi_edit_salt[$key]) . "')";
             }
 
+            // php Implementation of password_hash() function
+            elseif ($multi_edit_funcs[$key] == "PASSWORD_HASH") {
+                $current_value = password_hash($current_value,PASSWORD_DEFAULT);
+                return '\'' . $current_value . '\'';
+            }
+
             return $multi_edit_funcs[$key] . '(' . $current_value . ')';
         }
 
--- a/libraries/classes/Types.php
+++ b/libraries/classes/Types.php
@@ -501,6 +501,7 @@ class Types
                     'MONTHNAME',
                     'OLD_PASSWORD',
                     'PASSWORD',
+                    'PASSWORD_HASH',
                     'QUOTE',
                     'REVERSE',
                     'RTRIM',


williamdes (approved):
:100:

williamdes:
Looks good to me, good enhancement! 

@sudo-bot Deploy pr

ibennetch:
This looks great. I have two thoughts and I'm not certain about either of them:

1) Since PHP can take arguments for this function, should we prompt for them like we do for `AES_ENCRYPT`? If we do this, it probably will require two extra fields (one for the algorithm, one for the options), but I'm not certain whether we should add this.

2) I wonder if we should indicate in some way that this is the PHP function, rather than a MySQL built-in. I'm thinking an informational icon with hover text like "This is a PHP function" or some such.

kartik1000:
@ibennetch, I am not certain about the first one as well, because generally  i have used it for single field only, which is input password and create its hash using the function. Second one I think should be done, because some user might try to use the function in SQL Query after seeing it in list, so it will be good to inform them earlier that they cannot use it using SQL Query. Should I start working for Second one? What do you think @williamdes and @ibennetch ?

kartik1000:
@williamdes, @ibennetch I have added an alert related to it. Please look if it the position looks ok to you :)

kartik1000:
@williamdes can we think on adding this enhancement?

williamdes:
@kartik1000 If you can rebase onto latest master version ?

I want to implement this feature but you need to do something about https://github.com/phpmyadmin/phpmyadmin/pull/15344#issuecomment-504443907

Maybe you can add subcategory for php functions ?
https://www.w3schools.com/tags/tag_optgroup.asp

kartik1000:
@williamdes I have already added an alert message which @ibennetch suggested in 2nd option

kartik1000:
I will first rebase so you can deploy and see it.

kartik1000:
@williamdes  I have rebased with latest master. Please review. In case you don't see the alert-box to be fit then we can discuss some other option maybe subcategory for php functions.

js/table/change.js
```diff
@@ -184,6 +184,15 @@ function verificationsAfterFieldChange (urlField, multiEdit, theType) {
         $('#salt_' + target.id).remove();
     }
 
+    // Alert Message for PASSWORD_HASH to alert it as a PHP function and not SQL function
+    var new_alert_box = '<br><p name=salt[multi_edit][' + multi_edit + '][' + urlField + ']' +
+        ' id=salt_' + target.id + ' placeholder=\'' + PMA_messages.strEncryptionKey + '\'+ style="background-color:#FEF3CD;color:#8E6F29;padding:20px;">' + 'ALERT: PASSWORD_HASH is a PHP function and not a SQL function' + '<p>';
```
williamdes: I would like that another solution is found:
- The place holder is strange `'strEncryptionKey' => __('Encryption key'),`
- The string is not translated
- CSS is inlined
- HTML is injected

@kartik1000 can you please improve this ?

libraries/classes/InsertEdit.php
```diff
@@ -2613,6 +2613,12 @@ public function getCurrentValueAsAnArrayForMultipleEdit(
                     . $this->dbi->escapeString($multi_edit_salt[$key]) . "')";
             }
 
+            // php Implementation of password_hash() function
+            elseif ($multi_edit_funcs[$key] == "PASSWORD_HASH") {
```
williamdes:
```suggestion
            elseif ($multi_edit_funcs[$key] === "PASSWORD_HASH") {
```

libraries/classes/InsertEdit.php
```diff
@@ -2613,6 +2613,12 @@ public function getCurrentValueAsAnArrayForMultipleEdit(
...
+                $current_value = password_hash($current_value,PASSWORD_DEFAULT);
```
williamdes:
```suggestion
                $current_value = password_hash($current_value, PASSWORD_DEFAULT);
```

williamdes (changes requested):
I added some comments, thank you for the rebase

js/table/change.js
kartik1000: Yes, sure It was just a proposal so that you could know how I am thinking of the alert box. Do I need to modify this alert box or are we planning something different? @williamdes

mauriciofauth:
Thanks for your contribution @kartik1000.

PR #15348, opened 2019-06-22
Add Charset and Collation value objects

MauricioFauth:
### Description

Introduces Charset and Collation value objects. This way we can have a well defined structure instead of an array of strings, witch is prone to errors.

### DIFF
--- a/libraries/classes/Charsets.php
+++ b/libraries/classes/Charsets.php
@@ -9,8 +9,8 @@ declare(strict_types=1);
 
 namespace PhpMyAdmin;
 
-use PhpMyAdmin\DatabaseInterface;
-use PhpMyAdmin\Util;
+use PhpMyAdmin\Charsets\Charset;
+use PhpMyAdmin\Charsets\Collation;
 
 /**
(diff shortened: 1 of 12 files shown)


mauriciofauth:
@williamdes Why my commit has an invalid Signed-Off-By line?

See: https://github.com/phpmyadmin/phpmyadmin/pull/15348#commitcomment-34042130

williamdes:
@mauriciofauth I think this is because https://github.com/phpmyadmin/scripts/blob/23993f4de420b741668672f1f49807b5eeb4a90f/hooks/commits.php#L121 uses a line break in the pattern

Try to push a commit that has a line break before the signoff

mauriciofauth:
> Try to push a commit that has a line break before the signoff

My commit message has a line break. See:

Screenshot from 2019-06-22 23-45-50

williamdes:
@mauriciofauth I found out why, the regex was missing the unicode modifier.
Fixed by https://github.com/phpmyadmin/scripts/commit/153613e5362f301c07c206744d560977daf50e1c

Issue #14507, opened 2018-07-21
Bug report: Width of columns are never saved


diyoyo:
### Steps to reproduce
1.  Select a table to browse
2.  Adjust one of the column's width
3. Come back on that very same page later

### Expected behaviour
The width of the column should be set to the same as the latest modification

### Actual behaviour
The width is back to default behavior.

### Server configuration
**Operating system**:

**Web server:**

**Database:**

**PHP version:**

**phpMyAdmin version:**
4.8.2

### Client configuration
**Browser:**
Firefox, Chrome
**Operating system:**

lem9:
Not a bug because the width is not supposed to be saved. Could be a feature request but this is open to discussion about what would be the best behavior.

bozhi91:
Hey guys. Can I implement this feature?

bozhi91:
williamdes, Can you assign me and my team to this feature. We are interested in implement it. 
We need this for a school project.

williamdes:
@bozhi91  I can not assign you or your team since you are not phpmyadmin members but you are now "assigned" on this enhancement

Please keep us updated if you are not interested anymore or are actively working on this issue :)

bozhi91:
Hi William,

Thanks for your reply.

My idea is to implement this improvement and then push the changes and hope
the one that's in charge will approve it.
Since I'm assigned on the enhancement, does that mean that we can start
with the implementation of the feature now?

Regards,

williamdes:
You can start the implementation, be sure to use latest master branch as base for your work and to target master when creating the pull-request
Please use the  - -sign-off option of git when creating commits
I will review the pull-request and decide if more work needs to be done or if I can merge your enhancement
Other team members will maybe review the pull-request

bozhi91:
Many thanks.

I'll will inform you and the other guys about the progress. Also we'll
probably need your help at some
point. I guess we can count on you, right?

Regards.

williamdes:
Yes, I will help you if you need :)

bozhi91:
Hi William,
I want to comment you something about the project.
We started the development and we decided to store the table's width in cookies instead of using another tables for saving all the data. This way could be more comfortable for the users because only the users using the same computer may see the changes. If someone connects remotely, will not see the changes because does not have the data stored in cookies.
Also will be more easy to implement this functionality because we will apply a minor changes in the code.
Hope this will not be a problem.

Regards,

williamdes:
Seems okay :+1: 
[...]

PR #15360, opened 2019-06-25
#14507 - Save width of columns

bozhi91:
### Description

We fixed the issue number:  #14507 ( Save width of columns.)
Currently, the phpmyadmin is putting a column width by default. But when the user changes the column width and refresh the page, the changes are lost.

### Fixes 
 Our task was to fix this little issue and store the column width when the user modifies it.
When the page is reloaded or the session is closed and opened again, the changes are saved and the column width is loaded as the user set it the last time.

### DIFF
--- a/js/makegrid.js
+++ b/js/makegrid.js
@@ -14,6 +14,74 @@
  * @param enableVisib Optional, if false, show/hide column feature will be disabled
  * @param enableGridEdit Optional, if false, grid editing feature will be disabled
  */
+ //line 247
+function storeTable(n,nw){
+
+var url_string = window.location.href;
+var url   = new URL(url_string);
+var db    = url.searchParams.get("db");
+var table = url.searchParams.get("table");
+var key   = db+'-'+table;
+
+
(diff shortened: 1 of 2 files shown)


readme.txt
```diff
@@ -0,0 +1 @@
+This is a test
```
williamdes: must be removed

js/makegrid.js
```diff
@@ -2168,10 +2236,10 @@ function PMA_makegrid (t, enableResize, enableReorder, enableVisib, enableGridEd
     g.tableCreateTime = $(g.o).find('.table_create_time').val();
 
     // assign the hints
-    g.sortHint = Messages.strSortHint;
```
williamdes: seems to have a merge issue

williamdes (changes requested):

williamdes:
Thank you for contributing!, for now it can not be merged as is and has conflicts
If you want to improve the PR you can ask that we re-open it

PR #15376, opened 2019-07-06
Fix/minor improvements

Tithugues:
### Description

This MR includes many improvements on many topics:
* merge isset / unset
* use constants in some places
* use relevant static calls when needed

### DIFF
--- a/libraries/classes/Config.php
+++ b/libraries/classes/Config.php
@@ -1337,7 +1337,7 @@ class Config
             $is_https = true;
         } elseif (strtolower(Core::getenv('HTTPS')) == 'on') {
             $is_https = true;
-        } elseif (substr(strtolower(Core::getenv('REQUEST_URI')), 0, 6) == 'https:') {
+        } elseif (strtolower(substr(Core::getenv('REQUEST_URI'), 0, 6)) == 'https:') {
             $is_https = true;
         } elseif (strtolower(Core::getenv('HTTP_HTTPS_FROM_LB')) == 'on') {
             // A10 Networks load balancer
--- a/libraries/classes/Config.php
+++ b/libraries/classes/Config.php
@@ -447,7 +447,7 @@ class Config
         if (false !== strpos($ref_head, '/')) {
             // remove ref: prefix
             $ref_head = substr(trim($ref_head), 5);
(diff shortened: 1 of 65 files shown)


libraries/classes/Config.php
```diff
@@ -439,7 +439,7 @@ public function checkGitRevision(): void
         }
 
         if ($common_dir_contents = @file_get_contents($git_folder . '/commondir')) {
-            $git_folder = $git_folder . DIRECTORY_SEPARATOR . trim($common_dir_contents);
+            $git_folder .= trim($common_dir_contents) . DIRECTORY_SEPARATOR;
```
williamdes: You inverted the operation

libraries/classes/Export.php
```diff
@@ -589,7 +589,7 @@ public function exportServer(
         string $separate_files
     ): void {
         if (! empty($db_select)) {
-            $tmp_select = implode($db_select, '|');
+            $tmp_select = implode('|', $db_select);
```
williamdes: Good catch!

williamdes (changes requested):
LGTM
I left one comment

libraries/classes/Config.php
Tithugues: You're right, sorry for that.
This should be fixed now.

MauricioFauth (approved):

libraries/classes/Util.php
```diff
@@ -3652,7 +3651,7 @@ public static function getGISFunctions(
             'type' => 'int',
         ];
 
-        $geom_type = trim(mb_strtolower((string) $geom_type));
+        $geom_type = mb_strtolower(trim((string)$geom_type));
```
mauriciofauth: The coding standard requires 1 space after the cast statement.

```suggestion
        $geom_type = mb_strtolower(trim((string) $geom_type));
```

williamdes (approved):

libraries/classes/Util.php
Tithugues: Sorry. Fixed.

mauriciofauth:
Merged in ffdc2cf7bf66b1e7f1090b9d05794483bc3f4df0. Thanks for your contribution.

PR #15378, opened 2019-07-08
Refactoring move.js of Designer

Bournvita1998:
Signed-off-by: Mohit Kuri <mohit.kuri@research.iiit.ac.in>

### Description

Refactored move.js of the designer.

Fixes #

### DIFF
--- a/js/designer/move.js
+++ b/js/designer/move.js
@@ -101,7 +101,7 @@ DesignerMove.mouseDown = function (e) {
     }
 
     if (curClick !== null) {
-        $('#canvas')[0].style.display = 'none';
+        $('#canvas').css("display", 'none');
         curClick.style.zIndex = 2;
     }
 };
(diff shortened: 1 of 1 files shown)


Bournvita1998:
For testing, I have tested this whole thing in general. Here is an example of the same:

Using - document.getElementById('selector')

YO

Using - $('#selector')[0]

YO2

Since this worked, I replicated the same here.

@devenbansod Please review and provide feedback for the same. Will proceed further once the work till here is approved and we are good to go ahead.

js/designer/move.js
```diff
@@ -101,7 +101,7 @@ DesignerMove.mouseDown = function (e) {
     }
 
     if (curClick !== null) {
-        document.getElementById('canvas').style.display = 'none';
+        $('#canvas')[0].style.display = 'none';
```
devenbansod: @Bournvita1998 this would work for the elements selected with IDs only. I'd ideally use `.css()` method to set the styles instead of manually selecting the first element and setting the native properties. IMO (even though it works) this current approach sort of doesn't use JQuery's power at all.

Using `.css()` would auto-loop over all the selected elements, so would be easier to read when we're selecting based on input types, classes etc.

devenbansod (changes requested):
@Bournvita1998 I've left a comment. It applies to all the changes.

js/designer/move.js
shucon: I agree with @devenbansod . Don't try to hard-code things. As you're trying to refactor the code, try to do it in such a way that would let other developers work on the code without much changes to your code. It gets extremely difficult to manage such things on such a large scale.

Bournvita1998: Okay, I will make these changes asap. Thanks @devenbansod, @shucon

Bournvita1998:
@devenbansod @shucon Please have a look now

Bournvita1998:
I am done refactoring move.js once. Please share your reviews so that I can make the possible changes

PR #15403, opened 2019-07-26
Fix missing variable on insert page

Tithugues:
Since the use of routes, the context changed because "entrypoints" are not called directly anymore, but are called in closures, so the context is not shared. So if you want to use a global variable that was in the parent context, you now need to explicitely ask for the global variable.
In our case, "$is_upload" wasn't shared anymore.


MauricioFauth (approved):

mauriciofauth:
Thank you for fixing this.

PR #15405, opened 2019-07-27
Fix PHP 7.4 deprecation: accessing offset on null/bool

jfcherng:
`null['xxx']` or `false['xxx']` emits a notice in PHP 7.4. Previously, they just return `null` directly.

I kinda clicked all possible links/buttons on my phpMyAdmin and found these.
This PR may be considered dirty but at least they indicate where problems are.

Problematic pages:

- phpMyAdmin/db_tracking.php?db=DB_NAME
- phpMyAdmin/server_privileges.php?adduser=1
- some other pages using POST method...
  image

https://github.com/php/php-src/blob/0e6e2297fcb27103cc5d550ee8434680ff1e879e/UPGRADING#L25-L28

---

There is still another deprecation, which is the most annoying actually, from upstream though.

```
Deprecation Notice in .\vendor\symfony\cache\Adapter\ArrayAdapter.php#48
 Unbinding $this of closure is deprecated
```

**Update:** The fix [PR](https://github.com/symfony/symfony/pull/32796/files) has been merged.

### DIFF
--- a/libraries/classes/Display/Results.php
+++ b/libraries/classes/Display/Results.php
@@ -1183,6 +1183,8 @@ class Results
         $number_of_columns = $this->__get('fields_cnt');
 
         for ($j = 0; $j < $number_of_columns; $j++) {
+            // PHP 7.4 fix for accessing array offset on bool
+            $col_visib_current = is_array($col_visib) ? $col_visib[$j] : null;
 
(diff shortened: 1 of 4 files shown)


williamdes (approved):

williamdes:
Merged, thanks @jfcherng !

PR #15411, opened 2019-07-29
Fix #14291 Incorrect linkage from one table's value to another table

### LINKED ISSUE #14291: Incorrect linkage from one table's value to another table.
### Steps to reproduce
Please see video:
https://cdn.beamstyle.com.hk/phpmyadmin-bug.mp4

### Expected behaviour
When clicked, should be able to go to the table with correct query with full value, and not with the "ellipsis" (...) behind it.

### Actual behaviour
Goes to the table with wrong query (the query has ellipsis "..." behind it)

(But if you select +Options, and display as "Full Text" then it works)

### Server configuration
**Web server:**
Apache

**Database:**
5.7

**PHP version:**
7.0
**phpMyAdmin version:**
4.7.4

Tithugues:
Fixes: #14291

### DIFF
--- a/libraries/classes/Display/Results.php
+++ b/libraries/classes/Display/Results.php
@@ -5197,6 +5197,7 @@ class Results
      *                                                     field
      * @param array                 $map                   the list of relations
      * @param string                $data                  data
+     * @param string                $displayedData         data that will be displayed (maybe be chunked)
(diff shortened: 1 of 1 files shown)


williamdes:
Merged as 40ed1557758d36d64957b4f88cc745be26f63f55

PR #15415, opened 2019-08-03
Tables imp

murex971:
### Description

Table classes added to database and table pages.

### DIFF
--- a/themes/pmahomme/scss/_common.scss
+++ b/themes/pmahomme/scss/_common.scss
@@ -644,8 +644,8 @@ table tr {
   }
 }
 
-.table-hover tbody tr:hover th,
-.table-hover tbody tr:not(.disabled):hover td {
+.table-hover tbody tr:hover td,
+.table-hover tbody tr:not(.disabled):hover th {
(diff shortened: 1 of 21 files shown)


themes/pmahomme/scss/_common.scss
```diff
@@ -958,8 +971,8 @@ div.tools a:hover,
 
 tr.disabled td,
 td.disabled {
-  background-color: #f3f3f3;
-  color: #aaa;
+  background-color: #f3f3f3 !important;
+  color: #aaa !important;
```
mauriciofauth: It is not recommended to use `!important`. It is possible to do it in another way?

MauricioFauth (changes requested):
You also need to change the CSS of the others themes, not only _pmahomme_.

themes/pmahomme/scss/_common.scss
murex971: I found this as the most easy and effective way to add the things i wanted to use.
without adding !important it will become difficult to implement those property over bootstrap default css property.

PR #15428, opened 2019-08-11
Password String escaped correctly Fix#15202

### LINKED ISSUE #15202: Creating user with single quote in password results in no password user
Tried on demo.phpmyadmin 4.8.5 and 4.8.6-dev
Go to User accounts tab, Add user account
Username: test
Password: test'123
And submit/create...afterwards go back to User accounts tab and the new user will be listed as no password.

kartik1000:

### Description

The commit ensures that password string is escaped correctly while creating a new user account and adding password. It uses PHP addslashes() function to correctly escape string. 

Fixes #15202 

### DIFF
--- a/libraries/classes/Server/Privileges.php
+++ b/libraries/classes/Server/Privileges.php
@@ -5130,6 +5130,7 @@ class Privileges
      */
     public static function getHashedPassword($password)
     {
+        $password = $GLOBALS['dbi']->escapeString($password);
         $result = $GLOBALS['dbi']->fetchSingleRow(
             "SELECT PASSWORD('" . $password . "') AS `password`;"
         );


libraries/classes/Server/Privileges.php
```diff
@@ -5131,7 +5131,7 @@ public static function addUserAndCreateDatabase(
     public static function getHashedPassword($password)
     {
         $result = $GLOBALS['dbi']->fetchSingleRow(
-            "SELECT PASSWORD('" . $password . "') AS `password`;"
+            "SELECT PASSWORD('" . addslashes($password) . "') AS `password`;"
```
williamdes: You could use `addcslashes($password, '"');` ? (I did not try it but seems more appropriate since only `"` must be escaped

https://www.php.net/manual/en/function.addcslashes.php

```suggestion
            "SELECT PASSWORD('" . addcslashes($password, '"') . "') AS `password`;"
```

kartik1000: @williamdes, I checked by making the change, but it is not working as the issue is not being resolved and it shows Password as No in User Accounts

williamdes: @kartik1000 okay, I removed my suggestion
```sql
SELECT PASSWORD(''),PASSWORD('"'),PASSWORD('\''),PASSWORD("'"),PASSWORD("\"")
```

libraries/classes/Server/Privileges.php
```diff
@@ -5131,7 +5131,7 @@ public static function addUserAndCreateDatabase(
     public static function getHashedPassword($password)
     {
```
williamdes:
```suggestion
     {
     $password = $this->dbi->escapeString($password);
```

libraries/classes/Server/Privileges.php
williamdes: @kartik1000 here is a better one: https://github.com/phpmyadmin/phpmyadmin/pull/15428/files#r312753043

williamdes (changes requested):

libraries/classes/Server/Privileges.php
kartik1000: image
The error I am getting after making the change.

williamdes: Use `$GLOBALS['dbi']->` instead :)

kartik1000: This one is working @williamdes :) Shall I make the change then?

williamdes: Yes, absolutely :)

williamdes (approved):

williamdes (approved):
Thanks Kartik, merged !

PR #15437, opened 2019-08-17
Fixed url.php crashing because it's not loading the DatabaseInterface service

AbelianKraun:
### Description

The url.php script was crashing because not finding the DatabaseInterface service in the container. This happend because the `PMA_MINIMUM_COMMON ` is defined and the
`$containerBuilder->set(DatabaseInterface::class, DatabaseInterface::load());` 
is not called.

The exception appear in the CheckUserPrivilages construct when the Response::getInstance() is called.

### Fixes

Added the DatabaseInterface service in the container in the url.php script.

### DIFF
--- a/url.php
+++ b/url.php
@@ -10,6 +10,7 @@ declare(strict_types=1);
 use PhpMyAdmin\Core;
 use PhpMyAdmin\Response;
 use PhpMyAdmin\Sanitize;
+use PhpMyAdmin\DatabaseInterface;
 
 if (! defined('ROOT_PATH')) {
     define('ROOT_PATH', __DIR__ . DIRECTORY_SEPARATOR);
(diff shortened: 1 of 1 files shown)


williamdes:
@mauriciofauth I think you are the right person to review this fix ;)

williamdes:
@AbelianKraun I would like to merge your work, can you help me reproduce the issue ?

AbelianKraun:
@williamdes I have just:

1) Cloned the git repository
2) Installed dependecies (node and composer)
3) Changed the config to point to my mysql server
4) When I opened a DB, phpmyadmin suggested to create it's config table in the database and a link to the docs. When i clicked the link..

image

williamdes (commented):
Thanks for the explanation, I could reproduce it when clicking on a documentation button (i)

I rebased the branch on `QA_5_0` where the bug was also there.

Merged :tada:

PR #15445, opened 2019-08-19
Fix NULL wrongly checked on field change

Tithugues:
When a field was requested to be changed, the "null" field was compared to its previous value. Unfortunately, both fields don't have the same value, so a change was always identified.


williamdes (approved):
LGTM

PR #15452, opened 2019-08-21
Display an error when trying to import in designer a table that's already imported

Tithugues:
It's currently possible to add many times the same table in the designer. This should not be allowed.

I used an existing message… Please tell me if a new one should be used and how to do it.

Thank you.

### DIFF
--- a/js/messages.php
+++ b/js/messages.php
@@ -751,6 +751,9 @@ $js_messages['strStrong'] = __('Strong');
 $js_messages['strU2FTimeout'] = __('Timed out waiting for security key activation.');
 $js_messages['strU2FError'] = __('Failed security key activation (%s).');
 
+/* Designer */
+$js_messages['strTableAlreadyExists'] = __('Table %s already exists!');
+
 echo "var PMA_messages = new Array();\n";
 foreach ($js_messages as $name => $js_message) {
     Sanitize::printJsValue("PMA_messages['" . $name . "']", $js_message);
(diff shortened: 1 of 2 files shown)


williamdes (approved):
LGTM :100:

js/messages.php
```diff
@@ -751,6 +751,9 @@ function () {
 $js_messages['strU2FTimeout'] = __('Timed out waiting for security key activation.');
 $js_messages['strU2FError'] = __('Failed security key activation (%s).');
 
+/* Designer */
+$js_messages['strTableAlreadyExists'] = __('Table %s already exists!');
```
williamdes:
```suggestion
$js_messages['strTableAlreadyExists'] = _pgettext('The table already exists in the designer and can not be added once more.', 'Table %s already exists!');
```

williamdes (changes requested):
I added a suggestion to add context for translators, do you agree ?

js/messages.php
Tithugues: Honnestly, I don't really know the impact of this change. I looked at the `_pgettext` function, but don't really this the need for the context.
So I'll rely on you and commit your suggestion.

williamdes: The impact is that the translators know what they are translating
https://github.com/phpmyadmin/phpmyadmin/blob/a2308ee965b70d77d74f3570f60cc6fba9580637/po/phpmyadmin.pot#L912
image
https://hosted.weblate.org/translate/phpmyadmin/4-9/fr/?checksum=&offset=1&q=Table&search=exact&type=translated&ignored=&date=&only_user=&exclude_user=

PR #15466, opened 2019-09-10
Refactor the server databases entry point

MauricioFauth:
Related to #15377


williamdes (approved):

mauriciofauth:
@williamdes Do you think we should handle the response in the routes.php, inside de controller's action or somewhere else? In this pull request it is inside the routes.

williamdes:
> Do you think we should handle the response in the routes.php, inside de controller's action or somewhere else?

@mauriciofauth Inside de controller's action seems to be the best solution for now :)

mauriciofauth:
I agree that we should handle the response inside the controller, but I think we should refactor the Response class first.

Issue #15161, opened 2019-04-07
[UI] Text area overflows its parent element


AghaSaad04:
### Describe the bug

Text Area is overflowing its parent div

### To Reproduce

1. Select a DB
2. Click on 'Query'
3. Scroll down
4. See text area overflowing

### Expected behavior

It should not overflow its parent

### Screenshots

![image](https://user-images.githubusercontent.com/36513474/55687654-df66ee80-5988-11e9-99f2-76b6d1617563.png)

### Server configuration

- Operating system: Windows
- Web server: XAMPP
- PHP version: 7.3.0
- phpMyAdmin version: 5.0.0-dev

### Client configuration

- Browser: Firefox
- Operating system: Windows

asprazz:
Hie @williamdes Sir, I am not able to reproduce the bug as it seems fine with me in local as well as remote server :)
![Screenshot from 2019-05-30 20-49-37](https://user-images.githubusercontent.com/19568165/58643223-7f534f80-831c-11e9-9cb2-3cb94d98a777.png)
![Uploading Screenshot from 2019-05-30 20-49-14.png…]()

falconfx:
I am still trying to validate the PR so that it passes all the tests, but here I have made a possible approach to the solution #15342

PR #15342, opened 2019-06-17
Fix #15161 - Text area-overflows parent element

falconfx:
### Description

I was able to replicate the behavioral error in #15161 and fix them

Before submitting pull request, please review the following checklist:

- [x] Make sure you have read our [CONTRIBUTING.md](https://github.com/phpmyadmin/phpmyadmin/blob/master/CONTRIBUTING.md) document.
- [x] Make sure you are making a pull request against the correct branch. For example, for bug fixes in a released version use the corresponding QA branch and for new features use the _master_ branch. If you have a doubt, you can ask as a comment in the bug report or on the mailing list.
- [x] Every commit has proper `Signed-off-by` line as described in our [DCO](https://github.com/phpmyadmin/phpmyadmin/blob/master/DCO). This ensures that the work you're submitting is your own creation.
- [x] Every commit has a descriptive commit message.
- [x] Every commit is needed on its own, if you have just minor fixes to previous commits, you can squash them.
- [x] Any new functionality is covered by tests.

PR #15478, opened 2019-09-27
FIX: [UI] Text area overflows its parent element #15161

ecordoba24:
FIX: [UI] Text area overflows its parent element #15161

### DIFF
--- a/libraries/classes/Database/Qbe.php
+++ b/libraries/classes/Database/Qbe.php
@@ -1809,7 +1809,7 @@ class Qbe
             );
         $html_output .= '</legend>';
         $text_dir = 'ltr';
-        $html_output .= '<textarea cols="80" name="sql_query" id="textSqlquery"'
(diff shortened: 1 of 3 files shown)


libraries/classes/Database/Qbe.php
```diff
@@ -1809,7 +1809,7 @@ public function getSelectionForm()
             );
         $html_output .= '</legend>';
         $text_dir = 'ltr';
-        $html_output .= '<textarea cols="80" name="sql_query" id="textSqlquery"'
+        $html_output .= '<textarea style="width: 98%;" cols="80" name="sql_query" id="textSqlquery"'
```
mauriciofauth: Could you move this to a CSS file instead of using inline styles?

MauricioFauth (changes requested):

williamdes:
Sorry but I merged #15342 since this PR does not have a valid Signed-Off-By line and has debug code.

williamdes:
Maybe you can try to solve another issue :)

PR #15487, opened 2019-10-02
Origin

ArrayMy:

### DIFF
--- a/config.sample.inc.php
+++ b/config.sample.inc.php
@@ -149,6 +149,13 @@ $cfg['SaveDir'] = '';
 //$cfg['SendErrorReports'] = 'always';
 
 /**
+ * Variable for select theme
+ *
+ * Default available options
+ * ('original' / 'metro' / 'pmahomme')
+ */
+//$cfg['ThemeDefault'] = 'original'
+/**
  * You can find more configuration options in the documentation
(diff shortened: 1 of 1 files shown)


williamdes (approved):

ibennetch:
Hello and thank you for submitting a pull request.

It appears as if you have not used your real name in the signoff message. For compliance reasons, we are required to ask you for your name. You can modify your existing commit with `git commit —amend` and pushing again to your repository (it may be necessary to `—force` push).

If you have any questions or concerns, please let us know.

ibennetch:
I'm not sure whether this is needed; the point of `config.sample.inc.php` is to keep a small example file that new users can refer to, and by the point they wish to modify the default theme they should probably refer to the manual or setup script to customize their installation. I'm just concerned about adding too much to the sample file and wonder what others think of this.

ArrayMy:
I'm working on adjusting the installation in which this change affects, I know, I probably do not in any case it may not be completely unnecessary.

williamdes:
Since the repo is deleted you can not update the PR, so we need to decide if we merge or not

ibennetch:
I think I'll close this due to the reasons I cited above; if you wish to change the theme there's still the dropdown on the main page, or to apply a default theme site-wide you can edit `config.inc.php` directly.

Thanks and best wishes.

PR #15492, opened 2019-10-06
Add initial suport for prepared statements

MauricioFauth:
Creates the DbiExtension::prepare method.

https://www.php.net/manual/en/mysqli.quickstart.prepared-statements.php


williamdes (approved):
:100: Good idea, this will avoid security issues

devenbansod (approved):
Great idea. 💯 🥇

Issue #12373, opened 2016-07-08
Theme selection for future releases


nijel:
I think it's time to reconsider which themes are shipped with phpMyAdmin. The Original theme is really showing it's age also it might be good idea to include Metro which is IMHO way nicer.

alextech:
As long as have a choice. Pmahomme at least should remain. I stopped updating frontend software in past year or moved to alternatives of those switching to flat UI like Metro.

Lyko93:
In my free time I was thinking about seeing what I can do with [Semantic UI](http://semantic-ui.com/) for a theme.

nijel:
This issue was a bit forgotten, but maybe it's not too late to reopen this for 4.7, given that 4.7.0-beta1 should be out soon. On the other side with beta out, the theme developers might have good chance to work on their themes to fix any possible issues with 4.7 branch.

However I'd really like to have new theme as current one is really showing it's age. On the other side we have huge amount of conservative users still preferring the original theme (which was replaced by pmahome as default more than six years ago, see #5959 and #6050), so I'm not really sure if we're going to add another theme (what would be additional maintenance burden) or replace one of the existing ones (and receive negative feedback about removing theme people are used to).

Overall there are quite some things to decide :-).

In the end what might be doable plan:

* Release 4.7.0 with current set of themes.
* Target 5.0.0 as next major version to be released in few months. This assumes no theme related changes will happen between these two releases.
* Open theme contest just after releasing 4.7.0-beta1.
* Vote for themes to include in 5.0.0 after releasing 4.7.0.

Given our current two months release schedule that could be:

* End of January
  * 4.7.0-beta1 is released
  * Theme contest is announced
* End of March
  * 4.7.0 is released
  * Theme submission deadline
  * Voting for themes is started
* End of April
  * Theme voting deadline
* End of May
  * 5.0.0 is released with new theme

What does @phpmyadmin/developers think?

smartcorestudio:
New theme is MUCH needed. The sooner - the better

alextech:
Its fine to do a refresh; new looks can be exciting, I used to look forward to visual improvements. Just don't make it mostly white space, too flat with sharp corners that I am afraid I'll cut myself if get too close to monitor, or icons that do not represent anything.

Eg. DbSchema looks nice https://www.colormango.com/development/screenshot/dbschema_112104.png
[...]

PR #15494, opened 2019-10-07
Remove Original theme

MauricioFauth:
Do not bundle the Original theme with phpMyAdmin, making it an optional
theme.


williamdes (approved):
pmahomme is better ^^

ibennetch:
I would personally prefer to continue to distribute both themes as part of the download. Is there a strong reason you have to wish for the removal?

mauriciofauth:
I opened this PR to start a discussion about theme removal.

A discussion started a while ago in #12373 about "theme selection for future releases". See:

> I think it's time to reconsider which themes are shipped with phpMyAdmin. The Original theme is really showing it's age also it might be good idea to include Metro which is IMHO way nicer.

The [decision](https://github.com/phpmyadmin/phpmyadmin/wiki/2017-04_Meeting) later was to include the Metro theme, but the Original theme removal was not really discussed.

Pmahomme has been the default theme since version 3.4 if I'm not mistaken.

PR #15509, opened 2019-10-17
Use gender-neutral pronouns

MauricioFauth:
Replaces gender-specific pronouns with gender-neutral pronouns.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -5,6 +5,7 @@ phpMyAdmin - ChangeLog
 - issue #15350 Change Media (MIME) type references to Media type
 - issue #15377 Add a request router
 - issue        Automatically focus input in the two-factor authentication window
+- issue #15509 Replace gender-specific pronouns with gender-neutral pronouns
 
 5.0.0 (not yet released)
 - issue #13896 Drop support for PHP 5.5, PHP 5.6, PHP 7.0 and HHVM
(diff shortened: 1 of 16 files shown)


libraries/classes/Operations.php
```diff
@@ -1968,7 +1968,7 @@ public function getWarningMessagesArray()
         foreach ($this->dbi->getWarnings() as $warning) {
             // In MariaDB 5.1.44, when altering a table from Maria to MyISAM
             // and if TRANSACTIONAL was set, the system reports an error;
-            // I discussed with a Maria developer and he agrees that this
```
ibennetch: I think this one should be left as-is

ibennetch (approved):
This seems to be appropriate modern usage, I approve.

libraries/classes/Operations.php
williamdes: You can remove "and he" and add "who"?

williamdes (approved):

libraries/classes/Operations.php
ibennetch: That would be acceptable, but unless this specific developer has stopped using the "he" pronoun, I feel this would be an unnecessary change.

mauriciofauth: Thank you!

PR #15524, opened 2019-10-25
Translated using Weblate (Croatian)

zeenayouhan:
Translated using Weblate (Croatian)


williamdes:
I closed this one since it is a mistake

PR #15535, opened 2019-10-27
fix #15491 Generate password complexity setting

### LINKED ISSUE #15491: Generate password complexity setting
Hello,

I always encounter problem when mysql server is set to medium password level it requires special character in passwords when creating user.

When you use generate password function it doesn't have any of special chars so you always get an error from mysql for incorrect password.

Solution was simple: add special chars to function suggestPassword in functions.js file, no commas, no dots, no quotes, just these special chars !@#$%^&*()

How about making a setting in phpmyadmin to let people choose password complexity without having to modify js files? :)

Tithugues:
Add symbols to possible characters for suggested passwords to have a chance to match the default value of validate_password_special_char_count when 	validate_password_policy is set to MEDIUM or higher.

Fixes: #15491

### DIFF
--- a/js/functions.js
+++ b/js/functions.js
@@ -483,7 +483,7 @@ Functions.suggestPassword = function (passwordForm) {
     // restrict the password to just letters and numbers to avoid problems:
     // "editors and viewers regard the password as multiple words and
     // things like double click no longer work"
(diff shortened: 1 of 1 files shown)


MauricioFauth (approved):

mauriciofauth:

PR #15546, opened 2019-11-01
Fix operators precedence possible issue

Tithugues:
As "==" has priority over "&", it was firstly checking the equality… which were always true…

To be honnest, I didn't test it, as I don't really know how to test it… I can only guess what was expected to be done.


MauricioFauth (approved):

mauriciofauth:
I agree with this change, however I think it's better to include this in `QA_4_9` branch, instead of master.

Tithugues:
> I agree with this change, however I think it's better to include this in `QA_4_9` branch, instead of master.

Done!

williamdes (approved):

MauricioFauth (approved):

mauriciofauth:

PR #15554, opened 2019-11-02
Fix #15316 cross join clause is removed on export

### LINKED ISSUE #15316: CROSS JOIN clause is removed on export
### Describe the bug

CROSS JOIN clause is removed from the query when trying to export query results to CSV, CSV for MS Excel or OpenDocument Spreadsheet (I haven't tested the other formats). Consequently query fails.

### To Reproduce

1. Create a table:
```sql
CREATE TABLE `crossjoin_test` (
  `id` INT UNSIGNED NOT NULL AUTO_INCREMENT,
  PRIMARY KEY (`id`)
);
```

2. Insert some rows:
```sql
INSERT INTO `crossjoin_test` (`id`) VALUES (NULL), (NULL), (NULL);
```

3. Run query with CROSS JOIN:
```sql
SELECT * FROM `crossjoin_test` AS `t1` CROSS JOIN `crossjoin_test` AS `t2` WHERE `t1`.`id` > 1 AND `t2`.`id` > 1;
```

4. Export query results in CSV, CSV for MS Excel or OpenDocument Spreadsheet format.

5. Exported file (in any of the mentioned formats) contains HTML code for the error. Error mentions that the query is:
```sql
SELECT * FROM `crossjoin_test` WHERE crossjoin_test.`id` > 1 AND `t2`.`id` > 1
```

### Expected behavior

Query should not be rewritten and query results should be normally exported.

### Server configuration

- Operating system: Windows
- Web server: Apache 2.4.39
- Database version: MySQL 8.0.16
- PHP version: 7.3.6
- phpMyAdmin version: 4.9.0.1 (also tested on demo server with master 9102b98)

### Additional context

If CROSS JOIN is replaced with INNER JOIN, export works as expected.

Tithugues:
Fixes: #15316

### DIFF
--- a/tbl_export.php
+++ b/tbl_export.php
@@ -53,33 +53,6 @@ if (! empty($sql_query)) {
     if ((!empty($parser->statements[0]))
         && ($parser->statements[0] instanceof PhpMyAdmin\SqlParser\Statements\SelectStatement)
     ) {
-
-        // Finding aliases and removing them, but we keep track of them to be
-        // able to replace them in select expression too.
-        $aliases = array();
-        foreach ($parser->statements[0]->from as $from) {
-            if ((!empty($from->table)) && (!empty($from->alias))) {
-                $aliases[$from->alias] = $from->table;
-                // We remove the alias of the table because they are going to
-                // be replaced anyway.
-                $from->alias = null;
-                $from->expr = null; // Force rebuild.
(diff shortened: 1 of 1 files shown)


williamdes:
Merged, thank you @Tithugues for your contribution !

Issue #15578, opened 2019-11-20
Selecting a range of tables leaves one table unselected


musicpanda:
When I choose custom mode while exporting tables I can select ranges by selecting the first table and then selecting the last one while keeping the shift key pressed. This is a very useful feature that I use a lot.

However, this feature has one little bug: the one before last table is not selected. The attached picture shows what happened after I selected a table deep down and then selected the ps_access table while keeping the shift pressed. As you see the one but last table (ps_accessory) is not selected.

![range](https://user-images.githubusercontent.com/13930107/69248755-24086880-0bad-11ea-99fd-0129d4cea3bb.jpg)

saulotoledo:
I confirm the issue. I will submit a fix until this weekend.

williamdes:
Duplicate of #15327

Issue #15327, opened 2019-06-12
Shift-click in Export misses checkbox


svivian:
### Describe the bug

Using Shift + click on the checkboxes in Export does not select the correct tables.

### To Reproduce

1. Go to Export
2. Click on Custom
3. Under Tables, deselect all tables using the two checkboxes at the top.
4. Click a checkbox for a table (left hand column).
5. Hold Shift and click another checkbox further down.

### Expected behavior

All tables between the first and second boxes (inclusive) to be checked.

But it misses the second box. For example you click the 2nd then Shift-click 7th, boxes 2, 4, 5, 6 and 7 are checked but not 3.

### Server configuration

- Operating system: Mac
- Web server: both Mac localhost and Windows server
- Database version: 5.7.24
- PHP version: 7.2
- phpMyAdmin version: 4.9.0.1

### Client configuration

- Browser: Chrome
- Operating system: macOS 10.14.5

falconfx:
I am still trying to validate the PR so that it passes all the tests, but here I have made a possible approach to the solution #15343

ibennetch:
This issue was fixed with the release of phpMyAdmin 4.9.2. Thank you for reporting this issue.

PR #15343, opened 2019-06-18
Shift-click-in-Export-misses-checkbox fixed

falconfx:
### Description

Fixed bug for #15327 issue

Before submitting pull request, please review the following checklist:

- [x] Make sure you have read our [CONTRIBUTING.md](https://github.com/phpmyadmin/phpmyadmin/blob/master/CONTRIBUTING.md) document.
- [x] Make sure you are making a pull request against the correct branch. For example, for bug fixes in a released version use the corresponding QA branch and for new features use the _master_ branch. If you have a doubt, you can ask as a comment in the bug report or on the mailing list.
- [x] Every commit has proper `Signed-off-by` line as described in our [DCO](https://github.com/phpmyadmin/phpmyadmin/blob/master/DCO). This ensures that the work you're submitting is your own creation.
- [x] Every commit has a descriptive commit message.
- [x] Every commit is needed on its own, if you have just minor fixes to previous commits, you can squash them.
- [x] Any new functionality is covered by tests.

PR #15579, opened 2019-11-20
Fix table select range with Shift key

saulotoledo:
Fixes issue when trying to select a range of items in a table by using the shift key.
The problem can be visualised in the custom mode while exporting tables, for example. Before the patch, clicking in the first line and then selecting the last one while pressing the shift key leads to one of the lines being unselected. There were also other unexpected situations while trying to select items in the inverse order.

Fixes https://github.com/phpmyadmin/phpmyadmin/issues/15578
Fixes https://github.com/phpmyadmin/phpmyadmin/issues/15327

### DIFF
--- a/js/functions.js
+++ b/js/functions.js
@@ -1046,7 +1046,7 @@ AJAX.registerOnload('functions.js', function () {
             last_click_checked = checked;
 
             // remember the last clicked row
-            last_clicked_row = last_click_checked ? $table.find('tr:not(.noclick)').index($tr) : -1;
+            last_clicked_row = last_click_checked ? $table.find('tbody tr:not(.noclick)').index($tr) : -1;
             last_shift_clicked_row = -1;
         } else {
             // handle the shift click
@@ -1072,7 +1072,7 @@ AJAX.registerOnload('functions.js', function () {
             }
 
             // handle new shift click
-            var curr_row = $table.find('tr:not(.noclick)').index($tr);
+            var curr_row = $table.find('tbody tr:not(.noclick)').index($tr);
             if (curr_row >= last_clicked_row) {
                 start = last_clicked_row;
                 end = curr_row;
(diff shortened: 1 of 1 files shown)


saulotoledo:
@williamdes For you to review. :)

williamdes:
Duplicate of #15343

williamdes (approved):

williamdes:
@saulotoledo I like your approach, can you explain why it is better than https://github.com/phpmyadmin/phpmyadmin/pull/15343/files ?

If you feel you can do it, hard reset on QA_4_9 then cherry-pick the commit you made and force push
Finally change the target with edit button on this page

If not, I will do it ;)

saulotoledo:
Hi @williamdes. Thanks for your return!
I did not notice that there was another PR already, but I still prefer my approach. The reasons:

- [Line 1084](https://github.com/phpmyadmin/phpmyadmin/blob/QA_4_9/js/functions.js#L1084) is really missing a  `+ 1`. One line is left behind, so it is a bug both PRs try to fix;

- Considering the other lines, the main problem is that `$table.find('tr...` return all the `<tr>` from the table, including those in the header. On the other hand, `$tr.parent()` returns the `<tbody>` tag. Thus, the results are unexpected when trying to subtract one set from another. The quick fix was forcing the calculations to be done with the `<tr>` tags inside `<tbody>` only. The other PR adds more logic to fix the problem, but that was not really necessary once the calculations are done within the same set of table rows.

I believe the original code worked in the past, but later changes added the `<tbody>` wrapping the table body and possibly a second row inside the header in the HTML only, leading to such wrong calculations.

I did the requested changes in the PR. Let me know if you need any new changes! :D

williamdes:
@sudo-bot deploy PR

williamdes:
Thank you @saulotoledo !
It works perfectly :100:

ibennetch:
Thanks, this was merged and fixed with reload 4.9.2.

PR #15596, opened 2019-11-26
Implement absolute URL for links

MauricioFauth:
Currently phpMyAdmin uses relative URLs for links. This pull request changes the relative URLs with absolute URLs.

For example:
```diff
- <a href="index.php?route=/server/databases">
+ <a href="http://localhost/index.php?route=/server/databases">

- <script src="js/server/databases.js"></script>
+ <script src="http://localhost:8000/js/server/databases.js"></script>
```

The way phpMyAdmin works, using relative URLs is not an issue, but it doesn't work when using URL rewriting, because the base of the URL is always changing.

So using absolute URLs is a necessary step to add support for URL rewriting.

### DIFF
--- a/libraries/classes/Twig/UrlExtension.php
+++ b/libraries/classes/Twig/UrlExtension.php
@@ -27,6 +27,10 @@ class UrlExtension extends AbstractExtension
     {
         return [
             new TwigFunction(
+                'base_url',
+                'PhpMyAdmin\Url::getBaseUrl'
+            ),
+            new TwigFunction(
                 'get_hidden_inputs',
                 'PhpMyAdmin\Url::getHiddenInputs',
                 ['is_safe' => ['html']]
--- a/libraries/classes/Url.php
+++ b/libraries/classes/Url.php
@@ -278,6 +278,32 @@ class Url
      */
     public static function getFromRoute(string $route, array $additionalParameters = []): string
     {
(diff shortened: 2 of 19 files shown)


ibennetch:
I'm nervous about this because in the past we have had a lot of difficulty with automatically detecting the URL. Users have installations behind traditional proxy servers, https/ssl appliances, various caches, and other devices that make it difficult to detect.

If we implement this, we should test it extensively and also consider a configuration directive to override the automatic value.

mauriciofauth:
I agree. I'm very cautious about this, because if done wrong it could break many things.
Looks like is common to have a configuration directive to replace the automatic value.
I'm looking for the best practices for this as well.

Tithugues:
I agree with @ibennetch .
For instance, in the current PR, you rely on `$_SERVER['HTTPS']` to define the protocol. Just to give an example: you may have a proxy requiring usage of `https` from outside, but that will then forward the requests in `http` once we are inside the company network. In this case, `$_SERVER['HTTPS']` wouldn't be reliable.

So, to know the protocol, you may rely on [X-Forwarded-Proto](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto).
You can see the way Symfony implemented it looking at methods called by [getScheme](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/Request.php#L894).

Not so easy… And this is only for the protocol…

Issue #15623, opened 2019-12-06
Range search issue


OlafvdSpek:
Select BETWEEN for rental_id operator, a dialog pops up.

1. If you Cancel the dialog, there's no simple way to re-open it.
2. The dialog has date pickers for int fields.
3. The dialog has unnecessary empty space.
4. The fields aren't big enough to show a date fully.

https://demo.phpmyadmin.net/master-config/index.php?route=/table/search&db=sakila&table=rental

![pma](https://user-images.githubusercontent.com/2218936/70329148-fa04a680-183a-11ea-8271-65f1998c88fc.PNG)

PR #15631, opened 2019-12-07
Display date/time picker only for date/time fields

Tithugues:


williamdes (approved):
Do you have an issue number or a commit hash that introduced the need for such a fix ?

Tithugues:
Sorry, I forgot to mantion it…
It is related to https://github.com/phpmyadmin/phpmyadmin/issues/15623 , second point.

PR #15666, opened 2019-12-17
Renders absolutely nothing when header is not enabled.

t123yh:
# Description
When downloading blob data from database, the header template will be rendered as `\n` which adds an extra 0x0A at the beginning of the blob data. This fixes the problem.

### DIFF
--- a/libraries/classes/Header.php
+++ b/libraries/classes/Header.php
@@ -402,8 +402,8 @@ class Header
      */
     public function getDisplay(): string
     {
-        if (! $this->_headerIsSent) {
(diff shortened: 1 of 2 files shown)


williamdes:
@t123yh thank you for the pull-request, I need to reproduce it and check the fix

williamdes:
Thank you @t123yh for your contribution !

I had to do some changes to your repo since you used master branch and not QA_5_0 as a base
If you want to contribute more I would recommend you do delete and re-fork the repository

Issue #15584, opened 2019-11-23
Message SSL is not being used even if server is not remote


aommundsen:
### Describe the bug
We are running both phpMyAdmin and MariaDB on the same server as the websites. Meaning that MariaDB is not remote, the database connections is done by using "localhost". But still phpMyAdmin display this message in the right column: "Server connection: SSL is not being used"

In phpMyAdmin 4.9.2 the message is in black text color, but in phpMyAdmin 5.1.0-dev it is displayed with red color. It is not fun when customers on shared hosting servers create support tickets and ask about this!

### Expected behavior
When the database server is not remote, the message should not be displayed at all in phpMyAdmin.

### Server configuration
- Operating system: CentOS 7 and 8
- Web server: Apache 2.4.41
- Database version: MariaDB 10.3.20
- PHP version: 7.2.25
- phpMyAdmin version: 4.9.2 and 5.1.0-dev

### Additional context
Please consider to do one of the following so solve this:
1: Remove the message about SSL not being used if the database server is local.
or
2: Give us a configuration setting that we can put in config.inc.php to disable the message, for example something like this:
$cfg['SSLNotBeingUsed_DisableWarning'] = true;

thk12:
I usually patch this out of index.php because it's really anoying for customers. This warning shouldn't even be displayed when PMA connects through a typical private IP range (like using VPN).

williamdes:
Okay, a configuration override must be added

Jaideep-97:
I pushed 2 commits into my forked repo but failed to add signoff as it was my first contribution and hence unable to open PR for this issue. How do I change that?

williamdes:
You should create a branch on your fork and cherry-pick the changes
After cherry-picking you can do `git commit --amend --no-edit --signoff` and then push the commit

Jaideep-97:
Opened a PR for this issue. (#15673 )

aommundsen:
As phpMyAdmin 4.9.3 and 5.0.0 was released today, but this issue did not make it into those versions when I check the demo page at https://www.phpmyadmin.net/try/

That is really to bad. Because now I don't want to upgrade because in both 4.9.3 and 5.0.0 the text "SSL is not being used" is with RED TEXT COLOR. That is a step bakcward to previous versions wich was on black text color.

I would have hoped you at least could revert the introduction of red text color on that message until the fix for the issue is implemented. But that seems to late now.

[...]

PR #15673, opened 2019-12-19
SSL message removed when server is local. (#15584)

Jaideep-97:
Signed-off-by: JAIDEEP DATTA <34120588+Jaideep-97@users.noreply.github.com>

### Description

SSL message removed when server is local.

Fixes #

15584


templates/home/index.twig
```diff
@@ -115,10 +115,13 @@
                 {% trans 'Server type:' %}
                 {{ database_server.type }}
               </li>
-              <li id="li_server_connection">
-                {% trans 'Server connection:' %}
-                {{ database_server.connection|raw }}
-              </li>
+
+          {%  if(database_server.host !="Localhost via UNIX socket") %}
```
ibennetch: I think this indentation looks to be not ideal; I think you should line up the `{%` in line with the `</li`> above it, then one space before `if` (the newline is fine, though).

@mauriciofauth what do you think?

ibennetch: I don't think this test will work due to translations. For example, in German this message is shown as "Server-Verbindung: SSL wird nicht verwendet"

ibennetch: I wonder what would happen with Windows using a named pipe. I don't have a Windows MySQL installation to check what message is shown, but we could hide the message in the case of a named pipe as well as UNIX socket.

templates/sql/query.twig
```diff
@@ -5,13 +5,13 @@
...
-  <input type="hidden" name="prev_sql_query" value="{{ query }}">
+  <input type="hidden" name="prev_sql_query"  value="{{ query }}">
 
   {% if display_tab == 'full' or display_tab == 'sql' %}
-    <a id="querybox"></a>
+    <a id="querybox" ></a>
     <div id="queryboxcontainer">
       <fieldset id="queryboxf">
-        <legend>{{ legend|raw }}</legend>
+        <legend style="font-size:0.7em;">{{ legend|raw }}</legend>
```
ibennetch: Looks like you've made a few unrelated changes to this file as well, please revert those.

ibennetch (changes requested):
Overall I think this is a good idea.

Your pull request has a lot of extra files that aren't needed as part of this submission, can you remove the files composer-setup.php, composer.phar, and yarn-error.log?

williamdes:
The submitter of the issue now agrees that the text should be black https://github.com/phpmyadmin/phpmyadmin/issues/15584#issuecomment-569140761

Thank you for the contribution.

PR #15674, opened 2019-12-19
Replace twig/extensions with phpmyadmin/twig-i18n-extension

### LINKED ISSUE #15671: Twig extensions package is abandoned
phpMyAdmin depends on the [twig/extensions](https://github.com/twigphp/Twig-extensions) package for the `trans` tag/filter in templates. But the package is now abandoned.

> WARNING: This repository is abandoned in favor of Twig Core Extra extensions.

It's suggested to use `trans` filter from the [Symfony Translator](https://symfony.com/doc/current/translation.html) package.

> I18nExtension: `trans` filter -> use the [trans](https://symfony.com/doc/current/reference/twig_reference.html#trans) filter from the Symfony Translator extension (symfony/twig-bridge)

Current `I18nExtension` is not enough for us. It doesn't have support for custom gettext functions and doesn't have support for gettext context. We extended the `I18nExtension` to support that, but it have some bugs.

We can fork the Twig's I18nExtension, but that will increase our maintenance burden.

Using Symfony Translator is also an option, but I don't know how difficult will be to implement, and how well it will work with [motranslator](https://github.com/phpmyadmin/motranslator), if not completely replace it.

MauricioFauth:
Fixes #15671.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -13,6 +13,7 @@ phpMyAdmin - ChangeLog
 - issue        Fixed wrong jQuery function call in table search page
 - issue #15761 Fix uncaught TypeError when using "$cfg['ServerDefault'] = 0;"
 - issue #15780 Fixed unexpected UI of action links (text only mode)
+- issue #15674 Replace twig/extensions with phpmyadmin/twig-i18n-extension
 
 5.0.1 (2020-01-07)
 - issue #15719 Fixed error 500 when browsing a table when $cfg['LimitChars'] used a string and not an int value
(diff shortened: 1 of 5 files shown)


williamdes (approved):

ibennetch:
👍

Should we package this on Packagist so it's available as a Composer package instead of needing to point to the vcs repository?

mauriciofauth:
> Should we package this on Packagist so it's available as a Composer package instead of needing to point to the vcs repository?

@ibennetch Yes. I added the VCS repository to be able to test the changes, so when the package is published, I'll remove the repository line. But I can't submit to Packagist myself. You need to do this.

williamdes (approved):

williamdes:
+ will you add a changelog entry ?

 I hope `#15791` will also be part of the release

Issue #14433, opened 2018-06-27
"You do not have privileges to manipulate with the users!" on root superadmin


freedmandil:
### Steps to reproduce
Created a local server of MariaDB via home-brew
Installed latest version of phpmyadmin
logged in as root with password
pma control user set up

No functionality to add users

### Expected behaviour
Should see "add user account" button

### Actual behaviour
No button and error:
"You do not have privileges to manipulate with the users!"

### Server configuration
**Operating system**:
MacOS X High Sierra 10.13.5 (17F77)

**Web server:**
Apache/2.4.33 (Unix) OpenSSL/1.0.2o PHP/7.1.18
Database client version: libmysql - mysqlnd 5.0.12-dev - 20150407 - $Id: 38fea24f2847fa7519001be390c98ae0acafe387 $
PHP extension: mysqliDocumentation curlDocumentation mbstringDocumentation
PHP version: 7.1.18

**Database:**
Server: 127.0.0.1 via TCP/IP
Server type: MariaDB
Server connection: SSL is not being used Documentation
Server version: 10.3.7-MariaDB - Homebrew
Protocol version: 10
User: root@localhost
Server charset: UTF-8 Unicode (utf8)
**PHP version:**

**phpMyAdmin version:**
Version information: 4.8.2 (up to date)

### Client configuration
**Browser:**
Safari: Version 11.1.1 (13605.2.8)
Chrome: Version 67.0.3396.99 (Official Build) (64-bit)

**Operating system:**
MacOS X High Sierra 10.13.5 (17F77)

ibennetch:
Could you please post here the output of this SQL command (which you can run from the "SQL" tab anywhere in phpMyAdmin)?

`SHOW GRANTS;`

freedmandil:
Grants for root@localhost |  
-- | --
GRANT ALL PRIVILEGES ON *.* TO 'root'@'localhost' IDENTIFIED BY PASSWORD '*' WITH GRANT OPTION
GRANT PROXY ON ''@'%' TO 'root'@'localhost' WITH GRANT OPTION

sabl0r:
I have the same issue running MySQL 5.6.39.
The issue is gone after downgrading PMA from 4.8.1 to 4.7.8.
Note: I did not try any pma releases in between.

williamdes:
I can not reproduce this issue, post output of (ADJUST the `'''root''@''%'''` part):
```sql
[...]

PR #15681, opened 2019-12-21
Fix control user issue

### LINKED ISSUE #15608: DisableIS is broken on pma 4.9.2 with controluser configured
### Describe the bug

There is a problem when `DisableIS` cfg option is true, and there is a `controluser` configured.

After logging in as root and clicking on a database name on the navigation tree (not to the "+", directly to the name), it pops up the error "An error has occurred while loading the navigation display". At that point, refreshing the page causes all databases except information_schema, mysql, and performance_schema to disappear.

I tried further investigating the issue, it is probably hiding a more serious bug related to the management of links for control user and logged in user. In particular, the problem started to appear from commit b8650115be7cc2d79d231723b7a6f1961e834d27 (which was buggy but it doesn't matter for this issue), the relevant change is the behaviour of instance member _current_user, which is no longer cached on error. This means that while before it was always returning ["",""], now it starts returning ["root","%"] when clicking on the navigation tree, which in turns triggers the bug.

I'm afraid there is something wrong with the whole `DatabaseInterface` class, it seems it is mixing up the control user link and the logged in user.

Please let me know if you need further information as I spent several hours troubleshooting this.

### Server configuration

- Operating system: Fedora 31
- Web server: Apache/2.4.41 (Fedora) OpenSSL/1.1.1d PHP/7.3.12
- Database version: 10.3.18-MariaDB-log - MariaDB Server
- PHP version: 7.3.12
- phpMyAdmin version: 4.9.2

MauricioFauth:
Partially reverts commit 2bc8afea5f47d0d409163754698ad13487410158.

Fixes #15608
Related to #14433

### DIFF
--- a/libraries/classes/DatabaseInterface.php
+++ b/libraries/classes/DatabaseInterface.php
@@ -2513,8 +2513,6 @@ class DatabaseInterface
             /* Run post connect for user connections */
             if ($target == DatabaseInterface::CONNECT_USER) {
                 $this->postConnect();
-            } elseif ($target == DatabaseInterface::CONNECT_CONTROL) {
-                $this->postConnectControl();
             }
             return $result;
         }
--- a/libraries/classes/DatabaseInterface.php
+++ b/libraries/classes/DatabaseInterface.php
@@ -2330,11 +2330,7 @@ class DatabaseInterface
     {
         if (count($this->_current_user) == 0) {
             $user = $this->getCurrentUser();
-            if ($user === '@') {// Request did not succeed, please do not cache
-                return ['', ''];
-            } else {
-                $this->_current_user = explode("@", $user);
-            }
+            $this->_current_user = explode("@", $user);
         }
         return $this->_current_user;
     }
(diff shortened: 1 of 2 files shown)


libraries/classes/DatabaseInterface.php
```diff
@@ -2330,11 +2330,7 @@ public function getCurrentUserAndHost()
     {
         if (count($this->_current_user) == 0) {
             $user = $this->getCurrentUser();
-            if ($user === '@') {// Request did not succeed, please do not cache
-                return ['', ''];
```
williamdes: A lot of code relies on the access of the 2 indexes, nothing broken ?

williamdes (approved):

libraries/classes/DatabaseInterface.php
mauriciofauth: Exploding `'@'` returns the same array: `explode('@', '@') == ['', '']`

williamdes:
:rocket: thank you !

williamdes:
Introduced #15724

PR #15701, opened 2019-12-26
(Issue #15491) Generate password complexity setting

rabeehrz:
### Description

Fixes issue #15491 

### Fixes

- [ ] Retrieve and share MySQL password policies with JS.
- [x] Use the variables to generate the password.


rabeehrz:
Need some help with retrieving the MySQL password policies from the database. 

Also, Need clarification on dictionary validation. Are they required?

Issue #15606, opened 2019-12-01
Documentation for creating controluser is outdated


ibennetch:
The documentation suggests creating a controluser with this SQL:

```GRANT SELECT, INSERT, UPDATE, DELETE ON <pma_db>.* TO 'pma'@'localhost'  IDENTIFIED BY 'pmapass';```

However, that doesn't work on modern MySQL. Probably it needs `CREATE USER '<pma>'@'localhost' IDENTIFIED WITH ....

The documentation should probably reference which version of the command applies to particular MySQL/MariaDB versions.

AbdullahiAbdulkabir:
Hello, how can I contribute to the documentation. I am kinda confused about where to send a pull request

ibennetch:
This one is actually not a good issue as I’ve started working on it
yesterday.

In general, you would use the GitHub fork feature to mirror our repository
in your workspace, push your commits there, then use the GitHub interface
to create a pull request. The GitHub help is quite useful for the details
of each step, I suggest you start there and let us know if you have
specific problems.

ibennetch:
This should have automatically closed with #15748: 

> <img width="235" alt="image" src="https://user-images.githubusercontent.com/1443710/72206526-1ceb3180-345d-11ea-81f4-e300e436a827.png">

Closing manually.

PR #15748, opened 2020-01-05
Update documentation for creating a controluser

ibennetch:
This now provides information for newer MySQL (newer than 8.0) and MariaDB servers.

Fixes: #15606

### DIFF
--- a/doc/setup.rst
+++ b/doc/setup.rst
@@ -788,9 +788,30 @@ You will also need to have a controluser
 with the proper rights to those tables. For example you can create it
 using following statement:
 
+And for any MariaDB version:
+
 .. code-block:: mysql
 
-   GRANT SELECT, INSERT, UPDATE, DELETE ON <pma_db>.* TO 'pma'@'localhost'  IDENTIFIED BY 'pmapass';
+   CREATE USER 'pma'@'localhost' IDENTIFIED VIA mysql_native_password USING 'pmapass';
+   GRANT SELECT, INSERT, UPDATE, DELETE ON `<pma_db>`.* TO 'pma'@'localhost';
+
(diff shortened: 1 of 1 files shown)


ibennetch:
I think I've gotten all of the scenarios correct here, but let me know if I missed some variation.

doc/setup.rst
```diff
@@ -788,9 +788,26 @@ You will also need to have a controluser
 with the proper rights to those tables. For example you can create it
 using following statement:
 
+For MySQL 8.0 and newer:
+
+.. code-block:: mysql
```
williamdes: Can you mention the php version restriction?

williamdes (approved):

doc/setup.rst
ibennetch: I'm not sure that PHP versions are relevant to creating the controluser. Besides, we already mention the PHP versions in the Requirements section, don't we?

williamdes: I would add a line so users can understand why it does not work when they use sha2 method

sha2 method can be used from a non sha2 user, the user does not understand that the control user will not be able to connect because the method is not supported

Maybe my sentence is not clear enough, I do not know how to explain well the issue

ibennetch:
Oh yes, now it is quite clear.

doc/setup.rst
mauriciofauth: Isn't it better to have that information on a FAQ? And a link to the FAQ instead of repeating that info in various places.

ibennetch: Proposed solution for the version problem is #15750.

williamdes (approved):
Thank you

ibennetch:
Thanks for the suggestions!

PR #15759, opened 2020-01-07
Remove replication.inc.php file

MauricioFauth:
Extracts the `replication.inc.php` code to the `ReplicationInfo` class and removes the file.


williamdes (approved):
:100:

PR #15779, opened 2020-01-12
Informing the user about the HEX conversion scheme of Integers

### LINKED ISSUE #15136: Hex converter giving invalid results

thesmallstar:
Signed-off-by: Manthan Surkar <manthan.surkar@gmail.com>

### Description

A note is shown to the user when he tries to apply HEX function on integer.
Cases, like adding multiple integer notes and removing them, is also taken care of. 

image

Fixes #15136

### DIFF
--- a/js/messages.php
+++ b/js/messages.php
@@ -504,6 +504,8 @@ $javascriptMessages = [
     'strYes' => __('Yes'),
     'strCopyEncryptionKey' => __('Do you want to copy encryption key?'),
     'strEncryptionKey' => __('Encryption key'),
+    /* l10n: Tip for HEX conversion of Integers */
+    'HexConversionInfo' => __('The HEX function will treat the integer as a string while calculating the hexadecimal value'),
 
     /* For Tip to be shown on Time field */
     'strMysqlAllowedValuesTipTime' => __(
--- a/js/table/change.js
+++ b/js/table/change.js
@@ -279,6 +279,18 @@ function verificationsAfterFieldChange (urlField, multiEdit, theType) {
         });
     }
 
+    if (target.value === 'HEX' && theType.substring(0,3) === 'int') {
+        // Add note when HEX function is selected on a int
+        var newHexInfo = '<br><p id="note' +  target.id + '">' + Messages.HexConversionInfo + '</p>';
+        if (!$('#note' + target.id).length) {
+            $thisInput.after(newHexInfo);
+        }
+    } else {
+        $('#note' + target.id)
+            .prev('br')
+            .remove();
+        $('#note' + target.id).remove();
+    }
     // Unchecks the corresponding "NULL" control
     $('input[name=\'fields_null[multi_edit][' + multiEdit + '][' + urlField + ']\']').prop('checked', false);
 


js/table/change.js
```diff
@@ -196,7 +196,26 @@ function verificationsAfterFieldChange (urlField, multiEdit, theType) {
...
     }
-
+    //Add note when HEX function is selected on a int
+//Add note when HEX function is selected on a int
+    var newHexInfo =
+     "<br><p id=note" +
+        target.id +
+       ">" +
+      "HEX function will treat the integer as String while calculating hexadecimal value" +
+      "</p>";
```
williamdes: Can you add the string to `js/messages.php` and use it here using `Messages.`

williamdes (changes requested):
I am not sure if adding html that way is the best way to do, but I also do not have a better solution today

js/table/change.js
```diff
@@ -197,6 +197,25 @@ function verificationsAfterFieldChange (urlField, multiEdit, theType) {
         });
     }
 
+    //Add note when HEX function is selected on a int
+    var newHexInfo =
```
williamdes: can you make it oneline

js/messages.php
```diff
@@ -541,6 +541,7 @@ function () {
 $js_messages['strYes'] = __('Yes');
 $js_messages['strCopyEncryptionKey'] = __('Do you want to copy encryption key?');
 $js_messages['strEncryptionKey'] = __('Encryption key');
+$js_messages['HexConversionInfo'] = __('HEX function will treat the integer as String while calculating hexadecimal value');
```
williamdes: `String` should not have a capital letter ?
```suggestion
/* l10n: Please explain here to translators what they are translating 
$js_messages['HexConversionInfo'] = __('HEX function will treat the integer as string while calculating hexadecimal value');
```

thesmallstar:
@williamdes, Since this is not an error we cannot put this string/note in any error code section.
The following way is referred from the same file, selecting AES_ENCRYPT adds HTML to provide similar functionality. 

I also couldn't find a better way :P, Hope this works.

williamdes (changes requested):
> I also couldn't find a better way :P, Hope this works.
> The following way is referred from the same file

Okay then :p

You are on the right path ;)
Be sure to run `yarn run js-lint`/`yarn run js-lint --fix`

thesmallstar:
@williamdes Thank you so much for the suggestions, helps us improve :)

js/messages.php
```diff
@@ -541,6 +541,8 @@ function () {
 $js_messages['strYes'] = __('Yes');
 $js_messages['strCopyEncryptionKey'] = __('Do you want to copy encryption key?');
 $js_messages['strEncryptionKey'] = __('Encryption key');
+/*Tip for HEX conversion of Integers*/
```
williamdes:
```suggestion
/* l10n: Tip for HEX conversion of Integers*/
```

williamdes (approved):

js/messages.php
williamdes: @thesmallstar If you remove the comment prefix `gettext` will not extract it

thesmallstar: on it :D

williamdes:
Looks nice, I would like a team review or point of view before merging

kartik1000:
@thesmallstar You can add the notice in <strong> tags to highlight it. Also can't we add a notice only once when the function is selected in more than one columns. Don't makes much sense to keep repeating it again and again. What do you feel @williamdes? Rest everything looks good 😃

williamdes:
What about adding a validator the adds the message if the input is not considered correct or risky ?

Like the error message when typing text in an int

kartik1000:
Yess @williamdes that would not confuse a new user in case the output is correct and we are still giving some notice.  That would be the best.

williamdes:
I think you will need to use master branch because it seems more like an enhancement to me

thesmallstar:
@willamdes okay, I am still trying to find the best way to inform the user, I will let you know. @kartik1000 do you have more suggestions?

kartik1000:
@thesmallstar I think if you can make the change @williamdes has mentioned that would be the best and it won't require any more changes

MauricioFauth (approved):
Thanks for your contribution @thesmallstar.

LGTM.
I think this notice should be added in the _master_ branch.

williamdes:
Thank you @thesmallstar !

williamdes:
You will need to force push QA_5_0 to the right commit using #15942

PR #15793, opened 2020-01-15
Fix uncaught TypeError

### LINKED ISSUE #15761: php error using $cfg['ServerDefault'] = 0; on 5.0
after copying an working existing config.inc.php from phpMyAdmin 4.9.4 to the phpMyAdmin 5.0.1 directory phpMyAdmin 5.0.1 shows only a white page.
Running _php index.php_ shows some error - perhaps that may help.

Running phpMyAdmin 5.0.1 without config works - generating a config from /setup leads to the same error...
```
[/share/Web/phpMyAdmin5] # /usr/local/apache/bin/php7.3 -v
PHP 7.3.7 (cli) (built: Dec  6 2019 05:00:41) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.3.7, Copyright (c) 1998-2018 Zend Technologies

[/share/Web/phpMyAdmin5] # /usr/local/apache/bin/php7.3 index.php
Fatal error: Uncaught TypeError: chdir() expects parameter 1 to be a valid path, null given in libraries/classes/Response.php:506
Stack trace:
#0 libraries/classes/Response.php(506): chdir(NULL)
#1 [internal function]: PhpMyAdmin\Response->response()
#2 {main}
  thrown in libraries/classes/Response.php on line 506
```

MauricioFauth:
> chdir() expects parameter 1 to be a valid path, null given

Partially reverts c642c29aa4304084c0064a15b4be63d6f53f6bd3.

Sets `$dbi` even if `$cfg['Server']` is empty.

Fixes #15761

### DIFF
--- a/libraries/common.inc.php
+++ b/libraries/common.inc.php
@@ -325,13 +325,10 @@ if (! defined('PMA_MINIMUM_COMMON')) {
 
     ThemeManager::getInstance()->setThemeCookie();
 
-    if (! empty($cfg['Server'])) {
-        /**
-         * Loads the proper database interface for this server
-         */
-        $containerBuilder->set(DatabaseInterface::class, DatabaseInterface::load());
-        $containerBuilder->setAlias('dbi', DatabaseInterface::class);
+    $containerBuilder->set(DatabaseInterface::class, DatabaseInterface::load());
+    $containerBuilder->setAlias('dbi', DatabaseInterface::class);
 
+    if (! empty($cfg['Server'])) {
         // get LoginCookieValidity from preferences cache
         // no generic solution for loading preferences from cache as some settings
         // need to be kept for processing in
--- a/libraries/classes/Header.php
+++ b/libraries/classes/Header.php
@@ -115,11 +115,6 @@ class Header
     private $template;
 
     /**
-     * @var Navigation
-     */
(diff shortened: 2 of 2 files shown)


libraries/common.inc.php
```diff
@@ -325,13 +325,10 @@
...
     ThemeManager::getInstance()->setThemeCookie();
 
-    if (! empty($cfg['Server'])) {
-        /**
-         * Loads the proper database interface for this server
-         */
-        $containerBuilder->set(DatabaseInterface::class, DatabaseInterface::load());
-        $containerBuilder->setAlias('dbi', DatabaseInterface::class);
+    $containerBuilder->set(DatabaseInterface::class, DatabaseInterface::load());
+    $containerBuilder->setAlias('dbi', DatabaseInterface::class);
```
mauriciofauth: I don't know if this is a good idea or not. The container builder complains that the DatabaseInterface is not defined, because its used by a lot of classes.

williamdes: The template object seems to load to many extensions that need dbi

williamdes (approved):

williamdes:
Thank you so much for the fix @mauriciofauth !

PR #15801, opened 2020-01-16
Fixed deprecated jquery selectors (Obvious changes)

### LINKED ISSUE #15792: jQuery positional selectors are deprecated
jQuery 3.4.0 [deprecated](https://blog.jquery.com/2019/04/10/jquery-3-4-0-released/) the positional selectors `:first`, `:last`, `:eq`, `:even`, `:odd`, `:lt`, `:gt`, and `:nth`.

It's possible to know where we use jQuery positional selectors running the following command:
```
yarn run js-lint
```
That command will return something similar to this:
```
phpmyadmin/js/config.js
  613:9  warning  Positional selector extensions are not allowed  no-jquery/no-sizzle
  621:9  warning  Positional selector extensions are not allowed  no-jquery/no-sizzle
```

yashrajbothra:
### Description

(Commit: b93ae50)
1. `:first` has been replaced to `.first()`
Ref: https://api.jquery.com/first-selector/

2. `:last` has been replaced to `.last()`
Ref: https://api.jquery.com/last-selector/

(Commit : 9888c50)
1. `:eq` has been replaced with `.eq()`
Ref: https://api.jquery.com/eq-selector/

Fixes: #15792

### DIFF
--- a/js/config.js
+++ b/js/config.js
@@ -615,10 +615,10 @@ function setupConfigTabs () {
                 e.preventDefault();
                 setTab($(this).attr('href').substr(1));
             })
-            .filter(':first')
+            .first()
             .parent()
             .addClass('active');
-        $this.find('div.tabs_contents fieldset').hide().filter(':first').show();
+        $this.find('div.tabs_contents fieldset').hide().first().show();
     });
 }
 
--- a/js/database/central_columns.js
+++ b/js/database/central_columns.js
@@ -80,7 +80,7 @@ AJAX.registerOnload('database/central_columns.js', function () {
     });
     $('#add_new').find('td').each(function () {
         if ($(this).attr('name') !== 'undefined') {
-            $(this).find('input,select:first').attr('name', $(this).attr('name'));
+            $(this).find('input,select').first().attr('name', $(this).attr('name'));
         }
     });
     $('#field_0_0').attr('required','required');
@@ -138,7 +138,7 @@ AJAX.registerOnload('database/central_columns.js', function () {
         var rownum = $(this).data('rownum');
         $('#f_' + rownum + ' td').each(function () {
             if ($(this).attr('name') !== 'undefined') {
-                $(this).find(':input[type!="hidden"],select:first')
+                $(this).find(':input[type!="hidden"],select').first()
                     .attr('name', $(this).attr('name'));
             }
         });
@@ -192,7 +192,7 @@ AJAX.registerOnload('database/central_columns.js', function () {
     });
     $('#table-select').on('change', function () {
         var selectValue = $(this).val();
-        var defaultColumnSelect = $('#column-select').find('option:first');
+        var defaultColumnSelect = $('#column-select').find('option').first();
         var href = 'index.php?route=/database/central-columns';
         var params = {
             'ajax_request' : true,
--- a/js/database/structure.js
+++ b/js/database/structure.js
@@ -50,7 +50,7 @@ DatabaseStructure.adjustTotals = function () {
     /**
      * @var $allTr jQuery object that references all the rows in the list of tables
      */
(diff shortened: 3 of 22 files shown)


yashrajbothra:
@williamdes Should i mark it as Draft PR because I am working on some more commits related to this PR ?

williamdes:
@yashrajbothra Draft is better if work is in progress but no problem

mauriciofauth:
If it is a work in progress PR and is not ready to merge, you can mark it as draft.

williamdes (approved):

yashrajbothra:
@williamdes @mauriciofauth just after few seconds i realised there is no way to go from open to draft

ibennetch:
That's okay, we'll know from the comments here that it's not yet ready for
merge :)

MauricioFauth (approved):

yashrajbothra:
should i change vendor files because they also contain some deprecated code?

williamdes:
> should i change vendor files because they also contain some deprecated code?

Never change vendor files ;)
But in another PR if you want want to upgrade them we can discuss it

williamdes:
let us know when the PR is ready to merge

yashrajbothra:
And how do i test all these fixes which i am doing ?

williamdes:
> And how do i test all these fixes which i am doing ?

To be honest, it will be difficult
You need to read the code, find the feature and do after and before tests and it takes time

yashrajbothra:
Should i use click event handler to filter the elements as in this implementation we can't use any methods on selectors and `:eq` is deprecated instead `.eq()` is recommended

`$(document).on('click', 'li.trigger div:eq(1) a.ajax img,' +
    ' li.event div:eq(1) a.ajax img', function (event) {`

williamdes:
@yashrajbothra for this can can you replace by adding a css class so you can use it in the selector?
Maybe you can do that in another PR

yashrajbothra:
But `:eq` is now deprecated and i have to fix all deprecated selectors in this PR. So i think this PR is relatable to the change.

williamdes:
> But :eq is now deprecated and i have to fix all deprecated selectors in this PR. So i think this PR is relatable to the change.

I agree, but it will make review more difficult in case of big changes

yashrajbothra:
I thought making more number of commits will help in reviewing.
Should i make more PRs or more commits in this PR only?

williamdes:
> Should i make more PRs or more commits in this PR only?

If can find a balance to only have some obvious changes in one PR and some that require more work in another PR it will help reviewing and merging.
Let us know when we can review :)

yashrajbothra:
Sounds great then this PR will be only for obvious changes. For rest of the stuff i will make another one.
I try to push a commit tommrow after that it will be ready to review 👍 .

yashrajbothra:
Sorry, I need to do force push because previously the commit was containing some vendor files so i have made amendment to unstage those changes. But i dont think Re-review of that commit is required.

yashrajbothra:
Can anyone suggest me which of the possible solution would be better to use in case of `nth-child` as it is deprecated 

[My Solution](https://jsfiddle.net/yashrajbothra/xn5ef3v1/35/)

And please suggest if there is better solution than these?

mauriciofauth:
> Can anyone suggest me which of the possible solution would be better to use in case of `nth-child` as it is deprecated
> 
> [My Solution](https://jsfiddle.net/yashrajbothra/xn5ef3v1/35/)
> 
> And please suggest if there is better solution than these?

Your solution looks good.

williamdes (approved):
Recent changes are okay and your solution is good

williamdes:
Thank you @yashrajbothra for your hard work on those 2 pull-requests !

PR #15805, opened 2020-01-17
Removed AutoCommit=0

### LINKED ISSUE #15802: Why is there a SET AUTOCOMMIT=0 in the SQL of a database export
The first 4 lines of a database export with default settings in phpMyAdmin are:
```
SET SQL_MODE = "NO_AUTO_VALUE_ON_ZERO";
SET AUTOCOMMIT = 0;
START TRANSACTION;
SET time_zone = "+00:00";
```

The last 4 lines contain:
```
COMMIT;

/*!40101 SET CHARACTER_SET_CLIENT=@OLD_CHARACTER_SET_CLIENT */;
/*!40101 SET CHARACTER_SET_RESULTS=@OLD_CHARACTER_SET_RESULTS */;
/*!40101 SET COLLATION_CONNECTION=@OLD_COLLATION_CONNECTION */;
```

I see that the AUTOCOMMIT is disabled but never enabled. I have noticed because we imported a SQL (generated by phpMyAdmin) and in the same session also executed some INSERT statements. These rows where not visible in the database (because they where never committed).

In the MySQL docs I read this:

> With START TRANSACTION, autocommit remains disabled until you end the transaction with COMMIT or ROLLBACK. The autocommit mode then reverts to its previous state.

This states that a `SET AUTOCOMMIT = 0` automatically when starting a transaction. Is there a reason why phpMyAdmin disables the autocommit and then never enables it? Is it possible to remove the `SET AUTOCOMMIT=0` or add a `SET AUTOCOMMIT=1` on the last line of the export?

kartik1000:

### Description

The PR Removes AutoCommit as part of export transaction  as it is managed by 'START TRANSACTION' SQL Keyword.  

Fixes #15802 

### DIFF
--- a/libraries/classes/Plugins/Export/ExportSql.php
+++ b/libraries/classes/Plugins/Export/ExportSql.php
@@ -756,8 +756,7 @@ class ExportSql extends ExportPlugin
         }
 
         if (isset($GLOBALS['sql_use_transaction'])) {
-            $head .= 'SET AUTOCOMMIT = 0;' . $crlf
-                . 'START TRANSACTION;' . $crlf;
+            $head .= 'START TRANSACTION;' . $crlf;
         }
 
         /* Change timezone if we should export timestamps in UTC */


williamdes (changes requested):
This line is very important, please read the MySQL documentation on its usage
You will also understand why we need to set it back to its original value at the end of the import

williamdes:
Or maybe I am wrong, and in this case please link documentation :)

kartik1000:
@williamdes I referred this https://mariadb.com/kb/en/start-transaction/  and I think since we are using START TRANSACTION it is already disabling the AUTOCOMMIT and so there is not need for us to disable it. In case I am wrong shall I then SET AUTOCOMMIT = 1 at the end?

kartik1000:
@williamdes So shall I add AutoCommit = 1 at end?

williamdes (approved):
After some reading of MariaDB and https://dev.mysql.com/doc/refman/8.0/en/commit.html
Removing the line seems like the best idea.

williamdes:
Merged, thank you for fixing this issue
I fixed the tests in 1ff5193f1331aa29c8e777d71ba198eb5fc17baf

PR #15808, opened 2020-01-17
Added LOCK=NONE to ALGORITHM=INPLACE

### LINKED ISSUE #15312: Option to Perform ALTER ONLINE (or ALGORITHM=INPLACE)
When editing the structure of a table, it would be nice to have an option next to the "Preview SQL" button to try an Online Transaction.

image

This would perform the ALTER TABLE with either:

ALTER ONLINE or ALGORITHM=INPLACE

Many times, I want to try to see if I can make a change in the database without locking the table. I could simply select this option and try. If it gives an error saying it's not possible to perform online, then I can try again when the traffic is low.

At the moment, the way I do it is by doing "Preview SQL", copy the generated SQL, and add the word "ONLINE" to it, and try the SQL manually.
Having this checkbox would be a nice feature to avoid these extra steps :)

Thank you.

kartik1000:

### Description

The PR is in addition to Enhancement#15420 to add LOCK=NONE. while doing an online transaction.

Fixes #15312 

### DIFF
--- a/libraries/classes/Controllers/Table/StructureController.php
+++ b/libraries/classes/Controllers/Table/StructureController.php
@@ -942,7 +942,7 @@ class StructureController extends AbstractController
             $sql_query = 'ALTER TABLE ' . Util::backquote($this->table) . ' ';
             $sql_query .= implode(', ', $changes) . $key_query;
             if (isset($_POST['online_transaction'])) {
-                $sql_query .= ', ALGORITHM=INPLACE';
+                $sql_query .= ', ALGORITHM=INPLACE, LOCK=NONE';
             }
             $sql_query .= ';';
 
--- a/libraries/classes/CreateAddField.php
+++ b/libraries/classes/CreateAddField.php
@@ -534,7 +534,7 @@ class CreateAddField
         $sqlQuery = 'ALTER TABLE ' .
             Util::backquote($table) . ' ' . $sqlStatement;
         if (isset($_POST['online_transaction'])) {
-            $sqlQuery .= ', ALGORITHM=INPLACE';
+            $sqlQuery .= ', ALGORITHM=INPLACE, LOCK=NONE';
         }
         $sqlQuery .= ';';
 


nunoperalta:
Yup - looks perfect to me!

kartik1000:
@williamdes Please Review.

williamdes (changes requested):
"To avoid accidentally making the table unavailable for reads, writes, or both, specify a clause on the ALTER TABLE statement such as LOCK=NONE (permit reads and writes) or LOCK=SHARED (permit reads). The operation halts immediately if the requested level of concurrency is not available."
https://dev.mysql.com/doc/refman/5.6/en/innodb-online-ddl.html

https://stackoverflow.com/a/57137660/5155484

I am not sure that forcing the options together is the best idea
Maybe adding another option could be an idea?

williamdes:
@nunoperalta what are your thoughts on this?

nunoperalta:
I think no change needed here.
What you quoted seems to be the expected.

We're asking phpMyAdmin to do an ONLINE transaction, so "LOCK=NONE" is the way to go.

If MySQL/MariaDB will be able to make the transaction without locking (allowing Reads and Writes), then it will proceed.
Otherwise, an error will be returned, saying that this ALTER TABLE cannot be done online.


The StackOverflow question you mentioned doesn't seem to be the expected behavior. Could be a bug in MySQL 5.6, or the person could be doing something else wrong.

LOCK=NONE should definitely never lock the table, and an error will be returned if the database doesn't allow the online transaction.

To me, this PR can be approved.

nunoperalta:
> I am not sure that forcing the options together is the best idea
> Maybe adding another option could be an idea?

I don't see why wouldn't it be a good idea.

MariaDB's "ALTER ONLINE" is equivalent to "LOCK=NONE".

MySQL's way of doing ALTER ONLINE is also with "ALGORITHM=INPLACE, LOCK=NONE"


This page is saying that exactly: https://dev.mysql.com/doc/refman/5.6/en/innodb-online-ddl.html

> ALTER TABLE tbl_name ADD PRIMARY KEY (column), ALGORITHM=INPLACE, LOCK=NONE;

> To avoid accidentally making the table unavailable for reads, writes, or both, specify a clause on the ALTER TABLE statement such as LOCK=NONE (permit reads and writes)

> To avoid tying up the server with an ALTER TABLE operation that copies the table, include ALGORITHM=INPLACE. The statement halts immediately if it cannot use the in-place mechanism.


Therefore, the safest thing to do is:

> ALGORITHM=INPLACE, LOCK=NONE;




https://stackoverflow.com/questions/35424543/alter-table-without-locking-the-entire-table

> "It should happen automatically, but to be sure set ALGORITHM=inplace and LOCK=none to your ALTER TABLE statement."

https://dev.mysql.com/doc/refman/5.6/en/innodb-online-ddl.html

williamdes (approved):
Okay, that you for explaining this to me

This solution is the right one

nunoperalta:
Awesome! Thank you both! :)

williamdes:
Thank you for your patience and help :tada: 
I always prefer to understand what I merge ;)

PR #15816, opened 2020-01-19
fixed font sizing

### LINKED ISSUE #15598: Text size too large?
https://demo.phpmyadmin.net/master-config/index.php?route=/database/sql&db=assigment3

The bold text seems too large / larger than normal.

pma

rahgurung:
Closes #15598 

Used a global styling 1em unit and set font-size to 82% for HTML (it was so in earlier versions)

Signed-off-by: Rahul Gurung <gurungrahul2@gmail.com>

### DIFF
--- a/themes/pmahomme/scss/_common.scss
+++ b/themes/pmahomme/scss/_common.scss
@@ -350,7 +350,7 @@ fieldset {
     -webkit-box-shadow: 3px 3px 15px #bbb;
     box-shadow: 3px 3px 15px #bbb;
     width: initial;
-    font-size: 1rem;
+    font-size: 1em;
   }
 }
 


williamdes (approved):
LGTM

gurrrung:
The latest commit is for fixing font-size in the login panel where we enter root as the username.

themes/pmahomme/scss/_common.scss
```diff
@@ -1,3 +1,11 @@
+html {
+  font-size: 82%;
+}
```
mauriciofauth: This is redundant. This value is already defined in `_variables.scss`.
https://github.com/phpmyadmin/phpmyadmin/blob/d292855d697f4fe52ad0d180b62bfafa05e90cc8/themes/pmahomme/scss/_variables.scss#L63

MauricioFauth (changes requested):

themes/pmahomme/scss/_common.scss
gurrrung: If I'm not wrong, `rem` values are relative to the root html element and that root HTML element size was missing here. Thus I added this.

gurrrung: Ping !

mauriciofauth: _0.82rem_ is the same as _82%_ and the `$font-size-base` is used in the `body` tag.
https://github.com/twbs/bootstrap/blob/v4.4.1/scss/_reboot.scss#L46-L49

gurrrung: Thanks, I got it. Reverted the changes.

MauricioFauth (approved):

williamdes:
Thank your for your contribution @gurrrung !

williamdes:
I added ece12c5fb311cf78410cb30ab72ee86be2ca513d

Issue #15817, opened 2020-01-19
New Table - layout issue on original theme


nunoperalta:
phpMyAdmin 5.0.1

See headings - Name, Type, etc...

![image](https://user-images.githubusercontent.com/487079/72681521-573b7b00-3abc-11ea-89a3-fbf330bdcbe8.png)

milosterzic:
@nunoperalta 

Hi, can you tell us what browser you are using?

nunoperalta:
Chrome, Windows 10

if you cannot reproduce, let me know

milosterzic:
@williamdes 

Can I take this one?

williamdes:
@milosterzic Yes, go ahead and you can use QA_5_0 as base branch if you think it is a UI regression between 4.9 and 5.0

Demo servers (`root`, no password)
- [`5.0.0-dev`](https://demo.phpmyadmin.net/QA_5_0/)
- [`4.9.x`](https://demo.phpmyadmin.net/QA_4_9/)

PR #15829, opened 2020-01-21
Fix for overlapping rows

milosterzic:
Fix for overlapping rows while creating new table.

Signed-off-by: Milos Terzic <terzic.milos.tm@gmail.com>

Fix for issue #15817


williamdes:
Hi @milosterzic 
We only include security fixes into 4.9.x so I will review the other PR :)

PR #15840, opened 2020-01-23
Procedure execute privileges

### LINKED ISSUE #14430: Execute button is disabled for routines
### Steps to reproduce
1. Create a stored procedure:
```sql
DELIMITER $$
CREATE DEFINER=`wordpress_12`@`localhost` PROCEDURE `Sconto su alloggio`(IN `percentuale` DECIMAL(4,2), IN `anno` SMALLINT, IN `alloggio` VARCHAR(16), IN `fascia` CHAR(5))
    NO SQL
    COMMENT 'Applica sconto percentuale alle fasce di un alloggio'
SELECT 1$$
DELIMITER ;
```

2. Grant EXECUTE to another user:
```sql
GRANT EXECUTE ON PROCEDURE `wordpress_f2`.`Sconto su alloggio` TO 'hwreport'@'localhost';
```

3. Login with that user.

### Expected behaviour
Execute button is enabled for the routine.

### Actual behaviour
Execute button is disabled for the routine.
image

### Server configuration

**Web server:** Apache 2.4.18

**Database:** MySQL 5.7.22

**PHP version:** 7.1.13

**phpMyAdmin version:** 4.8.2

### Client configuration
**Browser:** Chrome 67.0.3396.99 (64 bit)

**Operating system:** Windows 10.0.17134 build 17134

victorphoenix3:
Signed-off-by: Jayati Shrivastava <gaurijove@gmail.com>

Fixes #14430 .

### DIFF
--- a/libraries/classes/Rte/RteList.php
+++ b/libraries/classes/Rte/RteList.php
@@ -293,6 +293,8 @@ class RteList
             $routine['type'],
             $routine['name']
         );
+        $params = [];
+        $stmt = (object) [];
         if ($definition !== null) {
             $parser = new Parser($definition);
 
@@ -302,9 +304,11 @@ class RteList
             $stmt = $parser->statements[0];
 
             $params = Routine::getParameters($stmt);
+        }
 
-            if (Util::currentUserHasPrivilege('EXECUTE', $db)) {
-                $execute_action = 'execute_routine';
+        if (Util::currentUserHasPrivilege('EXECUTE', $db, null, $routine['name'])) {
+            $execute_action = 'execute_routine';
+            if ($definition !== null) {
                 for ($i = 0; $i < $params['num']; $i++) {
                     if ($routine['type'] == 'PROCEDURE'
                         && $params['dir'][$i] == 'OUT'
@@ -314,20 +318,19 @@ class RteList
                     $execute_action = 'execute_dialog';
                     break;
                 }
-                $query_part = $execute_action . '=1&amp;item_name='
-                    . urlencode($routine['name']) . '&amp;' . $type_link;
-                $retval .= '                <a class="ajax exec_anchor"'
-                                                 . ' href="db_routines.php'
-                                                 . $url_query
-                                                 . ($execute_action == 'execute_routine'
-                                                     ? '" data-post="' . $query_part
-                                                     : '&amp;' . $query_part)
-                                                 . '">' . $titles['Execute'] . "</a>\n";
-            } else {
-                $retval .= "                {$titles['NoExecute']}\n";
             }
+            $query_part = $execute_action . '=1&amp;item_name='
+                . urlencode($routine['name']) . '&amp;' . $type_link;
+            $retval .= '                <a class="ajax exec_anchor"'
+                                             . ' href="db_routines.php'
+                                             . $url_query
+                                             . ($execute_action == 'execute_routine'
+                                                 ? '" data-post="' . $query_part
+                                                 : '&amp;' . $query_part)
+                                             . '">' . $titles['Execute'] . "</a>\n";
+        } else {
+            $retval .= "                {$titles['NoExecute']}\n";
         }
-
         $retval .= "            </td>\n";
         $retval .= "            <td>\n";
         if ((Util::currentUserHasPrivilege('CREATE ROUTINE', $db)
--- a/libraries/classes/Util.php
+++ b/libraries/classes/Util.php
@@ -3967,19 +3967,24 @@ class Util
      *            // 'CREATE ROUTINE' privilege or, if not, checks if the
      *            // user has this privilege on database 'mydb'.
      *
-     * @param string $priv The privilege to check
-     * @param mixed  $db   null, to only check global privileges
-     *                     string, db name where to also check for privileges
-     * @param mixed  $tbl  null, to only check global/db privileges
-     *                     string, table name where to also check for privileges
+     * @param string      $priv    The privilege to check
+     * @param mixed       $db      null, to only check global privileges
+     *                             string, db name where to also check
+     *                             for privileges
+     * @param mixed       $tbl     null, to only check global/db privileges
+     *                             string, table name where to also check
+     *                             for privileges
+     * @param string|null $routine null,to check for database-specific routine privileges
      *
      * @return bool
      */
-    public static function currentUserHasPrivilege($priv, $db = null, $tbl = null)
+    public static function currentUserHasPrivilege($priv, $db = null, $tbl = null, ?string $routine = null)
     {
         // Get the username for the current user in the format
         // required to use in the information schema database.
         list($user, $host) = $GLOBALS['dbi']->getCurrentUserAndHost();
+
+        $escapedDbName = $GLOBALS['dbi']->escapeString($db);
 
         if ($user === '') { // MySQL is started with --skip-grant-tables
             return true;
@@ -4017,7 +4022,7 @@ class Util
                     'SCHEMA_PRIVILEGES',
                     $username,
                     $priv,
-                    $GLOBALS['dbi']->escapeString($db)
+                    $escapedDbName
                 )
             );
             if ($schema_privileges) {
@@ -4040,7 +4045,7 @@ class Util
                     'TABLE_PRIVILEGES',
                     $username,
                     $priv,
-                    $GLOBALS['dbi']->escapeString($db),
+                    $escapedDbName,
                     $GLOBALS['dbi']->escapeString($tbl)
                 )
             );
@@ -4048,8 +4053,21 @@ class Util
                 return true;
             }
         }
+        if ($routine !== null) {
+            $grants = $GLOBALS['dbi']->fetchResult(
+                'SHOW GRANTS FOR CURRENT_USER()'
+            );
+            foreach ($grants as $grant) {
+                if (strpos($grant, $priv)
+                 && strpos($grant, 'ON PROCEDURE')
+                 && mb_strpos($grant, $escapedDbName)
+                 && mb_strpos($grant, $routine)) {
+                    return true;
+                }
+            }
+        }
         // If we reached this point, the user does not
-        // have even valid table-wise privileges.
+        // have even valid routine privileges.
         return false;
     }
 
--- a/libraries/classes/Rte/Routines.php
+++ b/libraries/classes/Rte/Routines.php
@@ -671,7 +671,16 @@ class Routines
         // Get required data
         $retval['item_name'] = $routine['SPECIFIC_NAME'];
         $retval['item_type'] = $routine['ROUTINE_TYPE'];
-
+        
+        $retval['item_num_params']       = '';
+        $retval['item_param_dir']        = '';
+        $retval['item_param_name']       = '';
+        $retval['item_param_type']       = '';
+        $retval['item_param_length']     = '';
+        $retval['item_param_length_arr'] = '';
+        $retval['item_param_opts_num']   = '';
+        $retval['item_param_opts_text']  = '';
+        
         $definition
             = $this->dbi->getDefinition(
                 $db,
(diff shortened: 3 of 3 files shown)


victorphoenix3:
@williamdes  do test

williamdes (approved):

williamdes:
@victorphoenix3 can you base the branch on QA_5_0 without closing this PR ?
still WIP or is it finished ?

victorphoenix3:
@williamdes  I made some changes. The PR is complete and has been rebased on QA_5_0. Do review.

libraries/classes/Rte/RteList.php
```diff
@@ -293,6 +293,8 @@ public function getRoutineRow(array $routine, $rowclass = '')
             $routine['type'],
             $routine['name']
         );
+        $params= [];
```
williamdes:
```suggestion
        $params = [];
```

libraries/classes/Rte/RteList.php
```diff
@@ -293,6 +293,8 @@ public function getRoutineRow(array $routine, $rowclass = '')
...
+        $stmt = (object) [];
```
williamdes:
```suggestion
        $stmt = new \stdClass();
```
Would it be better ?

williamdes (approved):

libraries/classes/Rte/RteList.php
victorphoenix3: done

libraries/classes/Rte/RteList.php
victorphoenix3: i think using (object) [] would be better,It's been used in other files of the codebase too.

williamdes: Okay then, if the codebase uses it

williamdes:
Given the data in https://github.com/phpmyadmin/phpmyadmin/issues/14430#issue-335734750
I could not validate your fix, It adds the button but it is still disabled..

victorphoenix3:
@williamdes 
Before the fix, Even a user with all the privileges on a database could not execute the procedure.
The PR fixes that. It can be tested by creating a new user and granting him all privileges on just the specific database.

however when granting privileges on just routines,  some discrepancy is still present...I'll try to fix the issue.

williamdes:
> however when granting privileges on just routines, some discrepancy is still present...I'll try to fix the issue.

Thank you, that will cover the test case given in the issue.

victorphoenix3:
@williamdes  I think it is because of this 
`if (Util::currentUserHasPrivilege('EXECUTE', $db))
 {
            $execute_action = 'execute_routine';`  in RteList.php

only database and table privileges are being checked. The routine privileges should be checked too.

victorphoenix3:
@williamdes Do test now.

libraries/classes/Util.php
```diff
@@ -4045,8 +4045,17 @@ public static function currentUserHasPrivilege($priv, $db = null, $tbl = null)
                 return true;
             }
         }
+        if ($routine !== null) {
+            $grants =  $GLOBALS['dbi']->fetchResult(
+                "SHOW GRANTS FOR CURRENT_USER();");
+            foreach ($grants as $grant) {
+                if (strpos($grant,'EXECUTE') && strpos($grant,'ROUTINE ON PROCEDURE') && strpos($grant,$routine)) {
+                    return true;
```
victorphoenix3: @williamdes  is there a better way to check if the CURRENT_USER has routine privileges?

libraries/classes/Util.php
```diff
@@ -4045,8 +4045,17 @@ public static function currentUserHasPrivilege($priv, $db = null, $tbl = null)
                 return true;
             }
         }
+        if ($routine !== null) {
+            $grants =  $GLOBALS['dbi']->fetchResult(
+                "SHOW GRANTS FOR CURRENT_USER();");
+            foreach ($grants as $grant) {
+                if (strpos($grant,$priv) && strpos($grant,'ON PROCEDURE') && strpos($grant,$routine)) {
+                    return true;
```
victorphoenix3: @williamdes  if there is a better way to check 'routine' privileges for CURRENT_USER, do suggest.

victorphoenix3:
@williamdes PR ready for review.

libraries/classes/Util.php
williamdes: @victorphoenix3 could `CheckUserPrivileges::class->getItemsFromShowGrantsRow` be an alternative ?

victorphoenix3: @williamdes  `getItemsFromShowGrantsRow` will require the row number to be passed to it which we don't know.
Do check if the PR is covering all the test cases.

williamdes: @victorphoenix3 I tested your work and it does not cover the case where the user is only granted on the function
Can you find out why ?

victorphoenix3: @williamdes  could you tell me the set of sql queries you used so that i may add the required functionality?

victorphoenix3: with what initial privileges did you create the new user?

williamdes: I followed https://github.com/phpmyadmin/phpmyadmin/issues/14430#issue-335734750

root is the procedure definer
and I added the procedure specific privilege to my test users

I would recommend you to try to use the existing code to parse grands, maybe you will need to change it a bit.

victorphoenix3:
@williamdes  I found this on the mariadb documentation.

` If SQL SECURITY is DEFINER, the function body is always evaluated using the privileges of the definer account. DEFINER is the default. Thus, by default, users who can access the database associated with the stored routine can also run the routine, and potentially perform operations they wouldn't normally have permissions for.`
from-
https://mariadb.com/kb/en/stored-routine-privileges/#running-stored-routines

So the reason why your query is not working is because the user has execute privileges but not database access.
Do you think a workaround for this would be appropriate? The PR covers all the other test cases.

williamdes:
> So the reason why your query is not working is because the user has execute privileges but not database access.

I am not okay with this because I executed the function on my test users manually ;)

> Do you think a workaround for this would be appropriate? The PR covers all the other test cases.

What server do you use ?

williamdes:
@sudo-bot Deploy PR

victorphoenix3:
> 
> What server do you use ?

mariadb

williamdes:
You can see the issue here
https://phpmyadmin-pr-15840.wdes.eu.org/db_routines.php?db=public

user: root/public
user: public/public
user: u1/_no password_

williamdes:
I can run the procedure on u1 but the button is disabled.
image

victorphoenix3:
@williamdes  oh i'll try to fix this.

victorphoenix3:
@williamdes
 dbi->getDefinition uses 'SHOW CREATE' query to retrieve the routine definition.
However,this returns null.

`To use either statement, you must be the user named in the routine DEFINER clause or have SELECT access to the mysql.proc table. If you do not have privileges for the routine itself, the value displayed for the Create Procedure or Create Function field will be NULL.`
https://dev.mysql.com/doc/refman/5.5/en/show-create-procedure.html

is there any other way for other users to access the routine definition? I haven't been able to find an alternative

williamdes:
@victorphoenix3 for me show grants is the only way to go, can you try to use some existing code to achieve this?

victorphoenix3:
> @victorphoenix3 for me show grants is the only way to go, can you try to use some existing code to achieve this?

show grants does not return the routine definition.

victorphoenix3:
@williamdes  if the user is able to execute the query manually, he must be having the routine definition stored somewhere...How can we access it?

victorphoenix3:
@williamdes  presently we can execute procedures that do not require input parameters but we need the routine definition to execute procedures with inputs so that the user can be asked to provide the inputs through the execute_dialog box

williamdes:
@victorphoenix3 after some research I now understand, so we can not execute the prodecure/function because we do not know the definition

So maybe we are blocked for now for that specific case

victorphoenix3:
@williamdes I found this comment in RteList.php 

`Also, information_schema might be hiding the ROUTINE_DEFINITION
but a routine with no input parameters can be nonetheless executed.`

Earlier, even this was broken. In the PR, i have fixed the code to work for those procedures that do not require any input parameters. So routines with no inputs can be now executed.Do test it.

Also, the same bug affects 'function' routines.If you say, I could open a separate PR for that.

libraries/classes/Util.php
```diff
@@ -4045,8 +4045,18 @@ public static function currentUserHasPrivilege($priv, $db = null, $tbl = null)
                 return true;
             }
         }
+        if ($routine !== null) {
+            $grants =  $GLOBALS['dbi']->fetchResult(
+                "SHOW GRANTS FOR CURRENT_USER()"
+                );
+            foreach ($grants as $grant) {
+                if (strpos($grant, $priv) && strpos($grant, 'ON PROCEDURE') && strpos($grant, $routine)) {
```
williamdes: Do we check in case the privilege is given to a procedure having the same name but on another db?

williamdes: > @williamdes I found this comment in RteList.php
> 
> `Also, information_schema might be hiding the ROUTINE_DEFINITION but a routine with no input parameters can be nonetheless executed.`
> 
> Earlier, even this was broken. In the PR, i have fixed the code to work for those procedures that do not require any input parameters. So routines with no inputs can be now executed.Do test it.
> 
> Also, the same bug affects 'function' routines.If you say, I could open a separate PR for that.

I can not test until some days but I would say they are the same thing procedures vs functions 

Could we add a :information_source: button with a link to the faq that explains that we can not find a definition

I had another idea that would be to redirect the user to the sql query box with a pre-filled query without know parameters so the user can complete them?

victorphoenix3: I just added the check on the db.

williamdes: Thank you

libraries/classes/Util.php
```diff
@@ -4045,8 +4046,19 @@ public static function currentUserHasPrivilege($priv, $db = null, $tbl = null)
                 return true;
             }
         }
+        if ($routine !== null) {
+            $grants =  $GLOBALS['dbi']->fetchResult(
+                "SHOW GRANTS FOR CURRENT_USER()"
+                );
+            foreach ($grants as $grant) {
+                if (strpos($grant, $priv) && strpos($grant, 'ON PROCEDURE') && strpos($grant, $GLOBALS['dbi']->escapeString($db))
```
williamdes: I think it would be more optimal to store the quoted db name so it is not calculated on each iteration

libraries/classes/Util.php
```diff
@@ -3969,10 +3969,11 @@ public static function getFunctionsForField(array $field, $insert_mode, array $f
      *                     string, db name where to also check for privileges
      * @param mixed  $tbl  null, to only check global/db privileges
      *                     string, table name where to also check for privileges
+     * @param mixed  $routine  null,to check for database-specific routine privileges
```
williamdes:
```suggestion
     * @param string|null $routine  to check for database-specific routine privileges
```

libraries/classes/Util.php
```diff
@@ -3969,10 +3969,11 @@ public static function getFunctionsForField(array $field, $insert_mode, array $f
...
      *
      * @return bool
      */
-    public static function currentUserHasPrivilege($priv, $db = null, $tbl = null)
+    public static function currentUserHasPrivilege($priv, $db = null, $tbl = null, $routine = null)
```
williamdes:
```suggestion
    public static function currentUserHasPrivilege($priv, $db = null, $tbl = null, ?string $routine = null)
```

williamdes (approved):
The code LGTM

libraries/classes/Util.php
victorphoenix3: > 
> I had another idea that would be to redirect the user to the sql query box with a pre-filled query without know parameters so the user can complete them?

There is pre-existing code for that in routines.php but it doesnt work since the number of input parameters cannot be found without the definition.

williamdes: Yeah I understand but we could prefil a query without knowing the parameters?
That said a FAQ section is an easier implementation

libraries/classes/Util.php
victorphoenix3: have made the required changes

victorphoenix3:
open

williamdes:
@victorphoenix3 where are we on this PR?

victorphoenix3:
It fixes the test case of the bug.

victorphoenix3:
It will work for routines that do not take any input parameters.

williamdes:
I will have another look at it then, sorry for the delay
If you have other open PRs feel free to :ping_pong: on them

williamdes:
I tested it on my MariaDB and it does not work for some reason

williamdes:
I can not validate your work, sorry for that.
Thank you for your contribution

PR #15859, opened 2020-01-26
Fix #15024: Change page settings buttons to Bootstrap

### LINKED ISSUE #15024: Bootstrap button class missing
Bootstrap **btn btn-primary** class should be used in **APPLY** button as old class **ui-button** is used and **btn btn-secondary** class should be used in **CANCEL** button in **Navigation Panel setting modal**.

**Current View**

php

**Future View**
php2

I would like to work on the same

rabeehrz:
Signed-off-by: Mohammed Rabeeh <mohammadrabeeh@gmail.com>

### Description

Fixes #15024 

Fixes #

Changed buttons bootstrap classes. The buttons were generated using jquery-ui. Therefore a selector was initialised switching the classes to Bootstrap ones.

### DIFF
--- a/js/page_settings.js
+++ b/js/page_settings.js
@@ -18,6 +18,7 @@ function showSettings (selector) {
     };
 
     // Keeping a clone to restore in case the user cancels the operation
+
     var $clone = $(selector + ' .page_settings').clone(true);
     $(selector)
         .dialog({
@@ -33,6 +34,11 @@ function showSettings (selector) {
             },
             buttons: buttons
         });
+
+    var $buttonSet = $('.ui-dialog-buttonset');
+    $buttonSet.children().first().attr("class", "btn btn-primary");
+    $buttonSet.children().last().attr("class", "btn btn-secondary");
+
 }
 
 function showPageSettings () {


williamdes (changes requested):
Can you add the class to the `.ui-dialog-buttonset` in the file directly instead of adding JS code ?

rabeehrz:
> Can you add the class to the `.ui-dialog-buttonset` in the file directly instead of adding JS code ?

@williamdes That would mean I would have to edit the minified jQuery UI file. Do you want me to do it that way? That's gonna change how forms are generated using JQuery UI.

williamdes:
> That would mean I would have to edit the minified jQuery UI file. Do you want me to do it that way? 

Oh, no. Never edit vendor files ;)

They do not provide any option ?
Can we edit our css files to add their css class if no option is available ?

rabeehrz:
@williamdes 
I'll check whether they provide options.
We can technically edit our css files to force the JQuery classes to use Bootstrap styles, but then all the buttons that uses those jQuery classes will be affected. I've thought of this and realised that it might not be a great idea in case anyone ever wants to use the original jQuery UI classes.

rabeehrz:
@williamdes I have looked in to the options provided by JQuery UI. They allow us to add extra styles to the buttons, But they don't have the functionality to remove existing styles. I think this is the best way to do this.

williamdes:
> We can technically edit our css files to force the JQuery classes to use Bootstrap styles, but then all the buttons that uses those jQuery classes will be affected.

Yes, but I will let you know that for example in Debian for packaging a package they tend to replace all the vendor contents of packages by an available vendor one when possible.

So quick exemple:
- we change the vendor file
- on Debian side they do not know and find a hypothetical jquery package that contains the vendor file.
- they add a replacement at build time
- and here you go the change you have made will have disappeared

Here is an example case: https://salsa.debian.org/phpmyadmin-team/phpmyadmin/issues/41


> @williamdes I have looked in to the options provided by JQuery UI. They allow us to add extra styles to the buttons, But they don't have the functionality to remove existing styles. I think this is the best way to do this.

Okay, then your fix LGTM :)

rabeehrz:
> Yes, but I will let you know that for example in Debian for packaging a package they tend to replace all the vendor contents of packages by an available vendor one when possible.

@williamdes Ah, that makes a lot of sense! I hadn't thought about that. Thank you for explaining!

> Okay, then your fix LGTM :)

Cool! Hoping to see it approved and merged.  😄

williamdes (approved):

williamdes:
@rabeehrz can you have a look to https://api.jqueryui.com/dialog/#option-classes ?

Maybe you will have to fix the behavior in a lot if places. But for now I an quite undecided to merge this fix because it may not cover all cases and standard ways of using Jquery

rabeehrz:
> Maybe you will have to fix the behavior in a lot if places. But for now I an quite undecided to merge this fix because it may not cover all cases and standard ways of using Jquery

@williamdes You're right. All the dialogs that are being created using JQuery have this problem. A quick search shows me there are at around 56 such dialogs.

> @rabeehrz can you have a look to https://api.jqueryui.com/dialog/#option-classes ?

I did take a look at this. The maximum reach of the classes option is the parent of the buttons, and not the buttons it self. As for adding the classes to buttons, It doesn't come up for some reason.

Also, If you read closely [here](https://learn.jquery.com/jquery-ui/widget-factory/classes-option)

It says:

> This code also removes the existing default value which was "ui-corner-all"

But in my testing, it doesn't remove the existing default value, but only appends to it.

After seeing there are multiple dialogs having the same problem, I can think of two solutions:

1. Do the same thing that is done here in the PR to all the other dialogs.
2. Change the styles for buttons in our css to force them to use Bootstrap by using the !important keyword.

Also, I've noticed that these JQuery dialogs haven't been converted to use Bootstrap style modals (Even in the Boostrap theme) but I think that deserves an issue of its own.

Take a look at this:
[Dialog theming](https://learn.jquery.com/jquery-ui/widget-factory/classes-option/#theming)

I think this was overlooked when the UI was transformed to Bootstrap.

mauriciofauth:
> Also, I've noticed that these JQuery dialogs haven't been converted to use Bootstrap style modals (Even in the Boostrap theme) but I think that deserves an issue of its own.

Yes, the jQuery modals aren't converted to Bootstrap modals yet. That's why the buttons wasn't converted.

rabeehrz:
> Yes, the jQuery modals aren't converted to Bootstrap modals yet. That's why the buttons wasn't converted.

@mauriciofauth Is it something that needs to be done? Shall I open up an issue and work on it?

PR #15896, opened 2020-02-02
Replaced `0` to `NULL`  in primary key

### LINKED ISSUE #15187: editing a row and using 'insert as new row' uses primary key 0
### Describe the bug

If editing a row with an autoincrement primary key and using the option 'insert as new row', the new row will get the primary key 0 assigned.

### To Reproduce

Steps to reproduce the behaviour:

0. Have the database server using SQL mode `NO_AUTO_VALUE_ON_ZERO` (https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_no_auto_value_on_zero).
1. Create a table with an autoincrement primary key `id` and a string `name`.
2. Insert some rows to the table.
3. Click on one row and select `edit`.
4. The bottom dialogue by default has `save` selected. Select `insert as new row`.
5. The column `id` now is automatically set to `0`.
6. Press `go` to edit the row.

### Expected behaviour

My intention is to edit a row and insert it as a new row. This works if the automatically added `0` to the column `id` is manually removed. I expect to not have an automatically filled out `0` in the column `id` as soon as I select `insert as new row`.

### Server configuration

- Database version: MySQL 5.6 with `NO_AUTO_VALUE_ON_ZERO` enabled
- phpMyAdmin version: 4.8.5

### Additional context

The SQL mode `NO_AUTO_VALUE_ON_ZERO` needs to be enabled in order to not interpret the given primary key `0` as a literal zero instead of an instruction to create a new primary key.

yashrajbothra:

### Description

Replaced `0` to `null` for primary id when edited.

Fixes #15187

### DIFF
--- a/js/table/change.js
+++ b/js/table/change.js
@@ -473,7 +473,7 @@ AJAX.registerOnload('table/change.js', function () {
                     || thisElemSubmitTypeVal === 'insertignore'
                     || thisElemSubmitTypeVal === 'showinsert'
                 ) {
-                    $(valueField).val(0);
+                    $(valueField).val(null);
                 } else {
                     $(valueField).val(previousValue);
                 }


yashrajbothra:
I was enhancement so I took `master` as Base 👍

williamdes (approved):
LGTM

williamdes:
@yashrajbothra can you use QA_5_0 as base?
I will consider this one as a bug finally because it blocks the user when inserting new rows

williamdes:
@yashrajbothra I will write a guide to do such operations :)
So you will be able to try it

yashrajbothra:
> @yashrajbothra I will write a guide to do such operations :)
> So you will be able to try it

Ya, that would be better 👍 
For now i think i can do this 😄

williamdes:
> > @yashrajbothra I will write a guide to do such operations :)
> > So you will be able to try it
> 
> Ya, that would be better 
> For now i think i can do this

Sure
Follow #15942

yashrajbothra:
@williamdes  is this good now?

williamdes (approved):
Yes, thank you!

williamdes:
Thank you for your contribution @yashrajbothra :muscle:

PR #15941, opened 2020-02-09
Fixed urls in navigation

### LINKED ISSUE #15626: vsprintf(): Too few arguments
https://demo.phpmyadmin.net/master-config/index.php?route=/table/structure&db=sakila&table=z&goto=index.php%3Froute%3D%2Ftable%2Fstructure&back=index.php%3Froute%3D%2Ftable%2Fstructure&field=a&change_column=1

pma

### LINKED ISSUE #15893: Invalid Link of Column | Index in Nav Panel
### To Reproduce

1. Go to 'Navigation Panel'
2. Expand Database -> Table
3. See error (Invalid URL of link)

### Expected behavior

image

### Screenshots

image

### Additional context

Only in Master Branch

yashrajbothra:

### Description

concatinated url string and removed from array

Fixes: #15893 
Fixes: #15626

### DIFF
--- a/libraries/classes/Navigation/Nodes/NodeColumn.php
+++ b/libraries/classes/Navigation/Nodes/NodeColumn.php
@@ -33,18 +33,12 @@ class NodeColumn extends Node
         $this->links = [
             'text' => Url::getFromRoute('/table/structure', [
                 'server' => $GLOBALS['server'],
-                'db' => '%3\$s',
-                'table' => '%2\$s',
-                'field' => '%1\$s',
                 'change_column' => 1,
-            ]),
+            ]) . '&amp;db=%3$s&amp;table=%2$s&amp;field=%1$s',
             'icon' => Url::getFromRoute('/table/structure', [
                 'server' => $GLOBALS['server'],
-                'db' => '%3\$s',
-                'table' => '%2\$s',
-                'field' => '%1\$s',
                 'change_column' => 1,
-            ]),
+            ]) . '&amp;db=%3$s&amp;table=%2$s&amp;field=%1$s',
             'title' => __('Structure'),
         ];
     }
--- a/libraries/classes/Navigation/Nodes/NodeColumnContainer.php
+++ b/libraries/classes/Navigation/Nodes/NodeColumnContainer.php
@@ -25,14 +25,10 @@ class NodeColumnContainer extends Node
         $this->links = [
             'text' => Url::getFromRoute('/table/structure', [
                 'server' => $GLOBALS['server'],
-                'db' => '%2\$s',
-                'table' => '%1\$s',
-            ]),
+            ]) . '&amp;db=%2$s&amp;table=%1$s',
             'icon' => Url::getFromRoute('/table/structure', [
                 'server' => $GLOBALS['server'],
-                'db' => '%2\$s',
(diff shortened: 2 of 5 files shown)


williamdes (changes requested):
I am not sure this fix it a good solution, can you try to fix the routing function or improve it?

yashrajbothra:
Definately I will try to improve the function.

I pushed this fix becuz all other urls follow similar concatenation ⬇️
image

yashrajbothra:
Hey I tried multiple possible solution but all had some consequenses so only this one fits the best.

Can you suggest anything @williamdes or should we call other members to review

williamdes:
@yashrajbothra did you see 
https://github.com/phpmyadmin/phpmyadmin/blob/0032dff0d33b3f8d84ece0381e7bb437797ddf57/libraries/classes/Url.php#L288
calls

https://github.com/phpmyadmin/phpmyadmin/blob/0032dff0d33b3f8d84ece0381e7bb437797ddf57/libraries/classes/Url.php#L173-L175

maybe add a parameter so it calls https://github.com/phpmyadmin/phpmyadmin/blob/0032dff0d33b3f8d84ece0381e7bb437797ddf57/libraries/classes/Url.php#L208 directly

yashrajbothra:
Yes I was debugging and saw all of those but the problem is when we call `getCommonRaw` it uses 
https://github.com/phpmyadmin/phpmyadmin/blob/0032dff0d33b3f8d84ece0381e7bb437797ddf57/libraries/classes/Url.php#L229
so the return value cannot have any `$` and this results to error when we use `vsprint` to generate link ahead

williamdes:
> Yes I was debugging and saw all of those but the problem is when we call `getCommonRaw` it uses
> https://github.com/phpmyadmin/phpmyadmin/blob/0032dff0d33b3f8d84ece0381e7bb437797ddf57/libraries/classes/Url.php#L229
> 
> 
> so the return value cannot have any `$` and this results to error when we use `vsprint` to generate link ahead

htmlspecialchars seems to be what blocks the code

yashrajbothra:
'vsprintf` contains false because link doesnt contain instructed format.
image

And here we can see that format is changed by this function 
image

As a result we cant see any link 😕

williamdes:
> `vsprintf` contains false because link doesnt contain instructed format.
> image
> 
> And here we can see that format is changed by this function
> image
> 
> As a result we cant see any link

> htmlspecialchars seems to be what blocks the code

Did you remove the function htmlspecialchars to try ?
Did you try to change the https://www.php.net/manual/en/function.http-build-query.php enc_type param ?

yashrajbothra:
> Did you remove the function htmlspecialchars to try ?

Yes, I did try but it didnt worked.

> Did you try to change the https://www.php.net/manual/en/function.http-build-query.php enc_type param ?

Yes, I tried both encoding types but i dindn't help 🤔

williamdes (approved):
Best solution for now, thank you

MauricioFauth (approved):
I'm fine with this change if it fixes the issue. However I think we should find another solution that doesn't require to build the query string manually.

williamdes:
Thank you so much for the fix @yashrajbothra !

PR #15948, opened 2020-02-11
Make logo clickable

### LINKED ISSUE #15946: PhpMyAdmin Logo not clickable
### Describe the bug
The PhpMyAdmin logo is not clickable in the "Metro" theme. 
It is supposed to redirect the user to the landing page onclick.
Works as expected for other themes.

Screenshot from 2020-02-11 01-59-10

Working on this

victorphoenix3:
Fixes #15946

### DIFF
--- a/templates/navigation/main.twig
+++ b/templates/navigation/main.twig
@@ -13,6 +13,9 @@
             {% endif %}
             {% if logo.source is not empty %}
               <img id="imgpmalogo" src="{{ logo.source }}" alt="phpMyAdmin">
+              {% if theme == 'metro' %}
+                <h id="show">phpMyAdmin</h>
+              {% endif %}
             {% else %}
               <h1>phpMyAdmin</h1>
             {% endif %}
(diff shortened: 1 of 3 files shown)


victorphoenix3:
@williamdes  do test.

templates/navigation/main.twig
```diff
@@ -13,6 +13,9 @@
             {% endif %}
             {% if logo.source is not empty %}
               <img id="imgpmalogo" src="{{ logo.source }}" alt="phpMyAdmin">
+              {% if theme == 'metro' %}
```
mauriciofauth: I think it's better to find a solution that work for all themes without adding an exception for the Metro theme.

MauricioFauth (changes requested):

templates/navigation/main.twig
victorphoenix3: But its only for the "metro" theme thay we aren't using the official logo and hence the issue.

PR #15949, opened 2020-02-11
Replace kbd tag by code tag (Bootstrap theme)

### LINKED ISSUE #15882: Bad UI: buttons showing instead of text (bootstrap theme)
### Describe the bug

Buttons are showing instead of text in page-related-settings / navigation panel / logo link target ( bootstrap theme )

### To Reproduce

In bootstrap theme

1. Go to 'Navigation Panel'
2. Click on 'Setting (⚙️) icon '
3. Go to 'navigation panel' tab
4. In Logo link target -> (main) and (new) large button size as compared to text.

### Screenshots

Capture

### Expected behavior

issue@

### Server configuration

- Operating system:
- Web server: xampp
- PHP version: 7.3.8
- phpMyAdmin version: 5.1.0-dev

### Client configuration

- Browser: chrome
- Operating system: windows 10

victorphoenix3:
Fixes #15882 

https://github.com/phpmyadmin/phpmyadmin/issues/15882#issuecomment-584837957

Defined kbd css properties for bootstrap similar to that present in pmahomme,metro and original theme.
Earlier, the properties were being inherited from the bootstrap modules and thus unexpected rendering.

### DIFF
--- a/libraries/classes/Config/Descriptions.php
+++ b/libraries/classes/Config/Descriptions.php
@@ -488,8 +488,8 @@ class Descriptions
             'NavigationLogoLink_desc' => __('URL where logo in the navigation panel will point to.'),
             'NavigationLogoLink_name' => __('Logo link URL'),
             'NavigationLogoLinkWindow_desc' => __(
-                'Open the linked page in the main window ([kbd]main[/kbd]) or in a new one '
-                . '([kbd]new[/kbd]).'
+                'Open the linked page in the main window ([code]main[/code]) or in a new one '
+                . '([code]new[/code]).'
             ),
             'NavigationLogoLinkWindow_name' => __('Logo link target'),
             'NavigationDisplayServers_desc' => __('Display server choice at the top of the navigation panel.'),


victorphoenix3:
@williamdes  ready for review

MauricioFauth (commented):
I don't agree with this change. The Bootstrap theme already have style for the `kbd` tag. I think it is better to change the tag instead of changing the default style of this tag.

williamdes (changes requested):
I agree with @mauriciofauth 
Please change the tag

victorphoenix3:
@williamdes  @mauriciofauth  I didn't find any properties for kbd defined in the _common.scss file for the bootstrap theme. The properties are being inherited from the _code.scss bootstrap module.

In the scss files for all the other themes these are being overridden by defining kbd properties again.

Am I missing something here? Do please take a closer look.

mauriciofauth:
> Am I missing something here? Do please take a closer look.

The idea of the Bootstrap theme is to use the default Bootstrap style. The [`kbd` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd) is used for keyboard inputs and that's why, for this case, I think the `code` tag makes more sense.

victorphoenix3:
I understand it now :)
Have changed the `kbd` to `code`

williamdes (approved):

williamdes:
Thank you @victorphoenix3 !

PR #15953, opened 2020-02-12
Refactor the route definition file 

MauricioFauth:
- Move request and response handling from routes file to the controllers.
- Simplify the routes file by moving the controller creation to the `index.php` file.
- Other minor changes.

### DIFF
--- a/libraries/classes/Controllers/Database/StructureController.php
+++ b/libraries/classes/Controllers/Database/StructureController.php
@@ -111,10 +111,8 @@ class StructureController extends AbstractController
      * Index action
      *
      * @param array $parameters Request parameters
-     *
-     * @return string HTML
      */
-    public function index(array $parameters): string
+    public function index(array $parameters): void
     {
         global $cfg;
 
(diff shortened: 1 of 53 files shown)


libraries/classes/Controllers/Database/StructureController.php
```diff
@@ -219,22 +216,26 @@ public function addRemoveFavoriteTablesAction(array $parameters): ?array
...
-            return [];
+            return;
         }
         $changes = true;
         $titles = Util::buildActionTitles();
         $favoriteTable = $parameters['favorite_table'] ?? '';
         $alreadyFavorite = $this->checkFavoriteTable($favoriteTable);
 
-        if (isset($parameters['remove_favorite'])) {
+        if (isset($_REQUEST['remove_favorite'])) {
```
williamdes: _POST ?

williamdes (approved):
:100:

libraries/classes/Controllers/Database/StructureController.php
mauriciofauth: It was `$_REQUEST['remove_favorite']` in the `routes.php`, but I agree it should be POST. However I think it should be done in other PR. And I think it's better to create a new route for this as well.

williamdes: Okay, I agree :)
Nice PR !

Issue #12625, opened 2016-10-09
Broken Edit links in query results of JOIN query


teo1978:
### Steps to reproduce
1. import the example database attached at the bottom
2. Go to phpmyadmin's SQL tab and type the following query:
```
SELECT DISTINCT user.id, user.username FROM message JOIN user ON message.user_id = user.id WHERE 1
```
3. (shitty markdown is rewriting my "3." as "1.", not my fault) Click in the "Edit" link on any of the result rows

### Expected behaviour
Either the Edit link should not be present or it must work.

### Actual behaviour
You click on the Edit link and you get an error that a query (generated by Phpmyadmin itself) is invalid

### Server configuration

Whatever the current live demo's configuration is.

### Client configuration
Totally irrelevant, but anyway:

**Browser:**
Chrome
**Operating system:**
Ubuntu

### Test Example Database
```
-- phpMyAdmin SQL Dump
-- version 4.7.0-dev
-- https://www.phpmyadmin.net/
--
-- Host: 192.168.30.23
-- Generation Time: Oct 09, 2016 at 05:45 PM
-- Server version: 5.5.46-0+deb8u1
-- PHP Version: 5.6.24-0+deb8u1

SET SQL_MODE = "NO_AUTO_VALUE_ON_ZERO";
SET time_zone = "+00:00";


/*!40101 SET @OLD_CHARACTER_SET_CLIENT=@@CHARACTER_SET_CLIENT */;
/*!40101 SET @OLD_CHARACTER_SET_RESULTS=@@CHARACTER_SET_RESULTS */;
/*!40101 SET @OLD_COLLATION_CONNECTION=@@COLLATION_CONNECTION */;
/*!40101 SET NAMES utf8mb4 */;

--
-- Database: `test67`
--

-- --------------------------------------------------------

--
-- Table structure for table `message`
--

CREATE TABLE `message` (
  `id` int(11) NOT NULL,
  `text` text NOT NULL,
  `user_id` int(11) NOT NULL
) ENGINE=InnoDB DEFAULT CHARSET=latin1;

--
-- Dumping data for table `message`
--

INSERT INTO `message` (`id`, `text`, `user_id`) VALUES
(1, 'Message 1 by Bob', 1),
(2, 'Message 2 by Bob', 1),
(3, 'Message 1 by Alice', 2),
(4, 'Message 2 by Alice', 2);

-- --------------------------------------------------------

--
-- Table structure for table `user`
--

[...]

PR #15971, opened 2020-02-17
Fixed Action buttons on join query

### LINKED ISSUE #13488: Cannot inline edit when a join is used in 4.7.2, works fine in 4.2.5
### Steps to reproduce
1. Run a basic inner join on any two tables (both have a column id which is the PK)  e.g.
SELECT a.* FROM a JOIN b ON b.id = a.org_id WHERE b.title = 'something'
2. I ran the same query in two different versions of phpmyadmin that connect to the same database.  The edit/copy/delete options appear in 4.2.5, however the exact same query does not show these option in v4.7.2

### Expected behaviour
Edit/Copy/Delete should appear to the left of each id, it does not appear in version 4.7.2, but does appear in the older version.

### Actual behaviour
The following message is displayed: "Current selection does not contain a unique column. Grid edit, checkbox, Edit, Copy and Delete features are not available."  edit/copy/delete options appear does not appear.

### Server configuration
Linux
Centos 6
MySQL - 5.5.42
PHP version: 5.5
phpMyAdmin version: 4.7.2
Browser: Chrome & Firefox
Operating system: Windows 8

yashrajbothra:

### Description

Replaced `||` with `&&` so `just_one_table` is only false when there is more than one table

Fixes #13488

### DIFF
--- a/libraries/classes/Sql.php
+++ b/libraries/classes/Sql.php
@@ -1514,9 +1514,9 @@ class Sql
                 }
             }
 
-            if ($analyzed_sql_results['join']
-                || $analyzed_sql_results['is_subquery']
-                || count($analyzed_sql_results['select_tables']) !== 1
+            if (($analyzed_sql_results['join']
+                || $analyzed_sql_results['is_subquery'])
+                && count($analyzed_sql_results['select_tables']) !== 1
             ) {
                 $just_one_table = false;
             }


williamdes (approved):

williamdes (changes requested):
`SELECT message.* FROM message JOIN user ON message.user_id = user.id WHERE user.username != ''` https://github.com/phpmyadmin/phpmyadmin/issues/12625#issue-181894158

The fix you made would revert 6b256e89dd4d405b5c74cacdaa1bf11aa53ee6e1 and 8c357fab12b041b1dd67ce8505e7767c4a7a413e for #12625

williamdes:
For reference:

https://github.com/phpmyadmin/phpmyadmin/blob/7fdde9e1cd80de19eb0a2f9da32306bbd5483b88/libraries/classes/Sql.php#L1860-L1864

@mauriciofauth @ibennetch do you think this PR is right or wrong (see above posted comments) ?

williamdes:
> `SELECT message.* FROM message JOIN user ON message.user_id = user.id WHERE user.username != ''` [#12625 (comment)](https://github.com/phpmyadmin/phpmyadmin/issues/12625#issue-181894158)
> 
> The fix you made would revert [6b256e8](https://github.com/phpmyadmin/phpmyadmin/commit/6b256e89dd4d405b5c74cacdaa1bf11aa53ee6e1) and [8c357fa](https://github.com/phpmyadmin/phpmyadmin/commit/8c357fab12b041b1dd67ce8505e7767c4a7a413e) for #12625

Hi @devenbansod 
Could you give me some advice on this old commit you had done ?

williamdes:
cc @ibennetch @mauriciofauth would anyone like to help me unblock this PR ?

mauriciofauth:
As William mentioned earlier, this pull request will cause a regression of the bug #12625.

PR #16015, opened 2020-03-06
Fix error in NavigationTree where $key might be sent as an int instead of a str to urlencode

simonvik:
### Description
PHP can end up giving an error if a user creates two tables prefixed with number + double underscore like:

```
2020__abc
2020__cde
```

Phpmyadmin will group these tables on the key `2020` and call `urlencode` with an integer and not astring.

Fixes #
The above bug.


williamdes (approved):

williamdes:
Thank you @simonvik for your contribution !
I rebased it onto QA_5_0 to have your fix into next 5.0 version

PR #16026, opened 2020-03-12
Don't fire keyboard shortcuts while SQL query area is focused

### LINKED ISSUE #15224: Shortcuts make entering text impossible on mobile browsers
### Describe the bug

When trying to type a SQL query in a mobile browser (specifically, Safari or Chrome on an iPad), the keyboard shortcuts always fire. For example, if you select the SQL editor and start typing ‘select’, the first ‘s’ brings up the settings menu.

### To Reproduce

1. Use the mobile versions of Safari or Chrome on an iPad
2. Go to the SQL tab
3. Select the query box
4. Press ‘s’
5. The settings window opens instead of putting an ‘s’ in the query box

### Expected behavior

It should be possible to enter text into the SQL editor containing the letter ‘s’

### Server configuration

- Operating system: Linux 
- Web server: cpsrvd 11.78.0.21
- Database version: libmysql - 5.1.73 / 5.6.41-84.1 - Percona Server (GPL), Release 84.1, Revision b308619
- PHP version: 7.2.7
- phpMyAdmin version: 4.8.3

### Client configuration

- Browser: Safari / Chrome
- Operating system: iOS

### Additional context

Perhaps an option to simply disable shortcuts in the settings would be an OK solution?

JayEn84:
Fixes a bug where the keyboard shortcuts functions fire while typing in the SQL query area. The problem was that the SQL query area is not (always) a `<textarea>` but a `<div contenteditable="true">`

Fixes #15224

### DIFF
--- a/js/shortcuts_handler.js
+++ b/js/shortcuts_handler.js
@@ -25,7 +25,7 @@ $(function () {
     var keyC = 67;
     var keyBackSpace = 8;
     $(document).on('keyup', function (e) {
-        if (e.target.nodeName === 'INPUT' || e.target.nodeName === 'TEXTAREA' || e.target.nodeName === 'SELECT') {
+        if (e.target.nodeName === 'INPUT' || e.target.nodeName === 'TEXTAREA' || e.target.nodeName === 'SELECT' || $(':focus').prop('contenteditable') === true) {
             return;
         }
 
@@ -53,7 +53,7 @@ $(function () {
             Console.toggle();
         }
 
-        if (e.target.nodeName === 'INPUT' || e.target.nodeName === 'TEXTAREA' || e.target.nodeName === 'SELECT') {
+        if (e.target.nodeName === 'INPUT' || e.target.nodeName === 'TEXTAREA' || e.target.nodeName === 'SELECT' || $(':focus').prop('contenteditable') === true) {
             return;
         }
 


williamdes (approved):
Thank you for the fix @JayEn84 !
I will fix the git base branch for you :)

williamdes:
while testing I could not validate the fix because something seems wrong

```
console.log($(':focus').prop('contenteditable'), e.target, $(e.target).prop('contenteditable'));
```

when I type `s`  into the query box
```
inherit <a href=​"#Navi_panel">​
            Panneau de navigation
            ​</a>​ inherit
```
when I hit tab
```
true <div class=​"CodeMirror-code" role=​"presentation" autocorrect=​"off" autocapitalize=​"off" spellcheck=​"false" contenteditable=​"true" tabindex=​"100">​…​</div>​ true
```

JayEn84:
So it seems that this only fixes the problem in desktop browser that are sending mobile client header information.

I honestly don't know how to solve this at the moment, because it seems the wrong HTML element is passed to the function that checks if the shortcut functionalities should fire or not.

MauricioFauth (approved):

williamdes:
In fact:
- I had added the console.log on keyup and not on keydown
- `$(e.target).prop('contenteditable') === 'true'` works and not `$(e.target).prop('contenteditable') === true`

williamdes:
Merged as 646f83cbfde24f17ccab8652a126570b1ec2c836

williamdes:
Thank you for your contribution @JayEn84 !

Issue #16005, opened 2020-03-03
More performance, speed first


williamdes:
```
xdebug.profiler_enable=0
xdebug.profiler_enable_trigger=1
xdebug.profiler_append=1
xdebug.profiler_output_dir="/mnt/Dev/@phpmyadmin/theREALphpMyAdminREPO"
xdebug.profiler_output_nme="xdebug-profile.%t.%p"

xdebug.auto_trace=1
xdebug.trace_enable_trigger=1
xdebug.trace_options=1
xdebug.collect_params=4
xdebug.collect_return=0
xdebug.trace_format=0
xdebug.show_mem_delta=1
xdebug.trace_output_dir="/mnt/Dev/@phpmyadmin/theREALphpMyAdminREPO"
xdebug.trace_output_name="trace.%t.%p"

xdebug.file_link_format="xdebug://%f@%l"
xdebug.remote_log="/mnt/Dev/@phpmyadmin/theREALphpMyAdminREPO/xdebug.log"
```

Yes I know we are talking about milliseconds. This is the point of this issue.

Symfony DependencyInjection consumes to much time.

williamdes:
@mauriciofauth the DI yaml parsing is very very complicated and uses a lot of operations.
I would want to change it to a format that does not need complex parsing
Do you agree ?

mauriciofauth:
If you think it is a good improvement, go ahead. And if I remember correctly, this is the only place where we use the yaml dependency.

williamdes:
I am working on it, kind of complicated because I need to write some php code to process the array but I will end up making it work

mauriciofauth:
Isn't easier to switch the [file loader](https://symfony.com/doc/current/components/dependency_injection.html#setting-up-the-container-with-configuration-files)?

williamdes:
The XML parsing is better than the yaml one but using a php array in a file should even better

williamdes:
Done in #16007

williamdes:
This patch did slow down the export by A LOT: https://github.com/phpmyadmin/phpmyadmin/commit/e61b2135da52051bb158c37b9eed35d734bf441c

williamdes:
Performance finding: calling PHP getters a LOT of times (example: 50000 times) is not worth it. Use a public class attribute instead (and add `@readonly` as a phpdoc`).

NKarasek:
[...]

Issue #14604, opened 2018-09-04
Lots of temporary files left open aftzer


madmetz:
### Server configuration

- Operating system:
- Web server:
- Database version:
- PHP version:
- phpMyAdmin version:

### Client configuration

- Browser:
- Operating system:

lem9:
Please show us some sample names of temporary files left there.

ibennetch:
I don't use XAMPP so I'm not sure what's in the `./phpmyadmin/tmp` folder, but it could be the cache folder. When phpMyAdmin is run, some aspects of the pages are generated once and stored in the cache to speed up subsequent access. Those aren't really "temporary" files in the traditional sense, even though we do refer to the folder as "TempDir."

Probably this is the expected behavior of the caching functionality and unlikely to change, although if you can show us the sample names Marc's asked about we could confirm that for you.

madmetz:
@ibennetch 
It seems you are right with the cache files.
BTW Xampp is a server collection for Windows including all the servers you need as a base for a cms.
After browsing some databases with phpMyAdmin the number of (cache) files increases to about 40 or more.
.
_Look at one of the filenames on the right side..._
![phpmyadmin_tmp](https://user-images.githubusercontent.com/42972320/45169906-df3e1f00-b1fe-11e8-93b4-a385c57d5e8c.jpg)
.
To me it makes sense to keep these files as long as phpMyAdmin is running.
But wouldn't it be better to delete these files after termination of phpMyAdmin ?
I'm not sure if this cache is still up-to-date after not using for some time.

lem9:
There is no "termination" of phpMyAdmin as it's not a long running program. It runs just during the time of servicing a request.

madmetz:
> ... as it's not a long running program.

That's right, but when I have to make some changes to a database I start phpMyAdmin in my webbrowser and when I'm done I exit phpMyAdmin by the icon with the half open door on the left side.
May be 'termination' is not the correct word in this context but please be tolerant.
English is not my native language. I'm sooooo sorry for that. :-)

PR #16028, opened 2020-03-14
Use caching to speed up route parsing

williamdes:
Use caching for routing parsing.

I reduces the amounts of php function calls from ~1200 to ~3

Ref: #16005

### DIFF
--- a/index.php
+++ b/index.php
@@ -14,14 +14,6 @@ if (! defined('ROOT_PATH')) {
 
 global $route;
 
-$route = Routing::getCurrentRoute();
-
-if ($route === '/import-status') {
-    // phpcs:disable PSR1.Files.SideEffects
-    define('PMA_MINIMUM_COMMON', true);
-    // phpcs:enable
-}
-
 require_once ROOT_PATH . 'libraries/common.inc.php';
 
 $dispatcher = Routing::getDispatcher();
--- a/index.php
+++ b/index.php
@@ -7,7 +7,7 @@ declare(strict_types=1);
 use FastRoute\Dispatcher;
 use PhpMyAdmin\Message;
 use PhpMyAdmin\Response;
-use function FastRoute\simpleDispatcher;
+use function FastRoute\cachedDispatcher;
 
 if (! defined('ROOT_PATH')) {
     // phpcs:disable PSR1.Files.SideEffects
@@ -15,7 +15,7 @@ if (! defined('ROOT_PATH')) {
     // phpcs:enable
 }
 
-global $containerBuilder, $route;
+global $containerBuilder, $route, $cfg;
 
 /** @var string $route */
 $route = $_GET['route'] ?? $_POST['route'] ?? '/';
@@ -41,7 +41,12 @@ if ($route === '/import-status') {
 require_once ROOT_PATH . 'libraries/common.inc.php';
 
 $routes = require ROOT_PATH . 'libraries/routes.php';
-$dispatcher = simpleDispatcher($routes);
+/** @var \PhpMyAdmin\Config|null $config */
+$config = $GLOBALS['PMA_Config'];
+$dispatcher = cachedDispatcher($routes, [
+    'cacheFile' => $config !== null ? $config->getTempDir('routing') . '/routes.cache' : null,
+    'cacheDisabled' => ($cfg['environment'] ?? '') === 'development',
+]);
 $routeInfo = $dispatcher->dispatch(
     $_SERVER['REQUEST_METHOD'],
     rawurldecode($route)
--- a/index.php
+++ b/index.php
@@ -4,10 +4,7 @@
  */
 declare(strict_types=1);
 
-use FastRoute\Dispatcher;
-use PhpMyAdmin\Message;
-use PhpMyAdmin\Response;
-use function FastRoute\cachedDispatcher;
+use PhpMyAdmin\Routing;
 
 if (! defined('ROOT_PATH')) {
     // phpcs:disable PSR1.Files.SideEffects
@@ -15,22 +12,9 @@ if (! defined('ROOT_PATH')) {
     // phpcs:enable
 }
 
-global $containerBuilder, $route, $cfg;
+global $route;
 
-/** @var string $route */
-$route = $_GET['route'] ?? $_POST['route'] ?? '/';
-
-/**
- * See FAQ 1.34.
- *
- * @see https://docs.phpmyadmin.net/en/latest/faq.html#faq1-34
- */
-if (($route === '/' || $route === '') && isset($_GET['db']) && mb_strlen($_GET['db']) !== 0) {
-    $route = '/database/structure';
-    if (isset($_GET['table']) && mb_strlen($_GET['table']) !== 0) {
-        $route = '/sql';
-    }
-}
+$route = Routing::getCurrentRoute();
 
 if ($route === '/import-status') {
     // phpcs:disable PSR1.Files.SideEffects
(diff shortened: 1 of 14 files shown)


williamdes:
Yes, I added more code safe-guards than necessary so nobody reports blank pages

williamdes:
Before: [trace.1584184315.25.xt.txt](https://github.com/phpmyadmin/phpmyadmin/files/4332664/trace.1584184315.25.xt.txt)
After: [trace.1584186349.56.xt.txt](https://github.com/phpmyadmin/phpmyadmin/files/4332666/trace.1584186349.56.xt.txt)

index.php
```diff
@@ -41,7 +41,12 @@
 require_once ROOT_PATH . 'libraries/common.inc.php';
 
 $routes = require ROOT_PATH . 'libraries/routes.php';
-$dispatcher = simpleDispatcher($routes);
+/** @var \PhpMyAdmin\Config|null $config */
+$config = $GLOBALS['PMA_Config'];
+$dispatcher = cachedDispatcher($routes, [
+    'cacheFile' => $config !== null ? $config->getTempDir('routing') . '/routes.cache' : null,
```
mauriciofauth: I don't think we should use the temp dir. I think a `cache` dir should be created somewhere, and warm up the cache for the release. We can use the same cache dir later for the Twig cache (#14604).

MauricioFauth (commented):
:100:

index.php
mauriciofauth: What do you think of `libraries/cache/`?

williamdes: Good idea, I need to do more testing to know how changes are handled
We could add the cache file to GIT ? (not twig templates cache ^^)

williamdes: > What do you think of `libraries/cache/`?

Why not but it seems a strange place for twig templates cache

mauriciofauth: > We could add the cache file to GIT ?

I don't think so. We should create the file using the create release script.

> Why not but it seems a strange place for twig templates cache

I know, but since the root directory is public, I think is easier to use an existing dir.

williamdes: > I don't think so. We should create the file using the create release script.

Okay

> I know, but since the root directory is public, I think is easier to use an existing dir.

I agree

I will do the changes tomorrow

williamdes: d32807862b20017beba0ba080a5f3626143877a9
Done

williamdes:
@ibennetch I made some changes to the release script, are you okay with them ?

libraries/cache/.gitignore
```diff
@@ -0,0 +1,2 @@
+/*
+!/.gitignore
```
ibennetch: I suggest adding `/libraries/cache` to `/.gitignore`, rather than adding this file here.

We could add the empty cache directory to git but use gitignore to leave it empty in the repository, right?

scripts/create-release.sh
```diff
@@ -80,7 +80,7 @@ while [ $# -gt 0 ] ; do
                     exit 1
                 fi
             elif [ -z "$branch" ] ; then
-                branch=`echo $1 | tr -d -c '0-9A-Za-z_-'`
+                branch=`echo $1 | tr -d -c '/0-9A-Za-z_-'`
```
ibennetch: Is this change related to other changes you're making?

libraries/cache/.gitignore
ibennetch: By the way, the `libraries/cache` folder is fine for me. We should consider adding the path to the `/libraries/vendor_config.php` file, though, so distributions can move it as they see fit.

ibennetch (commented):
Looks good to me. I've added a few comments but think this is a good development.

ibennetch:
> I made some changes to the release script, are you okay with them ?

Sure, they look good to me.

scripts/create-release.sh
williamdes: I could not run the release script with my branch name `perf/cachedRouting`
Can I keep this change ?

libraries/cache/.gitignore
williamdes: Good idea I will do that

williamdes: > We could add the empty cache directory to git but use gitignore to leave it empty in the repository, right?

This is what I did, adding it to the global ignore will not allow me to add the folder to git unfortunately

Is it okay then ?

libraries/classes/Command/RoutingCacheWarmupCommand.php
```diff
@@ -0,0 +1,28 @@
...
+
+use PhpMyAdmin\Routing;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+final class RoutingCacheWarmupCommand extends Command
+{
+    /** @var string */
+    protected static $defaultName = 'routing:cache:warmup';
```
mauriciofauth: What about `cache:routing-warmup`? Or maybe use `cache:warmup` for both warmups (twig and routing) and add cli options to run specific warmups.

scripts/create-release.sh
```diff
@@ -243,7 +240,7 @@ if [ ! -d libraries/tcpdf ] ; then
     cp composer.json composer.json.backup
     echo "* Running composer"
     composer config platform.php "$PHP_REQ"
-    composer update --no-dev
+    composer update --no-dev --optimize-autoloader
```
mauriciofauth: Will this work with customs plugins?

MauricioFauth (changes requested):

scripts/create-release.sh
williamdes: Good catch, I will try to be sure of that

libraries/classes/Command/RoutingCacheWarmupCommand.php
williamdes: I merged the commands, good suggestion :)

libraries/cache/.gitignore
williamdes: Done in 4d2800daa49819b85aceba9089f6bec3da254608

libraries/vendor_config.php
```diff
@@ -79,3 +79,8 @@
  * is used)
  */
 define('K_PATH_IMAGES', ROOT_PATH);
+
+/**
+ * Define the cache directory for routing
+ */
+define('ROUTING_CACHE_DIR', ROOT_PATH . 'libraries/cache/');
```
mauriciofauth: I think just `CACHE_DIR` is better, because we can use this for other types of cache later.

libraries/classes/Command/CacheWarmupCommand.php
```diff
@@ -42,13 +43,39 @@ final class CacheWarmupCommand extends Command
...
+    {
+        if ($input->getOption('twig') === true && $input->getOption('routing') === true) {
+            $output->writeln('Please specify --twig or --routing');
+            return 1;
+        } elseif ($input->getOption('twig') === true) {
+            return $this->warmUpTwigCache($output);
+        } elseif ($input->getOption('routing') === true) {
+            return $this->warmUpRoutingCache($output);
+        } else {
+            $output->writeln('Please specify --twig or --routing');
```
mauriciofauth: I think it could be less strict here and allow both options or none, and in both cases run both warmups.

MauricioFauth (approved):
Good job. :100:

libraries/classes/Command/CacheWarmupCommand.php
```diff
@@ -42,13 +43,39 @@ final class CacheWarmupCommand extends Command
...
+            return 1;
+        }
+    }
+
+    public function warmUpRoutingCache(OutputInterface $output): int
```
mauriciofauth: This should be private, shouldn't it?

libraries/vendor_config.php
williamdes: Done 39b3f3f5e8a9f8149b15726b440d290d328803de
I agree with you

libraries/classes/Command/CacheWarmupCommand.php
williamdes: f79c26f986b9b915e3ad70fadc6296236a302194

libraries/classes/Command/CacheWarmupCommand.php
williamdes: I would prefer not to make it private to allow some testing to be done some day
Do you agree ?

libraries/common.inc.php
```diff
@@ -95,6 +96,14 @@
 }
 require_once AUTOLOAD_FILE;
 
+$route = Routing::getCurrentRoute();
```
williamdes: @mauriciofauth the alternative was fd912545cc27e23d97c156316cf4cd0606046c72
Still okay ?

scripts/create-release.sh
williamdes: Custom plugins load as expected:
- tested a custom auth plugin
- tested a custom export plugin

libraries/classes/Command/CacheWarmupCommand.php
williamdes: c3b99d7baa2a3f91fa9828c30f06bc72fa408857

Issue #15515, opened 2019-10-19
Package on Ubuntu 19.10


nlkduy:
Hi guys,

I installed Ubuntu 19.10, but I didn't find phpmyadmin on this link: https://launchpad.net/ubuntu/eoan/+source/phpmyadmin.

Thank you for seeing!

williamdes:
Hi @nlkduy have a look to #15236, you will find a link to our PPA here (https://github.com/phpmyadmin/phpmyadmin/issues/15236#issuecomment-502392098) 
We are working on packaging for Debian and Ubuntu will update after Debian

Am I correct @Blaimi?

Blaimi:
Yep.

I'm not sure if the ppa is working currently for eoan because it doesn't have php-symfony-polyfill.

I'll update a new package without the dependency in the next days along with an update to 4.9.1

anburocky3:
Tried the **PPA** `sudo add-apt-repository ppa:phpmyadmin/ppa`, didn't work for **Ubuntu 19.10**

anburocky3:
Do we have any timeline deliveries for this package?

williamdes:
@Blaimi Can you please publish the package for the disco series on the PPA ?

williamdes:
"since Eoan is past feature freeze and Debian import freeze, phpmyadmin won't be included in the 19.10 release."
From: https://bugs.launchpad.net/ubuntu/+source/phpmyadmin/+bug/1837775

But we will update the PPA to offer support for Ubuntu 19.10

Blaimi:
I'll do my best to find the time on the coming weekend to update the ppa to 4.9.1 for eoan as well as bionic and inbetweeners.

(If I don't, feel free to disturb 😋)

anburocky3:
Waiting for eoan release. (Y)

Blaimi:
the ppa ist up-to-date now :) tested on eoan, bionic and debian buster.

nlkduy:
![Screenshot from 2019-11-02 10-01-54](https://user-images.githubusercontent.com/16206786/68065173-e35cc280-fd57-11e9-9708-48a07f24b23e.png)
Its work! Thank you very much!

anburocky3:
Here is the PPA rep link: 

>**sudo add-apt-repository ppa:phpmyadmin/ppa**

OGJAYYOU:
how do i open the phpmyadmin dashboard im pretty new to ubuntu

[...]

Issue #15236, opened 2019-04-30
Own Debian repository


Ninos:
Hey there,

since some months the phpmyadmin package for debian stretch is not updated, so it's still vulnerable for CVE-2018-19968:
https://security-tracker.debian.org/tracker/CVE-2018-19968

Because of that I recommend using an own repository for phpmyadmin packages. Any plans for that? :-)

nijel:
The Debian package was originally maintained by me and there used to be a PPA with recent packages. I don't have the time for that and nobody was willing to help (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=879741 was filed one and half year ago), so I've given up and dropped the PPA (see https://twitter.com/mcihar/status/1103290188938264576).

There is https://salsa.debian.org/phpmyadmin-team where some initial work was done for packaging libraries, but I have not clue what state is there (besides those being outdated by almost year now).

mauriciofauth:
I'm interested in helping with maintenance of the Debian package. I've been wanting to do this for a while, but I always end up leaving that aside.

Is it interesting for us to have a `debian` repository? The same way we have for docker and composer.

ibennetch:
Having our own repository would allow us to offer better updates than the
Debian hierarchy allows, but we'd have to be careful about dependency
versions (for instance, if we no longer support MySQL prior to version 8,
but most Debian repositories only contain version 5.7). I'm neutral on this
idea, I'm not really in favor or against it.

nijel:
There are repositories to bring recent PHP or MariaDB versions on Debian, those people would use having recent phpMyAdmin as well. For example see https://github.com/oerdnj/deb.sury.org/issues/1131

The reason for having own repo is that Debian Buster will ship without phpMyAdmin and people will look for installing it there. The alternative approach might be to introduce it there through backports.

Ninos:
For such applications an own repository makes sense, because of faster releasing of new versions. phpmyadmin is non-critical for the system and normally can be updated to newest stable releases. Gitlab is also another good example for having an own repository.

e-gaulue:
Debian is rather known to be serious but not really reactive. By serious, I mean also rather secure and usable for a production environment. 

[...]

PR #16039, opened 2020-03-23
Parenthesis problem

Mageistral:
### Description
On my PHP 7.2 / Ubuntu, leads to an error displaying tables. There is obviously an error because count() is counting a boolean.

Fixes #
Displaying tables without error.

Signed-off-by: Florent Lartet <florent.lartet@drozdi.fr>

### DIFF
--- a/libraries/sql.lib.php
+++ b/libraries/sql.lib.php
@@ -610,7 +610,7 @@ function PMA_isRememberSortingOrder($analyzed_sql_results)
             || $analyzed_sql_results['is_analyse'])
         && $analyzed_sql_results['select_from']
         && ((empty($analyzed_sql_results['select_expr']))
-            || (count($analyzed_sql_results['select_expr'] == 1)
+            || ((count($analyzed_sql_results['select_expr']) == 1)
                 && ($analyzed_sql_results['select_expr'][0] == '*')))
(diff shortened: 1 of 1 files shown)


Mageistral (commented):
Unable to edit the initial commit to add the following line as I don't have any repo on my computer.

Signed-off-by: Florent Lartet <florent.lartet@drozdI.fr>

williamdes:
Hi @Mageistral 
As you can see on https://www.phpmyadmin.net/downloads/ 4.6 had an end of life on April 1, 2017

Please try latest 5.0 version of if you are using Debian or Ubuntu refer to #15236 or #15515

williamdes:
Let me know if you have any more questions :)

ibennetch:
It looks like this has already been fixed since 2017, starting in version 4.7.2.

https://github.com/phpmyadmin/phpmyadmin/pull/13413

https://github.com/phpmyadmin/phpmyadmin/blob/QA_4_9/libraries/classes/Sql.php#L652

Unfortunately, phpMyAdmin 4.6 is so old that we won't be publishing a fix for it. If a vendor is still supplying that version, the fix is quite small for them to backport in a bugfix release and this should be handled through their bug reporting system.

Thank you for your work.

Issue #16060, opened 2020-04-02
Use prepared queries to avoid security issues


williamdes:
Ref: #15492

In order to reduce all possible SQL injections ~~I am thinking about making a query builder class.~~ I am thinking about using prepared queries.

PR #16061, opened 2020-04-02
Add a query builder

williamdes:
Pleas let me know what you think about this proposal, should I continue implementing it?

Ref: #16060

### DIFF
--- a/libraries/classes/Query/Builder.php
+++ b/libraries/classes/Query/Builder.php
@@ -13,7 +13,7 @@ class Builder
      * Make a new SELECT query
      * @param string[] $selectExpression The columns to select
      */
-    public static function select(array $selectExpression = []): SelectQuery
+    public static function select(array $selectExpression = ['*']): SelectQuery
     {
         return new SelectQuery($selectExpression);
     }
--- a/libraries/classes/Query/Builder.php
+++ b/libraries/classes/Query/Builder.php
@@ -1,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpMyAdmin\Query;
+
+/**
+ * Handles making new requests from strings
+ */
+class Builder
+{
+    /**
+     * Make a new SELECT query
+     * @param string[] $selectExpression The columns to select
+     */
+    public static function select(array $selectExpression = []): SelectQuery
+    {
+        return new SelectQuery($selectExpression);
+    }
+}
--- a/libraries/classes/Query/Constraints.php
+++ b/libraries/classes/Query/Constraints.php
@@ -1,0 +1,97 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpMyAdmin\Query;
+
+/**
+ * Represents some constraints
+ */
+trait Constraints
+{
+    /**
+     * @var array<int,array<string|int|float,string|int|float,true>>
+     */
+    private $constraintsExpressions = [];
+
+    /**
+     * Add a normal safe where
+     * @param string           $key      The key name
+     * @param string           $operator The operator (<,>,=)
+     * @param string|int|float $value    The value to bind
+     */
+    public function where(string $key, string $operator, $value): self
+    {
+        $this->constraintsExpressions[] = [$key, $operator, $value];
+        return $this;
+    }
+
+    /**
+     * Add a simple safe where
+     * @param string|int|float $value The value to bind
+     */
+    public function whereSimple($value): self
+    {
+        $this->constraintsExpressions[] = [$value];
+        return $this;
+    }
+
+    /**
+     * Add a raw where
+     * @param string           $key      The key name
+     * @param string           $operator The operator (<,>,=)
+     * @param string|int|float $value    The value to bind
+     */
+    public function whereRaw(string $key, string $operator, $value): self
+    {
+        $this->constraintsExpressions[] = [$key, $operator, $value, true];
+        return $this;
+    }
+
+    /**
+     * @return array<array<int,string|int|float,true>>
+     */
+    public function getConstraintsExpressions(): array
+    {
+        return $this->constraintsExpressions;
+    }
+
+    public function hasConstraintsExpressions(): bool
+    {
+        return count($this->constraintsExpressions) > 0;
+    }
+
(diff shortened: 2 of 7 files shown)


devenbansod:
@williamdes this looks good, definitely a great idea. :+1: 

I just had a couple of doubts:
1. It's not exactly replicating sql-parser's functionality end to end, but could we probably explore if we can use sql-parser's functionality to do this? It may not be possible as it is, but might worth a look. On the top of my head, I think we would have to create the instances of the required Statements and its Components (they generally have public properties to set and get) and then we could re-use their existing build procedures. 

2. Also, if we choose to go the above-suggested way, do you think this might be a good fit to be added the sql-parser instead?

williamdes:
> @williamdes this looks good, definitely a great idea. 
> 
> I just had a couple of doubts:
> 
> 1. It's not exactly replicating sql-parser's functionality end to end, but could we probably explore if we can use sql-parser's functionality to do this? It may not be possible as it is, but might worth a look. On the top of my head, I think we would have to create the instances of the required Statements and its Components (they generally have public properties to set and get) and then we could re-use their existing build procedures.
> 2. Also, if we choose to go the above-suggested way, do you think this might be a good fit to be added the sql-parser instead?

Thank you !

I would like to to add more complexity to sql-parser and maybe keep our classes or if they are very big or complex move them to another package so the world could use our query builder.

For now I would like to avoid using the sql-parser classes as much as possible, maybe could we talk about that when I finish the implementation and it gets merged ?


My idea it to have a fast, simple, secure query builder with minimal code.
100% tested of course

I hope that I did respond to all of your questions ;)

mauriciofauth:
This is a good idea.

What about start using this and replacing the queries when possible and then add a new functionality and start replacing again? Instead of writing the full builder first.
I think this way the Builder could be as simple as possible without unused functionalities.

What do you think?

williamdes:
> This is a good idea.
> 
> What about start using this and replacing the queries when possible and then add a new functionality and start replacing again? Instead of writing the full builder first.
> I think this way the Builder could be as simple as possible without unused functionalities.
> 
> What do you think?

I agree, I will add most of the needed functionalities and begin the work in a local stash to see if I cover most cases :)

And you have a good point: no unused functionalities.

Anyway I will ping you to review when I think the work is "finished"

williamdes:
```diff
diff --git a/libraries/classes/Server/Privileges.php b/libraries/classes/Server/Privileges.php
index cb07fe52a8..a933b2336f 100644
--- a/libraries/classes/Server/Privileges.php
+++ b/libraries/classes/Server/Privileges.php
@@ -12,6 +12,8 @@ use PhpMyAdmin\Display\ChangePassword;
 use PhpMyAdmin\Html\Generator;
 use PhpMyAdmin\Html\MySQLDocumentation;
 use PhpMyAdmin\Message;
+use PhpMyAdmin\Query\Builder;
+use PhpMyAdmin\Query\SelectQuery;
 use PhpMyAdmin\Relation;
 use PhpMyAdmin\RelationCleanup;
 use PhpMyAdmin\Response;
@@ -516,27 +518,28 @@ class Privileges
      * @param string $username username for database connection
      * @param string $hostname hostname for database connection
      *
-     * @return string sql query
+     * @return SelectQuery
      */
-    public function getSqlQueryForDisplayPrivTable($db, $table, $username, $hostname)
+    public function getSqlQueryForDisplayPrivTable($db, $table, $username, $hostname): SelectQuery
     {
         if ($db == '*') {
-            return 'SELECT * FROM `mysql`.`user`'
-                . " WHERE `User` = '" . $this->dbi->escapeString($username) . "'"
-                . " AND `Host` = '" . $this->dbi->escapeString($hostname) . "';";
+            return Builder::select()
+                        ->database('mysql')->table('user')
+                        ->where('User', '=', $username)
+                        ->where('Host', '=', $hostname);
         } elseif ($table == '*') {
-            return 'SELECT * FROM `mysql`.`db`'
-                . " WHERE `User` = '" . $this->dbi->escapeString($username) . "'"
-                . " AND `Host` = '" . $this->dbi->escapeString($hostname) . "'"
-                . " AND '" . $this->dbi->escapeString(Util::unescapeMysqlWildcards($db)) . "'"
-                . ' LIKE `Db`;';
-        }
-        return 'SELECT `Table_priv`'
-            . ' FROM `mysql`.`tables_priv`'
-            . " WHERE `User` = '" . $this->dbi->escapeString($username) . "'"
-            . " AND `Host` = '" . $this->dbi->escapeString($hostname) . "'"
-            . " AND `Db` = '" . $this->dbi->escapeString(Util::unescapeMysqlWildcards($db)) . "'"
-            . " AND `Table_name` = '" . $this->dbi->escapeString($table) . "';";
+            return Builder::select()
+                        ->database('mysql')->table('db')
+                        ->where('User', '=', $username)
+                        ->where('Host', '=', $hostname)
+                        ->where('Db', 'LIKE', Util::unescapeMysqlWildcards($db));
+        }
+        return Builder::select(['Table_priv'])
+                    ->database('mysql')->table('tables_priv')
+                    ->where('User', '=', $username)
+                    ->where('Host', '=', $hostname)
+                    ->where('Db', '=', Util::unescapeMysqlWildcards($db))
+                    ->where('Table_name', '=', $table);
     }
 
     /**
@@ -658,13 +661,13 @@ class Privileges
         if (isset($GLOBALS['username'])) {
             $username = $GLOBALS['username'];
             $hostname = $GLOBALS['hostname'];
-            $sql_query = $this->getSqlQueryForDisplayPrivTable(
+            $sqlQuery = $this->getSqlQueryForDisplayPrivTable(
                 $db,
                 $table,
                 $username,
                 $hostname
             );
-            $row = $this->dbi->fetchSingleRow($sql_query);
+            $row = $sqlQuery->get($this->dbi);
         }
         if (empty($row)) {
             if ($table == '*' && $this->dbi->isSuperuser()) {

```


Her is an example of a migration, does someone have comments ?

williamdes:
Closing this as I concluded this would reduce code readability and it would be better to introduce prepared queries instead.
Prepared queries would do the same as now but would bring security, that was the point of this PR

PR #16088, opened 2020-04-19
swapping long lat if mysql version >= 8.1

### LINKED ISSUE #16012: Spatial Data does not render correctly in MySQL 8 with geographic SRIDs
### Describe the bug

After upgrading to MySQL 8, using data with a geographic SRID (specifically, 4326) causes phpMyAdmin to behave incorrectly when visualizing GIS data.

### To Reproduce

1. Have a table with properly formatted spatial data in a geographic SRS.
2. Select some data from the table.
3. Click the "Visualize GIS Data" button below.
4. The map may temporarily draw something that appears to be a shape, but then blanks out.

### Expected behavior

One would expect the shape to be rendered correctly.

### Server configuration

- Operating system: CentOS 8
- Web server: Apache 2.4
- Database version: 8.0.18
- PHP version: 7.3.13
- phpMyAdmin version: 5.0.1

### Client configuration

- Browser: Safari 13.0.5
- Operating system: Mac OS X 10.15.3

### Additional context

This happens specifically when using a geographic SRS. It's worth noting that MySQL 8 uses and analyzes the SRID much more so than 5.7 did. In MySQL 5.7, the SRID was essentially ignored, so that if one wanted to store the location of the Topeka Zoo in Topeka, KS, for example, you would store it as `ST_GeomFromText('POINT(-95.726760 39.055857)')`, with the X coordinate (longitude) first, followed by the Y coordinate (latitude). It treated all coordinate systems as Cartesian coordinates with SRID 0. MySQL 5.7 at least checked that all SRIDs involved in an operation were the same, but did not otherwise use it. MySQL 5.6 and earlier versions didn't even do that.

MySQL 8 has a better understanding of SRSes, and when using an SRS system that represents geographic coordinates, rather than Cartesian coordinates, it expects points in the regular format: latitude, then longitude. So the Topeka Zoo would be created as `ST_GeomFromText('POINT(39.055857 -95.726760)', 4326)`. (If you were to try to store geometric data using the 5.7 method of X, then Y, MySQL 8 complains that it's not valid GIS data for the spatial reference system specified, because there's no such thing as 95°S.) But however phpMyAdmin is retrieving the data and sending it to the map, I don't think it understands this distinction. It's appears to be treating MySQL 8 geographic data as Cartesian data, which is probably why it starts drawing a shape, gets to a point that's more than 90°W, then craps out because that doesn't make sense, all while pointing the map at Antarctica. It's getting the latitude and longitude flipped.

ankush-0x90:
Signed-off-by: Ankush Patil <aspraz2658@gmail.com>

### Description
used ST_SwapXY when its MYSQL 8

Fixes #16012 

### DIFF
--- a/libraries/classes/Gis/GisVisualization.php
+++ b/libraries/classes/Gis/GisVisualization.php
@@ -166,10 +166,16 @@ class GisVisualization
         $modified_query = 'SELECT ';
         $spatialAsText = 'ASTEXT';
         $spatialSrid = 'SRID';
+        $axisOrder = '';
 
         if ($this->_userSpecifiedSettings['mysqlVersion'] >= 50600) {
             $spatialAsText = 'ST_ASTEXT';
             $spatialSrid = 'ST_SRID';
+        }
+
+        // If MYSQL version >= 8.1 override default axis order
+        if ($this->_userSpecifiedSettings['mysqlVersion'] >= 80010) {
+            $axisOrder = ', \'axis-order=long-lat\'';
         }
 
         // If label column is chosen add it to the query
@@ -182,7 +188,7 @@ class GisVisualization
         // Wrap the spatial column with 'ST_ASTEXT()' function and add it
         $modified_query .= $spatialAsText . '('
             . Util::backquote($this->_userSpecifiedSettings['spatialColumn'])
-            . ') AS ' . Util::backquote(
+            . $axisOrder . ') AS ' . Util::backquote(
                 $this->_userSpecifiedSettings['spatialColumn']
             )
             . ', ';


libraries/classes/Gis/GisVisualization.php
```diff
@@ -167,9 +167,15 @@ private function _modifySqlQuery($sql_query, $rows, $pos)
...
         $spatialSrid = 'SRID';
 
+        $spatialSwap = '';
+        $spatialSwapEnd = '';
+
         if ($this->_userSpecifiedSettings['mysqlVersion'] >= 50600) {
             $spatialAsText = 'ST_ASTEXT';
             $spatialSrid = 'ST_SRID';
+
+            $spatialSwap = 'ST_SwapXY(';
```
williamdes, line 177: Any MySQL version >= 5.6 is affected ?

asprazz: I have tested it on MYSQL8

Do we require it to affect 5.6 too?
ST_GeomFromText() was released in 5.6

williamdes: `if ($this->_userSpecifiedSettings['mysqlVersion'] >= 50600) {` will affect all 5.6+ version ;)

Could you find if this bug was always there or did it start in a particular version of MySQL ?

williamdes: https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-1.html

"A new spatial function, ST_SwapXY(), accepts a geometry argument and swaps the X and Y values of each coordinate pair within the geometry. For more information, see Geometry Format Conversion Functions."

So we need 8.0.1 at a minimum

asprazz: So We should only do this for versions  >= 80010

williamdes: I agree with you

williamdes (approved):
Good diff, maybe the argument to as text would be better to have a smaller diff

asprazz:
> Good diff, maybe the argument to as text would be better to have a smaller diff

Just wait a minute.. :)

asprazz:
please test it against versions <8.1

libraries/classes/Gis/GisVisualization.php
```diff
@@ -205,7 +211,7 @@ private function _modifySqlQuery($sql_query, $rows, $pos)
                 $modified_query .= $rows;
             }
         }
-
+        var_dump($modified_query);
```
williamdes, line 214: Oops you missed to remove this one
Please amend your commit :)

asprazz: and also having problems with versions < 8.1

libraries/classes/Gis/GisVisualization.php
```diff
@@ -166,12 +166,18 @@ private function _modifySqlQuery($sql_query, $rows, $pos)
...
+        $axisOrder = '';
 
         if ($this->_userSpecifiedSettings['mysqlVersion'] >= 50600) {
             $spatialAsText = 'ST_ASTEXT';
             $spatialSrid = 'ST_SRID';
         }
 
+        // If MYSQL version >= 8.1 override default axis order
+        if ($this->_userSpecifiedSettings['mysqlVersion'] >= 80010) {
+            $axisOrder = '\'axis-order=long-lat\'';
```
williamdes, line 178: I would include the comma here to have a 0% diff on the Sql query other user run on their servers

libraries/classes/Gis/GisVisualization.php
williamdes: This is strange 🤔
What kind of issues?

libraries/classes/Gis/GisVisualization.php
asprazz: Thanks that comma was causing problems for version lower than 8.0.1 
I wouldnt have noticed it easily

williamdes (approved):
:100:

williamdes:
I will test it then merge it ASAP

williamdes:
Thank you @asprazz !
I added 7b009216a9c2690e6e1ce3f708886d8539ffe852 to have automatic tests

williamdes:
@sudo-bot /deploy

williamdes:
@sudo-robot /deploy

williamdes:
@sudo-robot Deploy PR

williamdes:
@asprazz This PR broke MariaDB rendering

williamdes:
Fixed by 40d01944a3fde3d13f599a3c474c991b7789ce76

asprazz:
> @asprazz This PR broke MariaDB rendering

I didn't think of MariaDB.

PR #16181, opened 2020-06-09
Use button formaction attribute for table structure bulk links

MauricioFauth:
Uses [button formaction attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-formaction) for table structure bulk links.

It is a IE10+ feature.

### DIFF
--- a/js/table/structure.js
+++ b/js/table/structure.js
@@ -51,7 +51,7 @@ AJAX.registerTeardown('table/structure.js', function () {
     $(document).off('click', '#move_columns_anchor');
     $(document).off('click', '#printView');
     $(document).off('submit', '.append_fields_form.ajax');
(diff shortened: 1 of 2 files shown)


williamdes (approved):
We should document our compatibility minimums, I think some standardized files exist for that

mauriciofauth:
> We should document our compatibility minimums, I think some standardized files exist for that

@williamdes Bootstrap uses [Browserslist](https://github.com/browserslist/browserslist).

williamdes:
> > We should document our compatibility minimums, I think some standardized files exist for that
> 
> @williamdes Bootstrap uses [Browserslist](https://github.com/browserslist/browserslist).

I was looking for it, thank you
Can you add it?

mauriciofauth:
> Can you add it?

Sure!

PR #16195, opened 2020-06-15
Remove the blacklist/whitelist terminology

MauricioFauth:
Replaces blacklist with exclude list and whitelist with allow list.


williamdes (approved):
:100:

ibennetch:
This was great, thank you for the work on it.

PR #16208, opened 2020-06-20
Lock inactive closed issues with GitHub actions

MauricioFauth:
Uses https://github.com/marketplace/actions/lock-threads.

### DIFF
--- a/.github/workflows/lock.yml
+++ b/.github/workflows/lock.yml
@@ -1,0 +1,19 @@
+name: 'Lock threads'
+
+on:
+  schedule:
+    - cron: '0 0,12 * * *'
+
+jobs:
+  lock:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: dessant/lock-threads@v2
+        with:
+          github-token: ${{ github.token }}
+          issue-exclude-created-before: '2018-01-01T00:00:00Z'
+          issue-lock-comment: >
+            This issue has been automatically locked since there
+            has not been any recent activity after it was closed.
+            Please open a new issue for related bugs.
+          process-only: 'issues'


.github/workflows/lock.yml
```diff
@@ -0,0 +1,19 @@
+name: 'Lock threads'
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
```
mauriciofauth, line 5: Maybe we can use `- cron: '0 * * * *'` to run once a hour instead once a day, to clear the backlog more quickly.

.github/workflows/lock.yml
```diff
@@ -0,0 +1,19 @@
...
+
+jobs:
+  lock:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: dessant/lock-threads@v2
+        with:
+          github-token: ${{ github.token }}
+          issue-exclude-created-before: '2018-01-01T00:00:00Z'
```
mauriciofauth, line 14: Do not lock issues older than 2018. I added this to reduce the size of the initial backlog.

mauriciofauth:
I don't have experience with GitHub Actions, so I don't really know how it will work.

.github/workflows/lock.yml
williamdes: I would prefer one a day to have the notifications at the same time and not have to mark as read one by one if you understand what I mean 😉

williamdes:
I agree with the idea, can we configure how long is considered as inactive?
I would say a month

.github/workflows/lock.yml
williamdes: Also we have only 3000 mins of quota, @ibennetch could you confirm that?

.github/workflows/lock.yml
williamdes: Does this only lock closed issues?

mauriciofauth: Yes. It's set in `process-only: 'issues'`.

mauriciofauth:
> can we configure how long is considered as inactive?

Yes. The default is 365 days.

.github/workflows/lock.yml
mauriciofauth: It can be configured to also lock PRs as well.

.github/workflows/lock.yml
mauriciofauth: It only locks 50 issues at each run, that's why I suggested to run it once an hour until it clears the initial backlog.

mauriciofauth:
Searching for [`is:issue is:closed is:unlocked updated:<2019-06-20 created:>2018-01-01`](https://github.com/phpmyadmin/phpmyadmin/issues?q=is%3Aissue+is%3Aclosed+is%3Aunlocked+updated%3A%3C2019-06-20+created%3A%3E2018-01-01) returned 519 inactive issues.

.github/workflows/lock.yml
williamdes: Can we bump the limit and close them 100 by 100 ?

williamdes:
> Yes. The default is 365 days.

Perfect !

.github/workflows/lock.yml
mauriciofauth: > Can we bump the limit and close them 100 by 100 ?

No. The documentation says is "to avoid triggering abuse prevention mechanisms on GitHub".
We can run it twice a day: `- cron: '0 0,12 * * *'`. This way it will lock 100 issues per day.

mauriciofauth: > Also we have only 3000 mins of quota

I don't think this action is time consuming. We don't need to worry about it.

williamdes (approved):
Let's try it

PR #16236, opened 2020-06-27
✨ Use TextArea for JSON columns.

### LINKED ISSUE #16139: Use textareas for JSON columns

The `json` field is currently a small, simple input field. It is very frustrating to work with when I want to inspect or edit `json` values that are longer than 60 characters. Textareas, like they are used for `longtext`, would be the more practical choice for this data type. Especially considering that the [MySQL storage requirements](https://dev.mysql.com/doc/refman/5.7/en/storage-requirements.html#data-types-storage-reqs-json) explicitly state that they use roughly the same amount of storage:

> In general, the storage requirement for a JSON column is approximately the same as for a LONGBLOB or LONGTEXT column; that is, the space consumed by a JSON document is roughly the same as it would be for the document's string representation stored in a column of one of these types.

Make fields of type `json` resizable `<textarea>`s.

Increasing the number of columns and rows for input fields in the settings panel. However, this change applies for `char` and `varchar` as well and not just `json`, and it still does not make the input resizable.

### Additional context

- phpMyAdmin version: 5.0.1

image

ABGEO:
### Description

Hello,

It's my first contribution here 👋

Fixes #16139

### DIFF
--- a/libraries/classes/InsertEdit.php
+++ b/libraries/classes/InsertEdit.php
@@ -758,11 +758,9 @@ class InsertEdit
                 $foreignData,
                 $readOnly
             );
-        } elseif (
-            (
+        } elseif ((
                 $GLOBALS['cfg']['LongtextDoubleTextarea']
-                && mb_strstr($column['pma_type'], 'longtext')
-            )
+                && mb_strstr($column['pma_type'], 'longtext'))
             || mb_strstr($column['pma_type'], 'json')
         ) {
             $html_output .= $this->getTextarea(
--- a/libraries/classes/InsertEdit.php
+++ b/libraries/classes/InsertEdit.php
@@ -758,8 +758,12 @@ class InsertEdit
                 $foreignData,
                 $readOnly
             );
-        } elseif ($GLOBALS['cfg']['LongtextDoubleTextarea']
-            && mb_strstr($column['pma_type'], 'longtext')
+        } elseif (
+            (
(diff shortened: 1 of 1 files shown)


williamdes:
Hi @ABGEO 
Thank you for your contribution !
Could you run  `composer run phpcbf` in order to fix https://travis-ci.org/github/phpmyadmin/phpmyadmin/jobs/702700558#L415 ?

Anyway, only one last thing: could you use your real name (this option is an agreement to our [policy](https://github.com/phpmyadmin/phpmyadmin/blob/master/CONTRIBUTING.md#patches-submission)) for the signoff line ?

ABGEO:
Hi @williamdes 

> Could you run `composer run phpcbf`

Sure, I'll fix it.

> could you use your real name

When I try to use my real name, the DOC check fails because I committed with a different name

mauriciofauth:
> > could you use your real name
> 
> When I try to use my real name, the DCO check fails because I committed with a different name

Using the real name in the signed-off-by line is more important than making the DCO check pass.

MauricioFauth (approved):

williamdes:
Tested against https://github.com/phpmyadmin/phpmyadmin/issues/16139#issuecomment-653199186
Congratulations for your first contribution ! :clap:

ABGEO:
@williamdes thanks :pray: Glad to work with you.

PR #16251, opened 2020-07-08
Implement PSR-7 for HTTP messages handling

### LINKED ISSUE #15507: HTTP handling should implement PSR-7
The HTTP request and HTTP response handling should be refactored to implement the [PSR-7](https://www.php-fig.org/psr/psr-7/) standard.

From the PSR-7 standard:
> This document describes common interfaces for representing HTTP messages as described in [RFC 7230](http://tools.ietf.org/html/rfc7230) and [RFC 7231](http://tools.ietf.org/html/rfc7231), and URIs for use with HTTP messages as described in [RFC 3986](http://tools.ietf.org/html/rfc3986).

MauricioFauth:
This PR uses [nyholm/psr7](https://github.com/Nyholm/psr7), but it could change in the future.

Fixes #15507.

### DIFF
--- a/test/classes/Controllers/TransformationOverviewControllerTest.php
+++ b/test/classes/Controllers/TransformationOverviewControllerTest.php
@@ -7,6 +7,8 @@ declare(strict_types=1);
 
 namespace PhpMyAdmin\Tests\Controllers;
 
+use Nyholm\Psr7\Factory\Psr17Factory;
+use Nyholm\Psr7Server\ServerRequestCreator;
 use PhpMyAdmin\Controllers\TransformationOverviewController;
 use PhpMyAdmin\Template;
 use PhpMyAdmin\Tests\AbstractTestCase;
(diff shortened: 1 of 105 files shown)


williamdes (approved):
Could we have classes that extend the vendor classes to limit the exposure of a vendor in the codebase?
One of the ideas is that we could change vendors using a bunch of lines or a patch for some strange reasons or for packaging reasons :)

mauriciofauth:
> Could we have classes that extend the vendor classes to limit the exposure of a vendor in the codebase?

PSR interfaces are great for that. For Nyholm classes, I just use it for now to make it work. The idea is to be able to switch between any PSR-7 implementation.

test/classes/Controllers/TransformationOverviewControllerTest.php
```diff
@@ -36,17 +38,22 @@ protected function setUp(): void
 
     public function testIndexAction(): void
     {
```
williamdes, line 40: this needs to be moved to the abstract test class I created

mauriciofauth: I agree, but creating full instances is probably is not a good idea anyway.

PR #16272, opened 2020-07-18
Create Export\TemplateModel class

MauricioFauth:
Extracts the database calls from the controller to the model.

### DIFF
--- a/libraries/classes/Export/TemplateModel.php
+++ b/libraries/classes/Export/TemplateModel.php
@@ -1,0 +1,170 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpMyAdmin\Export;
+
+use PhpMyAdmin\DatabaseInterface;
+use PhpMyAdmin\Util;
+use function sprintf;
+
+final class TemplateModel
+{
+    /** @var DatabaseInterface */
+    private $dbi;
+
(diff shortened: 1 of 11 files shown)


libraries/classes/Export/TemplateModel.php
```diff
@@ -0,0 +1,170 @@
...
+            'INSERT INTO %s.%s (`username`, `export_type`, `template_name`, `template_data`)'
+                . ' VALUES (\'%s\', \'%s\', \'%s\', \'%s\');',
+            Util::backquote($db),
+            Util::backquote($table),
+            $this->dbi->escapeString($template->getUsername()),
+            $this->dbi->escapeString($template->getExportType()),
+            $this->dbi->escapeString($template->getName()),
+            $this->dbi->escapeString($template->getData())
+        );
+        $result = @$this->dbi->tryQuery(
```
williamdes, line 34: why do we need `@` ?

williamdes (approved):

libraries/classes/Export/TemplateModel.php
mauriciofauth: It suppress errors. It comes from `Relation::queryAsControlUser`. I will investigate why this operator is there though.

williamdes (approved):

PR #16288, opened 2020-07-31
Add an exception handler

MauricioFauth:


williamdes (approved):
Could you provide an example?

PR #16322, opened 2020-08-28
Create dedicated pages for table maintenance actions

MauricioFauth:
Creates dedicated pages instead of only showing the query results. Also creates a new controller, model and new routes.

The commit c112bfd6d6a16f783d1b1a59494b2ea7d22ca2fb should have been included in this pull request as well.


williamdes (approved):
Very good 💯

PR #16323, opened 2020-08-30
Remove the advisory rules file parser

MauricioFauth:
Uses the parsed PHP array directly instead of parsing a text file into an array every time.


williamdes (approved):
Very good!! 
Ref: #16005

williamdes (approved):
So much simple now

PR #16331, opened 2020-09-02
Add Babel JavaScript compiler

MauricioFauth:
Babel converts modern JavaScript into backwards compatible JavaScript that runs on older browsers.


williamdes (approved):

williamdes:
https://travis-ci.org/github/phpmyadmin/phpmyadmin/jobs/727123865#L596 `/bin/sh: 1: babel: not found`

mauriciofauth:
@williamdes It's because we run `yarn install --production`, but Babel is in `devDependencies`. I think we should install without the `--production` flag to be able to build, and then run `yarn install --production` later. What do you think?

williamdes:
> @williamdes It's because we run `yarn install --production`, but Babel is in `devDependencies`. I think we should install without the `--production` flag to be able to build, and then run `yarn install --production` later. What do you think?

I do not think it should be in devDependencies because it is the same topic as for scss, it is needed to make a production ENV work fine.

For me all `devDependencies` should be tools nobody needs to make phpMyAdmin run :) 
Don't you agree ?

or we should drop some kind of workflow with `STABLE` etc..

PR #16346, opened 2020-09-13
Fix for exporting JSON blobs as hex

realcc:
Apply base64_encode when exporting table that contains blob column

Signed-off-by: Andreas Kurniawan <andreas@kurniawan.ceo>

### DIFF
--- a/libraries/classes/Plugins/Export/ExportJson.php
+++ b/libraries/classes/Plugins/Export/ExportJson.php
@@ -20,6 +20,7 @@ use function bin2hex;
 use function explode;
 use function json_encode;
 use function stripslashes;
+use function strpos;
 
 /**
  * Handles the export for the JSON format
@@ -268,8 +269,8 @@ class ExportJson extends ExportPlugin
             $data = [];
 
             for ($i = 0; $i < $columns_cnt; $i++) {
-                if ($fieldsMeta[$i]->type === 'geometry') {
-                    // export GIS types as hex
+                if ($fieldsMeta[$i]->type === 'geometry' || strpos($fieldsMeta[$i]->type, 'blob') !== false) {
+                    // export GIS and blob types as hex
                     $record[$i] = '0x' . bin2hex($record[$i]);
                 }
                 $data[$columns[$i]] = $record[$i];


realcc (commented):
Apply base64_encode when exporting table that contains blob column

Signed-off-by: Andreas Kurniawan <andreas@kurniawan.ceo>

williamdes:
Could you provide a use case, I looked to your fix for the second time and am not sure to well understand ?

realcc:
> Could you provide a use case, I looked to your fix for the second time and am not sure to well understand ?

Try to create a table that contains a blob column, and then populate with some records. After that, try to export that table into a JSON file, and it does not work properly.

williamdes:
Hi @realcc 

I tested your work and this is the final diff, could you give me access to the repository so I can push the final change ?

```patch
commit 2ba35d78d3890d0b8779e92f8c44dc903a6d5f36 (json-export-blob)
Author: realcc <andreas@kurniawan.ceo>
Date:   Sun Sep 13 20:50:01 2020 +0700

    Fix Incomplete JSON Blob Export
    
    Apply base64_encode when exporting table that contains blob column
    
    Signed-off-by: Andreas Kurniawan <andreas@kurniawan.ceo>

diff --git a/libraries/classes/Plugins/Export/ExportJson.php b/libraries/classes/Plugins/Export/ExportJson.php
index a89918f77f..6fecee0f6c 100644
--- a/libraries/classes/Plugins/Export/ExportJson.php
+++ b/libraries/classes/Plugins/Export/ExportJson.php
@@ -16,10 +16,12 @@ use PhpMyAdmin\Properties\Options\Items\HiddenPropertyItem;
 use PhpMyAdmin\Properties\Plugins\ExportPluginProperties;
 use const JSON_PRETTY_PRINT;
 use const JSON_UNESCAPED_UNICODE;
+use function base64_encode;
 use function bin2hex;
 use function explode;
 use function json_encode;
 use function stripslashes;
+use function strpos;
 
 /**
  * Handles the export for the JSON format
@@ -271,6 +273,8 @@ class ExportJson extends ExportPlugin
                 if ($fieldsMeta[$i]->type === 'geometry') {
                     // export GIS types as hex
                     $record[$i] = '0x' . bin2hex($record[$i]);
+                } elseif (strpos($fieldsMeta[$i]->type, 'blob') !== false) {
+                    $record[$i] = base64_encode($record[$i]);
                 }
                 $data[$columns[$i]] = $record[$i];
             }

```

@ibennetch Should we warn our users about using base64 instead of plain broken binary in the JSON export ?

realcc:
Hi @williamdes

Access granted to the repository.

Many thanks.

williamdes (approved):

williamdes:
> Hi @williamdes
> 
> Access granted to the repository.
> 
> Many thanks.

Thank you, I did push the changes

libraries/classes/Plugins/Export/ExportJson.php
```diff
@@ -271,6 +273,8 @@ public function exportData(
                 if ($fieldsMeta[$i]->type === 'geometry') {
                     // export GIS types as hex
                     $record[$i] = '0x' . bin2hex($record[$i]);
+                } elseif (strpos($fieldsMeta[$i]->type, 'blob') !== false) {
+                    $record[$i] = base64_encode($record[$i]);
```
mauriciofauth, line 277:
```suggestion
                    $record[$i] = base64_encode($record[$i] ?? '');
```

mauriciofauth: Why not use `bin2hex` as well?

MauricioFauth (changes requested):

libraries/classes/Plugins/Export/ExportJson.php
realcc: base64_encode would be more useful when the user want to show the image stored in blob, such as using 'data:image/jpeg;base64,'

mauriciofauth: For general use, I think `bin2hex` makes more sense. What do you think @williamdes?

williamdes: Yes, I agree. In fact I hesitated when rebasing to change to hex.

mauriciofauth: @realcc Could you change to the bin2hex function?

williamdes (approved):

libraries/classes/Plugins/Export/ExportJson.php
williamdes: Done

williamdes:
Now blobs will be exported as hex, thank you for the pull-request @realcc !

PR #16359, opened 2020-09-17
Add Yarn build script that builds CSS and JS

MauricioFauth:
Related to https://github.com/phpmyadmin/phpmyadmin/pull/16331#issuecomment-692681030.

To do after merge:
- Update [pma-demo-update script](https://github.com/phpmyadmin/scripts/blob/81c51c9acdac1198edba65b85a4b31b32327419e/demo/bin/pma-demo-update#L73-L75) to include the build step.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,8 @@
     "lodash": "^4.17.19"
   },
   "scripts": {
-    "postinstall": "yarn run --silent css-compile --quiet --style=compressed && yarn run --silent js-compile --quiet",
+    "postinstall": "yarn run build",
+    "build": "yarn run css-compile --style=compressed && yarn run js-compile",
     "css-compile": "sass themes/pmahomme/scss:themes/pmahomme/css themes/original/scss:themes/original/css themes/metro/scss:themes/metro/css themes/bootstrap/scss:themes/bootstrap/css",
     "css-lint": "stylelint --syntax scss \"themes/**/scss/*.scss\"",
     "js-lint": "eslint js/src",


scripts/create-release.sh
```diff
@@ -295,6 +295,9 @@ fi
 
 if [ -f package.json ] ; then
     echo "* Running Yarn"
+    yarn install
+    yarn run --silent css-compile --quiet --style=compressed
```
williamdes, line 299: What about using build?

williamdes:
I am not too much in favor of this change but some documentation updates would convince me ;)

Users will complain about `--production` a specially the ones using Git STABLE branch

scripts/create-release.sh
mauriciofauth: I didn't use build because of the silent and quiet flags, but we're redirecting the output in the demo server update script anyway, so makes sense to use build here.

https://github.com/phpmyadmin/scripts/blob/81c51c9acdac1198edba65b85a4b31b32327419e/demo/bin/pma-demo-update#L73-L75

mauriciofauth:
Sass and Babel are only used at the build stage, so they are not required on a production environment, like jQuery or CodeMirror.

That is also described in the [npm documentation](https://nodejs.dev/learn/npm-dependencies-and-devdependencies):
> Development dependencies are intended as development-only packages, that are unneeded in production. For example testing packages, webpack or Babel.

We should distribute the release packages with JavaScript and CSS already compiled, there is no need to bundle Sass and Babel as well, because they are not used after the build process.

I think the STABLE branch is not the same as a released package. Using the STABLE branch is the same as using a RELEASE_X_Y_Z tag. Using a compiled software as an example, if you checkout a git tag, you will not find the compiled binaries there, you will still need to build them before use.

williamdes:
Okay then, we will need to have a compiled git version, users will complain for sure.
Perhaps it will be the composer one because we compile CSS files before we distribute them. 
What about having a Git repository with all vendor bundles as git branches @ibennetch?
Also with a stable one. 

This will end users using our development repository for production installs.

ibennetch:
For the reasons mentioned by William, I am hesitant about this. Users complain if we change the deployment process. Perhaps it's worthwhile regardless, but I'm pretty resistant to breaking changes.

ibennetch:
> What about having a Git repository with all vendor bundles as git branches

I'm not fond of this; I feel the Git repository should be for source code. Not for storing the built/compiled files that can be re-generated.

>  if you checkout a git tag, you will not find the compiled binaries there

This is a good point.

Perhaps I don't completely understand this. If a user runs `yarn install` without Babel, what happens? Does phpMyAdmin break or does it just slow down?

Can we put Babel in the production requirements (so that all users running from Git will have it), but remove it as part of the build process (so that we don't distribute the extra and unnecessary cruft)? Have I asked that question clearly enough to make sense?

Otherwise, I think this could be okay if we document the changes. We introduced yarn some time ago and users had to adjust their build process. I'm not fond of this but it seems we haven't come up with much of an alternative yet.

williamdes:
> I'm not fond of this; I feel the Git repository should be for source code. Not for storing the built/compiled files that can be re-generated.

I built https://github.com/code-lts/phpmyadmin-releases#readme to avoid having to do this on our side :)

> Perhaps I don't completely understand this. If a user runs yarn install without Babel, what happens? Does phpMyAdmin break or does it just slow down?

White screen, broken. phpMyAdmin without it's JS scripts does not work.

> Can we put Babel in the production requirements (so that all users running from Git will have it), but remove it as part of the build process (so that we don't distribute the extra and unnecessary cruft)? Have I asked that question clearly enough to make sense?

We do not need node_modules to be shipped so we can leave the process as it is. Just do the adjustment for GIT users

williamdes:
I did commit 590664d08a3ca2b4518c8bdb0ec4ca4d4b150859 to end the CI failures, this PR must be rebased and scss moved with babel.
This makes this PR a feature and not a bug fix ;)

mauriciofauth:
I still think that Babel and Sass should go to development dependencies. However, it is better to do this kind of change for phpMyAdmin 6.0 and not 5.1, since it is a big change for users that install directly from git.

PR #16366, opened 2020-09-20
Fix column definition form not showing default value

MauricioFauth:
To reproduce: 
Add a column that has an user defined default value to the central columns, then try to bulk edit that central column. The input field for the default value will be empty.

Introduced by 5d0fd81e674464c4ba79f40a84974b1129b350e4.


williamdes (approved):
Thanks

What is the scenario to see this bug exists in case someone reports it?

mauriciofauth:
> What is the scenario to see this bug exists in case someone reports it?

Add a column that has an user defined default value to the central columns, then try to bulk edit that central column. The input field for the default value will be empty.

PR #16368, opened 2020-09-20
Fix #16365 Foreign keys not showing

petrduda:
Signed-off-by: Petr Duda <petrduda@seznam.cz>

### Description

Refresh cachedResult variable after loading table details.

Fixes #16365

Before submitting pull request, please review the following checklist:

- [x] Make sure you have read our [CONTRIBUTING.md](https://github.com/phpmyadmin/phpmyadmin/blob/master/CONTRIBUTING.md) document.
- [x] Make sure you are making a pull request against the correct branch. For example, for bug fixes in a released version use the corresponding QA branch and for new features use the _master_ branch. If you have a doubt, you can ask as a comment in the bug report or on the mailing list.
- [x] Every commit has proper `Signed-off-by` line as described in our [DCO](https://github.com/phpmyadmin/phpmyadmin/blob/master/DCO). This ensures that the work you're submitting is your own creation.
- [x] Every commit has a descriptive commit message.
- [x] Every commit is needed on its own, if you have just minor fixes to previous commits, you can squash them.
- [ ] Any new functionality is covered by tests.

PR #16383, opened 2020-10-03
Use refreshed cache data

woodspire:
Clean install from git.
InnoDB table but cannot see "foreign key constraints" section in table "relation view".

The problem is that the cache is refreshed but the local copy of the cache wasn't refreshed.
So calling getStatusInfo('Engine') was returning null, instead of "InnoDB"


woodspire:
Travis errors seems to be unrelated.

williamdes (changes requested):
For some reason, this is a duplicate of #16368

williamdes:
Duplicate of #16368

PR #16403, opened 2020-10-12
Extract controllers from AjaxController

MauricioFauth:
Extracts the config set and get to the ConfigController, the databases list to the DatabaseController, the tables list to the TableController and the columns list to the ColumnController.


williamdes (approved):
:100:

PR #16440, opened 2020-10-28
Adding comments on test selenium.

masitings:
This probably the most common things that people must know. But other than that, hope this help.

### DIFF
--- a/test/bootstrap-phpstan.php
+++ b/test/bootstrap-phpstan.php
@@ -25,6 +25,8 @@ require_once AUTOLOAD_FILE;
 $GLOBALS['cfg'] = $cfg;
 $GLOBALS['server'] = 0;
 $GLOBALS['PMA_Config'] = new Config();
+
+// Defining PhpMyAdmin Configuration
 define('PMA_VERSION', $GLOBALS['PMA_Config']->get('PMA_VERSION'));
 define('PMA_MAJOR_VERSION', $GLOBALS['PMA_Config']->get('PMA_MAJOR_VERSION'));
 define('PROXY_URL', '');
@@ -32,6 +34,7 @@ define('PROXY_USER', '');
 define('PROXY_PASS', '');
 define('PMA_PATH_TO_BASEDIR', '');
 
+// Defining user agent
 $_SERVER['HTTP_USER_AGENT'] = 'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/67.0.3396.87 Safari/537.36';
 $GLOBALS['PMA_Config']->checkClient();
 $GLOBALS['PMA_Config']->checkWebServerOs();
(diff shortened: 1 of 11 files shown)


ibennetch:
Thanks for your interest in contributing, but I don't find most of these particularly helpful. I believe the code should be self-documenting where possible, so comments like most of these just give an opportunity for the comments and code to eventually diverge.

test/selenium/LoginTest.php
```diff
@@ -50,6 +50,7 @@ public function testSuccessfulLogin()
      */
     public function testLoginWithWrongPassword()
     {
+        // With username as "Admin" and the password "Admin"
```
ibennetch, line 53: This is an example of a comment that is basically documented in the code itself; anyone looking at the next line should be able to guess at the credentials. A completely made up alternative comment that would be better could be `The foo() function requires arguments in a strange order: site name, group name, and optionally a username and password` although even that is mostly made unnecessary by a good IDE these days which can often look up and hint at function arguments.

test/selenium/SqlQueryTest.php
```diff
@@ -48,13 +48,15 @@ protected function setUp(): void
...
         $this->waitForElement('partialLinkText', 'SQL')->click();
+        // Let the User Accounts page load
         $this->waitAjax();
 
         $this->typeInTextArea(
             'SET @t1=1, @t2=2, @t3:=4;'
             . 'SELECT 1 as `id`,  @t1, @t2, @t3, @t4 := @t1+@t2+@t3;'
         );
         $this->byId('button_submit_query')->click();
+        // Let the User Accounts page load
```
ibennetch, line 59: This is a moderately helpful type of comment, because the immediate context of what we're waiting for isn't clear, but the need to use it often kind of negates the usefulness.

test/bootstrap-phpstan.php
```diff
@@ -25,13 +25,16 @@
...
+
+// Defining PhpMyAdmin Configuration
 define('PMA_VERSION', $GLOBALS['PMA_Config']->get('PMA_VERSION'));
 define('PMA_MAJOR_VERSION', $GLOBALS['PMA_Config']->get('PMA_MAJOR_VERSION'));
 define('PROXY_URL', '');
 define('PROXY_USER', '');
 define('PROXY_PASS', '');
 define('PMA_PATH_TO_BASEDIR', '');
 
+// Defining user agent
```
ibennetch, line 37: This is another example where the code explains *what* we're doing, but a good comment could explain *why* we're doing it; a made up example could be `// Define the user agent as Chrome on Linux because Opera fails to properly render the animated cursor, which causes the testsuite to forkbomb and set the server on fire`

ibennetch:
Thanks again for your interest, we certainly would love to help you get involved with the project in some other capacity. One way could be with trying to reproduce some older bugs that may not be relevant.

PR #16438, opened 2020-10-27
Auto Save Query database and table independent (fix SQL query auto save)

shucon:
Signed-off-by: Saksham Gupta <shucon01@gmail.com>

### Description

Please describe your pull request.

Fixes #15997

Before submitting pull request, please review the following checklist:

- [ ] Make sure you have read our [CONTRIBUTING.md](https://github.com/phpmyadmin/phpmyadmin/blob/master/CONTRIBUTING.md) document.
- [ ] Make sure you are making a pull request against the correct branch. For example, for bug fixes in a released version use the corresponding QA branch and for new features use the _master_ branch. If you have a doubt, you can ask as a comment in the bug report or on the mailing list.
- [ ] Every commit has proper `Signed-off-by` line as described in our [DCO](https://github.com/phpmyadmin/phpmyadmin/blob/master/DCO). This ensures that the work you're submitting is your own creation.
- [ ] Every commit has a descriptive commit message.
- [ ] Every commit is needed on its own, if you have just minor fixes to previous commits, you can squash them.
- [ ] Any new functionality is covered by tests.

PR #16456, opened 2020-11-04
Enhancement - Get auto-saved query UX

### LINKED ISSUE #15997: Auto saved query no longer loading
### Describe the bug

The get auto-saved query button always shows the message

"No previously auto-saved query is available. Loading default query."

Rather than loading a previous query

### To Reproduce

1. Login to PMA
2. Go to the SQL tab for a particular table
3. Write and run an SQL statement
4. Return to the SQL tab and press "Get auto-saved query"
5. Message prompt "No previously auto-saved query is available. Loading default query."

### Expected behavior

On pressing Get auto-saved query it should load the previous query that was created and ran

### Screenshots

Initial Error
image

I can see the SQL being written into localStorage as I type
image
image

But on submitting a query I can see this value is removed in /js/vendor/sql.js
image

### Server configuration

- Operating system: Linux
- Web server: Apache/2.4.38 (Debian) 
- PHP version:  7.4.1 
- phpMyAdmin version:  5.0.1

### Client configuration

- Browser: Firefox 73.0.1 (64-bit)

### LINKED ISSUE #16046: Improvement to the "Get auto-saved query" UX

When returning to a SQL query if you have a query Saved in local storage you get a popup on screen for ~5 seconds

image

While not actually preventing input - it does have a loading feel, so implies a wait and obscures the editing area. The user may or not want to load a previous SQL at this point in their journey.

Similarly the get auto saved query button is always available regardless of if you do or don't have one saved, if you don't then upon pressing gives you the ~5 second popup to tell you there isn't one saved.

image

A better UX could be had with a simple change I have tested locally. 

- First removed the initial message - allowing the user flow to be not affected by the message 
- Second only show the Get auto saved query button if one is available - so user knows one is available and is not given a option to access a feature that's actually not available to them. 

### Additional context

To achieve seems relatively simple

https://github.com/phpmyadmin/phpmyadmin/blob/b2d1615176c7c9533dded6fdd961c987595cb66c/js/sql.js#L954

needs to be something like
 
```
Sql.checkSavedQuery = function () {
    if (!(isStorageSupported('localStorage') && window.localStorage.autoSavedSql !== undefined)) {
    {
        $("#saved").hide()
    }
};
```
The result

No more ~5seconds message boxes

When there is no auto-saved query available
image

When there is a auto saved query available
image

this would be great and tie in nicely once https://github.com/phpmyadmin/phpmyadmin/issues/15997 is resolved 👍

shucon:
### Description

Fixes https://github.com/phpmyadmin/phpmyadmin/issues/16046
~~Fixes https://github.com/phpmyadmin/phpmyadmin/issues/15997~~

Related to https://github.com/phpmyadmin/phpmyadmin/pull/16438

### DIFF
--- a/js/src/sql.js
+++ b/js/src/sql.js
@@ -988,12 +988,11 @@ Sql.getAutoSavedKey = function () {
 Sql.checkSavedQuery = function () {
     var key = Sql.getAutoSavedKey();
 
-    if (isStorageSupported('localStorage') &&
-        typeof window.localStorage.getItem(key) === 'string') {
-        Functions.ajaxShowMessage(Messages.strPreviousSaveQuery);
-    } else if (Cookies.get(key)) {
-        Functions.ajaxShowMessage(Messages.strPreviousSaveQuery);
+    if ((isStorageSupported('localStorage') && typeof window.localStorage.getItem(key) === 'string')
+        || (typeof Cookies.get(key) === 'string')) {
+        return;
     }
+    $('#saved').attr('disabled', true);
 };
 
 AJAX.registerOnload('sql.js', function () {
--- a/js/src/functions.js
+++ b/js/src/functions.js
@@ -1201,8 +1201,6 @@ Functions.insertQuery = function (queryType) {
             Functions.setQuery(window.localStorage.getItem(key));
         } else if (Cookies.get(key)) {
             Functions.setQuery(Cookies.get(key));
-        } else {
-            Functions.ajaxShowMessage(Messages.strNoAutoSavedQuery);
         }
         return;
     }
--- a/libraries/classes/Controllers/JavaScriptMessagesController.php
+++ b/libraries/classes/Controllers/JavaScriptMessagesController.php
@@ -393,10 +393,6 @@ final class JavaScriptMessagesController
             'strDelete' => __('Delete'),
             'strNotValidRowNumber' => __('%d is not valid row number.'),
             'strBrowseForeignValues' => __('Browse foreign values'),
-            'strNoAutoSavedQuery' => __('No previously auto-saved query is available. Loading default query.'),
-            'strPreviousSaveQuery' => __(
-                'You have a previously saved query. Click Get auto-saved query to load the query.'
-            ),
             'strBookmarkVariable' => __('Variable %d:'),
 
             /* For Central list of columns */


MauricioFauth (approved):

williamdes (approved):
So I have read the code once more and it seems like it is per table and not globally anymore.
The user is aware of that?

shucon:
> So I have read the code once more and it seems like it is per table and not globally anymore.
> The user is aware of that?

No, they don't. But the query gets erased once the user moves to another table before the patch, so I think it's nothing to be concerned about. What do you think?

shucon:
The Lint tests failed unexpectedly, give me a minute.

williamdes:
> > So I have read the code once more and it seems like it is per table and not globally anymore.
> > The user is aware of that?
> 
> No, they don't. But the query gets erased once the user moves to another table (before the patch), so I think it's nothing to be concerned about. What do you think?

Oh okay, then add some clue that is is per db or per db and table in the "get an autosaved" query no match found error.
What do you thing about that ?

shucon:
> > > So I have read the code once more and it seems like it is per table and not globally anymore.
> > > The user is aware of that?
> > 
> > 
> > No, they don't. But the query gets erased once the user moves to another table (before the patch), so I think it's nothing to be concerned about. What do you think?
> 
> Oh okay, then add some clue that is is per db or per db and table in the "get an autosaved" query no match found error.
> What do you thing about that ?

Yeah, we could do that. But we decided to remove that error altogether and only show the `Get auto-saved query` button incase a saved query is available.
Instead, we could change the button text to `Get auto-saved table query`?

williamdes:
> Yeah, we could do that. But we decided to remove that error altogether and only show the Get auto-saved query button incase a saved query is available.
Instead, we could change the button text to Get auto-saved table query?

Seems like I missed part of the diff on my mobile ..

`Get table auto-saved query` and `Get db auto-saved query` seem like a good option.
That said I think that an auto saved query can exist for the db and if one exists for the table the user needs to go to the db to get it.If this is right it is not very easy to use is case of db + table > db only. Maybe 2 buttons ?

cc @ibennetch

shucon:
> 
> `Get table auto-saved query` and `Get db auto-saved query` seem like a good option.
> That said I think that an auto saved query can exist for the db and if one exists for the table the user needs to go to the db to get it.If this is right it is not very easy to use is case of db + table > db only. Maybe 2 buttons ?
> 
> cc @ibennetch

Yes, two buttons would be a nice enhancement. Should we open an issue and discuss there?

williamdes:
> Yes, two buttons would be a nice enhancement. Should we open an issue and discuss there?

Here seems fine it is only a simple change, right ?

js/src/sql.js
```diff
@@ -986,12 +986,12 @@ Sql.getAutoSavedKey = function () {
...
     var key = Sql.getAutoSavedKey();
 
-    if (isStorageSupported('localStorage') &&
-        typeof window.localStorage.getItem(key) === 'string') {
-        Functions.ajaxShowMessage(Messages.strPreviousSaveQuery);
-    } else if (Cookies.get(key)) {
-        Functions.ajaxShowMessage(Messages.strPreviousSaveQuery);
+    if (isStorageSupported('localStorage') && typeof window.localStorage.getItem(key) === 'string') {
+        return;
+    } else if (typeof Cookies.get(key) === 'string') {
```
mauriciofauth, line 991: This `else if` can be merged with previous `if`.

js/src/sql.js
```diff
@@ -986,12 +986,12 @@ Sql.getAutoSavedKey = function () {
...
+        return;
     }
+    $('#saved').hide();
```
mauriciofauth, line 994: Add the `d-none` CSS class instead of using the `.hide()` function.

MauricioFauth (changes requested):
This PR should target the `master` branch instead of `QA_5_1`.

williamdes (changes requested):
Code looks okay, but I am not in favor of this because it hides a feature for the user.

js/src/sql.js
```diff
@@ -988,12 +988,11 @@ Sql.getAutoSavedKey = function () {
...
-    if (isStorageSupported('localStorage') &&
-        typeof window.localStorage.getItem(key) === 'string') {
-        Functions.ajaxShowMessage(Messages.strPreviousSaveQuery);
-    } else if (Cookies.get(key)) {
-        Functions.ajaxShowMessage(Messages.strPreviousSaveQuery);
+    if ((isStorageSupported('localStorage') && typeof window.localStorage.getItem(key) === 'string')
+        || (typeof Cookies.get(key) === 'string')) {
+        return;
     }
+    $('#saved').addClass('d-none');
```
williamdes, line 995: it removes the button, right ?
if yes, can it disable it ?

shucon: Showing the button but disabling it might cause some confusion.

williamdes: Removing the button may also cause confusion :/

williamdes:
Replaced by #19408

PR #16466, opened 2020-11-08
Implement MysqlSslWarningSafeHosts to fix #15584

### LINKED ISSUE #15584: Message SSL is not being used even if server is not remote

williamdes:
### Description 

Set text to black to avoid users spamming their support.

Fixes #15584

MysqlSslWarningSafeHosts

### Before
image

### After
image

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -1607,6 +1607,27 @@ Generic settings
         have to set :config:option:`$cfg['PmaAbsoluteUri']` for correct
         redirection.
 
+.. config:option:: $cfg['MysqlSslWarningSafeHosts']
+
+    :type: array
+    :default: ``['127.0.0.1', 'localhost']``
+
+    This search is case-sensitive and will match the exact string only.
+    If your setup does not use SSL but is safe because you are using a
+    local connection or private network, you can add your hostname or :term:`IP` to the list.
+    You can also remove the default entries to only include yours.
+
+    This check uses the value of :config:option:`$cfg['Servers'][$i]['host']`.
+
+    .. versionadded:: 5.1.0
+
+    Example configuration
+
+    .. code-block:: php
+
+        $cfg['MysqlSslWarningSafeHosts'] = ['127.0.0.1', 'localhost', 'mariadb.local'];
+
+
 .. config:option:: $cfg['ExecTimeLimit']
 
     :type: integer [number of seconds]
--- a/libraries/classes/Html/Generator.php
+++ b/libraries/classes/Html/Generator.php
@@ -301,7 +301,7 @@ class Generator
         $class = 'caution';
         if (! $server['ssl']) {
             $message = __('SSL is not being used');
-            if (! empty($server['socket']) || $server['host'] === '127.0.0.1' || $server['host'] === 'localhost') {
+            if (! empty($server['socket']) || in_array($server['host'], $GLOBALS['cfg']['MysqlSslWarningSafeHosts'])) {
                 $class = '';
             }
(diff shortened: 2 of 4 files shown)


doc/config.rst
```diff
@@ -1616,6 +1616,26 @@ Generic settings
...
+.. config:option:: $cfg['SslNotUsedSafeHostNames']
+
+    :type: array
+    :default: ``['127.0.0.1', 'localhost']``
+
+    If your setup does not use SSL but is safe because you are using a
+    local connection you can add your hostname or :term:`IP` to the list.
+    You can also remove the default entries to only include yours.
+
+    This check uses the value of :config:option:`$cfg['Servers'][$i]['host']`.
```
ibennetch, line 1628: I would add another sentence "This search is case-sensitive and will match the exact string only."

doc/config.rst
```diff
@@ -1616,6 +1616,26 @@ Generic settings
         have to set :config:option:`$cfg['PmaAbsoluteUri']` for correct
         redirection.
 
+.. config:option:: $cfg['SslNotUsedSafeHostNames']
```
ibennetch, line 1619: I'm not sure about the directive name, could we make it more clear?

Would MysqlSslWarningSafeHosts be any better?

I don't feel strongly about this, if we can't think of anything better than what you have proposed is fine.

ibennetch (approved):
Looks great. I have two suggestions but they are minor.

doc/config.rst
```diff
@@ -1616,6 +1616,26 @@ Generic settings
...
+
+    :type: array
+    :default: ``['127.0.0.1', 'localhost']``
+
+    If your setup does not use SSL but is safe because you are using a
+    local connection you can add your hostname or :term:`IP` to the list.
```
ibennetch, line 1625: I would add a comma `,` after "local connection"

ibennetch (approved):

doc/config.rst
ibennetch: How about "local connection or private network, you can add..." ?

williamdes:
I did the changes, feel free to let me know if I can merge this one ;)

doc/config.rst
williamdes: I renamed it, good suggestion :)

doc/config.rst
williamdes: Good catch about private networks, I added it.

ibennetch:
Looks great, you can go ahead and merge it.

MauricioFauth (approved):

PR #16470, opened 2020-11-08
Remove deprecated FOUND_ROWS()

### LINKED ISSUE #15652: Warning: #1287 FOUND_ROWS() is deprecated
` Warning: #1287 FOUND_ROWS() is deprecated and will be removed in a future release. Consider using COUNT(*) instead.`

When clicking on Unique Values for any field.

https://demo.phpmyadmin.net/master-config/index.php?route=/table/structure&db=sakila&table=actor

shucon:

### Description

Fixes: https://github.com/phpmyadmin/phpmyadmin/issues/15652

### DIFF
--- a/libraries/classes/Sql.php
+++ b/libraries/classes/Sql.php
@@ -795,32 +795,26 @@ class Sql
                         ->countRecords(true);
                 }
             } else {
-                // The SQL_CALC_FOUND_ROWS option of the SELECT statement is used.
-
-                // For UNION statements, only a SQL_CALC_FOUND_ROWS is required
-                // after the first SELECT.
-
-                $count_query = Query::replaceClause(
-                    $analyzed_sql_results['statement'],
-                    $analyzed_sql_results['parser']->list,
-                    'SELECT SQL_CALC_FOUND_ROWS',
-                    null,
-                    true
+                $statement = $analyzed_sql_results['statement'];
+                $token_list = $analyzed_sql_results['parser']->list;
+                $replaces = [
+                    // Remove ORDER BY to decrease unnecessary sorting time
+                    [
+                        'ORDER BY',
+                        '',
+                    ],
+                    // Removes LIMIT clause that might have been added
+                    [
+                        'LIMIT',
+                        '',
+                    ],
+                ];
+                $count_query = Query::replaceClauses(
+                    $statement,
+                    $token_list,
+                    $replaces
                 );
-
-                // Another LIMIT clause is added to avoid long delays.
-                // A complete result will be returned anyway, but the LIMIT would
-                // stop the query as soon as the result that is required has been
-                // computed.
-
-                if (empty($analyzed_sql_results['union'])) {
-                    $count_query .= ' LIMIT 1';
-                }
-
-                // Running the count query.
-                $this->dbi->tryQuery($count_query);
-
-                $unlim_num_rows = $this->dbi->fetchValue('SELECT FOUND_ROWS()');
+                $unlim_num_rows = $this->dbi->numRows($this->dbi->tryQuery($count_query));
             }
         } else {// not $is_select
             $unlim_num_rows = 0;


libraries/classes/Sql.php
```diff
@@ -795,32 +795,13 @@ private function countQueryResults(
...
-                // Running the count query.
-                $this->dbi->tryQuery($count_query);
-
-                $unlim_num_rows = $this->dbi->fetchValue('SELECT FOUND_ROWS()');
+                $statement = $analyzed_sql_results['statement'];
+                $token_list = $analyzed_sql_results['parser']->list;
+                $clause = 'SELECT';
+                $count_query = Query::getClause($statement, $token_list, $clause, -1, false) . ' ';
+                $count_query .= $clause . ' ' . Query::getClause($statement, $token_list, $clause, 0) . ' ' ;
+                $count_query .= Query::getClause($statement, $token_list, $clause, 'ORDER BY', false);
```
williamdes, line 803: Could you move this to `Query\Generator` and write unit tests for this ?
Because just reading the code I do not understand what it does :/

williamdes (commented):
This is a way more clean fix, thank you !
I hope another team member can review this one

libraries/classes/Sql.php
```diff
@@ -795,32 +796,26 @@ private function countQueryResults(
...
-
-                if (empty($analyzed_sql_results['union'])) {
-                    $count_query .= ' LIMIT 1';
-                }
-
-                // Running the count query.
-                $this->dbi->tryQuery($count_query);
-
-                $unlim_num_rows = $this->dbi->fetchValue('SELECT FOUND_ROWS()');
+                $unlim_num_rows = mysqli_num_rows($this->dbi->tryQuery($count_query));
```
mauriciofauth, line 818: You should not use mysqli functions directly. You can use `$this->dbi->numRows($this->dbi->tryQuery($count_query))` instead.

MauricioFauth (changes requested):

libraries/classes/Sql.php
```diff
@@ -24,6 +24,7 @@
 use function is_array;
 use function is_bool;
 use function microtime;
+use function mysqli_num_rows;
```
mauriciofauth, line 27:
```suggestion
```

MauricioFauth (changes requested):

libraries/classes/Sql.php
shucon: Ahh, sorry. Generally, my IDE takes care of it so I didn't pay attention to it.

MauricioFauth (approved):

libraries/classes/Sql.php
```diff
@@ -795,32 +795,26 @@ private function countQueryResults(
...
-                    true
+                $statement = $analyzed_sql_results['statement'];
+                $token_list = $analyzed_sql_results['parser']->list;
+                $replaces = [
+                    // Remove ORDER BY to decrease unnecessary sorting time
+                    [
+                        'ORDER BY',
+                        '',
+                    ],
+                    // Removes LIMIT clause that might have been added
```
williamdes, line 806: @shucon I am unsure why you remove the LIMIT every time, do you know ?

The previous code was adding one for one use case
```php
                // Another LIMIT clause is added to avoid long delays.
                // A complete result will be returned anyway, but the LIMIT would
                // stop the query as soon as the result that is required has been
                // computed.

                if (empty($analyzed_sql_results['union'])) {
                    $count_query .= ' LIMIT 1';
                }

```

If this is resolved, LGTM for the diff

shucon: The LIMIT clause is removed because it was mentioned in the comments of the function description: `Function to count the total number of rows for the same 'SELECT' query without the 'LIMIT' clause that may have been programatically added` .

I think it was added to `SQL_CALC_FOUND_ROWS` for some performance enhancement as mentioned in the comment above it. Or it might be some MySQL issue with `SQL_CALC_FOUND_ROWS` as mentioned in https://stackoverflow.com/questions/47407046/sql-calc-found-rows-with-found-rows-always-returns-1 .

williamdes: Thank you :)

williamdes (approved):

fliespl:
This solution is not a direct replacement of found_rows() as one would expect. It generated random results even on such queries:

 Showing rows 0 - 49 (164056 total, 0 in query, Query took 0.1498 seconds.)
```select * from game_auth_logs l join ( select al.user_id, max(al.id) as id from game_auth_logs al where al.successfull = 1 group by al.user_id ) last_log on last_log.id = l.id;```

Showing rows 0 - 24 (267 total, Query took 0.1533 seconds.)
```select l.* from game_auth_logs l join ( select al.user_id, max(al.id) as id from game_auth_logs al where al.successfull = 1 group by al.user_id ) last_log on last_log.id = l.id;```

Notice that only change was start or alias select.



I think it should be configurable to allow how total rows is count.

Mysql as a replacement proposes running two queries:

```
The SQL_CALC_FOUND_ROWS query modifier and accompanying FOUND_ROWS() function are deprecated as of MySQL 8.0.17; expect them to be removed in a future version of MySQL. As a replacement, considering executing your query with LIMIT, and then a second query with COUNT(*) and without LIMIT to determine whether there are additional rows. For example, instead of these queries:
```

williamdes:
> This solution is not a direct replacement of found_rows() as one would expect. It generated random results even on such queries:
> 
> Showing rows 0 - 49 (164056 total, 0 in query, Query took 0.1498 seconds.)
> `select * from game_auth_logs l join ( select al.user_id, max(al.id) as id from game_auth_logs al where al.successfull = 1 group by al.user_id ) last_log on last_log.id = l.id;`
> 
> Showing rows 0 - 24 (267 total, Query took 0.1533 seconds.)
> `select l.* from game_auth_logs l join ( select al.user_id, max(al.id) as id from game_auth_logs al where al.successfull = 1 group by al.user_id ) last_log on last_log.id = l.id;`
> 
> Notice that only change was star or alias select.
> 
> I think it should be configurable to allow end user to decide how total rows is count.
> 
> Mysql as a replacement proposes running two queries:
> 
> ```
> The SQL_CALC_FOUND_ROWS query modifier and accompanying FOUND_ROWS() function are deprecated as of MySQL 8.0.17; expect them to be removed in a future version of MySQL. As a replacement, considering executing your query with LIMIT, and then a second query with COUNT(*) and without LIMIT to determine whether there are additional rows. For example, instead of these queries:
> ```

What version of phpMyAdmin are you running ?

fliespl:
5.1.1

PR #16478, opened 2020-11-12
Improve documentation about installing with Debian and Ubuntu

ibennetch:

### DIFF
--- a/doc/setup.rst
+++ b/doc/setup.rst
@@ -28,13 +28,16 @@ distribution and you will automatically get security updates from your distribut
 Debian and Ubuntu
 -----------------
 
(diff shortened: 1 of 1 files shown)


ibennetch:
I was not happy with the amount of space this information would have taken in the installation documents, so I put some of the details in the wiki. I'm also not pleased about fragmenting the documentation, so please let me know what you think — should what I added to the wiki be put here, or is it better as a link there?

williamdes (approved):
Perfect, I like the wiki usage because I can edit it as often as I want if I change a packaging detail

MauricioFauth (approved):

PR #16499, opened 2020-11-24
Prevent Directory Listing

saharavitan:
### Description

Index.php files have been added to prevent Directory Listing Attack.

Fixes #

- [X] Any new functionality is covered by tests.

### DIFF
--- a/libraries/index.php
+++ b/libraries/index.php
@@ -1,0 +1,1 @@
+<!-- Nothing to show -->
--- a/scripts/index.php
+++ b/scripts/index.php
@@ -1,0 +1,1 @@
+<!-- Nothing to show -->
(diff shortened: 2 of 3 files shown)


ibennetch (changes requested):
Thanks for your suggestion.

Hm, I'm not convinced of the need for this; there should be no sensitive information in these three specific folders (out of the over 400 subdirectories in my current phpMyAdmin folder). Any of the contained files would be part of any download, so I don't see that there's a security issue specifically with allowing them to be listed.

I also don't think we need to create a blank `index.php` for each subdirectory that exists in the tree; I think that would be unnecessary and adding complexity for no true benefit. In general, I understand the reasoning behind not listing directories, but don't feel there's anything to be gained from protecting these subdirectories.

williamdes:
I agree there is no true benefit, you should use some webserver configuration instead.

PR #16500, opened 2020-11-24
Fix #16487 - Add paging to Routines page

### LINKED ISSUE #16487: Routines page lacks paging controls which leads to high CPU usage with large number of procedures
### Describe the bug

If the database contains large number of stored procedures (>5K), the `Routines` page take a very long time to load and the CPU is pegged to 100% during the load. With even larger number of stored procedures, the page will fail to load.

Paging controls should be added to the Routines page.

### To Reproduce

1. Create many stored procedures (5K+)
2. Expand database on the navigation tree.
3. Click on the `Procedures` in the tree

### Expected behavior

Routines page is displayed quickly with paging controls.

### Screenshots

image

### Server configuration

- Operating system: Centos
- Web server: apache2
- Database version: MySQL 5.7
- PHP version: 7.2
- phpMyAdmin version: 4.9.7

### Client configuration

- Browser: Chrome
- Operating system: MacOS

petrduda:
Signed-off-by: Petr Duda <petrduda@seznam.cz>

### Description

Added paging controls to the Routines page

Fixes #16487 

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -1559,6 +1559,15 @@ Generic settings
     The maximum number of table names to be displayed in the main panel's
     list (except on the Export page).
 
+.. config:option:: $cfg['MaxRoutineList']
+
+    :type: integer
+    :default: 250
+
+    The maximum number of routine names to be displayed in the main panel's
+    routine list.
+
+
 .. config:option:: $cfg['ShowHint']
 
     :type: boolean
--- a/libraries/classes/Controllers/Database/RoutinesController.php
+++ b/libraries/classes/Controllers/Database/RoutinesController.php
@@ -9,6 +9,7 @@ use PhpMyAdmin\Core;
 use PhpMyAdmin\Database\Routines;
 use PhpMyAdmin\DatabaseInterface;
 use PhpMyAdmin\DbTableExists;
+use PhpMyAdmin\Html\Generator;
 use PhpMyAdmin\Response;
 use PhpMyAdmin\Template;
 use PhpMyAdmin\Url;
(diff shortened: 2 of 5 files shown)


doc/config.rst
```diff
@@ -1559,6 +1559,15 @@ Generic settings
     The maximum number of table names to be displayed in the main panel's
     list (except on the Export page).
 
+.. config:option:: $cfg['MaxRoutineList']
+
+    :type: integer
+    :default: 250
+
```
williamdes, line 1566:
```suggestion
    .. versionadded:: 5.1.0
```

williamdes (approved):
LGTM

roccivic:
I don't think this PR will actually fix the issue. I think it will reduce the network latency, but the lag will remain as the very same slow query will still be running here: https://github.com/phpmyadmin/phpmyadmin/blob/5d8b9288d1fa919fe7f9d58cdae54cc3547246de/libraries/classes/Query/Generator.php#L154

As @erolatreston notes [here](https://github.com/phpmyadmin/phpmyadmin/issues/16487#issuecomment-736874324), the correct solution would be to run another faster query to only retrieve the names of routines when displaying the list, and then run the slow query for a single item when it needs to be edited.

This issue, at least in theory, also affects events (and likely triggers).

petrduda:
Hello @roccivic and @erolatreston, thank you for the comments. Before creating the pull request, I made a quick test as what was really causing the slow response. Did it again now on the current master, it does not seem to me that the slowdown is caused by the above mentioned query in the getInformationSchemaRoutinesRequest() function, but rather by the subsequent rendering of the rows one by one (which includes checking the user privileges to edit the procedure etc. in libraries/classes/Database/Routines.php). Hence the suggested simple modification. If you don't think the solution is clean enough, please feel free to close this PR and work on a more robust redesign of the page, which I won't be able to in the nearest future.

The following screencaps illustrate what I mean, the edited files are libraries/classes/Controllers/Database/RoutinesController.php and templates/database/routines/index.twig, the times are for 1000 stored functions. I observed no significant impact of the stored function complexity on the first time, but some impact on the second one.

RoutinesController php
index twig
routines_page_times

williamdes:
I assigned this to 5.2 so more discussion can be done before merging or changing your work

MauricioFauth (approved):

I'll merge this into `master` branch.

PR #16508, opened 2020-11-29
Create Table\PartitionController class

MauricioFauth:
Extracts table partitioning related actions to the `Table\PartitionController` class.

### DIFF
--- a/libraries/classes/Controllers/Table/PartitionController.php
+++ b/libraries/classes/Controllers/Table/PartitionController.php
@@ -161,23 +161,25 @@ final class PartitionController extends AbstractController
 
     public function repair(): void
     {
-        global $containerBuilder, $sql_query;
-
         $partitionName = $_POST['partition_name'] ?? '';
 
         if (strlen($partitionName) === 0) {
             return;
         }
 
-        $sql_query = sprintf(
-            'ALTER TABLE %s REPAIR PARTITION %s;',
-            Util::backquote($this->table),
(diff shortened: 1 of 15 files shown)


libraries/classes/Controllers/Table/PartitionController.php
```diff
@@ -0,0 +1,28 @@
...
+    {
+        global $containerBuilder, $sql_query;
+
+        $partitionName = $_POST['partition_name'] ?? '';
+
+        if (strlen($partitionName) === 0) {
+            return;
+        }
+
+        $sql_query = 'ALTER TABLE ' . Util::backquote($this->table) . ' ANALYZE PARTITION ' . Util::backquote($partitionName);
```
williamdes, line 22: This could move to the query/generator class

templates/table/structure/display_partitions.twig
```diff
@@ -74,18 +74,84 @@
...
+                            'partition_name': partition.getName(),
+                          }) }}">
+                            {{ get_icon('b_search', 'Analyze'|trans) }}
+                          </a>
+                        </td>
+
+                        <td>
+                          <a id="partition_action_CHECK" class="ajax" href="{{ url('/table/structure/partitioning') }}" data-post="{{ get_common({
+                            'db': db,
+                            'table': table,
```
williamdes, line 91: Query built here is not the best practice :)

PR #16517, opened 2020-12-05
Add post-create-project-cmd Composer script command

MauricioFauth:
phpMyAdmin have a dedicated [Composer repository](https://github.com/phpmyadmin/composer) to allow users to install it and be ready to use.

To make it ready to use, we use the [`daily-composer` script](https://github.com/phpmyadmin/scripts/blob/master/website/daily-composer) that runs the needed install steps for the user.

To [install phpMyAdmin using Composer](https://docs.phpmyadmin.net/en/latest/setup.html#installing-using-composer), you have to run this command:
```
composer create-project phpmyadmin/phpmyadmin
```

And according to the [Composer documentation](https://getcomposer.org/doc/articles/scripts.md#command-events):
> **post-create-project-cmd**: occurs after the `create-project` command has been executed.

So we can add the needed install steps to the `post-create-project-cmd` command event instead of running them using `daily-composer` script. This could make the dedicated Composer repository obsolete.

The cons of this change is that the user will need to have Yarn installed on their machine as well.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -113,7 +113,11 @@
             "@phpunit"
         ],
         "update:baselines": "phpstan analyse --generate-baseline && psalm --set-baseline=psalm-baseline.xml",
-        "twig-lint": "php scripts/console lint:twig templates --ansi --show-deprecations"
+        "twig-lint": "php scripts/console lint:twig templates --ansi --show-deprecations",
+        "post-create-project-cmd": [
+            "yarn install --non-interactive --silent",
+            "scripts/generate-mo --quiet"
+        ]
     },
     "config":{
         "sort-packages": true


composer.json
```diff
@@ -113,7 +113,11 @@
             "@phpunit"
         ],
         "update:baselines": "phpstan analyse --generate-baseline && psalm --set-baseline=psalm-baseline.xml",
-        "twig-lint": "php scripts/console lint:twig templates --ansi --show-deprecations"
+        "twig-lint": "php scripts/console lint:twig templates --ansi --show-deprecations",
+        "post-create-project-cmd": [
+            "yarn install --non-interactive --silent",
```
williamdes, line 118:
```suggestion
            "yarn install --production --non-interactive --silent",
```

composer.json
```diff
@@ -113,7 +113,11 @@
...
+            "scripts/generate-mo --quiet"
```
williamdes, line 119: If understand well, this would not be a duplicate of https://github.com/phpmyadmin/scripts/blob/4eae099eeeb7ef3adf85ffe6e41f01cd7746c7a9/website/daily-composer#L38 ?

mauriciofauth: If we merge this pull request, we will also have to not run the equivalent commands with the daily-composer script.

williamdes: I agree to merge this PR, you can push commits to scripts afterwards

williamdes (approved):

williamdes:
Update ?
Target QA_5_1 ?

mauriciofauth:
We shouldn't add new things to 5.1. Don't have to hurry with this.

williamdes:
:ping_pong: This can be merged into 5.2 ?

MauricioFauth:
Requiring Yarn for this is a blocker for now.

PR #16521, opened 2020-12-08
Upgrade Bootstrap to version 5

MauricioFauth:
Bootstrap 5 has RTL support integrated. But removed support for Internet Explorer.

>  Dropped support for Internet Explorer 10 and 11
    Dropped support for Microsoft Edge < 16
    Dropped support for Firefox < 60
    Dropped support for Safari < 10
    Dropped support for iOS Safari < 10
    Dropped support for Chrome < 60
    Dropped support for Android < 6


package.json
```diff
@@ -13,8 +13,9 @@
     "@babel/cli": "^7.12.1",
     "@babel/core": "^7.12.3",
     "@babel/preset-env": "^7.12.1",
+    "@popperjs/core": "^2.5.4",
```
williamdes, line 16: This did not succeed to remove this?

williamdes (approved):
Could you edit the sync js vendor script if needed?

ibennetch:
I'm fine with dropping support for Internet Explorer (and the rest of this), but I wonder if we need to add a note in `doc/require.rst` at the "Web browser" section. I know IE is old, but some enterprise sites were slow to move to Edge and I think we should document the change somewhere.

ibennetch:
I might suggest holding this off until 5.2, also.

mauriciofauth:
> I might suggest holding this off until 5.2, also.

I agree.

PR #16530, opened 2020-12-13
Fix wrong logic introduced by #15823

CommanderRoot:

### Description

#15823 changed some selectors to avoid using deprecated jQuery functions. The issue here was that on some rows `document.querySelectorAll` was used in a filtered context which did not result in the expected, pre change, results. My change now reverts the code to the old expected behaviour but without using the deprecated functions.

### DIFF
--- a/js/src/server/status/monitor.js
+++ b/js/src/server/status/monitor.js
@@ -1051,10 +1051,10 @@ AJAX.registerOnload('server/status/monitor.js', function () {
         /* Calculate how much spacing there is between each chart */
         $('#chartGrid').html('<tr><td></td><td></td></tr><tr><td></td><td></td></tr>');
         chartSpacing = {
-            width: $('#chartGrid').find(document.querySelectorAll('td:nth-child(2)')).offset().left -
-                $('#chartGrid').find(document.querySelectorAll('td:nth-child(1)')).offset().left,
-            height: $('#chartGrid').find(document.querySelectorAll('td:nth-child(2) td:nth-child(2)')).offset().top -
-                $('#chartGrid').find(document.querySelectorAll('td:nth-child(1) td:nth-child(1)')).offset().top
+            width: $('#chartGrid').find('td').eq(1).offset().left -
+                $('#chartGrid').find('td').eq(0).offset().left,
+            height: $('#chartGrid').find('tr').eq(1).find('td').eq(1).offset().top -
+               $('#chartGrid').find('tr').eq(0).find('td').eq(0).offset().top
         };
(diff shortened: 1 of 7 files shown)


williamdes (approved):
This looks like a very good patch, that said how did you see it was wrong?
Is there any UI bug that I can see before and after?

CommanderRoot:
Examples where behaviour is fixed:
[https://demo.phpmyadmin.net/master-config/index.php?route=/server/status/monitor](https://demo.phpmyadmin.net/master-config/index.php?route=/server/status/monitor) (doesn't work but does with this patch)
Also when quick editing a row multiple times, currently the original query gets removed. With this patch it's alwas visible and just the `Update` query gets added below and updated.

williamdes:
Thank you, your fix is good because afterwards it works fine.
But can you help me understand what is wrong with querySelectorAll ?
I made this codepen: https://codepen.io/williamdes/pen/VwKPrmR

CommanderRoot:
`querySelectorAll` is not always bad. The issue comes when it's used when filtering already filtered results.
`querySelectorAll` scans the full page but most of the time that's not really what you want, you just want to scan the results you already have and reduce them further. Depending on the page structur this can be the same thing but that's not always the case

williamdes:
> `querySelectorAll` is not always bad. The issue comes when it's used when filtering already filtered results.
> `querySelectorAll` scans the full page but most of the time that's not really what you want, you just want to scan the results you already have and reduce them further. Depending on the page structur this can be the same thing but that's not always the case

Thank you, it is clear for me now
Do you want to fix the `js/src/functions.js` occurrences ?

Merged :rocket:

CommanderRoot:
Sure. At first I thought these wouldn't be an issue due to them being global before anyway but clearly there is a difference. The screenshots in #16531 don't lie.

js/src/server/status/monitor.js
```diff
@@ -1807,7 +1807,7 @@ AJAX.registerOnload('server/status/monitor.js', function () {
             };
 
             // We just assume the sql text is always in the second last column, and that the total count is right of it
-            $('#logTable').find(document.querySelectorAll('table tbody tr td:nth-child(' + (runtime.logDataCols.length - 1) + ')')).each(function () {
+            $('#logTable').find('table tbody tr').children('td').eq(runtime.logDataCols.length - 2).each(function () {
```
williamdes, line 1810: this seems not to work

williamdes: Fixed by 3aa80636e71601c1bc58843b20fa0cde56139702

PR #16567, opened 2021-01-13
Redesign the Routines page

MauricioFauth:
- adds a heading
- moves Create New Routine button
- moves Filter input field
- redesigns the check all line

Before:

Screenshot from 2021-01-12 23-12-30

After:

Screenshot from 2021-01-12 23-11-54


williamdes (approved):

williamdes:
🤩 💯

PR #16569, opened 2021-01-13
Added IF EXISTS clause to DROP EVENT statement when editing event.

### LINKED ISSUE #16565: Updating events should use 'DROP EVENT IF EXISTS'
Apparently, when updating an event in phpMyAdmin, it issues a DROP EVENT \`event_name\` statement before the CREATE EVENT \`event_name\`.

This can cause problems in replication in Google CloudSQL migrations as they don't migrate SQL-Events to read-replica instances.
Thereby the replication fails beyond repair as it runs into an SQL error.

It would be more appropriate if phpMyAdmin will use  DROP EVENT I**F EXISTS** \`event_name\`

roccivic:
### Description

Fixes #16565

### DIFF
--- a/libraries/classes/Database/Events.php
+++ b/libraries/classes/Database/Events.php
@@ -101,7 +101,7 @@ class Events
                         'EVENT',
                         $_POST['item_original_name']
                     );
-                    $drop_item = 'DROP EVENT '
+                    $drop_item = 'DROP EVENT IF EXISTS '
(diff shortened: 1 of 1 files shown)


roccivic:
Screenshot 2021-01-13 at 18 55 14

williamdes (approved):
LGTM, I will test your work and merge it ;)

Issue #11505, opened 2015-09-27
Notice on htmlspecialchars()


lem9:
htmlspecialchars() expects parameter 1 to be string, array given

This report is related to user submitted report [#15042](https://reports.phpmyadmin.net/reports/view/15042) on the phpmyadmin error reporting server.

PR #16581, opened 2021-01-18
Fix "PHP message: PHP Fatal error:  Uncaught TypeError: htmlspecialch…

tarjei:
…ars() expects parameter 1 to be string, bool given in /usr/share/phpmyadmin/libraries/classes/Config/FormDisplayTemplate.php:306

During login to PHPMyadmin we got the white screen of death. In our logs we found:
`
"PHP message: PHP Fatal error:  Uncaught TypeError: htmlspecialchars() expects parameter 1 to be string, bool given in /usr/share/phpmyadmin/libraries/classes/Config/FormDisplayTemplate.php:306`

The patch above solves the problem, but we are unsure if there is a better way to handle this.

Ref: https://github.com/phpmyadmin/phpmyadmin/issues/11505

### DIFF
--- a/libraries/classes/Config/FormDisplayTemplate.php
+++ b/libraries/classes/Config/FormDisplayTemplate.php
@@ -277,7 +277,7 @@ class FormDisplayTemplate
                 // As seen in the reporting server (#15042) we sometimes receive
                 // an array here. No clue about its origin nor content, so let's avoid
                 // a notice on htmlspecialchars().
-                if (! is_array($value)) {
+                if (is_string($value)) {
                     $htmlOutput .= '<input type="text" size="25" ' . $nameId
                     . $fieldClass . ' value="' . htmlspecialchars($value)
                     . '">';


ibennetch:
Thanks for looking in to this. Could you please let us k ow which phpMyAdmin and PHP versions you were using when the error occurred?

tarjei:
@ibennetch  This was using 5.0.4+dfsg2-1 on Ubuntu 18.04 w/PHP 7.4. PHPMyadmin installed from the PPA.

ibennetch:
@tarjei I'm having trouble reproducing this, can you walk me through the exact steps?

Would you be willing to try [5.1.0-rc1](https://www.phpmyadmin.net/downloads/) to see if the problem still exists in that version?

I'm trying to determine if it might have been fixed in the meantime or is specific to the Debian packaging.

tarjei:
@ibennetch hmm, thank you for quick followup. 

I think this happens when a failed login occurs on an instance with multiple hosts. I haven' had time to look into the exact reason why it happens, as I do not know the code well, but thought the fix easy enough to provide a patch - also it seemed to me that there were not sideeffects from the change.

williamdes:
Hi @tarjei 
This is fixed on 5.1 as this code confirms:
https://github.com/phpmyadmin/phpmyadmin/blob/aa9ca81b19a4e749a01703a8b550bc64e549a72c/libraries/classes/Config/FormDisplayTemplate.php#L308-L309

Let us know if you still have this issue after modifying the line manually on your installed version.

williamdes:
Hi @tarjei 
Could you let me know how you create this issue on your side ?
I can not reproduce it

MauricioFauth (approved):

mauriciofauth:
I didn't reproduce this error, but this will not occur on `master` since we use the Twig escaping instead of `htmlspecialchars` directly.

williamdes:
I made a more coherent fix, only string can work for htmlspecialchars so I simplified the if to only allowing a string.

PR #16584, opened 2021-01-20
Improve the version handling

williamdes:
This PR is to improve the Version handling and what goes with it.


scripts/create-release.sh
```diff
@@ -120,24 +120,33 @@ mark_as_release() {
...
-    CONFIG_LIB=libraries/classes/Config.php
-elif git cat-file -e $branch:libraries/Config.php 2> /dev/null ; then
-    CONFIG_LIB=libraries/Config.php
-else
-    CONFIG_LIB=libraries/Config.class.php
-fi
+VERSION_FILE=libraries/classes/Version.php
+
+fetchReleaseFromFile() {
+    php -r "require_once('libraries/classes/Version.php'); echo \PhpMyAdmin\Version::phpMyAdminVersion();"
```
mauriciofauth, line 126: I think `VERSION_SUFFIX` will never be defined here.

libraries/classes/Version.php
```diff
@@ -0,0 +1,62 @@
...
+    public static function phpMyAdminVersion(): string
+    {
+        $versionRaw = self::VERSION_MAJOR . '.' . self::VERSION_MINOR . '.' . self::VERSION_PATCH;
+
+        if (self::IS_DEV) {
+            return $versionRaw . '-dev';
+        }
+
+        if (self::IS_PRE_RELEASE) {
+            return $versionRaw . '-' . self::PRE_RELEASE_NAME . self::PRE_RELEASE_NUMBER;
```
mauriciofauth, line 53: I think we could remove `IS_DEV` and `IS_PRE_RELEASE`, then add `dev` to `PRE_RELEASE_NAME` and only concatenate `PRE_RELEASE_NUMBER` if it is different than 0. Or drop `PRE_RELEASE_NUMBER` completely and add the number to  `PRE_RELEASE_NAME`. I think this will simplify the script.

scripts/create-release.sh
williamdes: Indeed, It is only for packaging people to not do this https://salsa.debian.org/phpmyadmin-team/phpmyadmin/-/blob/debian/latest/debian/rules#L47 but change the vendor constant

libraries/classes/Version.php
```diff
@@ -0,0 +1,69 @@
...
+
+    public const VERSION_MAJOR = 5;
+    public const VERSION_MINOR = 1;
+    public const VERSION_PATCH = 0;
+
+    // The version will be {major}.{minor}.{patch}-dev
+    public const IS_DEV = true;
+
+    // The version will be {major}.{minor}.{patch}-{PRE_RELEASE_NAME}
+    public const IS_PRE_RELEASE = false;
```
mauriciofauth, line 26: I started to think that `IS_DEV` and `IS_PRE_RELEASE` will only add unnecessary steps to the Release Manager (as we can see in the `create-release.sh` script). If we need those boolean values, we can determine they based on `PRE_RELEASE_NAME` value.

libraries/classes/Version.php
```diff
@@ -0,0 +1,69 @@
...
+        if (self::IS_DEV) {
+            return $versionRaw . '-dev';
+        }
+
+        if (self::IS_PRE_RELEASE) {
+            return $versionRaw . '-' . self::PRE_RELEASE_NAME;
+        }
+
+        if (defined('VERSION_SUFFIX')) {
+            return $versionRaw . VERSION_SUFFIX;
```
mauriciofauth, line 56: Is intended to only add `VERSION_SUFFIX` on released versions?

libraries/classes/Version.php
```diff
@@ -0,0 +1,69 @@
...
+    public const VERSION_MINOR = 1;
+    public const VERSION_PATCH = 0;
+
+    // The version will be {major}.{minor}.{patch}-dev
+    public const IS_DEV = true;
+
+    // The version will be {major}.{minor}.{patch}-{PRE_RELEASE_NAME}
+    public const IS_PRE_RELEASE = false;
+
+    public const PRE_RELEASE_NAME = '';
```
mauriciofauth, line 28:
```suggestion
    public const PRE_RELEASE_NAME = 'rc1';
```

https://github.com/phpmyadmin/phpmyadmin/blob/ffbebaa8a27e646e90d658515e703da9fbb2596b/libraries/classes/Config.php#L139

libraries/classes/Version.php
williamdes: Yes, it is only for people packaging the final version ;)

libraries/classes/Version.php
williamdes: I agree to in another hand this will make any simple code detection harder.
Let me know if you want me to proceed removing `IS_PRE_RELEASE` ?

But for `IS_DEV` it is too important to be removed (in a tools detection perspective)

My point is: people will build bad script parsing a version management they do not know, booleans are explicit ;p

libraries/classes/Routing.php
```diff
@@ -25,7 +25,7 @@ public static function getDispatcher(): Dispatcher
 
         return cachedDispatcher($routes, [
             'cacheFile' => CACHE_DIR . 'routes.cache.php',
-            'cacheDisabled' => ($cfg['environment'] ?? '') === 'development',
+            'cacheDisabled' => ($cfg['environment'] ?? '') === 'development' || Version::isDev(),
```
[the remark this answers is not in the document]
williamdes: Yes, devs, contributors, testers are not setting `environment` and fall into caching issues we have to handle afterwards on the GitHub issues.
It want to short circuit the problem: if DEV = routes can change = no cache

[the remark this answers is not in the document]
mauriciofauth: We should improve the cache cleanup and warmup instead of disabling the cache for everyone. This way we can not enable cache in development environments.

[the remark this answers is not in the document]
williamdes: For sure !
See https://github.com/nikic/FastRoute/issues/224 upstream

libraries/classes/Version.php
ibennetch: I would tend to agree with @mauriciofauth about this, but @williamdes has a valid point, too. I would leave IS_DEV in, but remove IS_PRE_RELEASE and refactor that check to see whether `PRE_RELEASE_NAME` is empty instead of also having the IS_PRE_RELEASE boolean. It's pretty minor in the scheme of things, so I won't insist on changing it.

ibennetch:
🤷‍♂️

I like the refactoring in general, but this seems to me to be a little complex. Still, it's an improvement, and I don't have any better idea, so go for it!

libraries/classes/Version.php
williamdes: This will have to be filled in when releasing, for now it should be empty

williamdes:
> I like the refactoring in general, but this seems to me to be a little complex. Still, it's an improvement, and I don't have any better idea, so go for it!

Thank you for the review !, I simplified slightly the code as you 2 suggested


@mauriciofauth do you agree that I merge this PR or do you have some more comments to do on it ?

mauriciofauth:
> do you agree that I merge this PR or do you have some more comments to do on it ?

@williamdes I just thought that we can have all those constants, but instead of manually update them, we can create a `scripts/console` command to auto-generate this class. This way we can have the version strings as constants as well.

ibennetch:
I was going to rework scripts/set-versions.sh to make this a bit less of a
manual process, but you’re welcome to jump on it too.

williamdes:
That seems like a good idea, not sure how I will do the automation but that will help for sure
Will work on that in some days more or less

libraries/classes/Routing.php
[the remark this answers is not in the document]
mauriciofauth: I really don't like this change and the Template class as well. Having both sides of the Or operator doesn't make sense to me, because you should not set `$cfg['environment']` to `'development'` if you are not in a development version.

We should improve the documentation about the `$cfg['environment']` setting and improve the cache cleanup and warmup as I said before.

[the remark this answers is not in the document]
williamdes: I understand and will revert the two changes
I agree we should try to fix this warm-up issue in another better way

mauriciofauth:
> I was going to rework scripts/set-versions.sh to make this a bit less of a manual process, but you’re welcome to jump on it too.

We could migrate this script to that new command as well.

> Will work on that in some days more or less

I can help you with that if you want.

williamdes:
> I can help you with that if you want.

Sure, you can push commits there or create a new PR

After a night of sleep I will have a day of driving so meanwhile you can do what seems best to you :) 

Just asking myself how we will manage the input of giving the version to the script to be parsed and re-built into constants

ibennetch:
> We could migrate this script to that new command as well.

I'm rather ignorant about the advantages of using scripts/console over the
shell script; is there something that makes that a better choice than a
shell script?

mauriciofauth:
> I'm rather ignorant about the advantages of using scripts/console over the shell script; is there something that makes that a better choice than a shell script?

Actually, we can add the `./scripts/console set-version 5.1.0-dev` line to the `set-version.sh` script and avoid having to migrate a shell script to PHP.

mauriciofauth:
I pushed the commit that adds the class generator. I renamed some constants to remove the redundant version word and I added the ID constant as well but I don't know if we'll going to use it or not.

libraries/classes/Command/SetVersionCommand.php
```diff
@@ -0,0 +1,105 @@
...
+
+    private function writeGeneratedClassFile(string $generatedClass): bool
+    {
+        $handle = fopen(ROOT_PATH . 'libraries/classes/Version.php', 'w');
+        if ($handle === false) {
+            return false;
+        }
+        fwrite($handle, $generatedClass);
+        fclose($handle);
+
```
williamdes, line 102: why not using file_put_contents ?

williamdes:
I like the changes you made
We need to:
- Have a lot of tests on the generation
- Correctness of the file, `php -l`
- Be sure to run the test suite **after** the version file has been generated

> I renamed some constants to remove the redundant version word and I added the ID constant as well but I don't know if we'll going to use it or not.

I think an important point is not if we will use it but if someone else will use it ;)

> I'm rather ignorant about the advantages of using scripts/console over the
shell script; is there something that makes that a better choice than a
shell script?

Nothing is better than a shell script and my point for that are:
- It never dies
- Is never deprecated
- Does not require any language to be setup
- Nobody does breaking changes to it (as far as I know)
- Does not require anybody to know a higher level of programming language and its specifics behaviors

libraries/classes/Command/SetVersionCommand.php
mauriciofauth: Good idea. For this case is a better idea to use it.

scripts/create-release.sh
```diff
@@ -120,24 +120,29 @@ mark_as_release() {
...
```
mauriciofauth, line 126:
```suggestion
    php -r "require_once('libraries/classes/Version.php'); echo \PhpMyAdmin\Version::VERSION;"
```

scripts/create-release.sh
```diff
@@ -504,11 +511,20 @@ Todo now:
...
+        - in $VERSION_FILE Version class:
+            - for a dev cycle
+                - set IS_DEV to true
+                - check that PRE_RELEASE_NAME is empty
+            - for a normal release
+                - check that IS_DEV is false
+                - check that PRE_RELEASE_NAME is empty
+            - for a -rc release
+                - check that IS_DEV is false
+                - change PRE_RELEASE_NAME to "rc1"
```
mauriciofauth, line 523: This should be updated.

scripts/create-release.sh
```diff
@@ -120,24 +120,29 @@ mark_as_release() {
...
-     - in $CONFIG_LIB Config::__constructor() the line
-          " \$this->set('PMA_VERSION', '$version'); "
+     - in $VERSION_FILE Version class:
+        - check that VERSION_MAJOR, VERSION_MINOR and VERSION_PATCH are correct.
+        - for a normal release
+            - check that IS_DEV is false
+            - check that PRE_RELEASE_NAME is empty
+        - for a -rc release
+            - check that IS_DEV is false
+            - change PRE_RELEASE_NAME to "rc1"
```
mauriciofauth, line 142: This should be updated.

MauricioFauth (changes requested):

scripts/create-release.sh
williamdes: Done

MauricioFauth (approved):

williamdes:
@ibennetch The work for the version handling is ready for a last review since you are the one that uses release scripts more
Will merge this as soon as you feel all this is okay :)

williamdes:
@ibennetch can I merge this one ?

libraries/classes/Command/SetVersionCommand.php
```diff
@@ -0,0 +1,102 @@
...
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        /** @var string $version */
+        $version = $input->getArgument('version');
+
+        $generatedClass = $this->getGeneratedClass($version);
+
+        if (! $this->writeGeneratedClassFile($generatedClass)) {
+            // failure
+            return 1;
```
mauriciofauth, line 65:
```suggestion
            return Command::FAILURE;
```

libraries/classes/Command/SetVersionCommand.php
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        $output->writeln('PhpMyAdmin\Version class successfully generated!');
+
+        // success
+        return 0;
```
mauriciofauth, line 71:
```suggestion
        return Command::SUCCESS;
```

MauricioFauth (approved):

libraries/classes/Command/SetVersionCommand.php
mauriciofauth: Forget about that, it will only work on `master`.

libraries/classes/Command/SetVersionCommand.php
mauriciofauth: Forget about that, it will only work on `master`.

williamdes: Good idea, I will add them on master
See: https://github.com/symfony/console/commit/27327a59875896d8b8e577ccddc35d92d7654664
it was added in console 5.1

williamdes: 66255ffbf91b1542eca8d64fa7b3cb5154bfd878

PR #16586, opened 2021-01-21
Fix #16585 - BLOB JPG preview thumbnail

### LINKED ISSUE #16585: BLOB to JPG transformation broken
Configuring a BLOB column to have browser transformation set to Inline JPG and clicking on the (also broken) thumbnail will result in the following warning:

```
Fatal error: Uncaught TypeError: Argument 1 passed to PhpMyAdmin\Core::downloadHeader() must be of the type string, null given, called in /var/www/html/transformation_wrapper.php on line 131 and defined in /var/www/html/libraries/classes/Core.php:631 Stack trace:
#0 /var/www/html/transformation_wrapper.php(131): PhpMyAdmin\Core::downloadHeader(NULL, 'Image/JPEG')
#1 {main} thrown in /var/www/html/libraries/classes/Core.php on line 631
```
Open new phpMyAdmin window
```
Notice in ./transformation_wrapper.php#131
 Undefined variable: cn
```

transformation_wrapper.php is expecting "cn" to be passed on the query string, which is not the case at the moment. Manually appending &cn= fixes the issue.

joltcoke:
Thumbnails are broken because imagecreatetruecolor() is expecting int but the params are either strings (from query string) or float. This PR fixes both issues.

Addresses the broken thumbnail mentioned in #16585 

Fixes: #16585

### DIFF
--- a/libraries/classes/Controllers/TransformationWrapperController.php
+++ b/libraries/classes/Controllers/TransformationWrapperController.php
@@ -168,10 +168,10 @@ class TransformationWrapperController extends AbstractController
                 . ($mime_options['charset'] ?? '');
         }
 
-        Core::downloadHeader($cn, $mime_type);
+        Core::downloadHeader($cn ?? '', $mime_type ?? '');
 
         if (! isset($_REQUEST['resize'])) {
-            if (stripos($mime_type, 'html') === false) {
+            if (stripos($mime_type ?? '', 'html') === false) {
                 echo $row[$transform_key];
             } else {
                 echo htmlspecialchars($row[$transform_key]);
--- a/libraries/classes/Controllers/TransformationWrapperController.php
+++ b/libraries/classes/Controllers/TransformationWrapperController.php
@@ -24,6 +24,7 @@ use function imagesx;
 use function imagesy;
 use function in_array;
 use function intval;
+use function round;
 use function str_replace;
 use function stripos;
 use function substr;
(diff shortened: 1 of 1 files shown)


williamdes:
Hi @joltcoke 
Could you rebase onto QA_5_1 ?
And change the PR target using the edit button ?

joltcoke:
Hi @williamdes 

i've rebased to QA_5_1 and converted the casts into intval() conversions to be consistent with the other code in the same file. This also fixed the linting issues.

williamdes:
I tested your fix and it works great !
But I added one more commit because the `cn` parameter seems not to be sent and it did throw and error.

libraries/classes/Controllers/TransformationWrapperController.php
```diff
@@ -194,11 +195,11 @@ public function index(): void
             $ratioHeight = $srcHeight / $_REQUEST['newHeight'];
 
             if ($ratioWidth < $ratioHeight) {
-                $destWidth = $srcWidth / $ratioHeight;
-                $destHeight = $_REQUEST['newHeight'];
+                $destWidth = intval(round($srcWidth / $ratioHeight));
```
williamdes, line 198: Let me know if https://stackoverflow.com/a/58712745/5155484 could be an issue here also

MauricioFauth (approved):

PR #16587, opened 2021-01-21
Added pydata-sphinx-theme in conf.py

SavvyOsive:
Added **pydata-sphinx-theme** for the rst documentation of phpmyadmin

Whenever someone goes to any phpmyadmin's documentation page, it's difficult to track "**Table of Content**" for that page 

It will be great to introduce "pydata-sphinx-theme" which offers **"On this page" feature**  in the right side of the page which will be helpful for anyone traversing the documentation and finding or looking for the features of phpmyadmin 

Reference Link for theme : https://github.com/pandas-dev/pydata-sphinx-theme

Screenshot from 2021-01-21 19-55-58

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -98,7 +98,7 @@ pygments_style = 'sphinx'
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
-html_theme = 'default'
+html_theme = 'pydata_sphinx_theme'
 
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the


ibennetch:
First, I should make sure you're talking about the local documentation, not what's hosted by ReadTheDocs at https://docs.phpmyadmin.net/en/latest/ right? Seems we're on the same page but I wanted to make sure.

I personally think the pydata-sphinx-theme wastes a lot of space (especially in the header), but agree that it's an improvement over what we currently have. One slight negative is that it introduces additional build requirements, but I don't consider that a blocker.

I'm fine with this, how do any other developers feel about it?

williamdes:
> I'm fine with this, how do any other developers feel about it?

Having a better theme is cool 

> One slight negative is that it introduces additional build requirements, but I don't consider that a blocker.

This is bad and can be a blocker for packaging people
Having an option to set back the theme to default would be great. And ENV ?

mauriciofauth:
I'm fine with this as well.

> Having an option to set back the theme to default would be great. And ENV ?

How can we make this optional?

williamdes:
Closing this one since no feedback was given

PR #16590, opened 2021-01-24
Make the console setup async

williamdes:
Make the console setup async to avoid hanging the page waiting for the config to be received

The console can wait.

### DIFF
--- a/js/src/console.js
+++ b/js/src/console.js
@@ -61,8 +61,16 @@ var Console = {
             return;
         }
 
-        Console.config = Functions.configGet('Console', false);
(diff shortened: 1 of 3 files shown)


MauricioFauth (approved):
Since it's not fixing a bug, I think it should go to the master branch.

williamdes:
> Since it's not fixing a bug, I think it should go to the master branch.

It is fixing slowness for our users and this was reported many times, still okay that I merge this into 5_1 ?

mauriciofauth:
> It is fixing slowness for our users and this was reported many times, still okay that I merge this into 5_1 ?

Go ahead.

PR #16593, opened 2021-01-25
Major performance improvements on browsing rows

williamdes:
I made some major improvements to the performance that browsing rows had.

Using: xhprof

Ref: #16005

Explained:
- The first change is to just move a variable to avoid it to be re-built
- The second one is splitting in 3 functions the description array to avoid wasting time fetching translations that will not be used (99% are not used, only one is picked up) (yes I built a script to check all the keys and values exist and are the same than before, diff should be okay)

Dataset and method:
- A table of more that 1.1 Gio of data
- 7 973 078 rows
- show rows by 500 on browse page

### Before (14 229ms)
image

#### Call refresh button (14 981ms)
image

### After (7 524 ms)
image

#### Call refresh button (6 673 ms)
image

### DIFF
--- a/libraries/classes/Config/Descriptions.php
+++ b/libraries/classes/Config/Descriptions.php
@@ -13,13 +13,11 @@ use function sprintf;
 use function str_replace;
 
 /**
- * Base class for forms, loads default configuration options, checks allowed
- * values etc.
+ * Class to get name or description for a configuration paths
  */
 class Descriptions
 {
     /**
-     * Return
      * Return name or description for a configuration path.
      *
      * @param string $path Path of configuration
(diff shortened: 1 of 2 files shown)


libraries/classes/Config/Descriptions.php
```diff
@@ -54,666 +52,337 @@ public static function get($path, $type = 'name')
...
-    /**
-     * Return name or description for a cleaned up configuration path.
-     *
-     * @param string $path Path of configuration
-     * @param string $type Type of message, either 'name', 'cmt' or 'desc'
-     *
-     * @return string|null Null if not found
-     */
-    public static function getString($path, $type = 'name')
+    private static function getCmt(): array
```
mauriciofauth, line 55:
```suggestion
    private static function getComments(): array
```

libraries/classes/Config/Descriptions.php
```diff
@@ -997,6 +995,27 @@ public static function getString($path, $type = 'name')
...
+     * @param string $type Type of message, either 'name', 'cmt' or 'desc'
+     *
+     * @return string|null Null if not found
+     */
+    public static function getString($path, $type = 'name'): ?string
+    {
+        $descriptions = [];
+
+        if ($type === 'cmt') {
+            $descriptions = self::getCmt();
```
mauriciofauth, line 1013:
```suggestion
            $descriptions = self::getComments();
```

MauricioFauth (approved):
Really nice. :1st_place_medal:

ibennetch:
I'm very impressed with the speed improvements you've made here. Nice work!

ibennetch:
I'd like to get this in 5.1 before I release 5.1.0rc2.

williamdes:
> I'm very impressed with the speed improvements you've made here. Nice work!

Thank you @ibennetch !

> Really nice. 1st_place_medal:

Thank you @mauriciofauth !

> I'd like to get this in 5.1 before I release 5.1.0rc2.

I will merge that right now, could you please review #16584 ?


NB: some discussion is going on at https://github.com/code-lts/welcome/issues/6 about xhprof, I intend to make it usable for other people so our users can send us xhprof files

Issue #12392, opened 2016-07-18
Replace TCPDF with tc-lib-pdf 


nijel:
Once tc-lib-pdf is stable, we should switch to it as TCPDF will not receive further maintenance.

See https://tcpdf.org/ and https://github.com/tecnickcom/tc-lib-pdf for details.

udan11:
The new library does not look like it will be ready any time soon (or at all). Should we consider alternatives?

nijel:
I think we can still stick with tcpdf without any problems, it still seem to receive some maintenance, see https://github.com/tecnickcom/tcpdf

williamdes:
Versions have been released of https://github.com/tecnickcom/tcpdf

williamdes:
Or by https://github.com/mpdf/mpdf

williamdes:
Most probably with https://github.com/code-lts/WarnockPDF the fork of TCPDF

williamdes:
See: https://github.com/tecnickcom/TCPDF/issues/867

Repo: https://github.com/tecnickcom/tc-lib-pdf

@MauricioFauth I think it is time that we move to the new library ?

MauricioFauth:
> I think it is time that we move to the new library ?

For sure!

PR #16596, opened 2021-01-26
Fix #16429 - Use php8.0 fixed version for TCPDF

### LINKED ISSUE #16429: PHP 8 compatibility
When using PHP 8.0.0rc1 provided by the Sury repository, I get the following message when running `composer update`:

```  Problem 1
    - php-webdriver/webdriver 1.8.3 requires php ^5.6 || ~7.0 -> your PHP version (8.0.0rc1) does not satisfy that requirement.
    - php-webdriver/webdriver 1.8.2 requires php ^5.6 || ~7.0 -> your PHP version (8.0.0rc1) does not satisfy that requirement.
    - php-webdriver/webdriver 1.8.1 requires php ^5.6 || ~7.0 -> your PHP version (8.0.0rc1) does not satisfy that requirement.
    - php-webdriver/webdriver 1.8.0 requires php ^5.6 || ~7.0 -> your PHP version (8.0.0rc1) does not satisfy that requirement.
    - php-webdriver/webdriver 1.8.3 requires php ^5.6 || ~7.0 -> your PHP version (8.0.0rc1) does not satisfy that requirement.
    - Installation request for php-webdriver/webdriver ^1.8 -> satisfiable by php-webdriver/webdriver[1.8.0, 1.8.1, 1.8.2, 1.8.3].
```

williamdes:
Fix #16429

Use php8.0 fixed version for TCPDF

I did use a fixed commit because upstream did not release the recent changes

See full diff: https://github.com/tecnickcom/TCPDF/compare/6.3.5...456b794f1fae9aee5c151a1ee515aae2aaa619a3

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -92,7 +92,7 @@
         "symfony/console": "^4.4",
         "symfony/finder": "^4.4",
         "symfony/twig-bridge": "^4.4",
-        "tecnickcom/tcpdf": "^6.3",
+        "tecnickcom/tcpdf": "dev-main#456b794f1fae9aee5c151a1ee515aae2aaa619a3",
         "vimeo/psalm": "^4.4.0"
     },
     "extra": {
--- a/composer.json
+++ b/composer.json
@@ -116,6 +116,7 @@
         "twig-lint": "php scripts/console lint:twig templates --ansi --show-deprecations"
     },
     "config":{
-        "sort-packages": true
+        "sort-packages": true,
+        "discard-changes": true
     }
 }
--- a/scripts/create-release.sh
+++ b/scripts/create-release.sh
@@ -243,7 +243,7 @@ if [ ! -d libraries/tcpdf ] ; then
     cp composer.json composer.json.backup
     echo "* Running composer"
     composer config platform.php "$PHP_REQ"
-    composer update --no-dev --optimize-autoloader
+    composer update --no-interaction --no-dev --optimize-autoloader
 
     # Parse the required versions from composer.json
     PACKAGES_VERSIONS=''
@@ -256,7 +256,7 @@ if [ ! -d libraries/tcpdf ] ; then
     do
         PACKAGES_VERSIONS="$PACKAGES_VERSIONS $PACKAGES:`awk "/require-dev/ {printline = 1; print; next } printline" composer.json | grep "$PACKAGES" | awk -F [\\"] '{print $4}'`"
     done
-    composer require --optimize-autoloader --update-no-dev $PACKAGES_VERSIONS
+    composer require --no-interaction --optimize-autoloader --update-no-dev $PACKAGES_VERSIONS
 
     mv composer.json.backup composer.json
     echo "* Cleanup of composer packages"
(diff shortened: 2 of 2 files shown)


williamdes:
Only the release script does fail.
PHP 8.0 CI does not have any PHP 8.0 deprecation errors: https://github.com/phpmyadmin/phpmyadmin/pull/16596/checks?check_run_id=1772586211#step:11:10

williamdes (commented):
Thank you @ibennetch for the release fix!

ibennetch:
I pushed two commits to fix the build that can be squashed together before merging.

ibennetch:
Correct me if I'm wrong, but this fixes deprecation warnings that are shown by the testsuite but aren't normally visible to the end user, right? I'm not certain if it's better to continue to allow those warnings for now or merge this fix to pin us to a specific commit.

Just to bring it up, another alternative idea is to fork tcpdf and keep our own maintenance of it, at least until tc-lib-pdf is ready. I'm not terribly fond of that yet, because commits are still getting merged upstream so I'd rather leave well enough alone for now.

For now I do think this is the best solution.

ibennetch (approved):
We should squash my two commits together before merging this, otherwise LGTM.

williamdes:
> Correct me if I'm wrong, but this fixes deprecation warnings that are shown by the testsuite but aren't normally visible to the end user, right? I'm not certain if it's better to continue to allow those warnings for now or merge this fix to pin us to a specific commit.

Nope, as you can see the errors that where visible on the test suite are the same our users reported us seeing on their import/export (not sure witch one) page.

And for the fork idea I agree and it could be done at @code-lts

MauricioFauth (approved):

ibennetch:
If you want to host it at code-lts that's fine, otherwise I think we can wait a little longer to see what happens with the official repository. We should probably re-evaluate where this is when we release 5.2.0; see https://github.com/phpmyadmin/phpmyadmin/issues/16601

williamdes:
> If you want to host it at code-lts that's fine, otherwise I think we can wait a little longer to see what happens with the official repository. We should probably re-evaluate where this is when we release 5.2.0; see #16601

Sure, that would also start a new period where we could close https://github.com/phpmyadmin/phpmyadmin/issues/12392 and have @code-lts make tcpdf clean and maintained.

xPaw:
Supplying this package from git made the archives balloon in size because the entire `vendor\tecnickcom\tcpdf\.git\` folder is included which is almost 70mb.

Look at the archive sizes for phpMyAdmin 5.1.0-rc2 (and snapshots): https://www.phpmyadmin.net/downloads/

williamdes:
> Supplying this package from git made the archives balloon in size because the entire `vendor\tecnickcom\tcpdf\.git\` folder is included which is almost 70mb.
> 
> Look at the archive sizes for phpMyAdmin 5.1.0-rc2 (and snapshots): https://www.phpmyadmin.net/downloads/

Oh thank you !
Will push a commit for that

ibennetch:
Thanks @xPaw, you posted this just in time to get me to pause releasing 5.1.0 in order to resolve this first. Thanks for pointing this out.

williamdes:
rc2 also has coding standard and phpcs vendor libs, this is bad
Trying to fix it

williamdes:
> rc2 also has coding standard and phpcs & phpstan vendor libs, this is bad
> Trying to fix it

Strange that it does not do that effect when building on my workstation
Working on checking the bundle and the routes warmup

PR #16600, opened 2021-01-28
Create Twig\AssetExtension and image Twig function

MauricioFauth:
Removes the need to pass `$PMA_Theme->getImgPath()` to templates.

### DIFF
--- a/libraries/classes/Twig/AssetExtension.php
+++ b/libraries/classes/Twig/AssetExtension.php
@@ -1,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpMyAdmin\Twig;
+
+use PhpMyAdmin\Theme;
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFunction;
+
+final class AssetExtension extends AbstractExtension
+{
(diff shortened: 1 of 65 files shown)


libraries/classes/Twig/AssetExtension.php
```diff
@@ -0,0 +1,33 @@
...
+
+final class AssetExtension extends AbstractExtension
+{
+    /**
+     * @return TwigFunction[]
+     */
+    public function getFunctions()
+    {
+        return [
+            new TwigFunction('image', [$this, 'getImagePath']),
```
williamdes, line 19: theme_image could be easier to maintain and search for?

williamdes (approved):
Very good work

libraries/classes/Twig/AssetExtension.php
mauriciofauth: As all images are from the themes, I thought that `image` is simpler. I thought about using `image_asset` as well.

williamdes: I am convinced, you can merge when you want :)

PR #16616, opened 2021-02-03
Replace themes preview popup with a modal

MauricioFauth:
When clicking the Theme link in the home page, it opens a modal instead of a popup.


templates/home/index.twig
```diff
@@ -241,4 +239,44 @@
...
+                </div>
+              {% endfor %}
+            </div>
+          </form>
+        </div>
+        <div class="modal-footer">
+          <button type="button" class="btn btn-secondary" data-bs-dismiss="modal">{% trans 'Close' %}</button>
+          <a href="{{ 'https://www.phpmyadmin.net/themes/'|link }}#pma_{{ phpmyadmin_major_version|replace({'.': '_'}) }}" class="btn btn-primary" rel="noopener noreferrer" target="_blank">
+            {% trans 'Get more themes!' %}
+          </a>
```
williamdes, line 275: I would invert them to have the close button as usual

williamdes (approved):
Very good work !
First of all, I discovered a new feature in the UI !
Second, it is better like you designed it :100: 

3 things I would ask:
- loading the screenshots will use more bandwidth on each homepage load, can we delay the loading ?
- can the picture be larger ?
- can the themes have the same height on a row ?

image

templates/home/index.twig
mauriciofauth: I just followed the Bootstrap examples where all primary buttons are at right.

mauriciofauth:
> * loading the screenshots will use more bandwidth on each homepage load, can we delay the loading ?

I tried adding `loading="lazy"` but it didn't seen to work in this case. We could have an empty modal and load the content only when requested.

> * can the picture be larger ?

We could have three themes on each row, but then it will have two rows.

> * can the themes have the same height on a row ?

Yes, but doesn't seen to work very well with the image on the middle. I will try other design.

williamdes:
> We could have an empty modal and load the content only when requested.

Or just have lazy img src attribute and replace the lazy src attribute name when the modal opens

> We could have three themes on each row, but then it will have two rows.

Two per row could be great too

> Yes, but doesn't seen to work very well with the image on the middle. I will try other design.

Okay, could you post a screenshot when you update the code ?

williamdes (approved):
Looks perfect now
Will test it soon

williamdes (approved):
:1st_place_medal: I tested it and it is very good, you can merge when you want

PR #16623, opened 2021-02-06
Fix link HTML messed up in Advisor

### LINKED ISSUE #16621: Link HTML messed up in Advisor
Where it says "this blog entry", it appears it should be 2 links instead of 1, and the HTML for the second link is being seen as a whole link (possibly a problem with escaping):

image

Version 5.0.4

MauricioFauth:
Fixes #16621


williamdes (approved):

williamdes:
Could we have a new test to cover that ?

PR #16626, opened 2021-02-08
Fix 16506 Hide_db and Only_db does not work

### LINKED ISSUE #16506: hide_db seems to not work
The hide_db configuration directive seems to have no effect.

I have set `$cfg['Servers'][$i]['hide_db'] = 'sakila';` and yet I still see the sakila database listed in the navigation pane. I've also tried ``$cfg['Servers'][$i]['hide_db'] = '^(sakila|gis)$';`` again with no luck.

I've tested with MariaDB 10.3 and MySQL 8.0, and phpMyAdmin 4.9.1 and 5.1.0-dev from master.

shucon:

### Description

Fixes #16506

### DIFF
--- a/libraries/classes/UserPreferences.php
+++ b/libraries/classes/UserPreferences.php
@@ -118,6 +118,12 @@ class UserPreferences
         $cfgRelation = $this->relation->getRelationsParam();
         $server = $GLOBALS['server'] ?? $GLOBALS['cfg']['ServerDefault'];
         $cache_key = 'server_' . $server;
+        foreach ($config_array as $key => $value) {
+            if ($value != '') {
+                continue;
+            }
+            unset($config_array[$key]);
+        }
         if (! $cfgRelation['userconfigwork']) {
             // no pmadb table, use session storage
             $_SESSION['userconfig'] = [


libraries/classes/UserPreferences.php
```diff
@@ -118,6 +118,11 @@ public function save(array $config_array)
         $cfgRelation = $this->relation->getRelationsParam();
         $server = $GLOBALS['server'] ?? $GLOBALS['cfg']['ServerDefault'];
         $cache_key = 'server_' . $server;
+        foreach ($config_array as $key => $value) {
+            if ($value == "") {
```
shucon, line 122: We could also add a condition where `$key` can only be `hide_db` or `only_db` but I think it's not required.

williamdes (changes requested):
This will maybe have side effects on users wanting to remove the value set by the config

MauricioFauth (changes requested):
This is not the correct solution. The issue is that after changing a setting through the Settings page, the value in the `config.inc.php` is ignored.

shucon:
What do you propose? What should be the desired behaviour when there's a value in both?

PR #16632, opened 2021-02-10
Fix #14026 - Fix error messages and conditions for MD5 and AES_* functions

### LINKED ISSUE #14026: AES DECRYPT and ENCRYPT in edit view not working
### Steps to reproduce
1. take any table and go to into a record edit view and choose function AES_DECRYPT, this will result in an error. right of the function.

### Expected behaviour
no error

### Actual behaviour
having an error

### Server configuration
Ubuntu 16.04 and Windows too.

Database server

    Server: Localhost via UNIX socket
    Server type: MySQL
    Server version: 5.7.21-0ubuntu0.16.04.1 - (Ubuntu)
    Protocol version: 10
    User: root@localhost
    Server charset: UTF-8 Unicode (utf8)

Web server

    Apache
    Database client version: libmysql - mysqlnd 5.0.12-dev - 20150407 - $Id: b5c5906d452ec590732a93b051f3827e02749b83 $
    PHP extension: mysqliDocumentation curlDocumentation mbstringDocumentation
    PHP version: 7.0.25-0ubuntu0.16.04.1

phpMyAdmin

    Version information: 4.7.4

### Client configuration
Firefox  58.0.2

**Operating system:**
Ubuntu 16.04

williamdes:
Fixes #14026

### DIFF
--- a/js/src/table/change.js
+++ b/js/src/table/change.js
@@ -270,11 +270,24 @@ function verificationsAfterFieldChange (urlField, multiEdit, theType) {
         $('#salt_' + target.id).remove();
     }
 
-    if (target.value === 'AES_DECRYPT'
-            || target.value === 'AES_ENCRYPT'
-            || target.value === 'MD5') {
+    // Remove possible blocking rules if the user changed functions
+    $('#' + target.id).rules('remove', 'validationFunctionForMd5');
+    $('#' + target.id).rules('remove', 'validationFunctionForAesDesEncrypt');
+
+    if (target.value === 'MD5') {
         $('#' + target.id).rules('add', {
-            validationFunctionForFuns: {
+            validationFunctionForMd5: {
+                param: $thisInput,
+                depends: function () {
+                    return checkForCheckbox(multiEdit);
+                }
+            }
+        });
+    }
+
+    if (target.value === 'DES_ENCRYPT' || target.value === 'AES_ENCRYPT') {
+        $('#' + target.id).rules('add', {
+            validationFunctionForAesDesEncrypt: {
                 param: $thisInput,
                 depends: function () {
                     return checkForCheckbox(multiEdit);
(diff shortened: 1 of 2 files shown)


williamdes:
@ibennetch I would like to ship this one into 5.1 because the feature is kinda "broken" since ever for our users.
Would you mind leaving a review please ?

ibennetch:
Could you explain this a bit? The results I'm seeing don't match my expectations from reading the code, which probably means my jQuery is rusty, ha ha.

The original report talks about editing a field, but I don't see a difference in the interface between editing an existing record or inserting a new one, so for the sake of discussion I refer to it as the insert operation.

In the current code, when I try to insert to a varchar or char column using AES_ENCRYPT(), phpMyAdmin says "Error" which is not helpful at all, but at least gives a hint that something isn't right. With this pull request, phpMyAdmin doesn't say anything and lets me submit the insert, but after submitting the query, MariaDB and MySQL both throw errors like `#1366 - Incorrect string value: '\xA1Z\xCE\xFB\xEF\x02...' for column 'charreg' at row 1`. I'd rather see us change "Error" to a message like "These functions are meant to return a binary result; to avoid inconsistent results you should store it in a BINARY, VARBINARY, or BLOB column."

Does that make sense or am I looking at the wrong thing here?

williamdes:
> I'd rather see us change "Error" to a message like "These functions are meant to return a binary result; to avoid inconsistent results you should store it in a BINARY, VARBINARY, or BLOB column."
Does that make sense or am I looking at the wrong thing here?

I think you are right, but I did not have this error while trying to store a value into a varchar field. And could decrypt it perfectly.
Let me know what table structure you did use so I can adjust what to allow the user to use :)

And this PR is for:
- Make the message clear to indicate it is a type selection error
- Relax the restrictions to allow types that "work" but are not very made for it
- Also fix and split out the MD5 validation logic message

ibennetch:
> Make the message clear to indicate it is a type selection error

I never see your new error/warning message.

> Relax the restrictions to allow types that "work" but are not very made for it

Seems reasonable, we should warn but not restrict the user.

> Also fix and split out the MD5 validation logic message

Yes, quite good.

ibennetch:
This is my table structure, so I can test inserting in to a CHAR, CHAR with binary option, or VARBINARY all from the same insert page.

```
CREATE TABLE `aes` (
  `id` int(11) NOT NULL,
  `charreg` varchar(255) NOT NULL,
  `charbin` varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL,
  `vbin` varbinary(255) NOT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;


ALTER TABLE `aes`
  ADD PRIMARY KEY (`id`);

ALTER TABLE `aes`
  MODIFY `id` int(11) NOT NULL AUTO_INCREMENT;
COMMIT;
```

williamdes:
> Seems reasonable, we should warn but not restrict the user.

I agree, but I have no idea if it is already possible in our JS code

> I never see your new error/warning message.

Try to encrypt a value when inserting a row, you will need to use a timestamp field for example

williamdes:
@sudo-robot deploy

williamdes:
@ibennetch here is the message :)
You can have fun on all the deployed servers, root has the same password

image

williamdes:
I updated the message and am merging this, we should warn and not block the user so I will not shorten the possible columns where the user can use the functions

PR #16633, opened 2021-02-10
Fix #16614 - Correct the dataseek offset on MySQLI extension

### LINKED ISSUE #16614: Argument #1 ($offset) must be greater than or equal to 0
5.1.0-rc1
PHP 8.0.1

When trying to browse an empty table:

```
PHP message: PHP Fatal error:  Uncaught ValueError: mysqli_result::data_seek(): Argument #1 ($offset) must be greater than or equal to 0 in /srv/phpMyAdmin-5.1+snapshot-english/libraries/classes/Dbal/DbiMysqli.php:308
Stack trace:
#0 /srv/phpMyAdmin-5.1+snapshot-english/libraries/classes/Dbal/DbiMysqli.php(308): mysqli_result->data_seek()
#1 /srv/phpMyAdmin-5.1+snapshot-english/libraries/classes/DatabaseInterface.php(2036): PhpMyAdmin\Dbal\DbiMysqli->dataSeek()
#2 /srv/phpMyAdmin-5.1+snapshot-english/libraries/classes/Display/Results.php(4339): PhpMyAdmin\DatabaseInterface->dataSeek()
#3 /srv/phpMyAdmin-5.1+snapshot-english/libraries/classes/Display/Results.php(4070): PhpMyAdmin\Display\Results->getSortedColumnMessage()
#4 /srv/phpMyAdmin-5.1+snapshot-english/libraries/classes/Sql.php(1356): PhpMyAdmin\Display\Results->getTable()
#5 /srv/phpMyAdmin-5.1+snapshot-english/libraries/classes/Sql.php(1157): PhpMyAdmin\Sql->getHtmlForSqlQueryResultsTable()
#6 /srv/phpMyAdmin-5.1+snapshot-english/librarie
```

williamdes:
Fixes #16614

### DIFF
--- a/libraries/classes/Display/Results.php
+++ b/libraries/classes/Display/Results.php
@@ -4336,7 +4336,10 @@ class Results
         );
 
         // fetch last row of the result set
-        $dbi->dataSeek($dt_result, $this->properties['num_rows'] - 1);
+        $dbi->dataSeek(
+            $dt_result,
+            $this->properties['num_rows'] > 0 ? $this->properties['num_rows'] - 1 : 0
+        );
         $row = $dbi->fetchRow($dt_result);
 
         // check for non printable sorted row data
@@ -4566,7 +4569,10 @@ class Results
         }
 
         // fetch last row of the result set
-        $dbi->dataSeek($dt_result, $this->properties['num_rows'] - 1);
+        $dbi->dataSeek(
+            $dt_result,
+            $this->properties['num_rows'] > 0 ? $this->properties['num_rows'] - 1 : 0
+        );
         $row = $dbi->fetchRow($dt_result);
 
         // @see DbiMysqi::fetchRow & DatabaseInterface::fetchRow


OlafvdSpek:
It masks the real problem, I wouldn't do it.

williamdes:
@mauriciofauth what are your thoughts on this PHP 8 fix ?

ibennetch:
I'm also quite curious why $offset would be less than 0.

On the surface this looks acceptable to me, but I'd really like to know more about what's gone wrong elsewhere in the code to call this.

williamdes:
> I'm also quite curious why $offset would be less than 0.
> 
> On the surface this looks acceptable to me, but I'd really like to know more about what's gone wrong elsewhere in the code to call this.

I did not check but we have operations like ` $this->properties['num_rows'] - 1` and if rows = 0 then 0-1 = -1

ibennetch:
Yeah, that makes sense. I think I'm okay with this. What one person considers masking the problem, another can consider "dealing with it in another code block" and it seems that any time we have a negative offset, it's meant to be zero anyway. I'm also curious what @mauriciofauth has to say but I would be okay with this.

williamdes:
@sudo-robot deploy

OlafvdSpek:
> Yeah, that makes sense. I think I'm okay with this. What one person considers masking the problem, another can consider "dealing with it in another code block" and it seems that any time we have a negative offset, it's meant to be zero anyway. I'm also curious what @mauriciofauth has to say but I would be okay with this.

Suppose I call dataSeek(res, -2). Without this PR, the bug would be caught and I could fix it. With this PR, the bug wouldn't be caught and I can't be sure I really meant dataSeek(res, 0). Maye I meant dataSeek(res, 2)..

libraries/classes/Dbal/DbiMysqli.php
```diff
@@ -305,7 +305,7 @@ public function fetchRow($result): ?array
      */
     public function dataSeek($result, $offset)
     {
-        return $result->data_seek($offset);
+        return $result->data_seek($offset < 0 ? 0 : $offset);
```
mauriciofauth, line 308: A better solution is to check for this earlier. For example, we can check if `$this->properties['num_rows'] > 0`.
```php
$dbi->dataSeek($dt_result, $this->properties['num_rows'] - 1);
```

williamdes: Indeed, I did that now

MauricioFauth (approved):

ibennetch:
👍

Issue #16164, opened 2020-06-05
[Severe] This bug did make me send a partial failed import asked by someone..


williamdes:
### Describe the bug

`#1242 - Subquery returns more than 1 row`

### To Reproduce

Have a table with many rows, make a query that can have for some reason more than one row in return.

Export the query and see a partial export with no errors.

Pagination allowed not to see the mistake in the query, because on page 1 no subquery error.
On page X out of 34735 rows there is an error.

Actual behavior exports the rows up to the failure and fails silently..

### Expected behavior

Export all lines or **abort the export**

### Server configuration

- phpMyAdmin version: 4.9, 5.0, 5.1

rajat315315:
How to export the query??

williamdes:
> How to export the query??

I will try to post an example, ping me back if I forget before monday

rajat315315:
Duplicate of #15317 , I guess.

williamdes:
> Possible duplicate of #15317.

Thank you, I think you are right 
Do you want to work on this one?

rajat315315:
Sure. Please assign to me.

rajat315315:
https://github.com/phpmyadmin/phpmyadmin/blob/b08b36e39a804d597e1ff34c7e2d2e1ac1918e44/libraries/classes/Dbi/DbiMysqli.php#L21
restricts MySQL to not report `#1242 - Subquery returns more than 1 row` this particular error in `QA_4_9` branch.
It needs to be changed to `mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT);` according to `https://www.php.net/manual/en/class.mysqli-sql-exception.php` User_contributed_notes.. but that would introduce further errors that we need to handle.

On a second note, from QA_5_0, we have moved to Symfony for DB related queries.

williamdes:
Thank you @rajat315315 for this precious details and debug !
@mauriciofauth what would you recommend ?

shucon:
🏓

PR #16636, opened 2021-02-11
Set MySQLi error reporting mode

### LINKED ISSUE #16589: MySQLi error reporting mode
Considering a recent RFC proposal I have made to PHP internals to [change the default mysqli error reporting mode](https://wiki.php.net/rfc/mysqli_default_errmode) I would like to ask phpMyAdmin project whether you are considering using the exception error reporting mode or continue using the silenced mode. I know that automatic error reporting has many advantages and can make the code simpler and cleaner, but I also understand that phpMyAdmin is a very specific project in that regards. phpMyAdmin needs to catch the mysqli errors and report them to the user, which is something that is not usually done in any other project. 

I can see that you are not setting the mysqli error reporting mode currently either way. When the default setting changes in PHP 8.1 you will have to make a decision whether to continue with silenced error reporting or enable exception mode. The setting should be maintained either way when using mysqli. If you decide against refactor and you would like to continue with the disabled error reporting then you can already include the following line of code right before `new mysqli()`:

    mysqli_report(MYSQLI_REPORT_OFF);

If you want to use automatic error reporting (proposed new default) then you can use this line instead:

    mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT);

Regardless of whether the RFC gets accepted, I believe that you should make it explicit in the project which mode you are using. I don't foresee any issues when the default setting changes as long as phpMyAdmin doesn't rely on the default value. I hope that with advance notice the project can make the right decision.

MauricioFauth:
Explicitly sets the MySQLi error reporting mode to the current default to avoid possibly future changes.

Fixes https://github.com/phpmyadmin/phpmyadmin/issues/16589

### DIFF
--- a/libraries/classes/Config/Validator.php
+++ b/libraries/classes/Config/Validator.php
@@ -9,9 +9,11 @@ namespace PhpMyAdmin\Config;
 
 use PhpMyAdmin\Core;
 use PhpMyAdmin\Util;
+use function mysqli_report;
 use const FILTER_FLAG_IPV4;
 use const FILTER_FLAG_IPV6;
 use const FILTER_VALIDATE_IP;
+use const MYSQLI_REPORT_OFF;
 use const PHP_INT_MAX;
 use function array_map;
 use function array_merge;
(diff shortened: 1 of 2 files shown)


williamdes (approved):
This change seems okay
But would you have a look to this issue and comment, maybe it will be fixed or can be fixed in this PR too

See: https://github.com/phpmyadmin/phpmyadmin/issues/16164#issuecomment-667686505

mauriciofauth:
> But would you have a look to this issue and comment, maybe it will be fixed or can be fixed in this PR too

It is better to change the report mode on the master branch.

PR #16642, opened 2021-02-13
Disable cache when the file is not writable

### LINKED ISSUE #16055: Caching requires webserver-user permissions on cache directory
Related to #16028, the caching requires that the webserver process has write access to the `libraries/cache` folder. By default, we don't do that (and can't know the local webserver user/group name).

With `tmp/`, this is solved/worked around by creating the tmp folder from within phpMyAdmin; in the case of caching the folder currently exists in the file system. Creating the folder externally (such as through yarn) probably will not give us sufficient information to give proper permissions, either.

My suggested solution is to find a way to create the `cache` folder directly from within phpMyAdmin before calling the caching library.

This currently affects master.

williamdes:
Disable cache when the file is not writable

Fixes: #16055

### DIFF
--- a/libraries/classes/Routing.php
+++ b/libraries/classes/Routing.php
@@ -11,6 +11,7 @@ use function htmlspecialchars;
 use function mb_strlen;
 use function rawurldecode;
 use function sprintf;
+use function is_writable;
 
 /**
  * Class used to warm up the routing cache and manage routing.
@@ -23,9 +24,16 @@ class Routing
 
         $routes = require ROOT_PATH . 'libraries/routes.php';
 
+        $cacheDisabled = ($cfg['environment'] ?? '') === 'development';
+        $cacheFile = CACHE_DIR . 'routes.cache.php';
+
+        if (! is_writable($cacheFile)) {
+            $cacheDisabled = true;
+        }
+
         return cachedDispatcher($routes, [
-            'cacheFile' => CACHE_DIR . 'routes.cache.php',
-            'cacheDisabled' => ($cfg['environment'] ?? '') === 'development',
+            'cacheFile' => $cacheFile,
+            'cacheDisabled' => $cacheDisabled,
         ]);
     }
 


ibennetch:
This looks innocent enough, but I'm still a little apprehensive about putting it in QA_5_1, do you think it belongs there instead of master for 5.2?

libraries/classes/Routing.php
```diff
@@ -23,9 +24,16 @@ public static function getDispatcher(): Dispatcher
 
         $routes = require ROOT_PATH . 'libraries/routes.php';
 
+        $cacheDisabled = ($cfg['environment'] ?? '') === 'development';
```
ibennetch, line 27: Nice use of `??` here.

williamdes:
This fix needs to go to 5.1 for sure, we will have too much reports of "broken" setups otherwise.

I did a setup from git for @sudo-robot and had this issue

ibennetch:
This looks good to me. However, I'm having a little trouble at the moment because I can't get the cache to write under any circumstances. It doesn't look like this pull request should be related but I don't have time to look in to the full details right now.

ibennetch:
> This fix needs to go to 5.1 for sure, we will have too much reports of "broken" setups otherwise.

I agree!

After I thought about it, I came to this conclusion on my own and deleted my comment asking about this. You probably responded at the same time I was doing that, which is why my question isn't there anymore :)

williamdes:
> > This fix needs to go to 5.1 for sure, we will have too much reports of "broken" setups otherwise.
> 
> I agree!
> 
> After I thought about it, I came to this conclusion on my own and deleted my comment asking about this. You probably responded at the same time I was doing that, which is why my question isn't there anymore :)

Yeah I was searching for the question here after viewing the email and decided to answer anyway thinking it was one more GH mobile bug ^^

I would wait for @mauriciofauth to review this one too before merging
But if the version must go out you can merge this one

MauricioFauth (approved):

PR #16653, opened 2021-02-15
Fix #15994 - Attribute bug when default current_timestamp used with on update current_timestamp

### LINKED ISSUE #15994: "Extra" column definition not editable
### Describe the bug

When editing a column "extra" definitions are not editable and not visible. Only AUTO_INCREMENT is visible and editable as a separate option, but, for example, ON UPDATE CURRENT_TIMESTAMP is not visible or editable. Furthermore, when you try to edit such a column, such extra definition will be lost. It is retained however when exporting the table with such column.

### To Reproduce

1. Add column with ON UPDATE CURRENT_TIMESTAMP (will require manually running an SQL) clause in its definition
2. Attempt to edit it through UI (say, add a comment)
3. Save changes
4. Check current definition of the column and see, that it's missing ON UPDATE CURRENT_TIMESTAMP

### Expected behavior

At least it should show extra definitions and preserve them when changing column through UI. Ideally - also allow to edit them.

### Screenshots

N/A

### Server configuration

- Operating system: Win10
- Web server: Apache 2.4
- Database version: 10.2.31-MariaDB
- PHP version: 7.4.3
- phpMyAdmin version: 5.0.1

### Client configuration

- Browser: MS Edge
- Operating system: Win10

### Additional context

N/A

iifawzi:

### Description
It seems like the bug in #15994 is occurred **ONLY** in the following case: 
-  when the default attribute is set to `CURRENT_TIMESTAMP` and an `on update CURRENT_TIMESTAMP` is used, this leads to generate the following **Extra** field: `DEFAULT_GENERATED ON UPDATE CURRENT_TIMESTAMP` as shown below (tested on the PMA demo with version 5.2.0-dev)
Screen Shot 2021-02-15 at 9 18 47 PM

which fails this check: 
https://github.com/phpmyadmin/phpmyadmin/blob/c1a66adcc7b5d3f1273f38c4f7ab898e0deeb4bf/templates/columns_definitions/column_attribute.twig#L4-L5

i've changed it to check if the extra field contains `on update CURRENT_TIMESTAMP` or not.

Fixes #15994

### DIFF
--- a/templates/columns_definitions/column_attribute.twig
+++ b/templates/columns_definitions/column_attribute.twig
@@ -1,7 +1,8 @@
 {% if submit_attribute is defined and submit_attribute != false %}
     {% set attribute = submit_attribute %}
+    {# MariaDB has additional parentheses #}
 {% elseif column_meta['Extra'] is defined
-    and column_meta['Extra'] == 'on update CURRENT_TIMESTAMP' %}
+    and ('on update CURRENT_TIMESTAMP' in column_meta['Extra'] or 'on update CURRENT_TIMESTAMP()' in column_meta['Extra']) %}
     {% set attribute = 'on update CURRENT_TIMESTAMP' %}
 {% elseif extracted_columnspec['attribute'] is defined %}
     {% set attribute = extracted_columnspec['attribute'] %}


williamdes (approved):
This looks like a good fix !

williamdes:
@sudo-robot deploy

williamdes:
Hi @iifawzi 

I could confirm that your fix works only for MySQL, could you make it MariaDB compatible ?

tested using the deployed server

iifawzi:
> Hi @iifawzi
> 
> I could confirm that your fix works only for MySQL, could you make it MariaDB compatible ?
> 
> tested using the deployed server

Hmm, it seems like that's because MariaDB generate the Extra field with additional Parentheses: 
`ON UPDATE CURRENT_TIMESTAMP()`

I will work on it

iifawzi:
i've added an `or` statement to check the case if it was with parentheses, I was thinking about using regex to delete the parentheses from the Extra column, but this would maybe cause an unexpected bugs/errors.

williamdes:
> i've added an `or` statement to check the case if it was with parentheses, I was thinking about using regex to delete the parentheses from the Extra column, but this would maybe cause an unexpected bugs/errors if there was other parentheses.

Maybe you could add a `starts with` https://twig.symfony.com/doc/3.x/templates.html#comparisons

And add a comment `{# MariaDB has additional parentheses #}`

What do you think ?
(could you squash your commits and force push ?)

iifawzi:
> > i've added an `or` statement to check the case if it was with parentheses, I was thinking about using regex to delete the parentheses from the Extra column, but this would maybe cause an unexpected bugs/errors if there was other parentheses.
> 
> Maybe you could add a `starts with` or `ends with` https://twig.symfony.com/doc/3.x/templates.html#comparisons
> 
> And add a comment `{# MariaDB has additional parentheses #}`
> 
> What do you think ?
> (could you squash your commits and force push ?)

I think `starts with` wouldn't work with MySql, since in MySql, the Extra field is generated as following (when default is used): `DEFAULT_GENERATED ON UPDATE CURRENT_TIMESTAMP`, am i missing something?

williamdes:
Indeed, very accurate, I seem to have missed that detail
I will merge your work tomorrow (16/02) after your squash ;)

iifawzi:
> Indeed, very accurate, I seem to have missed that detail
> I will merge your work tomorrow (16/02) after your squash ;)

Awesome, thanks for your time reviewing this.

williamdes (approved):

williamdes:
I added 5e5733dcb796b1e7b128a601b74b7da327bd3ca4 because it did not work on my MariaDB 10.4 instance, using a lowercase compare is safer

PR #16651, opened 2021-02-15
Refactor field flags to use an object an not strpos

williamdes:
Refactor field flags to use an object an not strpos

Ref: #16005

PR #16691, opened 2021-02-28
Fix Selenium\Table\BrowseTest::testSortRecords test

MauricioFauth:
Introduced by #16651.


williamdes:
Did I introduce a bug or did I fix something?

williamdes (approved):

mauriciofauth:
> Did I introduce a bug or did I fix something?

@williamdes I reverted the test and fixed the bug you introduced. 

I still prefer not having this special case and I don't know why it exists. But I kept it that way for now.

williamdes (commented):
I think I added a getter that simplifies stating all possible cases

PR #16696, opened 2021-03-02
Add Autoprefixer to the CSS build process

MauricioFauth:
Adding this fixes some minor visual issues.

From [Bootstrap documentation](https://getbootstrap.com/docs/5.0/getting-started/build-tools/#autoprefixer):

> Bootstrap uses [Autoprefixer](https://github.com/postcss/autoprefixer) (included in our build process) to automatically add vendor prefixes to some CSS properties at build time. Doing so saves us time and code by allowing us to write key parts of our CSS a single time while eliminating the need for vendor mixins like those found in v3.


williamdes (approved):
Good idea 💡

PR #16700, opened 2021-03-02
Fix Selenium failing tests

MauricioFauth:
This is not the better fix. The Console JavaScript is too slow for Selenium.


williamdes (commented):
Awesome 💯

PR #16710, opened 2021-03-04
Fix type error when exporting triggers to ODF

MauricioFauth:
I have not reproduced it, but `$aliases` is an array and `$dbi->getTriggers` is expecting a `$delimiter` (string) as the third parameter. 

Introduced by a78a7a0c9c0081f6bc19f9779cd4401bfa77a09d.

### DIFF
--- a/libraries/classes/Plugins/Export/ExportOdt.php
+++ b/libraries/classes/Plugins/Export/ExportOdt.php
@@ -727,7 +727,7 @@ class ExportOdt extends ExportPlugin
                 );
                 break;
             case 'triggers':
-                $triggers = $dbi->getTriggers($db, $table, $aliases);
+                $triggers = $dbi->getTriggers($db, $table);
                 if ($triggers) {
                     $GLOBALS['odt_buffer']
                     .= '<text:h text:outline-level="2" text:style-name="Heading_2"'


williamdes (approved):
This LGTM but if possible, add a doc block for dbi and phpstan/psalm prove the bug and this will be okay to validate your fix

libraries/classes/Plugins/Export/ExportOdt.php
```diff
@@ -727,7 +727,7 @@ public function exportStructure(
                 );
                 break;
             case 'triggers':
-                $triggers = $dbi->getTriggers($db, $table, $aliases);
+                $triggers = $dbi->getTriggers($db, $table);
```
williamdes, line 730: Would it make sense to use `$delimiter` like for SQL export ?

mauriciofauth: `ExportOdt::exportStructure` doesn't have a delimiter defined. So it will use the `$dbi->getTriggers` default (`'//'`).

williamdes: Oh okay, thanks
I am merging your work now

PR #16722, opened 2021-03-09
Remove the PMA_VERSION constant

MauricioFauth:
Uses the `PhpMyAdmin\Version::VERSION` constant instead.


libraries/classes/Setup/Index.php
```diff
@@ -162,9 +163,7 @@ public static function versionCheck()
             return;
         }
 
-        $version_local = $versionInformation->versionToInt(
-            $GLOBALS['config']->get('PMA_VERSION')
-        );
+        $version_local = $versionInformation->versionToInt(Version::VERSION);
```
williamdes, line 166: We have an int version const, right?

williamdes (approved):

PR #16726, opened 2021-03-10
Remove jQuery Fullscreen plugin dependency

MauricioFauth:
Uses the [Fullscreen API](https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API) instead.


williamdes (approved):
:100:

PR #16745, opened 2021-03-20
Add a flash messages service provider

MauricioFauth:
Used to persist messages from the current request to the next request.

### DIFF
--- a/libraries/classes/FlashMessages.php
+++ b/libraries/classes/FlashMessages.php
@@ -1,0 +1,54 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpMyAdmin;
+
+use RuntimeException;
+
+final class FlashMessages
+{
+    private const STORAGE_KEY = 'flashMessages';
+
+    /** @var array */
+    private $storage;
+
+    /** @var array<string, string[]> */
+    private $previousMessages = [];
+
+    public function __construct()
+    {
+        if (! isset($_SESSION)) {
+            throw new RuntimeException(__('Session not found.'));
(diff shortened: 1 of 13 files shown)


libraries/classes/FlashMessages.php
```diff
@@ -0,0 +1,54 @@
...
+        $this->storage = &$_SESSION;
+
+        if (isset($this->storage[self::STORAGE_KEY])) {
+            $this->previousMessages = $this->storage[self::STORAGE_KEY];
+        }
+
+        $this->storage[self::STORAGE_KEY] = [];
+    }
+
+    public function addMessage(string $key, string $message): void
```
williamdes, line 34: they could be static because the storage is based on a global value

williamdes (approved):

libraries/classes/FlashMessages.php
mauriciofauth: We could do that, but this will allow to call `addMessage` before constructing the object, causing that message to not be persisted to the next request.

templates/database/structure/index.twig
```diff
@@ -1,3 +1,11 @@
+{% for flash_key, flash_messages in flash() %}
```
williamdes, line 1: Could this be a template?

williamdes (approved):

templates/database/structure/index.twig
mauriciofauth: Definitely, I added this here to make this works, but I will find a better way of doing this.

PR #16795, opened 2021-04-07
Fix #16605 - Fix vertical alignment issue

### LINKED ISSUE #16605: Vertical alignment seems off
5.1.0-rc1

Vertical alignment of the table names seems off.
Same for the sum values of rows, size and overhead. Font size of these seems off too.

image

bradsi:
Signed-off-by: Brad Singleton <29055518+bradsingleton@users.noreply.github.com>

### Description

Small change in the relevant scss file to apply `vertical-align: middle;` to both `th` and `td` elements in the pmahomme theme. Previously `vertical-align: middle;` was only applied to `td` elements. Screenshot shown below.

This fixes:
- Vertical alignment of the table names.
- Vertical alignment of sum values for Rows, Size, and Overhead columns.

The issue linked to this fix mentions that the font size is off for the sum values stated above, I've not looked into this as the font size seems consistent with the values in the table so I'm assuming this was a design decision.

fix-16605-screenshot

Fixes #16605 

### DIFF
--- a/themes/pmahomme/scss/_tables.scss
+++ b/themes/pmahomme/scss/_tables.scss
@@ -2,6 +2,7 @@
   th,
   td {
     text-shadow: 0 1px 0 #fff;
+    vertical-align: middle;
   }
 
   th {
@@ -10,7 +11,6 @@
 
   td {
     touch-action: manipulation;
-    vertical-align: middle;
   }
 
   thead th {


OlafvdSpek:
Do the table names use th elements? That doesn't seem quite right does it?

bradsi:
@OlafvdSpek 
Yeah they're th elements, sum values as well. I think you're right, they should probably be td elements. I've changed the table names to td elements locally and it fixes the vertical alignment issue but then a `font-weight: bold` style needs to be applied. That might be a better fix.

OlafvdSpek:
Maybe the sum row doesn't even need to be bold.

bradsi:
I think the th elements on the table names are correct, the [example on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/th) uses a th for table names as well.

williamdes (approved):
Thank you, the fix looks good
I will test it

williamdes (approved):
Perfect, It looks better now

PR #16814, opened 2021-04-12
Fix #15566 Designer - support RTL mode

### LINKED ISSUE #15566: Designer RTL mode
Nothing more to say, if you need details :ping_pong: me.
designer

iifawzi:

### Description

Designer now supports RTL languages🎆  🎉 

pma

Fixes #15566

### DIFF
--- a/js/src/designer/move.js
+++ b/js/src/designer/move.js
@@ -55,6 +55,9 @@ DesignerMove.markUnsaved = function () {
     $('#saved_state').text('*');
 };
 
+var mainDirection = $('html').attr('dir') === 'rtl' ? 'right' : 'left';
+// Will be used to multiply the offsetLeft by -1 if the direction is rtl.
+var directionEffect = mainDirection === 'right' ? -1 : 1;
 var curClick = null;
 var smS           = 0;
 var smAdd         = 10;
@@ -130,13 +133,13 @@ DesignerMove.mouseMove = function (e) {
 
         var $curClick = $(curClick);
 
-        var curX = parseFloat($curClick.attr('data-left') || $curClick.css('left'));
+        var curX = parseFloat($curClick.attr('data-' + mainDirection) || $curClick.css(mainDirection));
         var curY = parseFloat($curClick.attr('data-top') || $curClick.css('top'));
 
-        var newX = curX - deltaX;
+        var newX = curX - directionEffect * deltaX;
         var newY = curY - deltaY;
 
-        $curClick.attr('data-left', newX);
+        $curClick.attr('data-' + mainDirection, newX);
         $curClick.attr('data-top', newY);
 
         if (onGrid) {
@@ -149,14 +152,14 @@ DesignerMove.mouseMove = function (e) {
         } else if (newY < 0) {
             newY = 0;
         }
-        $curClick.css('left', newX + 'px');
+        $curClick.css(mainDirection, newX + 'px');
         $curClick.css('top', newY + 'px');
     } else if (layerMenuCurClick) {
         if (menuMoved) {
             deltaX = -deltaX;
         }
         var $layerMenu = $('#layer_menu');
-        var newWidth = $layerMenu.width() + deltaX;
+        var newWidth = $layerMenu.width() + directionEffect * deltaX;
         if (newWidth < 150) {
             newWidth = 150;
         }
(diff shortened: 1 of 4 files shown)


williamdes (approved):
This looks amazing !

williamdes:
@sudo-robot deploy

williamdes:
Hi @iifawzi 
Using persian language I have a strange effect, see:
image

templates/database/designer/database_tables.twig
```diff
@@ -15,7 +15,7 @@
         cellpadding="0"
         cellspacing="0"
         class="pma-table designer_tab"
-        style="position:absolute; left:
+        style="position:absolute; {{ text_dir == 'rtl' ? 'right' : 'left' }}:
```
iifawzi, line 18: Hi @williamdes, I've noticed it also in the deployed version using Arabic, that's so weird and this shouldn't happen if this line works correctly, it seems like it always return 'left', I will try to investigate why this happening.

williamdes: The use of the global seems not very perfect, maybe you can find another access to the variable without the global ?

Sean4134 (approved):

templates/database/designer/database_tables.twig
iifawzi: I've changed it to the way used in other controllers, I think it should work correctly now.

williamdes:
@sudo-robot deploy

templates/database/designer/database_tables.twig
williamdes: I just re-deployed it and it seems not to work :/

iifawzi: That's so weird, the commit is working correctly in my local machine (even before the last change)
ezgif-2-5db5eab4c2a8

It seems to be a problem with the template cache in the deployed version. 
Screen Shot 2021-04-13 at 11 30 21 PM
Screen Shot 2021-04-13 at 11 34 22 PM

iifawzi: Hey @williamdes, What do you think? 👀  
If the bug is not related to the template cache; I will try to investigate more if you wish.

js/src/designer/move.js
```diff
@@ -55,6 +55,9 @@ DesignerMove.markUnsaved = function () {
     $('#saved_state').text('*');
 };
 
+var mainDirection =  $('html').attr('dir') === 'rtl' ? 'right' : 'left';
```
williamdes, line 58: `yarn run js-lint --fix`

templates/database/designer/database_tables.twig
williamdes: I just cleared cache on the container, did not work
changed the temp folder and the error is not shown anymore
but there is still a bug

The HTML direction attr seems good
Here are the rtl languages:
```php
    public function isRTL()
    {
        return in_array($this->code, ['ar', 'fa', 'he', 'ur']);
    }
```

williamdes: that said it works perfectly on my workstation, for some reason not completely on the demo server

iifawzi: This was driving me crazy, since i'm not able to re-produce. 
so, what might be the reason for this not to be working on the demo server? could I help?
I can see that the problem in the demo server is that the 'left' position is still there, where it shouldn't. I think it's cached somehow. 

Screen Shot 2021-04-15 at 9 18 27 PM

iifawzi: The only line who's responsible to set the left or right positions is the line marked above, and this can be observed if
we moved the tables, only the `right` position and `data-right` will be changed according to the moves. `left` position is initially there somehow.

williamdes: I am destroying the demo server and asking the bot to create it back 🤞🏻

williamdes:
@sudo-robot deploy

williamdes:
> @sudo-robot deploy

Even after destroy and re-rebuild there is a bug, I am open-sourcing the infrastructure and container to allow us to pin point the issue

williamdes:
> > @sudo-robot deploy
> 
> Even after destroy and re-rebuild there is a bug, I am open-sourcing the infrastructure and container to allow us to pin point the issue

Here it is: https://github.com/sudo-bot/gh-deployer-container#manual-use-of-the-phpmyadmin-container

I did run the deployed container on my workstation docker netowork and could not reproduce this issue
Sooo, thank you for your work !
I will probably wait for another review on it to merge it

iifawzi:
> > > @sudo-robot deploy
> > 
> > 
> > Even after destroy and re-rebuild there is a bug, I am open-sourcing the infrastructure and container to allow us to pin point the issue
> 
> Here it is: https://github.com/sudo-bot/gh-deployer-container#manual-use-of-the-phpmyadmin-container
> 
> I did run the deployed container on my workstation docker netowork and could not reproduce this issue
> Sooo, thank you for your work !
> I will probably wait for another review on it to merge it

Awesome, I will remove the extra space, and squash the commits. Thank you for your time reviewing this @williamdes

PR #16818, opened 2021-04-13
Remove useless htmlspecialchars on URL generation

### LINKED ISSUE #16792: Wrong links for COLUMN_NAME row values
### Describe the bug

Row values for `COLUMN_NAME` column, from `information_schema.COLUMNS` have wrong links. Instead of `&amp;` after `index.php?route=/table/structure/change`, is written as `&amp;amp;`.

### To Reproduce

1. Go to https://demo.phpmyadmin.net/master-config/index.php?route=/sql&db=information_schema&table=COLUMNS
2. Click on any COLUMN_NAME column values.
3. See error

### Expected behavior

A form to edit the column should appear.

### Screenshots

Right now:
Untitled

Expected:
expected

### Client configuration

- Browser: Firefox
- Operating system: Windows

williamdes:
Signed-off-by: William Desportes <williamdes@wdes.fr>

Fixes #16792

### DIFF
--- a/libraries/classes/Url.php
+++ b/libraries/classes/Url.php
@@ -151,31 +151,29 @@ class Url
      * // note the missing ?
      * echo 'script.php' . Url::getCommon($params);
      * // produces with cookies enabled:
-     * // script.php?myparam=myvalue&amp;db=mysql&amp;table=rights
+     * // script.php?myparam=myvalue&db=mysql&table=rights
      * // with cookies disabled:
-     * // script.php?server=1&amp;lang=en&amp;myparam=myvalue&amp;db=mysql
-     * // &amp;table=rights
+     * // script.php?server=1&lang=en&myparam=myvalue&db=mysql
+     * // &table=rights
      *
      * // note the missing ?
      * echo 'script.php' . Url::getCommon();
      * // produces with cookies enabled:
      * // script.php
      * // with cookies disabled:
-     * // script.php?server=1&amp;lang=en
+     * // script.php?server=1&lang=en
      * </code>
      *
-     * @param mixed  $params  optional, Contains an associative array with url params
-     * @param string $divider optional character to use instead of '?'
+     * @param array<string,int|string|bool> $params  optional, Contains an associative array with url params
+     * @param string                        $divider optional character to use instead of '?'
      *
      * @return string   string with URL parameters
      *
      * @access public
      */
-    public static function getCommon($params = [], $divider = '?')
+    public static function getCommon(array $params = [], $divider = '?')
     {
-        return htmlspecialchars(
-            self::getCommonRaw($params, $divider)
-        );
+        return self::getCommonRaw($params, $divider);
     }
 
     /**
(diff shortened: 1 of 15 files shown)


williamdes:
Hi @ibennetch @mauriciofauth @devenbansod 
I am requesting review from you because this could potentially lead to security issues. To avoid that I added a new test class `SecurityTest` where you will see that I tested all the cases I could think of.

Please let me know if this looks good to you or not, as far as I can see and did test there is no new security breach.

MauricioFauth (approved):

Sean4134 (approved):

Sean4134 (approved):

Sean4134 (approved):

williamdes:
Removes the code that was added in dce94f229f34cc8f7e26df3ee5e13b9c01df6201

Probably because `&` had to be escaped at this time, but could not find any proof of that

The migration to HTML5 was done here in 2012: https://github.com/phpmyadmin/phpmyadmin/commit/af0e6c71e72cc3bbcd9c8d35c64836e4e3b07b1e

Melnofil:
:-1:
Peut être que ça fonctionnerait très bien aujourd'hui. Mais d'après mon expérience, le point-virgule est optionnel sur certains navigateurs, si HTML5.2 ou HTML7 ajoute par exemple une entité `&table;` alors le code HTML suivant :

`data-post="db=test&table=test&where_clause=%60test%60.%60test%60+%3D+42&clause_is_unique=1&sql_query=SELECT+%2A+FROM+%60test%60+ORDER+BY+%60test%60.%60test%60+DESC&goto=sql.php&default_action=update"`

Va boguer dès que les navigateurs seront mis à jour, car « &table » sera alors transformé en ce caractère. Et de l'autre côté, je ne vois pas ce que la modification apporte de bénéfique.

williamdes:
Hi @Melnofil 

I am glad that you provide feedback, but could you give a real test case ?
We use HTML5

`&table;` this will never happen because queries are built by the PHP function and so it will always give the format
`{separator}{key}={value}` making `{separator}{key};{value}` impossible

Melnofil:
https://www.w3.org/TR/html50/introduction.html#common-pitfalls-to-avoid-when-using-the-scripting-apis

> Errors involving fragile syntax constructs
> 
> There are syntax constructs that, for historical reasons, are relatively fragile. To help reduce the number of users who accidentally run into such problems, they are made non-conforming.
> 
> For example, the parsing of certain named character references in attributes happens even with the closing semicolon being omitted. It is safe to include an ampersand followed by letters that do not form a named character reference, but if the letters are changed to a string that does form a named character reference, they will be interpreted as that character instead.
> 
> In this fragment, the attribute's value is `?bill&ted`:
> 
> `<a href="?bill&ted">Bill and Ted</a>`
> 
> In the following fragment, however, the attribute's value is actually `?art©`, not the intended `?art&copy`, because even without the final semicolon, `&copy` is handled the same as `&copy;` and thus gets interpreted as `©`:
> 
> `<a href="?art&copy">Art and Copy</a>`
> 
> To avoid this problem, all named character references are required to end with a semicolon, and uses of named character references without a semicolon are flagged as errors.
> 
> Thus, the correct way to express the above cases is as follows:
> 
> `<a href="?bill&ted">Bill and Ted</a> <!-- &ted is ok, since it's not a named character reference -->`
> 
> `<a href="?art&amp;copy">Art and Copy</a> <!-- the & has to be escaped, since &copy is a named character reference -->`

Voir par exemple le cas d'un dév qui se retrouve avec `&region=` interprété comme `&reg;ion=` (cas intéressant, car il est assez rare qu'un navigateur aille jusqu'à couper un mot) :
https://www.ipgirl.com/70752/pourquoi-reg-est-il-rendu-comme-sans-le-point-virgule.html

williamdes:
Hi @mauriciofauth 
See: https://jsfiddle.net/williamdes/zqkp15yx/2/

What should we do, consider this as an edge case ?

Melnofil:
Outch! I see the bug! I can't believe it!
Firefox 88.0.1 (64 bit) Ubuntu 20.10

PR #16844, opened 2021-04-21
Redesign the Charsets and Collations page

MauricioFauth:
Screenshot_2021-04-21 localhost 8000 localhost phpMyAdmin 5 2 0-dev


williamdes (approved):
😍 🤩

PR #16904, opened 2021-05-16
Replace REQUEST to POST in Routines.php

sider-bughunter:
REQUEST should be replaced to POST in this context.
For your reference, Events.php in the same directory has a similar code around Line number 297. It uses POST.

### Description

Fixes #


williamdes:
Hi @sider-bughunter 
Can you let us know how you checked that is was only using POST ?

sider-bughunter:
Hi @williamdes 
Thank you for your review.

First of all, this is not a bug. The original code should work.
I just think it is a bit inconsistent between Routines.php and Events.php.
I checked this based on a pattern analysis in the duplicate code block.

L470 - L486 in Routines.php is almost the same code as L283 - L299 in Events.php.

L297 in Events.php is:
`if (isset($_POST['item_type']) && $_POST['item_type'] === 'RECURRING') {`
POST is used.

On the other hand, L484 in Routines.php
`if (isset($_REQUEST['item_type']) && $_REQUEST['item_type'] === 'FUNCTION') {`
REQUEST is used.

So, either Events.php or Routines.php should be modified from the consistent code point of view.

williamdes:
I had a look and could confirm that it is only used in POST, maybe another dev could confirm. Because we have links that use this param name in URLs. See for example a right click > open in new table to a "Edit" button of a procedure or function

MauricioFauth (approved):

PR #16909, opened 2021-05-20
Add roave/security-advisories to composer.json

MauricioFauth:
https://github.com/Roave/SecurityAdvisories

> This package ensures that your application doesn't have installed dependencies with known security vulnerabilities.


williamdes (approved):
Very good idea !

PR #16910, opened 2021-05-21
Add DatabaseName value object

MauricioFauth:
The idea is to replace the `$db` string with the _DatabaseName_ value object to avoid having to validate the value multiple times.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,6 +58,7 @@
         "symfony/polyfill-ctype": "^1.17.0",
         "symfony/polyfill-mbstring": "^1.17.0",
         "twig/twig": "^3.0.1",
+        "webmozart/assert": "^1.10",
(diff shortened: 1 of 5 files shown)


williamdes (approved):
Seems like a good idea
Not fan of adding one more library dependency that said

mauriciofauth:
> Not fan of adding one more library dependency that said

I thought about using this library in more places. There is a lot of interesting assertions: https://github.com/webmozarts/assert#assertions

williamdes:
> > Not fan of adding one more library dependency that said
> 
> I thought about using this library in more places. There is a lot of interesting assertions: https://github.com/webmozarts/assert#assertions

Good idea, that will allow us to make more security checks on user inputs

PR #16913, opened 2021-05-24
Skip testImportOsm and testDoImport on 32-bit/s390x arch

utkarsh2102:
These two tests fail on s390x machines, looks like some
32-bit issues we already have. Probably worth skipping
these on 32-bit/s390x arch.

Test logs on Ubuntu: https://autopkgtest.ubuntu.com/results/autopkgtest-impish/impish/s390x/p/phpmyadmin/20210516_212523_5cd19@/log.gz


williamdes (approved):
:100:

MauricioFauth (approved):

PR #16944, opened 2021-06-05
Fix build-release action

MauricioFauth:


williamdes:
Ref: https://github.com/composer/composer/issues/9946

PR #16945, opened 2021-06-06
Add suport for the Mroonga engine

josepsanzcamp:
Add suport to the Mroonga engine

### Description

With this patch you can see the size of the Mroonga's tables

### DIFF
--- a/libraries/classes/Controllers/Database/StructureController.php
+++ b/libraries/classes/Controllers/Database/StructureController.php
@@ -22,6 +22,7 @@ use PhpMyAdmin\ReplicationInfo;
 use PhpMyAdmin\ResponseRenderer;
 use PhpMyAdmin\Sanitize;
 use PhpMyAdmin\Sql;
+use PhpMyAdmin\StorageEngine;
 use PhpMyAdmin\Table;
 use PhpMyAdmin\Template;
 use PhpMyAdmin\Tracker;
@@ -1054,6 +1055,17 @@ class StructureController extends AbstractController
             case 'SYSTEM VIEW':
                 // possibly a view, do nothing
                 break;
+            case 'Mroonga':
+                // The idea is to show the size only if Mroonga is available,
+                // in other case the old unknown message will appear
+                if (StorageEngine::hasMroongaEngine()) {
+                    [$currentTable, $formattedSize, $unit, $sumSize] = $this->getValuesForMroongaTable(
+                        $currentTable,
+                        $sumSize
+                    );
+                    break;
+                }
+                // no break, go to default case
             default:
                 // Unknown table type.
                 if ($this->isShowStats) {
@@ -1178,6 +1190,40 @@ class StructureController extends AbstractController
             /** @var int $tblsize */
             $tblsize = $currentTable['Data_length']
                 + $currentTable['Index_length'];
+            $sumSize += $tblsize;
+            [$formattedSize, $unit] = Util::formatByteDown(
+                $tblsize,
+                3,
+                ($tblsize > 0 ? 1 : 0)
+            );
+        }
+
+        return [
+            $currentTable,
+            $formattedSize,
+            $unit,
+            $sumSize,
+        ];
+    }
+
+    /**
+     * Get values for Mroonga table
+     *
+     * @param array $currentTable current table
+     * @param int   $sumSize      sum size
+     *
+     * @return array
+     */
+    protected function getValuesForMroongaTable(
+        array $currentTable,
+        int $sumSize
+    ): array {
+        $formattedSize = '';
+        $unit = '';
+
+        if ($this->isShowStats) {
+            /** @var int $tblsize */
+            $tblsize = $currentTable['Data_length'] + $currentTable['Index_length'];
             $sumSize += $tblsize;
             [$formattedSize, $unit] = Util::formatByteDown(
                 $tblsize,
--- a/libraries/classes/DatabaseInterface.php
+++ b/libraries/classes/DatabaseInterface.php
@@ -417,7 +417,7 @@ class DatabaseInterface implements DbalInterface
                         $tables[$one_database_name][$one_table_name]['Index_length'],
                     ] = StorageEngine::getMroongaLengths(
                         $one_database_name,
-                        $one_table_name,
+                        $one_table_name
                     );
                 }
             }
@@ -528,7 +528,7 @@ class DatabaseInterface implements DbalInterface
                         $each_tables[$table_name]['Index_length'],
                     ] = StorageEngine::getMroongaLengths(
                         $each_database,
-                        $table_name,
+                        $table_name
                     );
                 }
 
--- a/libraries/classes/DatabaseInterface.php
+++ b/libraries/classes/DatabaseInterface.php
@@ -400,6 +400,28 @@ class DatabaseInterface implements DbalInterface
                 $link
             );
 
+            // here, we check for Mroonga engine and compute the good data_length and index_length
+            // in the StructureController only we need to sum the two values as the other engines
+            foreach ($tables as $one_database_name => $one_database_tables) {
+                foreach ($one_database_tables as $one_table_name => $one_table_data) {
+                    if ($one_table_data['Engine'] !== 'Mroonga') {
+                        continue;
+                    }
+
+                    if (! StorageEngine::hasMroongaEngine()) {
+                        continue;
+                    }
+
+                    [
+                        $tables[$one_database_name][$one_table_name]['Data_length'],
+                        $tables[$one_database_name][$one_table_name]['Index_length'],
+                    ] = StorageEngine::getMroongaLengths(
+                        $one_database_name,
+                        $one_table_name,
+                    );
+                }
+            }
+
             if ($sort_by === 'Name' && $GLOBALS['cfg']['NaturalOrder']) {
                 // here, the array's first key is by schema name
                 foreach ($tables as $one_database_name => $one_database_tables) {
@@ -489,6 +511,26 @@ class DatabaseInterface implements DbalInterface
                 }
 
                 $each_tables = $this->fetchResult($sql, 'Name', null, $link);
+
+                // here, we check for Mroonga engine and compute the good data_length and index_length
+                // in the StructureController only we need to sum the two values as the other engines
+                foreach ($each_tables as $table_name => $table_data) {
+                    if ($table_data['Engine'] !== 'Mroonga') {
+                        continue;
+                    }
+
+                    if (! StorageEngine::hasMroongaEngine()) {
+                        continue;
+                    }
+
+                    [
+                        $each_tables[$table_name]['Data_length'],
+                        $each_tables[$table_name]['Index_length'],
+                    ] = StorageEngine::getMroongaLengths(
+                        $each_database,
+                        $table_name,
+                    );
+                }
 
                 // Sort naturally if the config allows it and we're sorting
                 // the Name column.
--- a/libraries/classes/StorageEngine.php
+++ b/libraries/classes/StorageEngine.php
@@ -189,12 +189,11 @@ class StorageEngine
                 unset($objectList[$mroongaName]);
             }
 
-            /**
-             * At this point, we can lost all mroonga_data because only need the mroongaName
-             */
+            // At this point, we can remove all the data because only need the mroongaName values
             Cache::set($cacheKey, array_keys($objectList));
         }
 
+        /** @var string[] */
         $objectList = Cache::get($cacheKey, []);
 
         $dataLength = 0;
--- a/libraries/classes/StorageEngine.php
+++ b/libraries/classes/StorageEngine.php
@@ -21,11 +21,16 @@ use PhpMyAdmin\Utils\SessionCache;
 
 use function __;
 use function array_key_exists;
+use function array_keys;
 use function explode;
 use function htmlspecialchars;
+use function in_array;
+use function json_decode;
 use function mb_stripos;
 use function mb_strtolower;
 use function sprintf;
+use function strlen;
+use function strncmp;
 
 /**
  * Library for extracting information about the available storage engines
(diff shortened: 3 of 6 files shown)


williamdes:
Hi @josepsanzcamp 
This was a very nice feature, can you push it back or do it on another branch ?

josepsanzcamp:
Hi @williamdes 

I have detected that the sort feature don't work because we must to put the correct size values inside the function getTablesFull located in the DatabaseInterface. I have another commit that solve the problem.

In a few minutes I will do another pull request with this improvement.

Josep.

josepsanzcamp:
I have fixed a bug when try to sort by rows, additionally I have created a function that returns the lengths of the data and index for the mroonga engine, this simplify the code and allow to be reused from other parts of the code.

libraries/classes/DatabaseInterface.php
```diff
@@ -560,6 +590,39 @@ static function ($a, $b) {
...
     }
 
+    /**
+     * Get the lengths of a table of database
+     *
+     * @param string $db_name    DB name
+     * @param string $table_name Table name
+     */
+    private function getMroongaLengths(string $db_name, string $table_name) {
+        static $object_list=[];
```
williamdes, line 600: I am not sure this `static` keyword is needed not is a good practice ;)
Could you run `composer run phpcs` and `composer run phpcbf` to fix the errors please ?

libraries/classes/DatabaseInterface.php
```diff
@@ -560,6 +590,39 @@ static function ($a, $b) {
         return $tables;
     }
 
+    /**
+     * Get the lengths of a table of database
+     *
+     * @param string $db_name    DB name
+     * @param string $table_name Table name
+     */
+    private function getMroongaLengths(string $db_name, string $table_name) {
```
williamdes, line 599:
```suggestion
     * @param string $table_name Table name
     * @return int[]
     */
    private function getMroongaLengths(string $db_name, string $table_name): array {
```

libraries/classes/DatabaseInterface.php
```diff
@@ -560,6 +590,39 @@ static function ($a, $b) {
...
+        }
+        $dataLength = 0;
+        $indexLength = 0;
+        foreach ($object_list[$db_name] as $mroonga_name => $mroonga_data) {
+            if(strncmp($table_name, $mroonga_name, strlen($table_name))==0) {
+                $result = $this->query("SELECT mroonga_command('object_inspect ${mroonga_name}')");
+                $row = $this->fetchRow($result);
+                $this->freeResult($result);
+                $temp = json_decode($row[0],true);
+                if(strncmp("${table_name}#${table_name}", $mroonga_name, strlen("${table_name}#${table_name}")) == 0) {
```
williamdes, line 616:
```suggestion
                if(strncmp($table_name . '#'. $table_name, $mroonga_name, strlen($table_name . '#'. $table_name)) === 0) {
```

libraries/classes/DatabaseInterface.php
```diff
@@ -560,6 +590,39 @@ static function ($a, $b) {
...
+                    $indexLength += $temp["disk_usage"];
+                } else {
+                    $dataLength += $temp["disk_usage"];
+                }
```
williamdes, line 620: I think this else could be avoided by adding a continue in the if

libraries/classes/Controllers/Database/StructureController.php
```diff
@@ -1194,6 +1200,40 @@ protected function getValuesForInnodbTable(
...
+     *
+     * @param array $currentTable current table
+     * @param int   $sumSize      sum size
+     *
+     * @return array
+     */
+    protected function getValuesForMroongaTable(
+        array $currentTable,
+        $sumSize
+    ) {
```
williamdes, line 1214:
```suggestion
     * @return int[]
     */
    protected function getValuesForMroongaTable(
        array $currentTable,
        int $sumSize
    ): array {
```

williamdes (changes requested):
Good work, some changes would make the code even nicer ;)
Could you post some SQL examples to help me reproduce and test your feature ?

libraries/classes/DatabaseInterface.php
josepsanzcamp: Hi @williamdes 

The idea of this static array is to store the result of the command object_list to prevent that the next executions require to re-execute it, is a cache optimization.

I have used a static global variable because this data is only valid for the usage of this function. If you want to not use this feature, the other maner to do it is by defining a global variable in the class, but the usage is very specific of this function and how no other parts of the code must to access to it, I vote for use it in this case.

What do you want to do???

Related to the composer commands: I have executed the two composer commands to solve all problems related to the code styling, you can see it in the last commit.

Thanks in advance.

Josep.

libraries/classes/DatabaseInterface.php
josepsanzcamp: OK, I have applied it in the last commit, thanks!!!

libraries/classes/DatabaseInterface.php
josepsanzcamp: OK, I have applied it in the last commit, thanks!!!

libraries/classes/DatabaseInterface.php
josepsanzcamp: I don't understand it, could you provide me an example??? the idea is to increment one variable or the other variable

libraries/classes/Controllers/Database/StructureController.php
josepsanzcamp: OK, I have applied it in the last commit, thanks!!!

josepsanzcamp:
Hi @williamdes 

This is a SQL example to create a Mroonga table:

`CREATE TABLE `idx_correo` (
  `search` mediumtext NOT NULL DEFAULT '',
  PRIMARY KEY (`id`),
  FULLTEXT KEY `idx_correo_search` (`search`)
) ENGINE=Mroonga AUTO_INCREMENT=1 DEFAULT CHARSET=utf8mb4
`

Is it right for your unit test???

Josep.

libraries/classes/Controllers/Database/StructureController.php
```diff
@@ -1194,6 +1200,40 @@ protected function getValuesForInnodbTable(
...
+     * @param array $currentTable current table
+     * @param int   $sumSize      sum size
+     *
+     * @return array
+     */
+    protected function getValuesForMroongaTable(
+        array $currentTable,
+        int $sumSize
+    ): array {
+        $formattedSize = $unit = '';
```
williamdes, line 1215:
```suggestion
        $formattedSize = '';
        $unit = '';
```

libraries/classes/DatabaseInterface.php
```diff
@@ -490,6 +509,22 @@ static function ($a, $b) {
 
                 $each_tables = $this->fetchResult($sql, 'Name', null, $link);
 
+                // here, we check for Mroonga engine and compute the good data_length and index_length
+                // in the StructureController only we need to sum the two values as the other engines
+                foreach ($each_tables as $table_name => $table_data) {
+                    if ($table_data['Engine'] != 'Mroonga') {
```
williamdes, line 515:
```suggestion
                    if ($table_data['Engine'] !== 'Mroonga') {
```

libraries/classes/DatabaseInterface.php
```diff
@@ -560,6 +595,47 @@ static function ($a, $b) {
...
+            $result = $this->query("SELECT mroonga_command('object_inspect " . $mroonga_name . "')");
+            $row = $this->fetchRow($result);
+            $this->freeResult($result);
+            $temp = json_decode($row[0], true);
+            $index_prefix = $table_name . '#' . $table_name;
+            if (strncmp($index_prefix, $mroonga_name, strlen($index_prefix)) == 0) {
+                $indexLength += $temp['disk_usage'];
+            } else {
+                $dataLength += $temp['disk_usage'];
+            }
```
williamdes, line 633:
```suggestion
            if (strncmp($index_prefix, $mroonga_name, strlen($index_prefix)) === 0) {
                $indexLength += $temp['disk_usage'];
                continue;
            }
            $dataLength += $temp['disk_usage'];
```

libraries/classes/DatabaseInterface.php
```diff
@@ -560,6 +595,47 @@ static function ($a, $b) {
...
+            $result = $this->query("SELECT mroonga_command('object_list')");
+            $row = $this->fetchRow($result);
+            $this->freeResult($result);
+            $object_list[$db_name] = json_decode($row[0], true);
+        }
+
+        $dataLength = 0;
+        $indexLength = 0;
+        foreach ($object_list[$db_name] as $mroonga_name => $mroonga_data) {
+            if (strncmp($table_name, $mroonga_name, strlen($table_name)) != 0) {
```
williamdes, line 620:
```suggestion
            if (strncmp($table_name, $mroonga_name, strlen($table_name)) !== 0) {
```

williamdes (approved):

libraries/classes/DatabaseInterface.php
williamdes: I posted a suggestion :)

williamdes:
@sudo-robot deploy

williamdes:
image
I just installed Mroonga on the MariaDB 10.5 server (mariadb:10.5 on docker hub) and had this error.
Followed: https://mariadb.com/kb/en/about-mroonga/

libraries/classes/DatabaseInterface.php
josepsanzcamp: OK, now I understand your previous comments, thanks!!!

libraries/classes/Controllers/Database/StructureController.php
josepsanzcamp: OK, applied in the next commit, thanks!!!

libraries/classes/DatabaseInterface.php
josepsanzcamp: OK, applied in the next commit, thanks!!!

libraries/classes/DatabaseInterface.php
josepsanzcamp: OK, applied in the next commit, thanks!!!

libraries/classes/DatabaseInterface.php
josepsanzcamp: OK, applied in the next commit, thanks!!!

josepsanzcamp:
Hi @williamdes 

I have added some checks before to use the Mroonga dependent code, with this, if Mroonga is not available the result will be the same that the previous release that doesn't contains the Mroonga support.

Josep.

williamdes:
Hi @josepsanzcamp
I can confirm that I could create the table only after I installed the plugin.
So the plugin is installed and working.
The function you use seems not to be available for some reason, I will try to find out why

josepsanzcamp:
Hi @williamdes 

This error is caused by a bad instalation.

Generally, when you install mroonga from a package (deb or rpm), at the end of the setup will be executed a post-setup script that executes some sql queries to install and define functions.

I don't know how you are installed the mroonga plugin, but to solve this puntual problem, you can try to do the follow sql sentence:

```
CREATE FUNCTION mroonga_command RETURNS STRING
  SONAME 'ha_mroonga.so';
```

More info here: https://mariadb.com/kb/en/creating-mroonga-user-defined-functions/

Josep.

libraries/classes/DatabaseInterface.php
```diff
@@ -560,6 +595,70 @@ static function ($a, $b) {
...
+    /**
+     * Returns if Mroonga is available to be used
+     *
+     * This is public to be used in the StructureComtroller, the first release
+     * of this function was looking Mroonga in the engines list but this second
+     *  method checks too that mroonga is installed succesfully
+     *
+     * @return int
+     */
+    public function hasMroongaEngine(): bool
```
williamdes, line 607: 🤔 returns bool :p ?

williamdes:
> CREATE FUNCTION mroonga_command RETURNS STRING
>   SONAME 'ha_mroonga.so';

Thank you, indeed I had a access denied to root user and not much more as an error so I just ignored it. Thank you for fixing the server and providing the query !
The code works fine

libraries/classes/Controllers/Database/StructureController.php
```diff
@@ -1054,6 +1054,17 @@ protected function getStuffForEngineTypeTable(
...
+                // The idea is to show the size only if Mroonga is available,
+                // in other case the old unknown message will appear
+                if ($this->dbi->hasMroongaEngine()) {
+                    [$currentTable, $formattedSize, $unit, $sumSize] = $this->getValuesForMroongaTable(
+                        $currentTable,
+                        $sumSize
+                    );
+                    break;
+                }
+                // no break
```
williamdes, line 1067:
```suggestion
                // no break, go to default case
```

libraries/classes/DatabaseInterface.php
```diff
@@ -400,6 +401,24 @@ public function getTablesFull(
                 $link
             );
 
+            // here, we check for Mroonga engine and compute the good data_length and index_length
+            // in the StructureController only we need to sum the two values as the other engines
+            foreach ($tables as $one_database_name => $one_database_tables) {
+                foreach ($one_database_tables as $one_table_name => $one_table_data) {
+                    if ($one_table_data['Engine'] !== 'Mroonga' || ! $this->hasMroongaEngine()) {
+                        continue;
+                    }
```
williamdes, line 410:
```suggestion
                    if ($one_table_data['Engine'] !== 'Mroonga') {
                        continue;
                    }
                    if (! $this->hasMroongaEngine()) {
                        continue;
                    }
```
This would be optimal and avoid wasting a query that would fail anyway, right ?

libraries/classes/DatabaseInterface.php
```diff
@@ -490,6 +509,22 @@ static function ($a, $b) {
 
                 $each_tables = $this->fetchResult($sql, 'Name', null, $link);
 
+                // here, we check for Mroonga engine and compute the good data_length and index_length
+                // in the StructureController only we need to sum the two values as the other engines
+                foreach ($each_tables as $table_name => $table_data) {
+                    if ($table_data['Engine'] !== 'Mroonga' || ! $this->hasMroongaEngine()) {
```
williamdes, line 515: same :)

libraries/classes/DatabaseInterface.php
josepsanzcamp: Sorry, copy-paste error!!!

libraries/classes/Controllers/Database/StructureController.php
josepsanzcamp: Done

libraries/classes/DatabaseInterface.php
josepsanzcamp: Done

libraries/classes/DatabaseInterface.php
josepsanzcamp: Done

libraries/classes/Controllers/Database/StructureController.php
```diff
@@ -1194,6 +1205,41 @@ protected function getValuesForInnodbTable(
         ];
     }
 
+    /**
+     * Get values for Mroonga table
+     *
+     * @param array $currentTable current table
+     * @param int   $sumSize      sum size
+     *
+     * @return array
```
williamdes, line 1214:
```suggestion
     * @return int[]
```

libraries/classes/DatabaseInterface.php
```diff
@@ -560,6 +603,70 @@ static function ($a, $b) {
...
+     *
+     * This is public to be used in the StructureComtroller, the first release
+     * of this function was looking Mroonga in the engines list but this second
+     *  method checks too that mroonga is installed succesfully
+     *
+     * @return bool true when the mroonga_command is found
+     */
+    public function hasMroongaEngine(): bool
+    {
+        static $bool = null;
```
williamdes, line 617: We have a Cache class in `$this->cache`, I think it would be the a solution to store the value
@mauriciofauth what do you think about this ?

williamdes (approved):
The code looks good, the `static $` should be fixed someway before merging

libraries/classes/DatabaseInterface.php
```diff
@@ -560,6 +606,66 @@ static function ($a, $b) {
...
+     *
+     * This is public to be used in the StructureComtroller, the first release
+     * of this function was looking Mroonga in the engines list but this second
+     *  method checks too that mroonga is installed succesfully
+     *
+     * @return bool true when the mroonga_command is found
+     */
+    public function hasMroongaEngine(): bool
+    {
+        if (! isset($this->object_list[__FUNCTION__])) {
```
williamdes, line 620: Sorry to be picky, could you move this to the Cache class ? and use a string key or just make it specific on the Cache class ?

libraries/classes/DatabaseInterface.php
```diff
@@ -138,6 +139,9 @@ class DatabaseInterface implements DbalInterface
     /** @var Cache */
     private $cache;
 
+    /** @var object_list Store all mroonga cache */
```
williamdes, line 142: I will explain why this is not very good:
- storing everything in the same array makes things not easy to analyse
- `@var object_list` is invalid and would be complicated to document since the stored objects/values and different
- it creates a property in a class that tries to stay small and slim
- having `__FUNCTION__` as a cache key is not very good, a string key would safer
- using this property will not allow tests to purge cache and will make testing difficult

Be sure I appreciate the effort you give and the code you make
I just try to give advice to make code as clean as possible

libraries/classes/DatabaseInterface.php
josepsanzcamp: Done

libraries/classes/DatabaseInterface.php
josepsanzcamp: Ok to all, but can you explain me what was the original problem with the static variable inside each function of the class??? why do you want to not use this feature of the language??? thanks in advance!!!

libraries/classes/DatabaseInterface.php
```diff
@@ -560,6 +603,70 @@ static function ($a, $b) {
...
+     *
+     * This is public to be used in the StructureComtroller, the first release
+     * of this function was looking Mroonga in the engines list but this second
+     *  method checks too that mroonga is installed succesfully
+     *
+     * @return bool true when the mroonga_command is found
+     */
+    public function hasMroongaEngine(): bool
+    {
+        $bool = $this->cache->getCachedTableContent(['mroonga_command']);
```
williamdes, line 617: looks good, I will just have to check if it does not interfere with some code :)

libraries/classes/DatabaseInterface.php
williamdes: To explain: using a static variable in a function makes it impossible to test correctly. You will not be able to reset the "cached" value once it is set. If you want you can try to write tests for your feature (see: DatabaseInterfaceTest) ?

If you put it in a class you can workaround this by using some hack code:
https://github.com/phpmyadmin/phpmyadmin/blob/cce4bc993cb3459c877a5e1bec89cf646be7b963/test/classes/AbstractNetworkTestCase.php#L98-L105

josepsanzcamp: OK, I understand now your recomendation, thanks!!!

libraries/classes/DatabaseInterface.php
josepsanzcamp: OK

MauricioFauth (approved):

williamdes:
Seems like I did once more the push mistake..
I re-created a branch for it on my fork with your commit: #16983
Sorry for the mistake, I think it is near ready to merge
I wrote some tests for the feature and discovered some details that could make the code slightly better and maybe also a bug

williamdes:
Thank you for pushing the change, I will fix up the branch and try not to make mistake once more :/

libraries/classes/StorageEngine.php
```diff
@@ -128,6 +133,96 @@ static function () use ($dbi) {
...
+            $result = $dbi->fetchSingleRow('SELECT mroonga_command(\'object_list\');', 'NUM');
+            $objectList = (array) json_decode($result[0] ?? '', true);
+            foreach ($objectList as $mroongaName => $mroongaData) {
+                /**
+                 * We only need the objects of table or column types, more info:
+                 * - https://groonga.org/docs/reference/commands/object_list.html#object-type
+                 * - https://groonga.org/docs/reference/commands/object_inspect.html#table-type-id
+                 * - https://groonga.org/docs/reference/commands/object_inspect.html#column-type-raw-id
+                 */
+                if (in_array($mroongaData['type']['id'], [48, 49, 50, 51, 64, 65, 72])) {
```
williamdes, line 185: Very good work @josepsanzcamp !
I like the links

williamdes (approved):

williamdes:
I think I will merge your work this week, just leaving open for other reviewers to comment if necessary

PR #16950, opened 2021-06-08
Add mutation tests with Infection

MauricioFauth:
https://infection.github.io/

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -9,6 +9,7 @@
 .weblate export-ignore
 codecov.yml export-ignore
 build.xml export-ignore
+infection.json.dist export-ignore
 phpcs.xml.dist export-ignore
 phpstan.neon.dist export-ignore
 phpstan-baseline.neon export-ignore
--- a/.github/workflows/mutation-tests.yml
+++ b/.github/workflows/mutation-tests.yml
@@ -1,0 +1,84 @@
+name: Mutation tests
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  tests:
+    name: Mutation tests with PHP ${{ matrix.php-version }}
+
+    runs-on: ${{ matrix.operating-system }}
+    if: "!contains(github.event.head_commit.message, '[ci skip]')"
+
+    strategy:
+      matrix:
+        php-version: ["8.0"]
+        operating-system: [ubuntu-latest]
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Fetch github.base_ref (for diffing)
+        if: ${{ github.base_ref != '' }}
+        run: |
+          git fetch --depth=1 origin +refs/heads/${{ github.base_ref }}:refs/remotes/origin/${{ github.base_ref }}
+
+      - name: Install Gettext
+        run: |
+          sudo apt-get install -y gettext
+
+      - name: Install PHP ${{ matrix.php-version }}
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php-version }}
+          coverage: pcov
+          ini-values: memory_limit=-1
+          tools: composer:v2, phive
+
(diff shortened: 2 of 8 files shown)


ibennetch:
This looks quite interesting to me.

.gitattributes
```diff
@@ -9,6 +9,7 @@
 .weblate export-ignore
 codecov.yml export-ignore
 build.xml export-ignore
+infection.json.dist export-ignore
```
williamdes, line 12: also probably needs to be removed in the release script
could you also add it to the forbidden files in the `check-release-excludes.sh` file ?

.github/workflows/mutation-tests.yml
```diff
@@ -0,0 +1,58 @@
...
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.composer-cache.outputs.dir }}
+          key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-composer-
+
+      - name: Install dependencies
+        run: |
+          composer update --no-interaction --no-progress --no-suggest
```
williamdes, line 48: `You are using the deprecated option "--no-suggest". It has no effect and will break in Composer 3. notice being printed in the output.`

.github/workflows/mutation-tests.yml
```diff
@@ -0,0 +1,58 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Install PHP ${{ matrix.php-version }}
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php-version }}
+          coverage: xdebug
```
williamdes, line 29: pcov is faster ;)

.github/workflows/mutation-tests.yml
```diff
@@ -0,0 +1,58 @@
...
+        run: |
+          composer update --no-interaction --no-progress --no-suggest
+
+      - name: Collect coverage report
+        run: |
+          php vendor/bin/phpunit --stop-on-failure
+
+      - name: Infection
+        run: |
+          php vendor/bin/infection -j2 --skip-initial-tests --coverage=build/logs --no-interaction --no-progress
```
williamdes, line 56: it would be nice to add this to composer.json and if the timeout if a problem then you could ask composer not to timeout

williamdes (approved):
Very nice idea !

.github/workflows/mutation-tests.yml
```diff
@@ -0,0 +1,70 @@
...
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php-version }}
+          coverage: pcov
+          ini-values: memory_limit=-1
+          tools: composer:v2, phive
+
+      - name: Install Infection
+        run: |
+          phive --no-progress install --target ./build/tools --trust-gpg-keys C5095986493B4AA0 infection
```
williamdes, line 39: Could you add a comment or a link where the key Id can be found ?

.gitattributes
williamdes: Just missing remove in the release script I think

scripts/create-release.sh
```diff
@@ -360,7 +360,7 @@ echo "* Removing unneeded files"
 rm -rf .github CODE_OF_CONDUCT.md DCO
 
 # Testsuite setup
-rm -f .travis.yml .scrutinizer.yml .jshintrc .weblate codecov.yml
+rm -f .travis.yml .scrutinizer.yml .jshintrc .weblate codecov.yml infection.json.dist
```
williamdes, line 363: I think it should be removed in the same line as the phpstan files
At: https://github.com/phpmyadmin/phpmyadmin/blob/55202a9532b584728d8fc2b78e89296a1136d282/scripts/create-release.sh#L483

williamdes (approved):
:100:

PR #16963, opened 2021-06-13
Run PHPUnit test in random order by default

MauricioFauth:
This prevents tests having hidden dependencies on each other.


williamdes (approved):
Are we ready or will tests fail very often until we fix all randomly failing tests?

mauriciofauth:
> Are we ready or will tests fail very often until we fix all randomly failing tests?

I will try to fix the failing seeds before merge but I don't think it will fail very often.

williamdes:
> > Are we ready or will tests fail very often until we fix all randomly failing tests?
> 
> I will try to fix the failing seeds before merge but I don't think it will fail very often.

I agree, maybe you can run them in process isolation locally to make some failing ones pop out?

mauriciofauth:
> I agree, maybe you can run them in process isolation locally to make some failing ones pop out?

Running in process isolation is different from running in random order. In random order you also catch cases where a test changes some global state of other test.

williamdes:
https://github.com/phpmyadmin/phpmyadmin/pull/16963/checks?check_run_id=2820763357#step:6:563

Good work, diff LGTM
Maybe you can use the repeat cli arg locally
Let me know if you need some help running tests until they fail

PR #16965, opened 2021-06-14
Run multi-arch tests with a fixed seed to investigate segfault

MauricioFauth:
Test on php 7 (arm64v8) with random seed 1623681052.

https://github.com/phpmyadmin/phpmyadmin/runs/2820763388


mauriciofauth:
It didn't fail this time.

PR #16972, opened 2021-06-18
Remove two symfony dev deps

williamdes:
We do not need them, it brings up too much code for only one class used !


MauricioFauth (commented):
It is a lot of code but it is for development only. Having the linter on our code base means that we will need to maintain it.

williamdes:
> It is a lot of code but it is for development only. Having the linter on our code base means that we will need to maintain it.

I agree, but have a look at the code, there is not much specific code. It's more wiring some code and show the results

I an other hand I would say it makes us more autonomous rather than relying on some code that is adapted to a specific Twig version range

MauricioFauth (changes requested):
The tests for the command are missing.

mauriciofauth:
I still don't like the idea of maintain the linter, but I'm not against merging this pull request.
Maybe we could work with Symfony/Twig to make the linter a standalone package.

williamdes:
I added some tests, let me know if it seems to be okay

MauricioFauth (approved):

PR #16984, opened 2021-06-29
Refactor `PhpMyAdmin\Plugins::getPlugins` method

MauricioFauth:


libraries/classes/Plugins.php
```diff
@@ -96,97 +98,66 @@ public static function getExport(string $type, bool $singleTable): array
...
-        global $skip_import;
-
-        $GLOBALS['plugin_param'] = $plugin_param;
-
-        $fullFsPathPluginDir = ROOT_PATH . $plugins_dir;
-
-        $handle = @opendir($fullFsPathPluginDir);
-        if (! $handle) {
+        try {
+            $files = new FilesystemIterator(ROOT_PATH . 'libraries/classes/Plugins/' . $type);
```
williamdes, line 126: I think you could use the folderopen way, we have a FileListing class but it has some useless regex code in this case
Anyway, good for now

It's just not the most performant method

libraries/classes/Plugins/Export/ExportXml.php
```diff
@@ -616,4 +603,12 @@ private function setTables(array $tables)
     {
         $this->tables = $tables;
     }
+
+    public function isAvailable(): bool
+    {
+        global $db;
+
+        // Can't do server export.
```
williamdes, line 611:
```suggestion
        // Can't do a server export without a database name.
```

composer.json
```diff
@@ -58,6 +58,7 @@
         "symfony/expression-language": "^5.2.3",
         "symfony/polyfill-ctype": "^1.17.0",
         "symfony/polyfill-mbstring": "^1.17.0",
+        "symfony/polyfill-php80": "^1.16",
```
williamdes, line 61: why is it needed ?

williamdes (approved):
Very good work on removing this global

composer.json
mauriciofauth: I used it for `str_starts_with`. We already have this package installed, I'm just using it.

```
composer why symfony/polyfill-php80
```

williamdes: Oh okay, if you plan to use it in some other places I agree 100% to use the polyfill

libraries/classes/Plugins.php
mauriciofauth: Using the Iterator was slower in my tests, but the difference was so small that it was insignificant.

libraries/classes/Plugins/Export/ExportXml.php
mauriciofauth: XML export is only available when exporting a database or a table, not when exporting the server.

PR #16989, opened 2021-07-03
Add support for PSR-7 request objects in controllers

MauricioFauth:
Related to https://github.com/phpmyadmin/phpmyadmin/issues/15507.


williamdes (approved):

libraries/classes/Routing.php
```diff
@@ -188,6 +187,6 @@ public static function callControllerForRoute(
 
         [$controllerName, $action] = $routeInfo[1];
         $controller = $container->get($controllerName);
-        $controller->$action($routeInfo[2]);
+        $controller->$action($request, $routeInfo[2]);
```
williamdes, line 190: Maybe there is a phpstan template way we could enforce the signature to be this on all public methods of a controller?
Or maybe just writing an unit test with some reflection could do it

mauriciofauth: I'll think a way of doing that. I'm planning in writing a wrapper around `nikic/fast-route`, maybe we could enforce it there.

williamdes (approved):
💯

PR #17005, opened 2021-07-11
Replace `scripts/fix-po-twig` with a Console command

MauricioFauth:


libraries/classes/Command/FixPoTwigCommand.php
```diff
@@ -0,0 +1,94 @@
...
+                return $replace[0] . ':0';
+            },
+            $pot
+        );
+        if ($pot === null) {
+            return Command::FAILURE;
+        }
+
+        /* Write fixed file */
+        $handle = fopen(self::POT_FILE, 'w');
```
williamdes, line 79: A file out contents would be quicker

williamdes (approved):
Nice change 💯

PR #16481, opened 2020-11-15
Fix wildcard database name

shucon:
### Description

Please describe your pull request.

Fixes https://github.com/phpmyadmin/phpmyadmin/issues/16477

Before submitting pull request, please review the following checklist:

- [ ] Make sure you have read our [CONTRIBUTING.md](https://github.com/phpmyadmin/phpmyadmin/blob/master/CONTRIBUTING.md) document.
- [ ] Make sure you are making a pull request against the correct branch. For example, for bug fixes in a released version use the corresponding QA branch and for new features use the _master_ branch. If you have a doubt, you can ask as a comment in the bug report or on the mailing list.
- [ ] Every commit has proper `Signed-off-by` line as described in our [DCO](https://github.com/phpmyadmin/phpmyadmin/blob/master/DCO). This ensures that the work you're submitting is your own creation.
- [ ] Every commit has a descriptive commit message.
- [ ] Every commit is needed on its own, if you have just minor fixes to previous commits, you can squash them.
- [ ] Any new functionality is covered by tests.

Issue #16477, opened 2020-11-12
No Option to enter TABLE specific permissions when database name contains _ (underscore)


Schwuuuuup:
### Describe the bug

I was used to grant permissions for single tables instead of permissions for whole databases. For this I opened the useraccount for editing, and klicked on the button "Database" and then on "Table" to select specific Tables for access.
The button "Table" is missing for all Databases with names containing an underscore.

### To Reproduce

1. Rename a database to have its name contain underscores like "test_with_underscore"
2. Try to grand table specific permissions:
User accounts->edit permissions->select "Database" button -> see that there does not appear a button "Table"

See Screenshots

### Expected behavior

An underscore in the name should not influence what permissions can be granted.
sie examlpes in the screenshots for "testwithoutunderscore"

### Screenshots
![selectdatabase_without_underscore](https://user-images.githubusercontent.com/16288510/98940972-2796ad00-24ec-11eb-9c63-39f598dafe33.png)
![selectdatabase_with_underscore](https://user-images.githubusercontent.com/16288510/98940975-28c7da00-24ec-11eb-8a1a-8b28f264fc6f.png)
![withoutunderscore_button_is_there](https://user-images.githubusercontent.com/16288510/98940976-28c7da00-24ec-11eb-8d65-3891d28356fb.png)
![with_underscore_button_is_missing](https://user-images.githubusercontent.com/16288510/98940977-29607080-24ec-11eb-8d69-c975b757b9a4.png)

### Server configuration

- Operating system: Red hat Enterprise Linux 7
- Web server: Apache httpd-2.4.6-95.el7.x86_64
- Database version: MariaDB-server-10.4.16-1.el7.centos.x86_64
- PHP version: PHP 7.3.23 (cli) (built: Oct  1 2020 20:38:17) ( NTS )
- phpMyAdmin version: 5.0.4

### Client configuration

- Browser: Vivaldi
- Operating system: MX-Linux

### Additional context

The Problem must have showed up some time in the past. I guess it was about half a year ago when I last used this feature and back then it worked fine.

williamdes:
Hi @Schwuuuuup !
I can confirm this issue, this is quite strange but I reproduced it.

@shucon would you agree to work on this one ?

shucon:
Sure!!

ibennetch:
[...]

PR #17010, opened 2021-07-13
Fix #16994 - Fixing the issue when assigning database to user with '_'

### LINKED ISSUE #16994: Issue assigning database to user with '_' in database name
### Describe the bug

When a database name has a '_' preceded by a digit in its name and a user is assigned to the database via the multiselect list on the 'database' tab of the 'Edit privileges' page, the name of the database is corrupted because of the '\\' needed to escape the database name in the multiselect list.  If additional databases with '_', but no digit preceding, are selected they are also affected, whereas if only a database with a '_' and no digit preceding is selected it is not affected.

### To Reproduce

1. Go to 'phpmyadmin/index.php?route=/server/privileges&username=temp&hostname=localhost'
2. Click on 'database' tab
3. Choose database with '<digit>_' in name
4. Click 'go'
5. View list of databases assigned to user

### Expected behavior

The database name assigned to user is the same as that chosen form the multiselect list

### Screenshots

Screen Shot 2021-07-06 at 6 33 07 AM
Screen Shot 2021-07-06 at 6 33 38 AM

### Server configuration

- Operating system: macos
- Web server: Apache/2.4.48
- Database version: 10.3.28-MariaDB
- PHP version: 7.4.20
- phpMyAdmin version: 5.1.1

### Client configuration

- Browser: Chrome
- Operating system: macos Big Sur

### Additional context

I have seen this bug for a while.  You can get around it by typing the database name in the box to the right of the multiselect list.

iifawzi:

### Description
The problem was that if multiple databases were chosen, they were not unescaped and will be added as they are to `mysql`.`db`, that's why they appear with the backslash in the privileges view

Fixes #16994

### DIFF
--- a/libraries/classes/Controllers/Server/PrivilegesController.php
+++ b/libraries/classes/Controllers/Server/PrivilegesController.php
@@ -229,7 +229,7 @@ class PrivilegesController extends AbstractController
                         ($username ?? ''),
                         ($hostname ?? ''),
                         ($tablename ?? ($routinename ?? '')),
-                        ($db_name ?? ''),
+                        (Util::unescapeMysqlWildcards($db_name) ?? ''),
                         $itemType
                     );
                 }


williamdes (approved):
Awesome !
I will test it and merge it with your second PR

williamdes:
Works fine, but the UI on the page after submit still escapes twice the char. I did not find a quick fix for that

libraries/classes/Controllers/Server/PrivilegesController.php
```diff
@@ -229,7 +229,7 @@ public function index(): void
                         ($username ?? ''),
                         ($hostname ?? ''),
                         ($tablename ?? ($routinename ?? '')),
-                        ($db_name ?? ''),
+                        (Util::unescapeMysqlWildcards($db_name) ?? ''),
```
williamdes, line 232: This would probably fail at some point because of a null value
The fix I pushed as 7c17889d780fe839bfe12c66b5a8fe01f68bf503 is:
```suggestion
                        (Util::unescapeMysqlWildcards($db_name ?? '')),
```

iifawzi:
> Works fine, but the UI on the page after submit still escapes twice the char. I did not find a quick fix for that

Ahh, I didn't notice it. I've found a fix for this, by unescaping the name of the databases in this else: 

https://github.com/phpmyadmin/phpmyadmin/blob/8500a6c5a15117b68355ba966aa976cb1a31b951/libraries/classes/Server/Privileges.php#L3265-L3273

To: 

```PHP
        if (! is_array($dbname) && strlen($dbname) > 0) {
            $_params['dbname'] = $dbname;
            if (strlen($tablename) > 0) {
                $_params['tablename'] = $tablename;
            }
        } else {
            foreach ($dbname as &$db_name) {
                $db_name = Util::unescapeMysqlWildcards($db_name);
            }
            unset($db_name);
            $_params['dbname'] = $dbname;
        }
```

It wouldn't affect anything else, we can use a separate array maybe called `unescaped_dbname` and pass it to twig, instead of mutating the `dbname` array if it's not preferred. 
I've tested and both solutions worked. What do you think? @williamdes

williamdes:
> I've tested and both solutions worked. 

Thank you !

> It wouldn't affect anything else, we can use a separate array maybe called unescaped_dbname and pass it to twig, instead of mutating the dbname array if it's not preferred.

I think this would be correct and avoid any possible side effects, using the unescaped_dbname option

iifawzi:
> > I've tested and both solutions worked.
> 
> Thank you !
> 
> > It wouldn't affect anything else, we can use a separate array maybe called unescaped_dbname and pass it to twig, instead of mutating the dbname array if it's not preferred.
> 
> I think this would be correct and avoid any possible side effects, using the unescaped_dbname option

Alright, should I create another PR for this, or I can add the commit to this PR?

williamdes:
> > > I've tested and both solutions worked.
> > 
> > 
> > Thank you !
> > > It wouldn't affect anything else, we can use a separate array maybe called unescaped_dbname and pass it to twig, instead of mutating the dbname array if it's not preferred.
> > 
> > 
> > I think this would be correct and avoid any possible side effects, using the unescaped_dbname option
> 
> Alright, should I create another PR for this, or I can add the commit to this PR?

You can add one more commit and I will craft a merge for it even if there is no PR ;)

iifawzi:
@williamdes, I've added the commit https://github.com/iifawzi/phpmyadmin/commit/b116bd2116eb26d23bd307213072c3b443f673bc, it works correctly, but there were two approaches to handle the case if `$dbname` is not an array, I've decided to make a check whether it's an array or not, if it's an array use `$unescaped_dbname` otherwise use `$dbname` 
```PHP
'database' => (is_array($dbname) && count($dbname) > 1) ? $unescaped_dbname : $dbname,
```

The other choice is to assign `$dbname` to `$unescaped_dbname` from the first place in the if statement,  but I think it might be misleading, thus I went with the first approach. 

Also, the statement I've added in the else can be avoided by converting the main `if` statement to ` else if `, feel free to commit the changes if you wish. 

Thank you for following up on this.

iifawzi:
I've figured a better approach that could solve both, the main issue and the UI issue, by unescaping the array elements as in the case of single db selected, from the beginning before distributing this array to other functions, and this could be done by escaping the array from ` getDataForDBInfo() ` that receives the selected elements from the first place.  

I've force pushed the changes https://github.com/iifawzi/phpmyadmin/commit/4056f24655aea22489e760c7013dfbcb9c5ecfe8 , and both issues solved 

Screen Shot 2021-07-14 at 2 51 22 AM
Screen Shot 2021-07-14 at 2 50 31 AM

Sorry for this mess, I just felt like we're doing something wrong be making multiple changes in multiple places, investigated more and found this fix :")

williamdes:
Thanks for your fix, but I can not confirm it passes my tests. Maybe you should open a PR because if I understand well the previous fix also needs to be reverted with this solution

### Before
image

### After

image


After there is no escaping at all and I think it is wrong
Maybe go for the twig solution ?

iifawzi:
Yes, if it's an accurate fix, the old one should be reverted, but I'm confused now, what do you mean by `there's no escaping at all` I think the backslash shown in the query is the wrong one - in the before image - , because this means that the grant is made on the table with the name that contains an escaping characters, which's wrong I think and will be added to `mysql`.`db` with the escaping characters. could you check the `mysql`.`db` table to check whether the db name is added correctly without the backslashes or not?  if it's shown without the backslashes so I'm mistaken. 

Anyway, the last change I've made is only unescaping the array elements, as in the case when single db name is selected (shown in the else statement), if you could test with single db selected, the same behavior is happening, the query will be shown unescaped. 

https://github.com/iifawzi/phpmyadmin/blob/4056f24655aea22489e760c7013dfbcb9c5ecfe8/libraries/classes/Server/Privileges.php#L2920-L2933

I'm asking because as shown in the code above, we were unescaping in the case if single db is selected, I've just unescaped also in the case of multiple db is selected, so logically it shouldn't make any difference. 

Here's the `mysql`.`db` table, it shows that the grant are added successfully on the correct db names with the current implementation. 
 
Screen Shot 2021-07-14 at 1 51 52 PM

before, will lead to this: 
Screen Shot 2021-07-14 at 1 56 14 PM
You will notice that the grant is added to incorrect db name, it's added to the escaped name, which's wrong as per my understand. 

I might be missing something though

williamdes:
> there's no escaping at all

I mean the last implementation you sent is not good according to 
> The _ and % wildcards are permitted when specifying database names in GRANT statements that grant privileges at the database level (GRANT ... ON db_name.*). This means, for example, that to use a _ character as part of a database name, specify it using the \ escape character as \_ in the GRANT statement, to prevent the user from being able to access additional databases matching the wildcard pattern (for example, GRANT ... ON `foo\_bar`.* TO ...).

Source: https://dev.mysql.com/doc/refman/8.0/en/grant.html

> Yes, if it's an accurate fix, the old one should be reverted

The actual fix you proposed is good (the one actually on QA branch). It only escapes `_` using `\` once.


Very nice avatar by the way :)

iifawzi:
> Source: https://dev.mysql.com/doc/refman/8.0/en/grant.html

Ohh, ok It make sense now, but doesn't this mean that there's a bug in the case when single db is selected? 
Screen Shot 2021-07-15 at 12 25 39 AM
it's not escaped at all too 👀 . 

Also, do you still prefer adding a commit to this repo? with the twig fix? or creating a separate PR?

> Very nice avatar by the way :)

Thank you! taken on the last day of my fourth college year, one year is remaining though 🤦🏻

williamdes:
> Also, do you still prefer adding a commit to this repo? with the twig fix? or creating a separate PR?

This is taking much more space than expected, to have a better overview a new PR would be better :)

> Ohh, ok It make sense now, but doesn't this mean that there's a bug in the case when single db is selected?

Yes, I think this is a bug, can you create an user to confirm the bug ?

iifawzi:
> Yes, I think this is a bug, can you create an user to confirm the bug ?

Yes, the same 
image

I think, this's the cause of the error, I didn't tested yet. 
https://github.com/phpmyadmin/phpmyadmin/blob/a0410aa77d903bec9ed72ec26e317094cd66d6e1/libraries/classes/Server/Privileges.php#L2929

> This is taking much more space than expected, to have a better overview a new PR would be better :)

Alright, I will create a separate PR with the twig fix discussed above, and the fix of the single db issue, if solved.

iifawzi:
Changes are almost done, but I've noticed something that I couldn't understand.
is `|escape_mysql_wildcards` intended?
https://github.com/phpmyadmin/phpmyadmin/blob/2f59c9b3bee42577edd09d659df03b7c0db3cc79/templates/server/privileges/privileges_summary.twig#L59

Why do we escape the already escaped names? if the database name is `test_db`, the code above will lead to `test_\\db` because it escaped twice, by the backend and by twig. 

if it's not intended and made by mistake, All the issues can be solved will less code, and without these changes everywhere. 
Thank you for your patience @williamdes, it's almost done, I've prepared the code for both ways :")

williamdes:
> Changes are almost done, but I've noticed something that I couldn't understand.
is |escape_mysql_wildcards intended?

Yes, I was searching for it, it is a twig extension we added (https://github.com/phpmyadmin/phpmyadmin/blob/QA_5_1/libraries/classes/Twig/UtilExtension.php#L165)
That could be super helpful meaning you would have less code to change

> if it's not intended and made by mistake, All the issues can be solved with less code and in a more cleaner way.
Thank you for your patience @williamdes, it's almost done, I've prepared the code for both ways :)

Let's go ;)

iifawzi:
I didn't get it :(, is it accurate to escape the db name twice? it's already escaped from the backend `escaped_databases[loop.index0]`, why escaping it again using the twig extension?
https://github.com/phpmyadmin/phpmyadmin/blob/2f59c9b3bee42577edd09d659df03b7c0db3cc79/templates/server/privileges/privileges_summary.twig#L59

williamdes:
> I didn't get it :(, is it accurate to escape the db name twice? it's already escaped from the backend `escaped_databases[loop.index0]`, why escaping it again using the twig extension?
> https://github.com/phpmyadmin/phpmyadmin/blob/2f59c9b3bee42577edd09d659df03b7c0db3cc79/templates/server/privileges/privileges_summary.twig#L59

Oh snap, I was thinking it was to "un" escape the wildcards
If you need to **un**escape you may add one more twig filter

> is it accurate to escape the db name twice?

No, it makes no sense. But maybe there is one case where it is not escaped 🤔 

Hey, look what I found when searching about this line: #16477
Maybe the mess comes from #16481 ..

iifawzi:
> No, it makes no sense. But maybe there is one case where it is not escaped
Hmm, if this's the case, it means that it will lead double escaping sometime, Anyway, I will test that my changes are working as expected for all the tests that I could think about, and then I will create a PR

> Maybe the mess comes from #16481.  
it seems like it made some conflicts, I will try to investigate more

williamdes:
> it seems like it made some conflicts, I will try to investigate more

Feel free to completely revert the fix and the make a fix for the fixed issue and the GUI + issue already fixed today by your first fix

Thank you so much for your work!

PR #17015, opened 2021-07-15
Fix: Fixing all the issues related to assigning privileges to databases with '_'

iifawzi:

### Description

This PR fixes all the bugs discussed at https://github.com/phpmyadmin/phpmyadmin/pull/17010 and the bug mentioned in #16994. 

Fixes: 

- The grant/revoke queries will be escaped once - Previously, when single db is selected, no escaping at all, and when multiple db are selected got escaped twice -  
- All the UI that's responsible to show the databases names are unescaped for the user to see the actual names. 
- All the links to edit/revoke are fixed. 
- Multiple select list will only show the databases that aren't already granted privileges. 

Screen Shot 2021-07-16 at 12 37 43 AM
Screen Shot 2021-07-16 at 12 37 06 AM
Screen Shot 2021-07-16 at 12 36 56 AM
Screen Shot 2021-07-16 at 12 47 11 AM

### DIFF
--- a/libraries/classes/Server/Privileges.php
+++ b/libraries/classes/Server/Privileges.php
@@ -514,8 +514,7 @@ class Privileges
             return 'SELECT * FROM `mysql`.`db`'
                 . " WHERE `User` = '" . $this->dbi->escapeString($username) . "'"
                 . " AND `Host` = '" . $this->dbi->escapeString($hostname) . "'"
-                . " AND '" . $this->dbi->escapeString(Util::unescapeMysqlWildcards($db)) . "'"
-                . ' LIKE `Db`;';
+                . " AND `Db` = '" . $this->dbi->escapeString($db) . "'";
         }
 
         return 'SELECT `Table_priv`'
@@ -1589,12 +1588,12 @@ class Privileges
         ];
         switch ($linktype) {
             case 'edit':
-                $params['dbname'] = Util::escapeMysqlWildcards($dbname);
+                $params['dbname'] = $dbname;
                 $params['tablename'] = $tablename;
                 $params['routinename'] = $routinename;
                 break;
             case 'revoke':
-                $params['dbname'] = Util::escapeMysqlWildcards($dbname);
+                $params['dbname'] = $dbname;
                 $params['tablename'] = $tablename;
                 $params['routinename'] = $routinename;
                 $params['revokeall'] = 1;
@@ -2926,7 +2925,7 @@ class Privileges
             } else {
                 $unescaped_db = Util::unescapeMysqlWildcards($dbname);
                 $db_and_table = Util::backquote($unescaped_db) . '.';
-                $return_db = $unescaped_db;
+                $return_db = $dbname;
             }
             if (isset($tablename)) {
                 $db_and_table .= Util::backquote($tablename);
--- a/libraries/classes/Controllers/Server/PrivilegesController.php
+++ b/libraries/classes/Controllers/Server/PrivilegesController.php
@@ -229,7 +229,7 @@ class PrivilegesController extends AbstractController
                         ($username ?? ''),
                         ($hostname ?? ''),
                         ($tablename ?? ($routinename ?? '')),
-                        Util::unescapeMysqlWildcards($db_name ?? ''),
+                        ($db_name ?? ''),
                         $itemType
                     );
                 }
--- a/templates/server/privileges/privileges_summary.twig
+++ b/templates/server/privileges/privileges_summary.twig
@@ -56,7 +56,7 @@
             {%- if databases is not empty %}
                 <select name="pred_dbname[]" multiple="multiple">
                     {% for database in databases %}
-                        <option value="{{ escaped_databases[loop.index0]|escape_mysql_wildcards }}">
+                        <option value="{{ escaped_databases[loop.index0] }}">
                             {{ database }}
                         </option>
                     {% endfor %}


williamdes:
Hi @iifawzi 
Your work is great, that said there is something wrong 
Let me explain: you need to show the escaping if there is some, but not escape the escaping as we can see when granting permission

Currently you removed all escaping, this is not right because as you can see on my example `a_%` is a wildcard example. And using your PR it would prevent distinguishing between literal char and escaped

So you need to add back escaping on DB list, edit page. That said the queries look good !

I have a theory about why there is two `\` used, maybe it is to escape the `\` at the insert moment, that would explain the actual GUI

### mysql.db

image

### Show dbs before
image

### Show dbs after

Sélection_2021071602:38:11001

iifawzi:
I can see that in your `mysql.db` table, the `a_%` and `a_hidden` are not escaped, I think this incorrect, and this's the actual bug at `QA_5_1`, right? if this screenshot is taken from using this PR implementation, there's something wrong then, but i'm not able to re-produce such a case, did you remove all the grants that were made before testing this PR? 
> a_% is a wildcard example. And using your PR it would prevent distinguishing between literal char and escaped

How can I encounter such an issue? what do you mean by `distinguishing between literal char and escaped?`

I'm asking because, if the queries look good, so it means the GUI looks good too, because i'm just unescaping the name used in the query, and used it in the GUI, if the. GUI can be misleading in any case, so it means there will be something wrong in the queries I think. how can the query looks good, and the GUI have bugs?

It took too long I know, thank your for your patience, I'm just wanna be sure that I understood your points before adding back the escaping on DB list, edit page GUIs.

Thank you again.

williamdes:
> I can see that in your mysql.db table, the a_% and a_hidden are not escaped, I think this's incorrect,

No, it's indended and is a wildcard example

> How can I encounter such an issue? what do you mean by distinguishing between literal char and escaped?

You can grant a user the use of `a_foo%`  using the box after the list of tables, If I am right it will mean `a` any char `foo` any string

I am not sure if this is more clear, but there is a case where `_` means one char and `\_` means literal `_`

williamdes:
So you only need to escape on database lists add to make `_` chars escaped, but if the user enters `_` in the box then keep it not escaped :)

iifawzi:
> You can grant a user the use of a_foo% using the box after the list of tables, If I am right it will mean a any char foo any string

Ohhh... I got it. Nice! I never thought that this's possible.

libraries/classes/Server/Privileges.php
```diff
@@ -514,8 +514,7 @@ public function getSqlQueryForDisplayPrivTable($db, $table, $username, $hostname
             return 'SELECT * FROM `mysql`.`db`'
                 . " WHERE `User` = '" . $this->dbi->escapeString($username) . "'"
                 . " AND `Host` = '" . $this->dbi->escapeString($hostname) . "'"
-                . " AND '" . $this->dbi->escapeString(Util::unescapeMysqlWildcards($db)) . "'"
-                . ' LIKE `Db`;';
+                . " AND `Db` = '" . $this->dbi->escapeString($db) . "'";
```
iifawzi, line 517: I've encountered another issue regarding the privileges table, if we have a privileges on `a_%` and `a_hidden` 
before, if we tried to edit the privileges on `a_hidden`, the privileges of `a_%` will be shown instead. (can be tested on https://demo.phpmyadmin.net/master/ ), I've fixed it by looking for an exact match instead of using `LIKE`, and removed the unescaping function, I've tested all the cases that crossed my mind, I hope that I'm not missing anything.

iifawzi (commented):
I've tested the mentioned cases in the PR's description and it worked, and the GUI now is working as expected, Hopefully :d.

williamdes (approved):
This all looks very nice, I will test it !

williamdes (approved):
Awesome !
It fixes all bugs

williamdes:
Just found something (#16994 is back)
image
Missing a tab
image

iifawzi:
Hmm, I think based on https://github.com/phpmyadmin/phpmyadmin/issues/16477#issuecomment-727145403, the issue was due to considering the `underscore` as wildcard - the bug mentioned here https://github.com/phpmyadmin/phpmyadmin/pull/17010#issuecomment-880260439 - 
since this commit fixed https://github.com/phpmyadmin/phpmyadmin/pull/17010#issuecomment-880260439, so #16477 is fixed too, here's an example granting a permissions on db with underscore `bar_1_prod` 

image

The issue you've mentioned is regarding granting a permissions on a wildcard example like `a_%` or `bar_1_prod`, logically it shouldn't be allowed I think, we have no actual tables. I've also tested in https://demo.phpmyadmin.net/master/index.php and it have thrown an error `table doesn't exist`, which make sense I think. 
image


the mess comes from that when https://github.com/phpmyadmin/phpmyadmin/pull/16481 is introduced, this bug https://github.com/phpmyadmin/phpmyadmin/pull/17010#issuecomment-880260439 wasn't fixed. which leaded to allowing the `table` tab to appear even for wildcards DBs `a_%` or `bar_1_prod`, which's is incorrect as per my understand, I might be missing something though

williamdes:
@sudo-robot deploy

williamdes:
Hmm, you are right. This was probably a bug, I could not replicate it on the deployed server.
And yes it makes sense that you can not add DBs to a wildcard privilege but that you can on a normal one.
Probably I made a mistake in my testing, I will let you know ;)

iifawzi:
let me know if you've found anything that could be fixed, and thank you for following up with this, your guidance, and your fast replies @williamdes, You're encouraging me to keep contributing

williamdes:
> let me know if you've found anything that could be fixed, and thank you for following up with this, your guidance, and your fast replies @williamdes, You're encouraging me to keep contributing

I am so happy to have so great and involved contributions ! ❤️ 

Conclusion: it works better when I use an escaped version 🤦🏻 

image

williamdes:
✅ Tests fixed: 25c2cfe699b96272a2380b6110ee4be84cb1eabb
✅ And one new test: 6d5c5e69d691637dd6d67bef917545d04ffd6efe

ℹ️ You can run tests with `./vendor/bin/phpunit --no-coverage --exclude-group=selenium`

PR #17021, opened 2021-07-18
Use negative length with mb_substr where possible

MoonE:
Use negative length parameter to simplify mb_substr calls.


williamdes (approved):
This really looks good, but do you have a clue why we where doing this this way 🤔 ?

MoonE:
I don't know a reason why it was done that way, maybe whoever implemented this did not know negative values can be used. It's been this way since the gis classes were added.

MauricioFauth (approved):

mauriciofauth:

PR #17032, opened 2021-07-23
Prevent default when scrolling in gis visualization

MoonE:
The page scrolls when zooming in GIS visualization via mouse wheel because preventDefault is not called, 
It was mistakenly removed in 2839b306d5a84b31339f83fa6f38c101f4573202 instead of using `{ passive: false }` with the addEventListener call. Doesn't seem like jquery has an interface for it, so this PR does not use jquery for the wheel event (https://github.com/jquery/jquery/issues/2871).

Passive event detection copied from https://github.com/openlayers/openlayers/blob/82956aaf5e5c8ced63612a89ef4b8e81052c4817/src/ol/has.js#L61-L79 (https://github.com/openlayers/openlayers/pull/10286)

Browser support for wheel event:
https://developer.mozilla.org/en-US/docs/Web/API/Element/wheel_event#browser_compatibility
Passive option, default passive for wheel event:
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#browser_compatibility

Please tell me if I should rebase.

### DIFF
--- a/js/src/config.js
+++ b/js/src/config.js
@@ -2,6 +2,8 @@
  * Functions used in configuration forms and on user preferences pages
  */
 
+/* exported PASSIVE_EVENT_LISTENERS */
+
 var configInlineParams;
 var configScriptLoaded;
 
@@ -877,3 +879,23 @@ function offerPrefsAutoimport () {
     });
     $cnt.show();
 }
+
+/**
+ * @type {boolean} Support for passive event listener option
+ */
+var PASSIVE_EVENT_LISTENERS = (function () {
+    var passive = false;
+    try {
+        var options = Object.defineProperty({}, 'passive', {
+            get: function () {
+                return (passive = true);
+            },
+        });
+
+        window.addEventListener('_', null, options);
+        window.removeEventListener('_', null, options);
+    } catch (error) {
+        // passive not supported
+    }
+    return passive;
+}());
--- a/js/src/table/gis_visualization.js
+++ b/js/src/table/gis_visualization.js
@@ -6,7 +6,7 @@
  * @requires    vendor/jquery/jquery.mousewheel.js
  */
 
-/* global drawOpenLayers */ // templates/table/gis_visualization/gis_visualization.twig
+/* global drawOpenLayers PASSIVE_EVENT_LISTENERS */ // templates/table/gis_visualization/gis_visualization.twig
 
 // Constants
 var zoomFactor = 1.5;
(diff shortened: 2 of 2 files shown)


MoonE:
Please tell me which branch to use as base.

williamdes:
> Please tell me which branch to use as base.

If this qualifies as a bug fix QA is the right branch, @mauriciofauth do you think it should go to QA ?

williamdes:
Hi @MoonE 
Could I revert https://github.com/phpmyadmin/phpmyadmin/commit/2839b306d5a84b31339f83fa6f38c101f4573202 on QA_5_1 and merge your work into master ?
Does it sound okay ?

MoonE:
Reverting 2839b306d5a84b31339f83fa6f38c101f4573202 will only fix this for browsers that are listed in [addEventListener compatibility](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#browser_compatibility) with  `options.passive parameter defaults to true for wheel and mousewheel events` as not  supported.
That would be Internet Explorer and Safari and outdated versions of the other browsers.

But I'm with merging this in whatever branch you prefer.

williamdes:
> Reverting [2839b30](https://github.com/phpmyadmin/phpmyadmin/commit/2839b306d5a84b31339f83fa6f38c101f4573202) will only fix this for browsers that are listed in [addEventListener compatibility](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#browser_compatibility) with `options.passive parameter defaults to true for wheel and mousewheel events` as not supported.
> That would be Internet Explorer and Safari and outdated versions of the other browsers.
> 
> But I'm with merging this in whatever branch you prefer.

Thank you so much for explaining this so clearly, I really appreciate
~~The browser compatibility file seems not to include them anyway~~ (not sure I understood in the right way the compatibility part according to the test on Chrome): https://github.com/phpmyadmin/phpmyadmin/blob/QA_5_1/.browserslistrc

~~So I will go for a revert on QA, that will avoid any unpredictable side effects ;)~~

After testing I found out that I maybe did not understand correctly what you explained, and Chrome was still scrolling

So I will probably go with a merge into QA 5.1 since Chrome is affect with or without the revert.

williamdes:
Hi @ahocevar  
Can we use your code for this ? (searched on GH and it seems like you are the only one to have this exact peace of code, It is not anything crazy/special but I still prefer to ask)

ahocevar:
Thanks for asking @williamdes. Please feel free to use that code.

williamdes (approved):
Awesome, it works perfectly !

PR #17036, opened 2021-07-25
Fix: Don't add multiple OpenLayers maps, remove listeners on dispose

MoonE:
With 5.1 the map is only created when it is first shown, but every second time
switching to the map a new instance is created because zero is returned when no new
map was created.

OpenLayers adds a resize listener to the window when a target is defined to
handle size adjustments, but this means setTarget(null) has to be called to
fully dispose of the added listener when the map is no longer needed.

### DIFF
--- a/js/src/table/gis_visualization.js
+++ b/js/src/table/gis_visualization.js
@@ -21,6 +21,8 @@ var scale = defaultScale;
 
 /** @type {SVGElement|undefined} */
 var gisSvg;
+/** @type {ol.Map|undefined} */
+var map;
 
 /**
  * Zooms and pans the visualization.
@@ -129,16 +131,14 @@ function initGISVisualization () {
     zoomAndPan();
 }
 
-function drawOpenLayerMap (openLayerCreate) {
+function drawOpenLayerMap () {
     $('#placeholder').hide();
     $('#openlayersmap').show();
     // Function doesn't work properly if #openlayersmap is hidden
-    if (!openLayerCreate) {
+    if (typeof map !== 'object') {
         // Draws openStreetMap with openLayers
-        drawOpenLayers();
-        return 1;
+        map = drawOpenLayers();
     }
-    return 0;
 }
 
 function getRelativeCoords (e) {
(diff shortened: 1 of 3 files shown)


js/src/table/gis_visualization.js
```diff
@@ -159,16 +160,15 @@ function initGISVisualization () {
...
     // Function doesn't work properly if #openlayersmap is hidden
-    if (!openLayerCreate) {
+    if (!map) {
         // Draws openStreetMap with openLayers
-        drawOpenLayers();
-        return 1;
+        map = drawOpenLayers();
     }
-    return 0;
+    return !!map;
```
williamdes, line 171:
```suggestion
    return !!map;
```
Could this syntax be improved for readability ?

MoonE: It's actually not needed anymore, I removed it now.

js/src/table/gis_visualization.js
```diff
@@ -210,18 +209,20 @@ AJAX.registerTeardown('table/gis_visualization.js', function () {
     $(document).off('click', '#up_arrow');
     $(document).off('click', '#down_arrow');
     $('.vector').off('mousemove').off('mouseout');
+    if (map) {
```
williamdes, line 212: Could you add a comment about why it is there and what it does ?

js/src/table/gis_visualization.js
```diff
@@ -19,6 +19,7 @@ var y = 0;
 var scale = 1;
 
 var svg;
+var map;
```
williamdes, line 22: Maybe a jsdoc block could add a type or just something saying object|undefined

js/src/table/gis_visualization.js
```diff
@@ -159,16 +160,14 @@ function initGISVisualization () {
     zoomAndPan();
 }
 
-function drawOpenLayerMap (openLayerCreate) {
+function drawOpenLayerMap () {
     $('#placeholder').hide();
     $('#openlayersmap').show();
     // Function doesn't work properly if #openlayersmap is hidden
-    if (!openLayerCreate) {
+    if (!map) {
```
williamdes, line 167:
```suggestion
    if (typeof map !== 'object') {
```

williamdes (approved):

williamdes (approved):

williamdes:
Confirmed
image

williamdes:
Hi @MoonE 

Could you please rebase to fix conflicts ?
(I might do it, depending on who is the quicker to do)

williamdes:
Hi @MoonE 
I tried your PR and the bug of this screenshot still exists: https://github.com/phpmyadmin/phpmyadmin/pull/17036#issuecomment-896021399

Just spam check,un-check the opensstreemaps checkbox

Maybe I did not properly test your fix ?

MoonE:
I can't reproduce it.
Are you sure you tested with correct changes?

williamdes:
Yup, let's ask my bot to deploy your work
@sudo-robot deploy

williamdes:
Indeed, on the deployed instance on 10.5 MariaDB I can not reproduce the issue
Will try once more locally

https://pr-17036-phpmyadmin.public-instance.eu/index.php?route=/sql&db=opengis&table=world_borders&pos=0&server=2

PR #17040, opened 2021-07-26
Fix #18287 - gis vis pos and limit

### LINKED ISSUE #18287: Number of lines ignored in GIS visualization after a search
The field "Number of lines" doesn't have any effect:
image
When I change it, it always shows the number of rows selected before in Search screen or Browse screen (25 for me).
Even lowering such value doesn't have any effect.

Changing the number of rows before clicking GIS visualization works:
image
image
And now, it finally shows more polygons.

It was very counterintuitive and took time to figure it out, I think this is an UI bug.

MoonE:
The pos and limit form fields in the gis visualization were ignored because they are read from $_GET instead of $_POST.

### DIFF
--- a/libraries/classes/Controllers/Table/GisVisualizationController.php
+++ b/libraries/classes/Controllers/Table/GisVisualizationController.php
@@ -107,15 +107,16 @@ final class GisVisualizationController extends AbstractController
             $visualizationSettings['spatialColumn'] = $spatialCandidates[0];
         }
 
+        // Download as PNG/SVG/PDF use _GET and the normal form uses _POST
         // Convert geometric columns from bytes to text.
-        $pos = $_GET['pos'] ?? $_SESSION['tmpval']['pos'];
-        if (isset($_GET['session_max_rows'])) {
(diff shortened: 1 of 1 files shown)


williamdes (approved):
Interesting !
If you can post a screenshot or the places you found this issue ?

MoonE:
As you can see number of rows is set to 1, but there are two points. And the start row is also ignored.
image

libraries/classes/Controllers/Table/GisVisualizationController.php
```diff
@@ -108,14 +108,14 @@ public function index(): void
         }
 
         // Convert geometric columns from bytes to text.
-        $pos = $_GET['pos'] ?? $_SESSION['tmpval']['pos'];
-        if (isset($_GET['session_max_rows'])) {
-            $rows = $_GET['session_max_rows'];
+        $pos = (int) ($_POST['pos'] ?? $_SESSION['tmpval']['pos']);
```
williamdes, line 111: This is correct but incorrect for the download links for PDF/PNG/SVG download button

PR #17043, opened 2021-07-28
Fix #14796 - Fixing the issue of hiding the structure options while exporting

### LINKED ISSUE #14796: Export Database page: UI prevents exporting procedures only
### Describe the bug

On the Export Database page. I found that in PMA 4.8.4 its not possible to export all procedures without exporting some table structure. This worked fine upto 4.7.0 and some newer releases.

### To Reproduce

1. Go to Export Database page
2. Uncheck all tables and structure.
3. Scroll down to option "add CREATE PROCEDURE / FUNCTION / EVENT statements"
4. See that these are missing now because all structure was unchecked.

Same story for other options that can pertain to procedures, such as "add DROP".

### Expected behavior

Being able to export all procedures. The "smart" UI is preventing this.

A workaround would be the export procedures button in the procedures list page. But this does not output to a file. (and does not sort the procedures by name - making it unfit for version control use, as mentioned in issue #14795)

### Additional context

I think this commit introduced this problem: https://github.com/phpmyadmin/phpmyadmin/commit/298ef434822fd294428029790480055cd6ff1952

iifawzi:

### Description
Fixes #14796
This PR will prevent the structure options from being hidden even if the structure checkbox is not selected

### DIFF
--- a/js/src/export.js
+++ b/js/src/export.js
@@ -506,19 +506,11 @@ Export.checkTableSelectAll = function () {
 };
 
 Export.checkTableSelectStructureOrData = function () {
-    var strChecked = $('input[name="table_structure[]"]:checked').length;
     var dataChecked = $('input[name="table_data[]"]:checked').length;
     var autoIncrement = $('#checkbox_sql_auto_increment');
 
     var pluginName = $('select#plugins').val();
     var dataDiv = '#' + pluginName + '_data';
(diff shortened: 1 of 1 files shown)


williamdes (approved):
It looks good but I will have to check once more if it has no hiding side effects

PR #17049, opened 2021-07-30
Fix gis visualization save file

MoonE:
- The GIS visualization can now be saved with a different label / column, before it always used the first spatial column and no label.
- Saving image as png works with label, the coordinates need to be `int` instead of `float` and should be rounded before casting to int
- If a label is just the number zero, it was treated as no label in the OpenLayers map.

### DIFF
--- a/libraries/classes/Gis/GisMultiPolygon.php
+++ b/libraries/classes/Gis/GisMultiPolygon.php
@@ -348,8 +348,8 @@ class GisMultiPolygon extends GisGeometry
             . 'fill: new ol.style.Fill(' . json_encode($fill_style) . '),'
             . 'stroke: new ol.style.Stroke(' . json_encode($stroke_style) . ')';
 
-        if ($label) {
-            $text_style = ['text' => $label];
+        if (trim($label) !== '') {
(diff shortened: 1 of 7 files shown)


williamdes (approved):
Very good fixes!

I might write tests for this PR before merging it

libraries/classes/Gis/GisMultiPolygon.php
```diff
@@ -168,8 +169,8 @@ public function prepareRowAsPng(
             imagestring(
                 $image,
                 1,
-                $points_arr[2],
-                $points_arr[3],
+                (int) round($label_point[0]),
+                (int) round($label_point[1]),
```
williamdes, line 173: Was the change to 0,1 intended ?

MoonE: It now uses the `$label_point` instead of `$points_array`, `$label_point` was unused before and is initialized just above
```js
                $label_point = [
                    $points_arr[2],
                    $points_arr[3],
                ];
```

williamdes: Indeed, thanks

PR #17058, opened 2021-08-07
Fix `DEL` character in generated random strings

### LINKED ISSUE #17054: Blowfish secret auto-generated by setup script includes non-printable characters
### Describe the bug

When the phpMyAdmin setup script (/phpmyadmin/setup) generated `$cfg['blowfish_secret']`, the `DEL` character (ASCII code 127, [Wikipedia](https://en.wikipedia.org/wiki/Delete_character)) was included as one of the 32 characters. When I attempted to copy the contents of the file from Windows to my server through SSH instead of transferring the config file itself, this character was automatically stripped, causing the secret to now only be 31 characters and too short. When I discovered that this was the cause of the issue, I transferred the file to the server instead of copying the contents (leaving the `DEL` character intact), and phpMyAdmin treats the secret as valid. However, I don't think that the inclusion of the non-printable `DEL` character is intended behavior.

### To Reproduce

1. Go to the phpMyAdmin setup script at /phpmyadmin/setup
2. Create a new server having authentication type 'cookie'

Once the server settings are saved, the secret will be automatically generated and viewable in 'Features' under the 'Security' tab. If the `DEL` character is included, it appears as a wide space in Google Chrome, `DEL` in Notepad++, and `^?` in nano.

### Expected behavior

The secret should be limited to printable characters.

### Screenshots

image
(The space in the secret is the `DEL` character.)

Generated line in config.inc.php: [generated_secret.txt](https://github.com/phpmyadmin/phpmyadmin/files/6942449/generated_secret.txt) (file attached because GitHub replaces the character)

### Server configuration

- Operating system: Ubuntu Server 21.04 (GNU/Linux 5.11.0-25-generic x86_64)
- Web server: Apache/2.4.46 (Ubuntu)
- Database version: MySQL 8.0.26-0ubuntu0.21.04.3
- PHP version: 8.0.9
- phpMyAdmin version: 5.1.1 (downloaded from [phpmyadmin.net](https://www.phpmyadmin.net/downloads/))

### Client configuration

- Browser: Vivaldi 4.1.2369.11
- Operating system: Windows 10 Education version 1909

### Additional context

I have not tried to see if the `DEL` character in the secret causes any problems with phpMyAdmin functionality.

lrb2:
### Description

Expands the prohibited ASCII characters for strings generated by `generateRandom()` to include the `DEL` character (127) in addition to the other control characters (0-31) and space (32).

Fixes #17054

### DIFF
--- a/libraries/classes/Util.php
+++ b/libraries/classes/Util.php
@@ -3017,8 +3017,8 @@ class Util
             // Get random byte and strip highest bit
             // to get ASCII only range
             $byte = ord((string) $random_func(1)) & 0x7f;
-            // We want only ASCII chars
-            if ($byte <= 32) {
+            // We want only ASCII chars and no DEL character (127)
+            if ($byte <= 32 || $byte === 127) {
                 continue;
             }
 


williamdes (approved):
Awesome !

libraries/classes/Util.php
```diff
@@ -3018,7 +3018,7 @@ public static function generateRandom(int $length, bool $asHex = false): string
             // to get ASCII only range
             $byte = ord((string) $random_func(1)) & 0x7f;
             // We want only ASCII chars
-            if ($byte <= 32) {
+            if ($byte <= 32 || $byte == 127) {
```
williamdes, line 3021:
```suggestion
            if ($byte <= 32 || $byte === 127) {
```

libraries/classes/Util.php
```diff
@@ -3018,7 +3018,7 @@ public static function generateRandom(int $length, bool $asHex = false): string
             // to get ASCII only range
             $byte = ord((string) $random_func(1)) & 0x7f;
             // We want only ASCII chars
```
williamdes, line 3020:
```suggestion
            // We want only ASCII chars and no DEL character (127)
```

williamdes (approved):
💯

MauricioFauth (approved):

PR #17059, opened 2021-08-07
Fix #17055 - missing navigation buttons on views

### LINKED ISSUE #17055: No page selector for system database tables
### Describe the bug

When browsing the system database tables (e.g. information_schema.COLUMNS), the navigation bar does not display any page selector or forward/backwards buttons to browse through the database table.
In non-system tables the page selector as well as the forward/backwards buttons are shown, if the number of records in the table exceeds the number of displayed rows.

### To Reproduce

1. Select a system table (e.g. information_schema.COLUMNS)
2. See that page selector and forward/backwards buttons are missing in the navigation bar

### Expected behavior

Page selector and forward/backwards buttons should be displayed, if the number of records in the table exceeds the number of displayed rows. It should be displayed regardless of whether it is a system table or not.

### Screenshots

Screenshot of system table:
Screenshot_system_tables

For comparison, screenshot of non-system table:
Screenshot_non_system_table

### Server configuration

- Operating system: Debian 10
- Web server: Apache2
- PHP version: 7.3.29
- phpMyAdmin version: 5.2.0-dev (clone of current Github master branch)

### Client configuration

- Browser: Firefox 90.0.2
- Operating system: Linux Manjaro

### Additional context

In phpMyAdmin v. 5.1.1 the forward/backwards buttons are displayed, but the page selector is also missing.

Edit:
I figured out where this issue originates from. I would really like to contribute to phpMyAdmin, so would you accept a pull request for it? It would be a one line change in getTableNavigation.

kroloq:
### Description

Corrects an if-condition in getTableNavigation. Property 'unlim_num_rows' cannot get -1. Instead 'unlim_num_rows' is set to false in countRecords(), if a VIEW was queried and the config 'MaxExactCountViews' is set to 0, which is the default setting.
So the if-condition results in a positive number (for tables) or false (for tables).

The correct if-condition lived in QA_5_1 for quite some time, but has never been merged into the master branch obviously.

Fixes #17055

This fix does not make appear the page selector for VIEWS. Displaying the page selector is only possible, if the number of records in the query result will be obtained in countRecords() - rather than skipping the counting (and returning false). After having checked the respective code section I conclude that it is intended to not show the page selector for views. Returning the actual number of records could bring down the server for complex views, the code comment says.

### DIFF
--- a/libraries/classes/Display/Results.php
+++ b/libraries/classes/Display/Results.php
@@ -870,7 +870,7 @@ class Results
         // Move to the next page or to the last one
         $moveForwardButtons = '';
         if (
-            $this->properties['unlim_num_rows'] === -1 // view with unknown number of rows
+            $this->properties['unlim_num_rows'] === false // view with unknown number of rows
             || (! $isShowingAll
             && intval($_SESSION['tmpval']['pos']) + intval($_SESSION['tmpval']['max_rows'])
                 < $this->properties['unlim_num_rows']


williamdes:
Hi @kroloq 

This is not true, see: ba50cbb91cdd6b9f06f2386b4ff5b1a46c0d644d

It was changed in 8f992a4fbfbca92b83a3e4f21d70c1fca1274e4c on June 15th by @mauriciofauth (https://github.com/phpmyadmin/phpmyadmin/commit/8f992a4fbfbca92b83a3e4f21d70c1fca1274e4c#diff-6a82aee55be3c2036240f864d5d4666c8379558893e70fe2484041a0c2deb57fR872)

@mauriciofauth could you check this PR please ?

kroloq:
Hi @williamdes,

thank you for having had a look into this PR. I see, it iwas changed on June 15th. In parallel QA_5_1 has the 'correct' if-condition for quite a long time. 
I have just seen that the QA_5_1 branch has been merged into master a few time sind June 15th. So I expected to see the thwo branhces aligned. But still the if-conditions are different in master and QA_5_1.
But I am very new to Git(hub), so my understanding is probably not correct.

Regarding the code itself, it is not intuitive to have a property 'unlim_num_rows' with a value of 'false'. The doc in the class says it could be either int|numeric-str. So @mauriciofauth's version is better I think. However, it would require countRows() to return -1 instead of false. And this probably would have a number of side effects on other functions/code. 

So I provided this fx, in order to make the application work as expected. May be, I should update the class doc to reflect the boolean value of the 'unlim_num_rows' property.
But in the longer term it seems preferable to change to purely int values for this property.

williamdes:
I was about to merge this one but our analyzers seem to not be convinced by the change so I will wait for another review as intended 

Sorry for the time this takes to merge

kroloq:
Hi @williamdes,
thank you for following up.
Let's wait for the review.

MauricioFauth (approved):

williamdes:
Hi @mauriciofauth 
Can you cherry pick this into QA_5_2 ?

mauriciofauth:
> Can you cherry pick this into QA_5_2 ?

@williamdes Done! I noticed it I forgot to change the base the moment after I merged it.

PR #17060, opened 2021-08-07
Fix #17017 - creation of unpartitioned tables

### LINKED ISSUE #17017: Can't submit create table form
### Describe the bug

The create table form can't be submitted when partitioning is unset.
> Please select a value that is no less than 2.

### To Reproduce

1. Go to https://demo.phpmyadmin.net/master-config/
2. Select a database
3. Go to 'Create table', input table name and click 'Go'
4. Fill out column form.
5. Click ~'Go'~ 'Save' at the bottom 
6. See Error: 'Please select a value that is no less than 2.' on the partition field

### Expected behavior

The table should be created.

### Server configuration

- phpMyAdmin version: RELEASE_5_1_1-1058-gf0bb3c42e0 

### Client configuration

- Browser: Firefox 90
- Operating system: Linux

lrb2:
### Description

To fix the inability to create an unpartitioned table, the default value for the partition count input is now 1 instead of empty (0). Therefore, the input's minimum value is now 1 (instead of 2, which caused #17017). Partitions can still be created by increasing the value to 2 or greater, and partitioning can be easily canceled by decreasing the value back to 1. In prior versions, partitioning could only be canceled by manually typing '0' into the input, but with this change it can also be canceled using the arrows built into the number input element.

This fix also applies to creating subpartitions.

I have not been able to find any issues arising from `partition_count` being 1 instead of empty, since all uses of `partition_count` or `subpartition_count` that I have found check if it is greater than 1 before using it.

Fixes #17017

### DIFF
--- a/templates/columns_definitions/partitions.twig
+++ b/templates/columns_definitions/partitions.twig
@@ -37,7 +37,7 @@
         <td><label for="partition_count">{% trans 'Partitions:' %}</label></td>
         <td colspan="2">
             <input name="partition_count" type="number" min="2"
-                value="{{ partition_details['partition_count'] }}">
+                value="{{ partition_details['partition_count'] ?: '' }}">
         </td>
     </tr>
     {% if partition_details['can_have_subpartitions'] %}
@@ -65,7 +65,7 @@
             <td><label for="subpartition_count">{% trans 'Subpartitions:' %}</label></td>
             <td colspan="2">
                 <input name="subpartition_count" type="number" min="2"
-                       value="{{ partition_details['subpartition_count'] }}">
+                    value="{{ partition_details['subpartition_count'] ?: '' }}">
             </td>
         </tr>
     {% endif %}


williamdes (changes requested):
This works but the UI is not adding a line for the value of 1
Before it was doing it because the box had no value and onChange it created the line

That said `min=1` seems okay

lrb2:
As far as I know, phpMyAdmin hasn't supported having only a single partition, at least it doesn't work on the demo server in 4.9 or 5.1. I was able to create a table with a single partition through manual SQL, and I can see the partition in phpMyAdmin, but I can't edit it. I could allow making a single partition (which would make the solution to this issue super easy), but I imagine there's a good reason not to just based on how many `> 1` checks there are, but I haven't really used partitions so I don't know. That said, I didn't think it was even possible to have only one partition so I'll try to implement a better solution.

lrb2:
I found that the issue was caused by 069d8e13db2958012e09ef8b3b56f3195e35568b in `master`, which caused the input to show 0 (which is considered less than 2) instead of empty. The input will now be empty if (`sub`-)`partition_count` is 0 (yields false). I'm certain I was reproducing the issue in 5.1 yesterday, even when the input showed empty, but maybe something was cached somehow from when I tried it on `master`? Anyway, the issue was on `master` so this should be the solution.

williamdes (approved):
🚢

PR #17061, opened 2021-08-07
Fix #16804 - table CSS for number alignment

### LINKED ISSUE #16804: Number alignment
Numbers seem to be left-aligned rather than right-aligned.

5.1.1-dev+20210408.b5829d01e0

lrb2:
### Description

The tables generated by the `makeGrid` JS function used the `pma_table` CSS class instead of `pma-table` as used throughout the CSS styles, so many CSS rules were not being applied to these tables. In addition, the only two `pma_table` rules in each theme were also changed to `pma-table`.

Fixes #16804

### DIFF
--- a/libraries/classes/Display/Results.php
+++ b/libraries/classes/Display/Results.php
@@ -2082,7 +2082,7 @@ class Results
             return;
         }
 
-        $th_class[] = 'right';
+        $th_class[] = 'text-right';
     }
 
     /**
@@ -3498,7 +3498,7 @@ class Results
     ) {
         if (! isset($column) || $column === null) {
             $cell = $this->buildNullDisplay(
-                'right ' . $class,
+                'text-right ' . $class,
                 $condition_field,
                 $meta,
                 ''
@@ -3508,7 +3508,7 @@ class Results
             $where_comparison = ' = ' . $column;
 
             $cell = $this->getRowData(
-                'right ' . $class,
+                'text-right ' . $class,
                 $condition_field,
                 $analyzed_sql_results,
                 $meta,
@@ -3525,7 +3525,7 @@ class Results
             );
         } else {
             $cell = $this->buildEmptyDisplay(
-                'right ' . $class,
+                'text-right ' . $class,
                 $condition_field,
                 $meta,
                 ''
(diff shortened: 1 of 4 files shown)


williamdes (approved):
very nice fix !

themes/bootstrap/scss/_common.scss
```diff
@@ -2374,7 +2374,7 @@ input#auto_increment_opt {
   top: 60px;
 }
 
-.pma_table {
+.pma-table {
```
williamdes, line 2377: I am not sure this is a wanted side effect of this PR but it breaks css styles see the block above the table is wrong

williamdes: image

lrb2: Ah, I looked right over that and didn't realize that those styles didn't need to be applied to the other `pma-table`'s, just the ones that had been `pma_table`'s. I just updated my commit to fix that!

js/src/makegrid.js
```diff
@@ -2190,7 +2190,7 @@ var makeGrid = function (t, enableResize, enableReorder, enableVisib, enableGrid
     g.table = $commonHiddenInputs.find('input[name=table]').val();
 
     // add table class
-    $(t).addClass('pma_table');
+    $(t).addClass('pma-table');
```
mauriciofauth, line 2193: The `.pma-table` class is not the same as the `.pma_table` class. `.pma-table` is the style that was applied to the `table` HTML tag and was extracted to a CSS class. The `pma-` prefix was added to avoid conflict with the Bootstrap's `.table` class.

MauricioFauth (changes requested):

lrb2:
Thanks for your comment @mauriciofauth! I found the solution that a few tables that use Bootstrap were still using the old `right` class instead of Bootstrap's `text-right`. I do think that it would be nice if the `pma_table` class was renamed to `pma-grid` (since it's made using the `makeGrid` function) or something because it's easy to confuse `pma-table` and `pma_table` right now. But if you're used to it then changing it might cause more harm than good.

MauricioFauth (approved):

williamdes:
For reference this was made on master and seems similar: d3c31fabd4012cd483ff268068d2399c36fb5b9f

PR #17066, opened 2021-08-09
Fix #17019 - table tabs in header when the table has not yet been created

### LINKED ISSUE #17019: ''Browse" button visible when creating a table from the database structure view
### Describe the bug

The browse button should only be displayed when an already existing table is selected.

### To Reproduce

1. Go to https://demo.phpmyadmin.net/master-config
2. Select a database ('Structure' tab)
3. In the main view go to 'Create table'
4. enter a name and click 'Go'
5. 'Browse' button is visible at the top, when clicked it shows an error

### Expected behavior

The button should not be visible when the table does not exist. 
This works when the table is created via 'New' in the navigation panel

### Server configuration
- Web server: nginx/1.14.2
- Database version: 8.0.18 - MySQL Community Server - GPL
- PHP version: 7.4.21

### Client configuration

- Browser: Firefox 90 
- Operating system: Linux

lrb2:
### Description

The `$table` global variable was being set to empty to try to fix the issue in #17019, but that only happens _after_ the menu was already created using the previous value of `$table`. This fix takes advantage of the existing `setTable` function to change the value in the existing menu.

Fixes #17019

### DIFF
--- a/libraries/classes/Controllers/Table/CreateController.php
+++ b/libraries/classes/Controllers/Table/CreateController.php
@@ -165,8 +165,8 @@ class CreateController extends AbstractController
             return;
         }
 
-        // This global variable needs to be reset for the header class to function properly
-        $table = '';
+        // Do not display the table in the header since it hasn't been created yet
+        $this->response->getHeader()->getMenu()->setTable('');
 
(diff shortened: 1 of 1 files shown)


williamdes (approved):
Thank you for this fix !

PR #17070, opened 2021-08-11
Fix #17062 - Pagination issues when working with identically named tables in separate databases

### LINKED ISSUE #17062: Pagination doesn't work as expected for identical table names in different databases
### Describe the bug

Pagination doesn't work as expected for identical table names in different databases. Count is greater than rows in table.
I have attached the SQL files for both databases.

### To Reproduce

1. Create two databases `db_test_1` and `db_test_2` (database names don't matter)
2. Create a table name `db_test_1`.`tbl_test` and insert 3 rows
2. Copy the table with structure and data in the second database
3. Insert 100-200 rows in `db_test_2`.`tbl_test`
4. Open `db_test_1`.`tbl_test` in a browser tab
5. Open `db_test_2`.`tbl_test` in another browser tab
6. Navigate to a page greater than 1 in `db_test_2`.`tbl_test`
7. Click on table name at the top in `db_test_1`.`tbl_test`
8. See error - `Showing rows 75 - 74 (75 total, Query took 0.0004 seconds.) ` even though the first table has 3 rows, and pagination is messed.

### Expected behavior

An accurate count should be done when clicking on table name at the top.

### Screenshots

First table

tbl1

Second table

tbl2

### Server configuration

- Operating system: Windows
- Web server: nginx/1.21.1
- Database version: 5.7.35 - MySQL Community Server (GPL)
- PHP version: 8.0.9
- phpMyAdmin version: RELEASE_5_1_1-1242-g3c2d727139

### Client configuration

- Browser: Firefox
- Operating system: Windows

### Additional context

[tables.zip](https://github.com/phpmyadmin/phpmyadmin/files/6949997/tables.zip)

lrb2:
### Description

Certain details, including the position in the table, are kept in `$_SESSION['tmpval']['query']` under the md5 hash of the SQL query. Since a query for any table named `test` (no matter which database) is `SELECT * FROM 'test'`, the query hash is identical for each table, even though they are in different databases. My change makes the hash dependent on the server and database as well, not just the query, so the hashes are no longer identical.

Fixes #17062

This PR should be merged into `master` because `$sqlMd5` is named `$sql_md5` in `QA_5_1`.

### DIFF
--- a/libraries/classes/Display/Results.php
+++ b/libraries/classes/Display/Results.php
@@ -1014,7 +1014,10 @@ class Results
 
         // required to generate sort links that will remember whether the
         // "Show all" button has been clicked
-        $sql_md5 = md5($this->properties['sql_query']);
+        $sql_md5 = md5($this->properties['server']
+            . $this->properties['db']
+            . $this->properties['sql_query']
+        );
         $session_max_rows = $is_limited_display
             ? 0
             : $_SESSION['tmpval']['query'][$sql_md5]['max_rows'];
@@ -3874,7 +3877,10 @@ class Results
      */
     public function setConfigParamsForDisplayTable()
     {
-        $sql_md5 = md5($this->properties['sql_query']);
+        $sql_md5 = md5($this->properties['server']
+            . $this->properties['db']
+            . $this->properties['sql_query']
+        );
         $query = [];
         if (isset($_SESSION['tmpval']['query'][$sql_md5])) {
             $query = $_SESSION['tmpval']['query'][$sql_md5];


williamdes:
> This PR should be merged into master because $sqlMd5 is named $sql_md5 in QA_5_1.

Could you please rebase it onto QA_5_1, I will do the merge and fix the variable before the merge of QA -> master commits ?

I am a bit afraid the db property is not always available, will have another look at that

lrb2:
I just rebased it. Note that the `providerSetConfigParamsForDisplayTable` function in ResultsTest.php (added in fa12894e5fd0d544a24820600111a3148cfaba96 on `master`) will have to be changed to use the new hash, which should be `27b1330f2076ef45d236f20839a92831`, instead of `0040c297a1a5a4cb1291b086b3ca965b`.

If there is no database specified, `properties['db']` is an empty string in master or null in QA_5_1. Even if it's null, the concatenation still works (at least in PHP 8). I haven't been able to imagine a possible combination of valid queries with and without a database where the resulting hash would be the same. However, when set, `properties['db']` uses the global value of `db` instead of the variable `$db` that was passed to the `executeQueryAndGetQueryResponse` function. I'm not sure when `$db` and `$GLOBALS['db']` are different, but if they are, there is a rare possibility that the hashes could be the same.

williamdes (changes requested):
Thank you!

williamdes (approved):

williamdes (approved):
💯

williamdes:
I copied one of your previous commits into master branch to fix the tests
See: 13807a1049c1eb58f4acd61a7740387a1114631c

PR #17074, opened 2021-08-12
Fix #17018 - Cannot save data from gis editor for spatial column insert

### LINKED ISSUE #17018: Cannot save data from gis editor for spatial column
### Describe the bug
Cannot save data from gis editor for spatial column.

### To Reproduce

1. Go to the insert form on a table with 'Point' column (or any other spatial column)
2. Click on 'Edit/Insert' for the Point column
3. Click 'Copy' to get `'POINT(0 0),' 0` into the form
4. Click 'Save'
5. It tries to execute an invalid query: ``INSERT INTO `test` (`id`, `point`) VALUES (NULL, '\'POINT(0 0)\',0')``
 Instead of ``INSERT INTO `test` (`id`, `point`) VALUES (NULL, ST_PointFromText('POINT(0 0)',0))``

### Expected behavior

A new row should be inserted without error

### Screenshots
image

### Server configuration

- Operating system:
- Web server: nginx/1.14.2 
- Database version: 8.0.18 - MySQL Community Server - GPL 
- PHP version: 7.4.21 
- phpMyAdmin version: RELEASE_5_1_1-1058-gf0bb3c42e0 

### Client configuration

- Browser: Firefox 90
- Operating system: Linux

### Additional context

Spatial functions are missing from the list of selectable functions when inserting data, may be related.

MoonE:
Fixes #17018

### DIFF
--- a/libraries/classes/Controllers/Table/ReplaceController.php
+++ b/libraries/classes/Controllers/Table/ReplaceController.php
@@ -173,10 +173,18 @@ final class ReplaceController extends AbstractController
             'PolyFromText',
             'MPolyFromText',
         ];
-
-        $gis_from_wkb_functions = [];
+        $gis_from_wkb_functions = [
+            'GeomFromWKB',
+            'GeomCollFromWKB',
+            'LineFromWKB',
+            'MLineFromWKB',
+            'PointFromWKB',
+            'MPointFromWKB',
+            'PolyFromWKB',
+            'MPolyFromWKB',
+        ];
         if ($this->dbi->getVersion() >= 50600) {
-            $gis_from_wkb_functions = [
+            $gis_from_text_functions = [
                 'ST_GeomFromText',
                 'ST_GeomCollFromText',
                 'ST_LineFromText',
@@ -185,6 +193,16 @@ final class ReplaceController extends AbstractController
                 'ST_MPointFromText',
                 'ST_PolyFromText',
                 'ST_MPolyFromText',
+            ];
+            $gis_from_wkb_functions = [
+                'ST_GeomFromWKB',
+                'ST_GeomCollFromWKB',
+                'ST_LineFromWKB',
+                'ST_MLineFromWKB',
+                'ST_PointFromWKB',
+                'ST_MPointFromWKB',
+                'ST_PolyFromWKB',
+                'ST_MPolyFromWKB',
             ];
         }
 
--- a/libraries/classes/Html/Generator.php
+++ b/libraries/classes/Html/Generator.php
@@ -49,6 +49,8 @@ use function str_replace;
 use function strlen;
 use function strncmp;
 use function strpos;
+use function strtoupper;
+use function substr;
 use function trim;
 use function urlencode;
 
@@ -284,6 +286,15 @@ class Generator
         $current_class = $dbi->types->getTypeClass($field['True_Type']);
         if (! empty($current_class) && isset($cfg['DefaultFunctions']['FUNC_' . $current_class])) {
             $default_function = $cfg['DefaultFunctions']['FUNC_' . $current_class];
+            // Change the configured default function to include the ST_ prefix with MySQL 5.6 and later.
+            // It needs to match the function listed in the select html element.
+            if (
+                $current_class === 'SPATIAL' &&
+                $dbi->getVersion() >= 50600 &&
+                strtoupper(substr($default_function, 0, 3)) !== 'ST_'
+            ) {
+                $default_function = 'ST_' . $default_function;
+            }
         }
 
         // what function defined as default?
@@ -340,9 +351,7 @@ class Generator
         $functions = $dbi->types->getAllFunctions();
         foreach ($functions as $function) {
             $retval .= '<option';
-            if (isset($foreignData['foreign_link']) && $foreignData['foreign_link'] !== false
-                && $default_function === $function
-            ) {
+            if ($function === $default_function && ! isset($foreignData['foreign_field'])) {
                 $retval .= ' selected="selected"';
             }
             $retval .= '>' . $function . '</option>' . "\n";
--- a/libraries/classes/InsertEdit.php
+++ b/libraries/classes/InsertEdit.php
@@ -2686,9 +2686,9 @@ class InsertEdit
             return "'" . $uuid . "'";
         }
 
-        if ((in_array($multi_edit_funcs[$key], $gis_from_text_functions)
-            && substr($current_value, 0, 3) == "'''")
-            || in_array($multi_edit_funcs[$key], $gis_from_wkb_functions)
+        if (
+            in_array($multi_edit_funcs[$key], $gis_from_text_functions) ||
+            in_array($multi_edit_funcs[$key], $gis_from_wkb_functions)
         ) {
             // Remove enclosing apostrophes
             $current_value = mb_substr($current_value, 1, -1);


libraries/config.default.php
```diff
@@ -2933,7 +2933,7 @@
     'FUNC_CHAR' => '',
     'FUNC_DATE' => '',
     'FUNC_NUMBER' => '',
-    'FUNC_SPATIAL' => 'GeomFromText',
+    'FUNC_SPATIAL' => 'ST_GeomFromText',
```
MoonE, line 2936: This won't work with MySQL 5.6.
This config is needed to pre-select the correct function in the insert form.

libraries/classes/InsertEdit.php
```diff
@@ -2686,9 +2686,9 @@ public function getCurrentValueAsAnArrayForMultipleEdit(
             return "'" . $uuid . "'";
         }
 
-        if ((in_array($multi_edit_funcs[$key], $gis_from_text_functions)
-            && substr($current_value, 0, 3) == "'''")
-            || in_array($multi_edit_funcs[$key], $gis_from_wkb_functions)
+        if (
+            in_array($multi_edit_funcs[$key], $gis_from_text_functions) ||
+            in_array($multi_edit_funcs[$key], $gis_from_wkb_functions)
```
MoonE, line 2691: What was the point of `substr($current_value, 0, 3) == "'''"`?

libraries/classes/Controllers/Table/ReplaceController.php
```diff
@@ -186,6 +186,16 @@ public function index(): void
                 'ST_PolyFromText',
                 'ST_MPolyFromText',
             ];
+            $gis_from_wkb_functions = [
```
MoonE, line 189: Is WKB used anywhere?

libraries/classes/Controllers/Table/ReplaceController.php
```diff
@@ -176,7 +176,7 @@ public function index(): void
 
         $gis_from_wkb_functions = [];
         if ($this->dbi->getVersion() >= 50600) {
-            $gis_from_wkb_functions = [
+            $gis_from_text_functions = [
                 'ST_GeomFromText',
```
MoonE, line 180: WKT functions in `$gis_from_wkb_functions` array was introduces in https://github.com/phpmyadmin/phpmyadmin/commit/fbd89fb882715cdd08cb24c67a6c8470e1c6285f#diff-50b557eedadea84c727650f37ac543980498bff75a0bc0ddbc28d4df23d039c6

libraries/config.default.php
williamdes: Maybe just keep your change and add a comment to use `GeomFromText` on older versions

libraries/classes/Types.php
```diff
@@ -676,7 +676,7 @@ public function getAllFunctions()
             $this->getFunctionsClass('CHAR'),
             $this->getFunctionsClass('NUMBER'),
             $this->getFunctionsClass('DATE'),
-            $this->getFunctionsClass('UUID')
+            $this->getFunctionsClass('SPATIAL')
```
williamdes, line 679: could you explain further this change ?

libraries/classes/InsertEdit.php
williamdes: This was moved 9 years ago: https://github.com/phpmyadmin/phpmyadmin/commit/7a7a18fd0e96e5eeb94d1ea9fec7a26d144d2fef

And most probably added 10 years ago in: https://github.com/phpmyadmin/phpmyadmin/commit/da6fe46a4d34c650e3d7f0b70d04ea22730f8e19

I think this diff will probably give you clues about why it was there
```
// Converting $val of the form '''POINT(12 15)'',145' to 'POINT(12 15)',145
```

libraries/classes/Html/Generator.php
```diff
@@ -284,6 +285,13 @@ public static function getDefaultFunctionForField(array $field, $insert_mode): s
         $current_class = $dbi->types->getTypeClass($field['True_Type']);
         if (! empty($current_class) && isset($cfg['DefaultFunctions']['FUNC_' . $current_class])) {
             $default_function = $cfg['DefaultFunctions']['FUNC_' . $current_class];
+            if (
+                $current_class === 'SPATIAL' &&
+                $dbi->getVersion() >= 50600 &&
+                mb_strtoupper(mb_substr($default_function, 0, 3)) !== 'ST_'
```
williamdes, line 291: You do not need multi byte support for the default function text, right ?

libraries/classes/Controllers/Table/ReplaceController.php
williamdes: Do you have a visual change that you could share us via screenshots ?

williamdes:
Thank your for your work @MoonE 
I added some comments

libraries/classes/Types.php
MoonE: The UUID function class does not exist. This call always returns an empty array.
The UUID function is in the CHAR function class.

libraries/classes/Controllers/Table/ReplaceController.php
MoonE: Not sure I understand what you are asking.

Currently the `$gis_from_wkb_functions` variable contains *FromText functions instead of the WKB equivalent, and it happend in the linked merge above.
Here is the current state in the master branch:
https://github.com/phpmyadmin/phpmyadmin/blob/c60b78691597cec8fe31f35f2dc085f2b119312d/libraries/classes/Controllers/Table/ReplaceController.php#L183-L206

libraries/classes/Html/Generator.php
MoonE: Probably not, it should only be a-z and _ characters.

I kind of hacked this part together to make it work, not quite happy with how it turned out.

libraries/classes/Types.php
williamdes: Thank you, at one point in history this made sense: https://github.com/phpmyadmin/phpmyadmin/blob/eebad9fbb9adf4ec41be79d0e4e65ccab8b86aeb/libraries/Types.class.php

I could reproduce this issue and made an issue for it #17079 and copied your fix as 14cd12ca1e2c5eb15ad7e6da40f9a83857c7b063 into QA

libraries/classes/Html/Generator.php
```diff
@@ -284,6 +286,15 @@ public static function getDefaultFunctionForField(array $field, $insert_mode): s
...
             $default_function = $cfg['DefaultFunctions']['FUNC_' . $current_class];
+            // Add ST_ prefix to configured default function if it does not already have it
+            // It needs to match the function listed in the select element.
+            if (
+                $current_class === 'SPATIAL' &&
+                $dbi->getVersion() >= 50600 &&
+                strtoupper(substr($default_function, 0, 3)) === 'ST_'
+            ) {
+                $default_function = 'ST_' . $default_function;
+            }
```
williamdes, line 297: could this be done in the function that provides the list ?
It would be more consistent with existing code :)

MoonE: Which function and what list are you referring to?

This changes the default function name from `GeomFromText` to `ST_GeomFromText` when mysql >=5.6 is used where all of the spatial functions are prefixed with ST_.  This way the current FUNC_SPATIAL default config remains compatible to all MySQL versions.

libraries/classes/Html/Generator.php
```diff
@@ -331,7 +342,6 @@ public static function getFunctionsForField(array $field, $insert_mode, array $f
         global $dbi;
 
         $default_function = self::getDefaultFunctionForField($field, $insert_mode);
-        $dropdown_built = [];
```
williamdes, line 334: I think you can remove this for QA branch, I pushed it to master branch 037c4c45bf33658bde27a8a2e4fbe59387c3ae96

libraries/classes/Html/Generator.php
williamdes: Oh, I understand the fix now !

It "hacks" the value of the DefaultFunctions config

Sorry for the misunderstanding, maybe can you add one more comment line about that it changes the config value ?

libraries/classes/Controllers/Table/ReplaceController.php
williamdes: 🤔 I am still totally confused about why the code seems to remove WKB functions
Will probably have to test your work to find out

MoonE: I don't see how you could enter WKB data in the textfield. WKB is binary data, not text. So it seems rather useless to have these functions.

libraries/classes/Html/Generator.php
```diff
@@ -341,12 +347,11 @@ public static function getFunctionsForField(array $field, $insert_mode, array $f
         foreach ($functions as $function) {
             $retval .= '<option';
             if (isset($foreignData['foreign_link']) && $foreignData['foreign_link'] !== false
-                && $default_function === $function
+                || $default_function === $function
```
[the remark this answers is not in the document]
williamdes: While investigating I found that `timestamp_seen` seems to lead to a global that is not switched anywhere
Will add another comment when I find out about the foreign key case

libraries/classes/Controllers/Table/ReplaceController.php
williamdes: Well, you can enter that data

```sql
INSERT INTO `gis_all` (`name`, `shape`) VALUES ('', ST_GeomFromWKB(0x01010000000000000000003e400000000000002440))
```

If the form did not add quotes it would work :)
Do you agree to revert this part ?

libraries/classes/Html/Generator.php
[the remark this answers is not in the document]
MoonE: I also noticed that, and I am unsure how the condition where it is used could ever be true:
https://github.com/phpmyadmin/phpmyadmin/blob/517479f8461766a6fa6b9a2ed9b30de4ecc6b9bd/libraries/classes/Html/Generator.php#L300-L315
Timestamps columns are treated rather special and I seem to be unable to create a timestamp column that is not  null and has no default value.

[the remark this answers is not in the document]
williamdes: > Timestamps columns are treated rather special and I seem to be unable to create a timestamp column that is not null and has no default value.

Makes sense, I could not find a possible case too
Not sure if an issue should be opened for that

libraries/classes/Controllers/Table/ReplaceController.php
williamdes: Maybe some users use it ^^

libraries/classes/Html/Generator.php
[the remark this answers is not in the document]
williamdes: It's because it makes no sense to auto select a function when the values are fixed (FK list of values)
image

libraries/config.default.php
MoonE: I reverted it to GeomFromText, and it is prefixed automatically when using MySQL >=5.6.

libraries/classes/Html/Generator.php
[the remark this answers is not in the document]
williamdes:
```diff
diff --git a/libraries/classes/Html/Generator.php b/libraries/classes/Html/Generator.php
index f5de9f8d39..8a2a396ab7 100644
--- a/libraries/classes/Html/Generator.php
+++ b/libraries/classes/Html/Generator.php
@@ -340,8 +340,9 @@ class Generator
         $functions = $dbi->types->getAllFunctions();
         foreach ($functions as $function) {
             $retval .= '<option';
-            if (isset($foreignData['foreign_link']) && $foreignData['foreign_link'] !== false
-                && $default_function === $function
+            if (
+                // Do not auto-select a function when the field value is a foreign key link (list of values)
+                ($foreignData['foreign_field'] ?? null) === null || $default_function === $function
             ) {
                 $retval .= ' selected="selected"';
             }

```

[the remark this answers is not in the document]
MoonE: So shouldn't the condition be like this.
```patch
diff --git a/libraries/classes/Html/Generator.php b/libraries/classes/Html/Generator.php
index 5bbf0d81d1..1e17b5bf5d 100644
--- a/libraries/classes/Html/Generator.php
+++ b/libraries/classes/Html/Generator.php
@@ -351,8 +351,8 @@ class Generator
         $functions = $dbi->types->getAllFunctions();
         foreach ($functions as $function) {
             $retval .= '<option';
-            if (isset($foreignData['foreign_link']) && $foreignData['foreign_link'] !== false
-                || $default_function === $function
+            if ($default_function === $function &&
+                (! isset($foreignData['foreign_link']) || $foreignData['foreign_link'] !== false)
             ) {
                 $retval .= ' selected="selected"';
             }
```

williamdes (approved):
Looks nice, I posted my patch solution for the FK code
When you think it is okay you can mark this PR as ready for review :)

libraries/classes/Html/Generator.php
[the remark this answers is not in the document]
williamdes: I would not rely on foreign_link but on foreign_field

On all my tests, Insert or Edit foreign_link was always false

williamdes:
@sudo-robot deploy

williamdes:
I added this config on the instance

```php
$cfg['DefaultFunctions'] = [
    'FUNC_CHAR' => 'DATE',
    'FUNC_DATE' => 'NOW',
    'FUNC_NUMBER' => 'DATE',
    'FUNC_SPATIAL' => 'GeomFromText',
    'FUNC_UUID' => 'UUID',
    'first_timestamp' => 'NOW',
];
```

And as you can see the value has the function auto filled when if I am editing the row
https://pr-17074-phpmyadmin.public-instance.eu/index.php?route=/sql&server=2&server=2&db=public&table=articles&pos=0

Not sure if this is due to this PR

```
![image](https://user-images.githubusercontent.com/7784660/129449639-5f73d8ef-d758-4d99-b73c-76b72741ae43.png)

File: libraries/classes/InsertEdit.php [hunk 10]
```diff
@@ -2686,9 +2686,9 @@ public function getCurrentValueAsAnArrayForMultipleEdit(
             return "'" . $uuid . "'";
         }
 
-        if ((in_array($multi_edit_funcs[$key], $gis_from_text_functions)
-            && substr($current_value, 0, 3) == "'''")
-            || in_array($multi_edit_funcs[$key], $gis_from_wkb_functions)
+        if (
+            in_array($multi_edit_funcs[$key], $gis_from_text_functions) ||
```
williamdes, line 2690: how do you trigger this code ?
I could not find the patch

MoonE: Submit the insert form for a table with a spatial column.

williamdes: Thank you, it works
Is the PR still a draft ?

libraries/classes/Html/Generator.php
```diff
@@ -340,9 +351,7 @@ public static function getFunctionsForField(array $field, $insert_mode, array $f
         $functions = $dbi->types->getAllFunctions();
         foreach ($functions as $function) {
             $retval .= '<option';
-            if (isset($foreignData['foreign_link']) && $foreignData['foreign_link'] !== false
-                && $default_function === $function
-            ) {
+            if ($function === $default_function && ! isset($foreignData['foreign_field'])) {
```
williamdes, line 354:
```suggestion
            if ($function === $default_function && ! isset($foreignData['foreign_field']) && $insert_mode) {
```
For me this would be the right operation but it would not match "Functions selected by default when inserting/changing row" from the documentation

PR #17075, opened 2021-08-12
Use int as srid type

MoonE:
Improved type checking, use int type for srid.


williamdes (approved):
👏🏻

PR #17068, opened 2021-08-11
Fix #16067 - Make select elements where multiple items can be selected resizable

lrb2:
### Description

`<select multiple>` elements can now be resized vertically to make it easier to select multiple items.

nNfrrIEBA4

Fixes #16067

PR #17076, opened 2021-08-13
Fix grouping of tables in navigation tree if `$cfg['NavigationTreeTableLevel'] > 1`

### LINKED ISSUE #17003: Weird database-name "nesting" in the left menu when underscore signs are used in their names
### Describe the bug
Before posting this as a "bug", I did a little research and found out that whenever you create 2 databases, in phpMyAdmin, with underscores in their names, like **db_one** and **db_two**, phpMyAdmin will display them "grouped" in the left menu as follows:
**db
    ->db_one
    ->db_two**

My impression was that - that is intentional. Ok, but when I tested creating databases with MULTIPLE underscores in their names that produced a WEIRD MULTIPLE NESTING. Just look at the screenshot I am posting here and you will see what I mean.

### To Reproduce
1. Go to [phpMyAdmin Demo](https://demo.phpmyadmin.net/master-config/) 
2. One by one create 2 databases with the names: db_with_more_underscores_1 and db_with_more_underscores_2 (no need to add any tables in them)

...and you should immediately see the weird multiple nesting seen in the screenshot

### Expected behavior
If that is not some intentional effect - I believe ALL databases should display in a perfect column - below one another - regardless of any underscore signs in their names. And no nesting should occur.

### Screenshots
underscore_issue

### Server configuration
- Operating system: ? (It's the Demo site)
- Web server: nginx/1.14.2
- Database version: MySQL
- PHP version: 7.4.20

### Client configuration
- Browser: Chrome, Firefox

### Additional context
I found similar issues with underscores dating back to 2013

lrb2:
### Description

The contents in the `if` statement were not being run for nodes that have a real parent (ex. a database) but do not have siblings in the navigation tree. This caused an issue where, if `$cfg['NavigationTreeTableLevel']` is set to a value greater than the default of 1 (so that nested groups can exist), group nodes in the navigation tree that do not have a sibling would not be shown. See the pictures below: note that the group `with` was incorrectly not being shown.

Before fix:
image

After fix:
image

Fixes: #17003

### DIFF
--- a/libraries/classes/Navigation/NavigationTree.php
+++ b/libraries/classes/Navigation/NavigationTree.php
@@ -1143,10 +1143,18 @@ class NavigationTree
         $controlButtons = '';
         $paths = $node->getPaths();
         $nodeIsContainer = $node->type === Node::CONTAINER;
-        $hasSiblingsOrIsNotRoot = $node->hasSiblings() || $node->realParent() === false;
         $liClasses = '';
 
-        if ($hasSiblingsOrIsNotRoot) {
+        // Whether to show the node in the tree (true for all nodes but root)
+        // If false, the node's children will still be shown, but as children of the node's parent
+        $showNode = $node->hasSiblings() || count($node->parents(false, true)) > 0;
+
+        // Don't show the 'Tables' node under each database unless it has 'Views', etc. as a sibling
+        if ($node instanceof NodeTableContainer && ! $node->hasSiblings()) {
+            $showNode = false;
+        }
+
+        if ($showNode) {
             $response = ResponseRenderer::getInstance();
             if ($nodeIsContainer && count($node->children) === 0 && ! $response->isAjax()) {
                 return '';
@@ -1262,7 +1270,7 @@ class NavigationTree
         return $this->template->render('navigation/tree/node', [
             'node' => $node,
             'class' => $class,
-            'has_siblings_or_is_not_root' => $hasSiblingsOrIsNotRoot,
+            'show_node' => $showNode,
             'has_siblings' => $node->hasSiblings(),
             'li_classes' => $liClasses,
             'control_buttons' => $controlButtons,
--- a/templates/navigation/tree/node.twig
+++ b/templates/navigation/tree/node.twig
@@ -1,4 +1,4 @@
-{% if has_siblings_or_is_not_root %}
+{% if show_node %}
   <li class="{{ li_classes }}">
     <div class="block">
       <i{{ class == 'first' ? ' class="first"' }}></i>


libraries/classes/Navigation/NavigationTree.php
```diff
@@ -1143,9 +1143,14 @@ private function renderNode(Node $node, string $class = ''): string
         $controlButtons = '';
         $paths = $node->getPaths();
         $nodeIsContainer = $node->type === Node::CONTAINER;
-        $hasSiblingsOrIsNotRoot = $node->hasSiblings() || $node->realParent() === false;
+        $hasSiblingsOrIsNotRoot = $node->hasSiblings() || count($node->parents(false, true));
```
williamdes, line 1146: could you add a comment and make the count check a bit more explicit with a `> 0` ?

libraries/classes/Navigation/NavigationTree.php
```diff
@@ -1143,9 +1143,14 @@ private function renderNode(Node $node, string $class = ''): string
...
         $liClasses = '';
 
+        // Don't show the 'Tables' node under each database unless necessary
+        if ($node instanceof NodeTableContainer && ! $node->hasSiblings()) {
+            $hasSiblingsOrIsNotRoot = false;
```
williamdes, line 1151:
```suggestion
            // It was not siblings and is a table container so it is self contained
            $hasSiblingsOrIsNotRoot = false;
```
Normally I am not too much of a commenting code person but in this class it is more than necessary for people that will fix issues with it in 5-10 years

I commented about how I understand this if ;)

lrb2:
The purpose of `$hasSiblingsOrIsNotRoot` was really confusing to me until I went through a bunch of trial-and-error, so I went ahead and renamed the variable to `$showNode`, since that better states what it does. Also, this fixes that issue in  does **not** fix #17003 - this only fixes the issue I casually mentioned at the top of https://github.com/phpmyadmin/phpmyadmin/issues/17003#issuecomment-897056868, though I do have a working solution for #17003 that I'm not completely satisfied with yet.

williamdes:
Did you intend to target QA in each of your PRs ?

williamdes (approved):
It's so much more clear with the new comments and variable rename, thank you !

lrb2:
> Did you intend to target QA in each of your PRs ?

This commit needs to be merged into `master` because the `renderNode` function was just refactored in 41c478edd722c19698db71872ec056eb40f18483, so there would have to be a separate fix only for the QA branch. Though I did switch #17068 to QA, and it looks like I can target QA for #17078, which helps mitigate this issue on its own.

MauricioFauth (approved):

mauriciofauth:

PR #17082, opened 2021-08-17
Fix #14502 - Triggered the on-change event on the input field to uncheck the "ignore" checkbox 

### LINKED ISSUE #14502: "Ignore" check box (in INSERT tab) does not uncheck when all the fields are edited through popup windows
### Steps to reproduce
1.  Create a table with let's say 2 columns with foreign key
2. In the related tables (those referred to by the keys), add many entries, so that it reaches the threshold of PhpMyAdmin where the < select > becomes a popup, for displaying the options.

### Expected behaviour
The "Ignore" checkbox should become unchecked for that entry, since fields are filled.

### Actual behaviour
The js listener does not understand that fields have been filled, and thus, the "Ignore" button is not set to the unchecked state!

### Server configuration
**phpMyAdmin version:** 4.8.3, 5.0.0-dev

anny21:
### Description
The "Ignore" checkbox doesn't uncheck when fields are filled using the pop up modal

Fixes: #14502

### DIFF
--- a/js/src/sql.js
+++ b/js/src/sql.js
@@ -937,6 +937,9 @@ Sql.browseForeignDialog = function ($thisA) {
             }
             // Set selected value as input value
             $input.val($(this).data('key'));
+            // Unchecks the Ignore checkbox for the current row
+            $input.trigger('change');
+
             $dialog.dialog('close');
         });
         $(formId).on('click', showAllId, function () {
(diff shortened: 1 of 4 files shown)


williamdes (approved):
Looks good, thank you !

js/src/sql.js
```diff
@@ -937,6 +937,9 @@ Sql.browseForeignDialog = function ($thisA) {
             }
             // Set selected value as input value
             $input.val($(this).data('key'));
+            // Unchecks the Ignore checkbox for the current row
+            $input.trigger("change");
```
williamdes, line 941: Tips: `yarn run js-lint --fix`

williamdes:
I applied the js-lint fix and added one more commit to handle GIS modal on save un-check action

PR #17085, opened 2021-08-18
Fix #17084 - ErrorHandler sliceErrors not returning errors

### LINKED ISSUE #17084: PHP session errors details are missing on error page
### Describe the bug

PHP session errors that cause phpmyadmin to fail to load don't display the underlying error details on the error page, making it very difficult to diagnose the problem. There were some issues with the PHP session already being started in my project, which resulted in the error page showing showing only the generic "Error during session start..." message and no further information.

### To Reproduce
This is difficult to reproduce my exact scenario, but this is generally how it to happen:
1. Disable `session.auto_start` setting in your php.ini
2. At the bottom of your config.inc.php file, add a line with `session_start();` to start the session prematurely
3. Try to load phpmyadmin, and it should fail to abort the session because of the session.autostart setting being off (which I suspect is an unrelated bug)
4. The specific PHP error message will be missing on the error page

### Expected behavior

The error page should display the message "session_start(): Ignoring session_start() because a session is already active".

### Screenshots
What the error page looks like:
Screen Shot 2021-08-18 at 1 45 22 PM

What it should look like:
Screen Shot 2021-08-18 at 1 45 48 PM

### Server configuration

- Operating system: Ubuntu Linux 18.04
- Web server: Apache 2.4.29
- Database version:  MariaDB 10.6.4
- PHP version: 8.0.9
- phpMyAdmin version:  5.1.1

### Client configuration

- Browser: Firefox 91.0
- Operating system: macOS 11.4 (Big Sur)

### Additional context
The underlying problem seems to be that the ErrorHandler's sliceErrors method doesn't return the errors properly, and instead returns an empty array.

cmmcneill:
### Description

ErrorHandler class' sliceErrors method was failing to return the correct errors from the error array according to the passed $count argument. This was due to an extra array_splice being performed on $errors, which has already had the errors before $count removed in the first array_splice, and so the index of the remaining errors had changed.

This probably wasn't noticed before because the test case only slices from index 0, and you need to sliceErrors from an index > 0 for the bug to occur. I've fixed this to return all leftover errors after $count index, and updated the ErrorHandlerTest's testSliceErrors to better test this method.

Fixes #17084

### DIFF
--- a/libraries/classes/ErrorHandler.php
+++ b/libraries/classes/ErrorHandler.php
@@ -151,13 +151,19 @@ class ErrorHandler
     /**
      * Pops recent errors from the storage
      *
-     * @param int $count Old error count
+     * @param int $count Old error count (amount of errors to splice)
      *
-     * @return Error[]
+     * @return Error[] The non spliced elements (total-$count)
      */
     public function sliceErrors(int $count): array
     {
+        // store the errors before any operation, example number of items: 10
         $errors = $this->getErrors(false);
+
+        // before array_splice $this->errors has 10 elements
+        // cut out $count items out, let's say $count = 9
+        // $errors will now contain 10 - 9 = 1 elements
+        // $this->errors will contain the 9 elements left
         $this->errors = array_splice($errors, 0, $count);
 
         return $errors;
--- a/libraries/classes/ErrorHandler.php
+++ b/libraries/classes/ErrorHandler.php
@@ -160,7 +160,7 @@ class ErrorHandler
         $errors = $this->getErrors(false);
         $this->errors = array_splice($errors, 0, $count);
 
-        return array_splice($errors, $count);
+        return $errors;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


williamdes:
The logic you are changing was added in https://github.com/phpmyadmin/phpmyadmin/pull/12193 to fix https://github.com/phpmyadmin/phpmyadmin/issues/12146

I will be testing just in case it reverts the fix for https://github.com/phpmyadmin/phpmyadmin/issues/12146

williamdes (approved):
This really makes sense, I will add one more test and comments to the code

williamdes:
Hi @cmmcneill 
I fixed your test by changing one 16 to 15 (line number, duplicate checking ignored it with the same value).
Added a bunch of comments and a new test case
Let me know if you understand the comments in the same was as the code works

cmmcneill:
> Hi @cmmcneill
> I fixed your test by changing one 16 to 15 (line number, duplicate checking ignored it with the same value).
> Added a bunch of comments and a new test case
> Let me know if you understand the comments in the same was as the code works

@williamdes Thanks for fixing my test and adding another one. I think your comments describe how sliceErrors works quite well. Everything looks good.

PR #17093, opened 2021-08-29
Make controllers callable

MauricioFauth:


libraries/classes/Controllers/Table/ReplaceController.php
```diff
@@ -664,7 +664,7 @@ public function index(): void
         if ($goto_include === '/database/sql') {
             /** @var DatabaseSqlController $controller */
             $controller = $containerBuilder->get(DatabaseSqlController::class);
```
williamdes, line 666:
```suggestion
            $containerBuilder->get(DatabaseSqlController::class)();
```
This could be an alternative, but is it a good one ?. I will let you decide ^^

williamdes (approved):

libraries/classes/Controllers/Table/ReplaceController.php
mauriciofauth: I also prefer this way. :)

PR #17096, opened 2021-09-02
Remove the `Core::isValid` method

MauricioFauth:
This improves the type checking and simplifies the code.


williamdes (changes requested):
It would probably be better to create validation methods in the abstract controller to keep a central point of control on validation
What do you think about it ?

I feel like this PR is good but will make pointing out validation steps impossible

mauriciofauth:
> It would probably be better to create validation methods in the abstract controller to keep a central point of control on validation
> What do you think about it ?
> 
> I feel like this PR is good but will make pointing out validation steps impossible

I agree we should improve validation, but this should be done in another pull request, as this PR is mostly an inline of the `Core::isValid` method.

williamdes (approved):
> I agree we should improve validation, but this should be done in another pull request, as this PR is mostly an inline of the Core::isValid method.

Okay

LGTM then :)

Issue #17095, opened 2021-09-01
Copy table name


nunoperalta:
Hi,

It used to be easy to copy the table name, by selecting the name from here:

![image](https://user-images.githubusercontent.com/487079/131700516-24d8cc8a-0ff7-4b3e-aedc-4b82e5fb503d.png)

However, now every time I attempt to select the table name there, it does reverse selection, or copies the whole line:

![image](https://user-images.githubusercontent.com/487079/131700717-e3cfc71e-1d9e-4726-8b3d-8f25c8e588ec.png)

This wouldn't be a problem if the table name was NOT hyperlinked. I don't see a reason for it to be, since that's the section I'm currently at anyway, and there's the "Browse" tab just below.

Is it possible to either:

1) remove the hyperlink   (so I can simply double click the table name to copy it)

or

2) have a button next to it to copy table name

?

Thank you very much.

williamdes:
Hi !
> have a button next to it to copy table name

This could be a nice option for non mobile users

Maybe there is a way to say what value should be copied

But thanks for reporting this

yogeshojha:
I will work on this @williamdes 🥳

williamdes:
> I will work on this @williamdes 🥳

Thank you, let us know if you need help

soysudhanshu:
Hey Yogesh,

I was going through browser compatibilty tables for Clipboard API support. I want to know about how we are planning to implement cross browser compatibilty. For eg Firefox only allows Clipboard API in secure contexts, for which we might need to fallback to deprecated `exec`commands. 

I was also wondering is Internet Explorer supported by PHPMyAdmin?

I would really like to help you out in this one.

P.S: I am a newbie here

nunoperalta:
The point (1) I suggested above is to just make the table name plain text (no hyperlink). I'd be very happy with that solution :)

(all I'd have to do is double click the table name and Ctrl + C)

soysudhanshu:
I have fixed the issue but I don't know how to file a PR for PHPMyAdmin. Can you guys help me out?

williamdes:
> I have fixed the issue but I don't know how to file a PR for PHPMyAdmin. Can you guys help me out?

You successfully did it: https://github.com/phpmyadmin/phpmyadmin/pull/17104 🎉

williamdes:
> I have fixed the issue but I don't know how to file a PR for PHPMyAdmin. Can you guys help me out?

[...]

PR #17104, opened 2021-09-11
Removed anchor from breadcrumb table name item for easier copy pasting

soysudhanshu:
Removes anchor for breadcrumb table/view item, and fixes color of the breadcrumb item after removal of anchor

Ref: #17095


williamdes (changes requested):
I always use this link, this is not the right solution IMO

But thanks for contributing

PR #17106, opened 2021-09-13
Do not generate code coverage from multi-arch tests

MauricioFauth:
Code coverage from multi-arch tests is not needed. Knowing if the tests passes is enough.

This reduces test time and memory usage.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -31,7 +31,6 @@ jobs:
               # tcpdf allowed memory exhausted needs the memory_limit workaround
               # musl-locales and musl-locales-lang are needed to run some locale specific tests
               # gettext is needed to run generate-mo
-              # pcov for code coverage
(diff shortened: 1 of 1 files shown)


williamdes:
I am kind of split on this, having coverage report allowed to know if the tests actually ran correctly and where not Volkswagen passing. Do your know why coverage did drop ?

mauriciofauth:
This reduces test time and memory usage. It's also an attempt to avoid some random failing tests.

> I am kind of split on this, having coverage report allowed to know if the tests actually ran correctly and where not Volkswagen passing.

We still have coverage for regular tests. How code coverage from a specific architecture could differ from regular code coverage?

> Do your know why coverage did drop ?

Seems to be unrelated. Probably related to failing tests.

williamdes:
> This reduces test time and memory usage. It's also an attempt to avoid some random failing tests.

Thanks

> How code coverage from a specific architecture could differ from regular code coverage?

Only 32 bit specific code could make that 🤔

PR #17107, opened 2021-09-13
Add `void` return type where possible

MauricioFauth:


examples/openid.php
```diff
@@ -37,10 +37,8 @@
  * Simple function to show HTML page with given content.
  *
  * @param string $contents Content to include in page
- *
- * @return void
  */
-function Show_page($contents)
+function Show_page($contents): void
```
williamdes, line 41:
```suggestion
function Show_page(string $contents): void
```

examples/openid.php
```diff
@@ -68,10 +66,8 @@ function Show_page($contents)
  * Display error and exit
  *
  * @param Exception $e Exception object
- *
- * @return void
  */
-function Die_error($e)
+function Die_error($e): void
```
williamdes, line 70:
```suggestion
function Die_error(\Exception $e): void
```
I know it's out of the scope of this PR but it could be changed too as example files do not change often

williamdes (approved):

examples/openid.php
mauriciofauth: I will open up more PRs like this with these changes.

PR #17120, opened 2021-09-21
Fix #17121 - remove single quote and escaped value before hashing password

### LINKED ISSUE #17121: password_hash function incorrectly adds single quotes to password before hashing
### Describe the bug

This was reported by Hidayatullah ayat.kyo@gmail.com at #17120.

The new (since PMA 5.1.0) feature to use PHP's `password_hash()` function when inserting data seems to hash with single quotes surrounding the `'`password`'`. This, of course, fails to match when run through `password_verify()`.

### To Reproduce

From the Insert tab, select "password_hash() PHP function" from the dropdown of functions.

Take the returned result and run it through password_verify() and see that the password requires single quotes around it to match.

### Server configuration

- Operating system: Debian Linux
- Web server: Apache
- PHP version: 8.0.3

### Client configuration

- Browser: Firefox

ayatkyo:
Signed-off-by: Hidayatullah <ayat.kyo@gmail.com>

### Description

Remove single quote and escaped value before hashing value using `password_hash`.

In current version, value with quote and escaped will be send to `password_hash` function, so `password_verify` will always return `false`.

For example my password is `super_secret` and select **The PHP function password_hash() with default options.** function, then **Go** to save it.

Using this code will return `false`
```php
password_verify("super_secret", "PASSWORD_HASH");
```

but using this code will return `true` because i wrap with `'`
```php
password_verify("'super_secret'", "PASSWORD_HASH");
```

Fixes #17121

### DIFF
--- a/libraries/classes/InsertEdit.php
+++ b/libraries/classes/InsertEdit.php
@@ -1634,6 +1634,10 @@ class InsertEdit
         }
 
         if ($multiEditFuncs[$key] === 'PHP_PASSWORD_HASH') {
+            // Remove single quote and escaped value
+            $currentValue = mb_substr($currentValue, 1, -1);
+            $currentValue = stripslashes($currentValue);
+
             /**
              * @see https://github.com/vimeo/psalm/issues/3350
              *


ibennetch:
I don't see a bug report about this behavior, so I opened #17121.

Thank you for your pull request.

@mauriciofauth could you please take a look at this?

ibennetch:
I don't believe this is the most effective way to solve this; it would be better to address it further up the stream when `$current_value` is first created. This might mean all the strings are hashing improperly.

ayatkyo:
Thanks for response.

Right, MD5 and PASSWORD use MySQL function where value must be wrapped in quote and escaped, whereas `password_hash` use PHP function which require original value, so there is no need to wrap it in quote and escaped.

For example if my value is `super'secret\slash` then `$currentValue` will be escape and quoted as `'super\'secret\\slash'` so we need to remove quote and slash before pass to PHP function. 


**MD5**

```sql
PASSWORD('super\'secret\\slash')
```

**PASSWORD**

```sql
MD5('super\'secret\\slash')
```

**password_hash()** before pull request

```sql
'$2y$10$z.SyP2hO0ewkgnVE0lBL5.ojZR5Dmy2xiHF3.ob3ovEUDu9tjwPEe'
```

Test using password_verify

```php
// Original input is -> super'secret\slash
password_verify('super\'secret\slash', '$2y$10$z.SyP2hO0ewkgnVE0lBL5.ojZR5Dmy2xiHF3.ob3ovEUDu9tjwPEe'); // FALSE

// Input must we write -> 'super\'secret\\slash'
password_verify("'super\\'secret\\\\slash'", '$2y$10$z.SyP2hO0ewkgnVE0lBL5.ojZR5Dmy2xiHF3.ob3ovEUDu9tjwPEe'); // TRUE
```

**password_hash()** after pull request

```sql
'$2y$10$cMG60hAmjPrLmoh15nY6m.Qhj.HumHg.Ch.1QqfwdVkSp.P.SRfO.'
```

Test using `password_verify`

```php
password_verify('super\'secret\slash', '$2y$10$cMG60hAmjPrLmoh15nY6m.Qhj.HumHg.Ch.1QqfwdVkSp.P.SRfO.'); // TRUE
```

williamdes (approved):

kamil-tekiela:
I am completely against this change and I believe it should not be merged. I agree that the current behaviour is buggy, but this "fix" will only break it more. I'll try to come up with a proper fix for this. 

Passwords should never be modified in any way. Whatever value is received from user input should be passed to `password_hash`. The issue here isn't the quotes, but rather escaping that is performed prior to hashing.

PR #17122, opened 2021-09-22
Get default config values from `Config\Settings` classes

MauricioFauth:
~~Removes the `libraries/config.default.php` file.~~

### DIFF
--- a/doc/setup.rst
+++ b/doc/setup.rst
@@ -519,7 +519,7 @@ To manually create the file, simply use your text editor to create the
 file :file:`config.inc.php` (you can copy :file:`config.sample.inc.php` to get
 a minimal configuration file) in the main (top-level) phpMyAdmin
 directory (the one that contains :file:`index.php`). phpMyAdmin first
-loads :file:`libraries/config.default.php` and then overrides those values
+loads the default configuration values and then overrides those values
 with anything found in :file:`config.inc.php`. If the default value is
 okay for a particular setting, there is no need to include it in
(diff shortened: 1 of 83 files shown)


mauriciofauth:
Hi, @ibennetch and @williamdes. I'm split about removing the `config.default.php` file. With this PR, it is not needed anymore by the code, however it is useful for copying config directives, but it is also redundant in its majority with the documentation. Should we removed it? Should we keep it for reference?

williamdes:
> Hi, @ibennetch and @williamdes. I'm split about removing the `config.default.php` file. With this PR, it is not needed anymore by the code, however it is useful for copying config directives, but it is also redundant in its majority with the documentation. Should we removed it? Should we keep it for reference?

I would really like to keep it, maybe just built a test to check that the default config file matches the default values in the classes ?

Or maybe we still should load it 🤔 
Because not removing it, people maybe will edit it 🤔 ?

ibennetch:
I agree with everything you've both said. Since it's redundant, it makes sense to remove it — but sometimes users edit the default file instead of creating their own `config.inc.php`.

I also refer to that file very regularly to see the definitive source of possible values. It's useful for users to see the exact syntax for a directive (whether it's a `$i` or a global directive, in particular).

I'm really uncertain about what is best with this.

ibennetch:
I also don't want to end up with _two_ places to maintain default values.

mauriciofauth:
I'm not going to remove the `libraries/config.default.php` file for now. We can figure out what to do later.

PR #17137, opened 2021-10-06
Add Table\Maintenance\Message value object

MauricioFauth:
Value object for table maintenance statements results (`ANALYZE TABLE`, `CHECK TABLE`, `OPTIMIZE TABLE` and `REPAIR TABLE`, `CHECKSUM TABLE` has a different result format).

- https://dev.mysql.com/doc/refman/en/table-maintenance-statements.html
- https://mariadb.com/kb/en/table-statements/


williamdes (commented):
Very cool

PR #17142, opened 2021-10-09
Add button next to table name to copy it to the clipboard

RVxLab:
Signed-off-by: RVxLab <richard@rvx.works>

### Description

This PR adds a button next to the table name so it can be easily copies, as was previously mentioned in #17095

From my end, there is 1 open question still, which is regarding the button's graphic. As I can't find a suitable icon in the current set and my severe lack of designing skills I opted to use `s_link.png` for the time being. Surely we don't want to keep it at this.

### DIFF
--- a/js/src/breadcrumbs.js
+++ b/js/src/breadcrumbs.js
@@ -1,0 +1,22 @@
+/**
+ * @fileoverview This file includes logic to deal with breadcrums
+ *
+ * @requires js/functions.js
+ * @requires js/ajax.js
+ * @requires jQuery
+ */
+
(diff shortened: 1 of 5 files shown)


js/src/breadcrumbs.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * @fileoverview This file includes logic to deal with breadcrums
```
ibennetch, line 2: *breadcrum**b**s* :)

williamdes (approved):
It probably should be rebased but the code looks good to me

PR #17143, opened 2021-10-09
Use `composer/ca-bundle` to manage the CA cert file

MauricioFauth:
Uses the [`composer/ca-bundle`](https://github.com/composer/ca-bundle) package to handle the `cacert.pem` file.

> Small utility library that lets you find a path to the system CA bundle, and includes a fallback to the Mozilla CA bundle.

The `openssl` PHP extension is now required.


williamdes (approved):
This is a brilliant idea!

PR #17183, opened 2021-11-10
Remove array support from `Util::backquote`

MauricioFauth:
Extracts the `foreach` from the `Util::backquote` method.


williamdes (approved):
Awesome !
Excellent idea !

PR #17199, opened 2021-11-21
Minor refactoring of InsertEdit.php

kamil-tekiela:
This is still minor refactoring despite many changes, as this class is in dire need of Value Objects and removal of globals, which I haven't touched in this PR. Static analysis and I are both restricted in the number of changes that can be made due to the fact that many of the values are typed as mixed. 

The goals of this refactoring:
- Removal of dead code
- Removal of redundant code
- Improving static analysis by providing type hints and type declarations
- Small refactorings to make code easier to read and spot bugs 

This PR does not:
- Change any functionality
- Add any new methods or classes
- Improve unit tests
- Change coding style

It seems that DbiDummy is broken (see the one test failure). The tests send an empty array where the `$result` object is expected, I fixed that by casting to an object. The mock methods seem to be returning an integer. I'm not sure what the correct behaviour of this should be, but it looks like it doesn't actually test the right functionality. Am I wrong?

This code contains ABI break, but I hope it won't be a problem. I changed some parameters of private and public methods. It will work the same but callers need to pass different arguments now. 

For the explanation of each change, see the description in commits.


libraries/classes/InsertEdit.php
```diff
@@ -690,26 +677,27 @@ private function getHtmlInput(
...
         }
 
-        $inputMinMax = false;
+        $inputMinMax = '';
         if (in_array($column['True_Type'], $this->dbi->types->getIntegerTypes())) {
             $extractedColumnspec = Util::extractColumnSpec($column['Type']);
             $isUnsigned = $extractedColumnspec['unsigned'];
             $minMaxValues = $this->dbi->types->getIntegerRange($column['True_Type'], ! $isUnsigned);
-            $inputMinMax = 'min="' . $minMaxValues[0] . '" '
+            $inputMinMax = ' min="' . $minMaxValues[0] . '" '
```
williamdes, line 685: Not sure if the space should be added there or where the variable is used

libraries/classes/InsertEdit.php
```diff
@@ -718,12 +706,12 @@ private function getHtmlInput(
     /**
      * Get HTML select option for upload
      *
-     * @param string $vkey   [multi_edit]['row_id']
-     * @param array  $column description of column in given table
+     * @param string $vkey      [multi_edit]['row_id']
+     * @param string $field_MD5 array index as an MD5 to avoid having special characters
```
williamdes, line 710:
```suggestion
     * @param string $fieldHashMd5 array index as an MD5 to avoid having special characters
```
This is maybe better for our coding standard

libraries/classes/InsertEdit.php
```diff
@@ -734,45 +722,46 @@ private function getSelectOptionForUpload($vkey, array $column)
                 . __('The directory you set for upload work cannot be reached.') . "\n";
         }
 
-        if (! empty($files)) {
+        if ($files) {
```
williamdes, line 725: This is a bit vague, could it be more precise?

libraries/classes/InsertEdit.php
```diff
@@ -899,17 +888,15 @@ private function getValueColumnForOtherDatatypes(
...
-     * @param array $extractedColumnspec associative array containing type,
-     *                                    spec_in_brackets and possibly enum_set_values
-     *                                    (another array)
+     * @param array  $column           description of column in given table
+     * @param string $spec_in_brackets text in brackets inside column definition
      *
      * @return int field size
      */
-    private function getColumnSize(array $column, array $extractedColumnspec)
+    private function getColumnSize(array $column, string $spec_in_brackets): int
```
williamdes, line 896: We use camel casing ;)

libraries/classes/InsertEdit.php
```diff
@@ -1119,21 +1103,21 @@ private function getSpecialCharsAndBackupFieldForExistingRow(
...
-        array $column,
-        $realNullValue
+        array $column
     ) {
         if (! isset($column['Default'])) {
             $column['Default'] = '';
             $realNullValue = true;
             $data = '';
         } else {
+            $realNullValue = false;
```
williamdes, line 1120: Is it a bug fix?

williamdes (changes requested):
Very nice work, thanks a lot! 
I did a bit of review from my mobile
Will have to have another look

libraries/classes/InsertEdit.php
kamil-tekiela: It doesn't matter really, but if it looks better then I will roll it back. In cases like this, I would usually use `implode()` rather than string concatenation but I didn't want to change too much here.

libraries/classes/InsertEdit.php
kamil-tekiela: I can make it a more strict comparison. I didn't want to change the functionality, but the string comes from Template render so it probably wouldn't hurt to check strictly for an empty string rather than a non-empty-string (which includes string `'0'`). I doubt that the template render would ever return string `'0'` so it should be safe, and we can make it more clear with the help of early return.

libraries/classes/InsertEdit.php
kamil-tekiela: I was not able to identify any buggy behaviour here. Technically it was possible to provide `true` here through the use of the parameter, but in all the uses I could find (including unit tests) it was always hardcoded to `false`. Was it a bug? I don't think so.

williamdes: Okay, thanks for the fix then ;)

libraries/classes/InsertEdit.php
```diff
@@ -1366,16 +1347,10 @@ public function executeSqlQuery(array $urlParams, array $query)
...
-                if ($tmp) {
-                    $totalAffectedRows += $tmp;
-                }
-
-                unset($tmp);
+                $totalAffectedRows += $this->dbi->affectedRows();
 
                 $insertId = $this->dbi->insertId();
-                if ($insertId !== false && $insertId != 0) {
+                if ($insertId) {
```
williamdes, line 1353: could it be more specific ?

williamdes (approved):
Did not see anything alarming, did post some comments on minor things

libraries/classes/InsertEdit.php
```diff
@@ -1785,94 +1758,91 @@ public function getCurrentValueForDifferentTypes(
...
-            // be possible if Javascript is disabled in the browser)
-            if (! empty($multiEditColumnsNull[$key]) && ($currentValue == "''" || $currentValue == '')) {
+        if ($type !== 'protected' && $type !== 'set' && strlen($currentValue) === 0) {
+            // best way to avoid problems in strict mode
+            // (works also in non-strict mode)
+            if (isset($multiEditAutoIncrement, $multiEditAutoIncrement[$key])) {
                 $currentValue = 'NULL';
+            } else {
+                $currentValue = "''";
             }
```
williamdes, line 1780: The elses could be avoided by setting the value above the if

libraries/classes/InsertEdit.php
```diff
@@ -2130,15 +2092,14 @@ private function getHtmlForInsertEditFormColumn(
...
-        $column = $tableColumns[$columnNumber];
         $readOnly = false;
 
         if (! isset($column['processed'])) {
             $column = $this->analyzeTableColumnsArray($column, $commentsMap, $timestampSeen);
         }
 
         $asIs = false;
-        if (! empty($repopulate) && ! empty($currentRow)) {
+        if ($repopulate && $currentRow) {
```
williamdes, line 2102: since currentRow is an array, could we narrow down what we are checking here and further ?

libraries/classes/InsertEdit.php
kamil-tekiela: Well... No. The value comes from `insertId()` that relies on undefined behaviour. It's unclear what the possible return values could be. The return value is type hinted as `int|false`, but I doubt that this is really the type of value that will be produced by this method. 

For the record, I removed the redundant check because `$insertId != 0` casts the value to an integer before comparison. This makes `$insertId !== false` redundant. The condition should therefore be `if($insertId != 0)`. But I think here is where I made a mistake/an assumption. If the returned value is an empty array (unlikely) or a non-empty non-numeric-string (possible) then the result of the comparison will be different. For this reason, I will add the `!= 0` back.

libraries/classes/InsertEdit.php
kamil-tekiela: Hmm, yeah. What are we checking here? 😀

kamil-tekiela: Ok, that's tough. `$repopulate` is a global. It has no defined structure. All we are really checking here is that both arrays are non-empty. I can only introduce `array_key_exists` here on `$repopulate`. 

I could remove the parameter altogether and replace it with the element that we need, but the problem is that I have absolutely no idea what this piece of code does. Unit tests don't cover this code and manual analysis of code flow leads me to madness... 

I will add `array_key_exists()` but if you have an idea on how to improve this piece of code, I would appreciate it.

MauricioFauth (approved):

mauriciofauth:
Merged. Thank you very much.

libraries/classes/InsertEdit.php
williamdes: Thank you for the `array_key_exists` ;)

PR #17204, opened 2021-11-24
Refactor DatabaseInterface::getColumns()

kamil-tekiela:
This PR introduces two new methods to DatabaseInterface.
- getColumn - public - split of from getColumns
- attachIndexInfoToColumns - private - curved out from getColumns. A method should do only one thing

This allows for cleaner code that doesn't need to pass `null` most of the time. This also helps static analysis significantly. We can properly type hint each return type now.

### DIFF
--- a/libraries/classes/Dbal/DbalInterface.php
+++ b/libraries/classes/Dbal/DbalInterface.php
@@ -174,21 +174,37 @@ interface DbalInterface
     ): array;
 
     /**
-     * Returns descriptions of columns in given table (all or given by $column)
+     * Returns description of a $column in given table
      *
      * @param string $database name of database
      * @param string $table    name of table to retrieve columns from
-     * @param string $column   name of column, null to show all columns
+     * @param string $column   name of column
      * @param bool   $full     whether to return full info or only column names
      * @param int    $link     link type
      *
-     * @return array array indexed by column names or,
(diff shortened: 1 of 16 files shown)


libraries/classes/Dbal/DbalInterface.php
```diff
@@ -174,21 +174,18 @@ public function getColumnsFull(
...
      * @param string $table    name of table to retrieve columns from
-     * @param string $column   name of column, null to show all columns
      * @param bool   $full     whether to return full info or only column names
      * @param int    $link     link type
      *
-     * @return array array indexed by column names or,
-     *               if $column is given, flat array description
+     * @return array<string, array> array indexed by column names
      */
     public function getColumns(
```
williamdes, line 186: Should getColumn be added to this interface ?

williamdes (approved):

libraries/classes/Dbal/DbalInterface.php
kamil-tekiela: Sorry, it seems like I forgot to add it. I added it now.

williamdes (approved):

MauricioFauth (approved):

mauriciofauth:
Merged. Thank you very much!

PR #17205, opened 2021-11-25
Fix tryMultiQuery signature

kamil-tekiela:
The signature of this method was misleading. Generally, it's not recommended to use this functionality, but since phpMyAdmin uses it for something then at least let's fix the signature.


williamdes (approved):

MauricioFauth (approved):

mauriciofauth:
Merged. Thank you very much.

PR #17207, opened 2021-11-26
Refactor OpenLayer visualization code generation

MoonE:
### Description
- Display coordinate of (0 0) which wasn't displayed before
- Removes some redundancy when createing OpenLayers js code and gets rid of some inefficencies in the generated js code.
- There is no need to construct intermediate geometry objects and the final geometry objects can be transformed in one call to `transform`

- [ ] Check that editing coordinates / incomplete coordinates in edit form still works


MoonE:
Replaced by #18151

PR #17209, opened 2021-11-26
Refactor DatabaseInterface.php

kamil-tekiela:
This PR improves code readability, performance and stability of DatabaseInterface.php. These are rather minor changes with the exception of adding a new internal method to simplify the code. 

For the explanation behind each change see commit descriptions.

### DIFF
--- a/libraries/classes/Database/CentralColumns.php
+++ b/libraries/classes/Database/CentralColumns.php
@@ -150,7 +150,7 @@ class CentralColumns
                 . 'LIMIT ' . $from . ', ' . $num . ';';
         }
 
-        $has_list = (array) $this->dbi->fetchResult($query, null, null, DatabaseInterface::CONNECT_CONTROL);
+        $has_list = $this->dbi->fetchResult($query, null, null, DatabaseInterface::CONNECT_CONTROL);
         $this->handleColumnExtra($has_list);
 
         return $has_list;
@@ -210,13 +210,13 @@ class CentralColumns
         if ($allFields) {
             $query = 'SELECT * FROM ' . Util::backquote($central_list_table) . ' '
                 . 'WHERE db_name = \'' . $this->dbi->escapeString($db) . '\' AND col_name IN (' . $cols . ');';
-            $has_list = (array) $this->dbi->fetchResult($query, null, null, DatabaseInterface::CONNECT_CONTROL);
+            $has_list = $this->dbi->fetchResult($query, null, null, DatabaseInterface::CONNECT_CONTROL);
             $this->handleColumnExtra($has_list);
         } else {
             $query = 'SELECT col_name FROM '
                 . Util::backquote($central_list_table) . ' '
                 . 'WHERE db_name = \'' . $this->dbi->escapeString($db) . '\' AND col_name IN (' . $cols . ');';
-            $has_list = (array) $this->dbi->fetchResult($query, null, null, DatabaseInterface::CONNECT_CONTROL);
+            $has_list = $this->dbi->fetchResult($query, null, null, DatabaseInterface::CONNECT_CONTROL);
         }
 
         return $has_list;
@@ -795,7 +795,7 @@ class CentralColumns
         }
 
         $this->dbi->selectDb($cfgCentralColumns['db'], DatabaseInterface::CONNECT_CONTROL);
-        $columns_list = (array) $this->dbi->fetchResult($query, null, null, DatabaseInterface::CONNECT_CONTROL);
+        $columns_list = $this->dbi->fetchResult($query, null, null, DatabaseInterface::CONNECT_CONTROL);
         $this->handleColumnExtra($columns_list);
 
         return $columns_list;
@@ -886,7 +886,7 @@ class CentralColumns
         $query = 'SELECT COUNT(db_name) FROM ' . Util::backquote($central_list_table) . ' '
             . 'WHERE db_name = \'' . $this->dbi->escapeString($db) . '\'' .
             ($num === 0 ? '' : 'LIMIT ' . $from . ', ' . $num) . ';';
-        $result = (array) $this->dbi->fetchResult($query, null, null, DatabaseInterface::CONNECT_CONTROL);
+        $result = $this->dbi->fetchResult($query, null, null, DatabaseInterface::CONNECT_CONTROL);
 
         if (isset($result[0])) {
             return (int) $result[0];
--- a/libraries/classes/Database/CentralColumns.php
+++ b/libraries/classes/Database/CentralColumns.php
@@ -14,7 +14,9 @@ use PhpMyAdmin\Template;
 use PhpMyAdmin\Util;
 
 use function __;
+use function array_diff;
 use function array_unique;
+use function array_values;
 use function bin2hex;
 use function ceil;
 use function count;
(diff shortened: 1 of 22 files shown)


libraries/classes/Database/CentralColumns.php
```diff
@@ -900,18 +902,9 @@ public function getColumnsNotInCentralList(string $db, string $table): array
...
-        foreach ($columnNames as $column) {
-            if (in_array($column, $existingColumns)) {
-                continue;
-            }
-
-            $columns[] = $column;
-        }
-
-        return $columns;
+        return array_values(array_diff($columnNames, $existingColumns));
```
williamdes, line 907: This is a bit hard to read, could you add a comment about what it does ?

libraries/classes/Plugins/Export/ExportMediawiki.php
```diff
@@ -299,7 +299,7 @@ public function exportData(
             $column_names = $dbi->getColumnNames($db, $table);
 
             // Add column names as table headers
-            if ($column_names !== null) {
+            if ($column_names) {
```
williamdes, line 302: maybe use count or `!= []` ?

libraries/classes/Database/CentralColumns.php
kamil-tekiela: I'm am of opinion that clean code should be self-explanatory and for this reason I prefer to keep code comments to the minimum. I am a little surprised that this simplification is still confusing. It does exactly what the method name says in 3 lines of code. There is one line `selectDb()` that is confusing to me and I can't explain its purpose. 

Nonetheless, I have added a comment right above `return` to make it as clear as possible that this function returns a simple list of column names except the ones in central list of columns. Please let me know if this makes things any clearer. 

P.S. The method `getFromTable()` is confusing, but that's a separate topic not related to this PR.

libraries/classes/DatabaseInterface.php
```diff
@@ -1306,6 +1260,25 @@ private function fetchValueOrValueByIndex($row, $value)
         return $value === null ? $row : $row[$value];
     }
 
+    /**
+     * returns array of rows with numeric or associative keys
+     *
+     * @param object $result result set identifier
+     * @param int    $mode   either self::FETCH_ROW or self::FETCH_ASSOC
```
mauriciofauth, line 1267:
```suggestion
     * @param int    $mode   either self::FETCH_ROW or self::FETCH_ASSOC
     * @psalm-param self::FETCH_ROW|self::FETCH_ASSOC $mode
```

MauricioFauth (changes requested):

libraries/classes/DatabaseInterface.php
kamil-tekiela: Thanks for this suggestion. I have thought about it and decided that the solution I came up with was too limited in scope. I decided to rewrite this commit and change the parameter type. I moved the constants to the interface and refactored `fetchSingleRow()`. I replaced literal strings with constants. Hopefully this makes the code cleaner and easier to understand.

libraries/classes/Database/CentralColumns.php
williamdes: It's very cool with the comment, that helps the brain clear the process between what values are kept from the diff

libraries/classes/DatabaseInterface.php
```diff
@@ -665,7 +656,7 @@ public function getDatabasesFull(
 
             // get table information from information_schema
             $sqlWhereSchema = '';
-            if (! empty($database)) {
+            if ($database) {
```
williamdes, line 659:
```suggestion
            if ($database !== null) {
```

libraries/classes/DatabaseInterface.php
```diff
@@ -786,24 +777,20 @@ public function getColumnMapFromSql(string $sql_query, array $view_columns = [])
         }
 
         $meta = $this->getFieldsMeta($result);
-        if ($meta === null) {
-            return [];
-        }
-
-        $nbFields = count($meta);
-        if ($nbFields <= 0) {
+        if (! $meta) {
```
williamdes, line 780: The previous syntax was not better and more strict ?

libraries/classes/DatabaseInterface.php
```diff
@@ -1119,7 +1100,7 @@ public function postConnect(): void
 
         /* Locale for messages */
         $locale = LanguageManager::getInstance()->getCurrentLanguage()->getMySQLLocale();
-        if (! empty($locale)) {
+        if ($locale) {
```
williamdes, line 1103: Given the non typed function at https://github.com/phpmyadmin/phpmyadmin/blob/70c43a98633dc475834a0a8d008d96e846dd2ec3/libraries/classes/Language.php#L111

I think it should be explicit
Please try to not widen the possible values that can make if conditions valid :)

libraries/classes/DatabaseInterface.php
```diff
@@ -1594,56 +1556,52 @@ public function getRoutines(
                 empty($name) ? null : $this->escapeString($name)
             );
             $result = $this->fetchResult($query);
-            if (! empty($result)) {
+            if ($result) {
```
williamdes, line 1559: should we check for false ?

libraries/classes/DatabaseInterface.php
kamil-tekiela: I haven't widened any condition. I have tried my best to not change any logic. In this case the condition remains the same. 

`empty()` is equivalent to `! isset($var) || ! $var`. See https://www.php.net/manual/en/function.empty.php
This means that `! empty()` is `! (! isset($var) || ! $var)`, which is `(isset($var) && $var)`.
Since we know that the variable is defined this becomes: `($var !== null && $var)` which can be simply written as `if($var)`. 

I have not widened the condition. The change I made has nothing to do with the type returned by the method. I am happy to narrow it down if you think it's safe, but as I said I tried not to change any logic.

libraries/classes/DatabaseInterface.php
kamil-tekiela: I loosened the condition per your suggestion. Thanks.

libraries/classes/DatabaseInterface.php
kamil-tekiela: I admit, this is one of the places where I took the liberty of changing the logic. It looks like I forgot to put the commit description. 

This was done because I believe the previous code was buggy. However, the change was related to few lines below it, where I changed it to `if ($nbColumns >= $i) {`. 

I think it is safe to revert back to previous syntax, which was a little bit more verbose, but clearer. According to the code, `getFieldsMeta()` can return `?array`. The condition can be `if($meta === null || $meta === [])`. The `|| $meta === []` is absolutely redundant as the code will behave exactly the same without it, so to make it really easy to read, let's keep it as `if($meta === null)`

libraries/classes/DatabaseInterface.php
kamil-tekiela: Thanks for pointing this out, This looks wrong indeed. Let me investigate.

kamil-tekiela: Thanks! The `if` statements were redundant. This function always returns an array. I removed them now. 

I then decided to add a return type to `fetchResult()` to catch similar issues. This led me to some Psalm fixes and errors in unit tests. All of this should be fixed in the latest commit.

MauricioFauth (approved):

mauriciofauth:
Merged. Thank you very much!

williamdes (commented):
:100:

PR #17210, opened 2021-11-27
Refactor relations.php

kamil-tekiela:
I did not change any logic. All the changes in this PR are only related to code cleanliness. Mostly removing dead and redundant code. I have added type declarations, type hints and type casts only in places where I could easily infer the proper type.

### DIFF
--- a/libraries/classes/Relation.php
+++ b/libraries/classes/Relation.php
@@ -1783,12 +1783,9 @@ class Relation
      * @param string $db     name of master table db.
      * @param string $table  name of master table.
      * @param string $column name of master table column.
-     *
-     * @return array
      */
-    public function getChildReferences($db, $table, $column = '')
+    public function getChildReferences($db, $table, $column = ''): array
     {
-        $child_references = [];
         if (! $GLOBALS['cfg']['Server']['DisableIS']) {
             $rel_query = 'SELECT `column_name`, `table_name`,'
                 . ' `table_schema`, `referenced_column_name`'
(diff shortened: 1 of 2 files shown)


libraries/classes/Relation.php
```diff
@@ -1480,9 +1477,9 @@ public function getForeignData(
                     . Util::backquote($foreign_table) . '.'
                     . Util::backquote($foreign_display);
 
-                $f_query_limit = ! empty($foreign_limit) ? $foreign_limit : '';
+                $f_query_limit = $foreign_limit ?: '';
 
-                if (! empty($foreign_filter)) {
+                if ($foreign_filter) {
```
williamdes, line 1482: Removing the not empty and using no defined condition really triggers me 😄 
Could you add a real condition ?

williamdes (approved):

libraries/classes/Relation.php
kamil-tekiela: I know how you feel. Type juggling triggers me in general. I can definitely add a strict check, but since the parameter isn't actually typed, I have no idea of knowing what the accepted values are. As you are aware, `! empty()` also does type juggling so when I removed it, I haven't changed the logic in any way. That `! empty()` was just redundant. 

I will add `$foreign_filter !== ''` and proper type declaration as it seems that the only possible value here is of type string. I hope this won't break anything.

williamdes (approved):

MauricioFauth (approved):

mauriciofauth:

PR #17215, opened 2021-11-29
Use array_keys whenever only keys are needed in foreach

kamil-tekiela:
This is a global search for all possible places where `array_keys` could be used to simplify code and improve static analysis. There are a couple more `UnusedForeachValue` which are either FP or could not be easily replaced. 

In a couple of places I had to change the code more. I replaced a loop with `array_merge()`. Built-in functions are usually faster and simpler than loops. I changed `is_iterable` to `is_array` as it looks like it was a mistake. I added `! is_array()` in one place for improved validation of input.

### DIFF
--- a/libraries/classes/BrowseForeigners.php
+++ b/libraries/classes/BrowseForeigners.php
@@ -8,6 +8,7 @@ declare(strict_types=1);
 namespace PhpMyAdmin;
 
 use function __;
+use function array_keys;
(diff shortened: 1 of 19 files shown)


williamdes (approved):

MauricioFauth (approved):

mauriciofauth:

PR #17217, opened 2021-11-30
Fix bug #17216

### LINKED ISSUE #17216: Cannot grant privileges on multiple databases at the same time
### Describe the bug

When assigning privileges on a single database, there is no issue. When assigning it on multiple at a time, there is a exception:
image

### To Reproduce

1. Go to User accounts
2. Click on Edit Privileges
3. Switch over to Database tab
4. Select multiple databases
5. Select one or more permissions
6. Press Ok

### Expected behavior

No error. 

### Server configuration

- PHP version: 8.1
- phpMyAdmin version: master

P.S. PR is on its way.

kamil-tekiela:
Fixes: #17216
I untangled this mess... 

During code review, I noticed a bug. I have reproduced the bug when running the software in the browser. I proceeded to fix this bug, but that code was very messy. In the end, I decided to rewrite this method almost completely. There were a couple of issues in this code, but the biggest one was a regression issue introduced via recent changes (Removal of `Core::isValid() 8f2f3ef via PR #17096). 

I have simplified the validation. It should still work as it did before. The code is much simpler, although in overall, it's still a mess. I have also fixed a bug with routine privileges. 
I thought about redesigning the whole functionality, but that is too much work. First, we need to get rid of globals which are probably the worst problem of this whole project.

### DIFF
--- a/libraries/classes/Controllers/Server/PrivilegesController.php
+++ b/libraries/classes/Controllers/Server/PrivilegesController.php
@@ -25,6 +25,7 @@ use function __;
 use function header;
 use function implode;
 use function is_array;
+use function is_string;
 use function ob_get_clean;
 use function ob_start;
 use function str_replace;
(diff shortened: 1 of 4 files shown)


williamdes (approved):
Thank you for the tests !

MauricioFauth (approved):

mauriciofauth:

PR #17220, opened 2021-12-01
Remove unused variables

kamil-tekiela:
This PR addresses most of `UnusedVariable` Psalm issues. I was not able to remove some of them as they are either FPs or I couldn't rule out side-effects.

### DIFF
--- a/libraries/classes/Plugins/Auth/AuthenticationSignon.php
+++ b/libraries/classes/Plugins/Auth/AuthenticationSignon.php
@@ -133,9 +133,6 @@ class AuthenticationSignon extends AuthenticationPlugin
         /* Session name */
         $session_name = $GLOBALS['cfg']['Server']['SignonSession'];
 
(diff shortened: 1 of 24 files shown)


libraries/classes/Plugins/Auth/AuthenticationSignon.php
```diff
@@ -146,7 +143,7 @@ public function readCredentials(): bool
         $single_signon_cfgupdate = [];
 
         /* Handle script based auth */
-        if (! empty($script_name)) {
+        if ($script_name) {
```
williamdes, line 146: I think you should check for non empty

williamdes (changes requested):

MauricioFauth (approved):

mauriciofauth:

williamdes (commented):
Thank you for all your work !

PR #17221, opened 2021-12-01
Psalm fixes 3

kamil-tekiela:
This is a collection of random fixes found through Psalm analysis

### DIFF
--- a/libraries/classes/Database/Qbe.php
+++ b/libraries/classes/Database/Qbe.php
@@ -231,7 +231,7 @@ class Qbe
      * Current search
      *
      * @access private
-     * @var SavedSearches
+     * @var SavedSearches|null
      */
     private $currentSearch = null;
 
@@ -245,12 +245,12 @@ class Qbe
     public $template;
 
     /**
-     * @param Relation          $relation        Relation object
-     * @param Template          $template        Template object
-     * @param DatabaseInterface $dbi             DatabaseInterface object
-     * @param string            $dbname          Database name
-     * @param array             $savedSearchList List of saved searches
-     * @param SavedSearches     $currentSearch   Current search id
+     * @param Relation           $relation        Relation object
+     * @param Template           $template        Template object
+     * @param DatabaseInterface  $dbi             DatabaseInterface object
+     * @param string             $dbname          Database name
+     * @param array              $savedSearchList List of saved searches
+     * @param SavedSearches|null $currentSearch   Current search id
      */
     public function __construct(
         Relation $relation,
@@ -293,7 +293,7 @@ class Qbe
     /**
      * Getter for current search
      *
-     * @return SavedSearches
+     * @return SavedSearches|null
      */
     private function getCurrentSearch()
     {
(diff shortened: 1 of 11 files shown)


libraries/classes/Database/Qbe.php
```diff
@@ -231,7 +231,7 @@ class Qbe
      * Current search
      *
      * @access private
-     * @var SavedSearches
+     * @var ?SavedSearches
```
williamdes, line 234:
```suggestion
     * @var SavedSearches|null
```

libraries/classes/Database/Qbe.php
```diff
@@ -250,7 +250,7 @@ class Qbe
      * @param DatabaseInterface $dbi             DatabaseInterface object
      * @param string            $dbname          Database name
      * @param array             $savedSearchList List of saved searches
-     * @param SavedSearches     $currentSearch   Current search id
+     * @param ?SavedSearches    $currentSearch   Current search id
```
williamdes, line 253:
```suggestion
     * @param SavedSearches|null    $currentSearch   Current search id
```

For now we stick to the phpdoc standard :)

libraries/classes/Utils/HttpRequest.php
```diff
@@ -102,7 +102,7 @@ private function handleContext(array $context)
             ];
             if (strlen($this->proxyUser) > 0) {
                 $auth = base64_encode($this->proxyUser . ':' . $this->proxyPass);
-                $context['http']['header'] .= 'Proxy-Authorization: Basic '
+                $context['http']['header'] = 'Proxy-Authorization: Basic '
```
williamdes, line 105: I already reviewed this while checking code mistakes
I am not sure it should be changed
There was something that made this concat sill have some sense
Not sure what

williamdes (changes requested):

libraries/classes/Utils/HttpRequest.php
kamil-tekiela: Ok, I will check it more carefully. I wasn't actually bothered to set up an environment for this to test as I need to disable CURL and enable `allow_url_fopen`. Let me see if it really throws an error as I'd expect it to.

kamil-tekiela: I checked it and when using concatenation it throws `Undefined array key "header"` error. Maybe in the past the code looked different and the concatenation made sense. As of now, I can't see any possibility that the code does not throw a warning.

libraries/classes/Database/Qbe.php
```diff
@@ -293,7 +293,7 @@ private function loadCriterias()
     /**
      * Getter for current search
      *
-     * @return SavedSearches
+     * @return ?SavedSearches
```
williamdes, line 296:
```suggestion
     * @return SavedSearches|null
```

libraries/classes/Utils/HttpRequest.php
williamdes: Thank you

If I remember maybe this code is a bug because it would erase all the other headers
Not sure anymore

williamdes (approved):

libraries/classes/Utils/HttpRequest.php
kamil-tekiela: I agree that this looks like a bug. But if it is a bug then I do not know what was the desired behaviour. The name of the method is `handleContext` which doesn't tell us much about what the method should do. Let's remove the concatenation for now, and if someone wants to study this code in the future, they can fix the logic.

MauricioFauth (approved):

mauriciofauth:

PR #17223, opened 2021-12-03
Make $subPart non-nullable and remove redundant ?? when it's not null

kamil-tekiela:
The whole global `$sub_part` looks fishy, but I wasn't able to identify what purpose it actually fulfils as a global. To remove static analysis issues I removed `??` in places where it was clearly defined and made the parameter non-nullable. According to my analysis it was set to null only in a single place where it was not used anyway. 

Side note: I have been running Psalm and PHPStan with PHP 8.0 until now but recent changes have made it unmanageable so I had to downgrade to PHP 7.4. 7d3c84a88163237211b9908978d473e91657ce3f has reversed psalm-baseline.xml update so when I run `-set-baseline=` again it added the same lines once more. I don't know if it was intentional or not.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1138,12 +1138,6 @@
       <code>$each_table</code>
       <code>$table_select</code>
     </MixedAssignment>
-    <RedundantCondition occurrences="1">
-      <code>$sub_part</code>
-    </RedundantCondition>
-    <TypeDoesNotContainNull occurrences="1">
-      <code>''</code>
-    </TypeDoesNotContainNull>
   </file>
   <file src="libraries/classes/Controllers/Database/ImportController.php">
     <MixedArgument occurrences="2">
@@ -1237,7 +1231,7 @@
       <code>$db</code>
       <code>$db</code>
     </MixedAssignment>
-    <RedundantCondition occurrences="11">
+    <RedundantCondition occurrences="10">
       <code>! $_error</code>
       <code>! $_error</code>
       <code>! $_error</code>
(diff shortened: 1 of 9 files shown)


psalm-baseline.xml
```diff
@@ -1,5 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <files psalm-version="4.13.1@5cf660f63b548ccd4a56f62d916ee4d6028e01a3">
+  <file src="config.inc.php">
```
mauriciofauth, line 3: Psalm reported errors in your local config file. This should be ignored.

MauricioFauth (changes requested):

psalm-baseline.xml
kamil-tekiela: Of course, I forgot about it. That was stupid of me, thanks for explaining.

mauriciofauth:
> The whole global `$sub_part` looks fishy, but I wasn't able to identify what purpose it actually fulfils as a global.

This looks like a leftover from when the router was introduced in phpMyAdmin 5.1. This was part of the entry point file names. For example: `server_export.php`, `db_export.php` and `tbl_export.php`.

> Side note: I have been running Psalm and PHPStan with PHP 8.0 until now but recent changes have made it unmanageable so I had to downgrade to PHP 7.4. [7d3c84a](https://github.com/phpmyadmin/phpmyadmin/commit/7d3c84a88163237211b9908978d473e91657ce3f) has reversed psalm-baseline.xml update so when I run `-set-baseline=` again it added the same lines once more. I don't know if it was intentional or not.

In fact, there are some issues running PHPStan with PHP 8.0+. If we make PHPStan green with PHP 7, PHPStan with PHP 8 will fail and vice versa.

kamil-tekiela:
> This looks like a leftover from when the router was introduced in phpMyAdmin 5.1. This was part of the entry point file names. For example: `server_export.php`, `db_export.php` and `tbl_export.php`.

Thanks. I won't change it anymore at the moment. I still need to understand more about this codebase before I can confidently do bigger changes. 

> In fact, there are some issues running PHPStan with PHP 8.0+. If we make PHPStan green with PHP 7, PHPStan with PHP 8 will fail and vice versa.

Ok, it's not a problem for me. I can run it using PHP 7.4.

williamdes (approved):

MauricioFauth (approved):

mauriciofauth:

PR #17227, opened 2021-12-06
Static analysis 3

kamil-tekiela:
This is another one of static analysis fixes. This time I have executed the optional PHPStan checks `checkAlwaysTrueCheckTypeFunctionCall` and `checkAlwaysTrueStrictComparison`. The fixes below are mostly from these two checks. I have left many others unfixed as they were either false positives or not so easy to fix. 

There are a couple of changes here that I am unsure about:
- `getTrackingTable()` looks like it might be having side effects, but its return value is not used meaningfully. I wasn't sure if it can be removed or not. 
- parameter `$field_fulltext` in `ColumnDefinition::displayForm()` has never been used in the code. Therefore the whole functionality looks to be unused. I removed it, but maybe it was a bug and the functionality is needed just not used currently. 
- on line 1330 in Export.php there is a `@var` annotation that isn't correct. It's possible that `Core::fatalError` should be marked as terminating in Psalm and PHPStan config and then the `@var` annotation should contain `null` type as well. I couldn't decide on the right choice of action and left it as it is now. 

Please let me know if I should revert some of these changes.

### DIFF
--- a/libraries/classes/Core.php
+++ b/libraries/classes/Core.php
@@ -869,10 +869,6 @@ class Core
      */
     public static function safeUnserialize(string $data)
     {
-        if (! is_string($data)) {
-            return null;
-        }
-
         /* validate serialized data */
         $length = strlen($data);
         $depth = 0;
--- a/libraries/classes/Core.php
+++ b/libraries/classes/Core.php
@@ -101,7 +101,7 @@ class Core
          * (this can happen on early fatal error)
          */
         if (
-            isset($dbi, $GLOBALS['config']) && $dbi !== null
+            isset($dbi, $GLOBALS['config'])
             && $GLOBALS['config']->get('is_setup') === false
             && ResponseRenderer::getInstance()->isAjax()
         ) {
@@ -494,8 +494,7 @@ class Core
         // inform the server that compression has been done,
         // to avoid a double compression (for example with Apache + mod_deflate)
         $notChromeOrLessThan43 = $GLOBALS['config']->get('PMA_USR_BROWSER_AGENT') !== 'CHROME' // see bug #4942
-            || ($GLOBALS['config']->get('PMA_USR_BROWSER_AGENT') === 'CHROME'
-                && $GLOBALS['config']->get('PMA_USR_BROWSER_VER') < 43);
+            || $GLOBALS['config']->get('PMA_USR_BROWSER_VER') < 43;
 
         if (str_contains($mimetype, 'gzip') && $notChromeOrLessThan43) {
             $headers['Content-Encoding'] = 'gzip';
--- a/libraries/classes/Core.php
+++ b/libraries/classes/Core.php
@@ -491,12 +491,10 @@ class Core
         }
 
         $headers['Content-Type'] = $mimetype;
+
         // inform the server that compression has been done,
         // to avoid a double compression (for example with Apache + mod_deflate)
-        $notChromeOrLessThan43 = $GLOBALS['config']->get('PMA_USR_BROWSER_AGENT') !== 'CHROME' // see bug #4942
-            || $GLOBALS['config']->get('PMA_USR_BROWSER_VER') < 43;
-
-        if (str_contains($mimetype, 'gzip') && $notChromeOrLessThan43) {
+        if (str_contains($mimetype, 'gzip') && $GLOBALS['config']->get('PMA_USR_BROWSER_AGENT') !== 'CHROME') {
             $headers['Content-Encoding'] = 'gzip';
         }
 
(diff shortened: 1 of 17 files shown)


libraries/classes/Core.php
```diff
@@ -494,8 +494,7 @@ public static function downloadHeader(
         // inform the server that compression has been done,
         // to avoid a double compression (for example with Apache + mod_deflate)
         $notChromeOrLessThan43 = $GLOBALS['config']->get('PMA_USR_BROWSER_AGENT') !== 'CHROME' // see bug #4942
-            || ($GLOBALS['config']->get('PMA_USR_BROWSER_AGENT') === 'CHROME'
-                && $GLOBALS['config']->get('PMA_USR_BROWSER_VER') < 43);
+            || $GLOBALS['config']->get('PMA_USR_BROWSER_VER') < 43;
```
mauriciofauth, line 497: phpMyAdmin does not support Chrome < 43 anymore. phpMyAdmin 5.1 supports Chrome >= 45 and phpMyAdmin 5.2 (master) supports Chrome >= 60. See the `.browserslistrc` file.

kamil-tekiela: Ok, then I understand that this check can be safely removed. I simplified it to just check for CHROME and nothing else.

libraries/classes/Display/Results.php
```diff
@@ -483,8 +483,7 @@ private function setDisplayPartsForShow(array $displayParts)
         $bIsProcessList = isset($which[1]);
         if ($bIsProcessList) {
             $str = ' ' . strtoupper($which[1]);
-            $bIsProcessList = $bIsProcessList
-                && strpos($str, 'PROCESSLIST') > 0;
+            $bIsProcessList = strpos($str, 'PROCESSLIST') > 0;
```
williamdes, line 486: I need to research why this was made this way

libraries/classes/Server/Privileges.php
```diff
@@ -623,7 +621,7 @@ public function getHtmlToDisplayPrivilegesTable(
                 $row = [];
                 if ($db === '*') {
                     $sqlQuery = 'SHOW COLUMNS FROM `mysql`.`user`;';
-                } elseif ($table === '*') {
+                } else {
                     $sqlQuery = 'SHOW COLUMNS FROM `mysql`.`db`;';
```
williamdes, line 625: moving this above the else will avoid using an else

libraries/classes/Tracker.php
```diff
@@ -85,9 +85,9 @@ public static function isActive(): bool
             return false;
         }
 
-        $table = self::getTrackingTable();
+        self::getTrackingTable();
```
williamdes, line 88: why are we not using the value ?
the call is now useless

kamil-tekiela: >getTrackingTable() looks like it might be having side effects, but its return value is not used meaningfully. I wasn't sure if it can be removed or not.

I think we could remove it, but I have no idea why it was there in the first place. All I can tell for sure is that the value was not used at all. Should I remove the call? Is it safe?

libraries/classes/Server/Privileges.php
kamil-tekiela: Or even better we can reduce repetition with a ternary.

libraries/classes/Server/Privileges.php
```diff
@@ -621,11 +619,7 @@ public function getHtmlToDisplayPrivilegesTable(
         if (empty($row)) {
             if ($table === '*' && $this->dbi->isSuperUser()) {
                 $row = [];
-                if ($db === '*') {
-                    $sqlQuery = 'SHOW COLUMNS FROM `mysql`.`user`;';
-                } elseif ($table === '*') {
-                    $sqlQuery = 'SHOW COLUMNS FROM `mysql`.`db`;';
-                }
+                $sqlQuery = 'SHOW COLUMNS FROM `mysql`.' . ($db === '*' ? '`user`;' : '`db`;');
```
williamdes, line 622:
```suggestion
                $sqlQuery = 'SHOW COLUMNS FROM `mysql`.' . ($db === '*' ? '`user`' : '`db`'). ';';
```

libraries/classes/Tracker.php
mauriciofauth: As `$relationParameters->trackingwork` is `true` at this point, `$relationParameters->db` and `$relationParameters->tracking` are not `null`, so the `self::getTrackingTable()` call can be safely removed. This is confusing, but I am working on it.

```php
return $relationParameters->trackingwork;
```

MauricioFauth (approved):

libraries/classes/Tracker.php
kamil-tekiela: Applied the suggestion and removed the `self::getTrackingTable()` call.

PR #17230, opened 2021-12-07
FR: Improve performance of Config classes

kamil-tekiela:
This PR is a feature request for a very simple performance improvement. It contains two unrelated changes:

1. In the Descriptions.php class there are 3 methods that are providing an array of descriptions in the currently selected language. They are calling `__()` internally which proves to be very expensive. It essentially calls a number of functions for each text multiple times. To avoid this, we call `__()` once for each string and buffer the results in a static variable. When the array is requested again, we provide buffered results. The assumption here is that the results of `__()` should not change in between. 
2. The functionality for flattening array is usually complex. It uses private property, multiple `array_walk`s, recursive functions, and unnecessary closures. I have considered using `RecursiveIteratorIterator` but due to the `! isset($value[0])` I couldn't use it, and it was also slower than current implementation. The next best thing was a simple loop with a recursive call. This proved to be 50%-90% faster than current implementation and much simpler. 

Altogether, these improvements reduced page load time on my development machine by 30% without any regressions. I hope the performance improves in other environments too.

Ref: #16005

### DIFF
--- a/libraries/classes/Config/ConfigFile.php
+++ b/libraries/classes/Config/ConfigFile.php
@@ -12,7 +12,6 @@ use PhpMyAdmin\Core;
 use function array_diff;
 use function array_flip;
 use function array_keys;
-use function array_walk;
 use function count;
 use function is_array;
 use function preg_replace;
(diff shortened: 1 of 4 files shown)


MauricioFauth (approved):

mauriciofauth:
Merged. Thank you very much! :100:

williamdes (commented):
💯 
Awesome !

PR #17233, opened 2021-12-08
Refactor CreateAddField

kamil-tekiela:
This PR removes one private method which wasn't helping with anything and replaces it with simple `foreach` loops. It also removes some temporary variables that didn't help with readability. Finally, it changes the output by removing unnecessary whitespace from the generated SQL. Whitespace is not important and what I removed seems to be completely unnecessary.

### DIFF
--- a/libraries/classes/Controllers/Table/AddFieldController.php
+++ b/libraries/classes/Controllers/Table/AddFieldController.php
@@ -107,16 +107,16 @@ class AddFieldController extends AbstractController
 
             $createAddField = new CreateAddField($this->dbi);
 
-            $sqlQuery = $createAddField->getColumnCreationQuery($table);
+            $sql_query = $createAddField->getColumnCreationQuery($table);
 
             // If there is a request for SQL previewing.
             if (isset($_POST['preview_sql'])) {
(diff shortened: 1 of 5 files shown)


libraries/classes/Controllers/Table/AddFieldController.php
```diff
@@ -107,16 +107,16 @@ public function __invoke(): void
 
             $createAddField = new CreateAddField($this->dbi);
 
-            $sqlQuery = $createAddField->getColumnCreationQuery($table);
+            $sql_query = $createAddField->getColumnCreationQuery($table);
```
mauriciofauth, line 110: Please use camel case where possible.

MauricioFauth (changes requested):

libraries/classes/Controllers/Table/AddFieldController.php
kamil-tekiela: It's a global variable. Is it safe to rename it? If so, I think it needs to be done globally, right?

MauricioFauth (approved):

mauriciofauth:

PR #17235, opened 2021-12-09
Fix var annotation for Charset and Collation

kamil-tekiela:
This PR attempts to fix types annotated in phpdoc for these two classes. Instead of putting `@var` in many places, specify the right return types and parameter types. I also changed Twig template to call getters directly. This avoids unnecessary loops in the controllers. 

While changing this, I have noticed that findCollationByName() contains a redundant `if` statement. The result of `explode()` will always be `non-empty-list<string>` so that `if` statement does not apply.

### DIFF
--- a/libraries/classes/Charsets.php
+++ b/libraries/classes/Charsets.php
@@ -27,7 +27,7 @@ class Charsets
     /**
      * MySQL charsets map
      *
-     * @var array
+     * @var array<string, string>
      */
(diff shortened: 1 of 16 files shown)


MauricioFauth (approved):

mauriciofauth:

williamdes (commented):
💯

PR #17236, opened 2021-12-09
Remove most `sleep` calls from Selenium tests

MauricioFauth:
Uses `instant` scrolls instead of `smooth` scrolls.


williamdes (commented):
💯

Issue #17239, opened 2021-12-09
Repeating headers are not working properly


kamil-tekiela:
### Describe the bug

phpMyAdmin has the functionality of repeating headers for results. By default only 25 results show per page and repeated headers are set to 100 so this functionality isn't used very often. However, when enabled the headers are not shown. 

### To Reproduce

1. Execute SELECT
2. Show more than 100 rows
3. Scroll down to 100th row
4. No repeated header

### Expected behavior

There should be table header after every 100 rows.

### Additional context

The problem is because of the following line in `libraries/classes/Display/Results.php`:

```
            // add repeating headers
            if (
                ($rowNumber != 0) && ($_SESSION['tmpval']['repeat_cells'] != 0)
                && ! $rowNumber % $_SESSION['tmpval']['repeat_cells']
            ) {
                $tableBodyHtml .= $this->getRepeatingHeaders($displayParams);
            }
```

The condition is always false and Psalm warns about it. The correct condition should be:
```diff
            // add repeating headers
            if (
-                ($rowNumber != 0) && ($_SESSION['tmpval']['repeat_cells'] != 0)
-                && ! $rowNumber % $_SESSION['tmpval']['repeat_cells']
+                ($rowNumber !== 0) && ($_SESSION['tmpval']['repeat_cells'] > 0)
+                && ($rowNumber % $_SESSION['tmpval']['repeat_cells']) === 0
            ) {
                $tableBodyHtml .= $this->getRepeatingHeaders($displayParams);
            }
````

I will fix this code, but as the proper fix demands changes to CSS/HTML I would like to keep this bug open.

mauriciofauth:
Closed by #17243

williamdes:
Did this affect 5.1?

williamdes:
> ```diff
> getRepeatingHeaders
> ```

Answer to myself: Yes, in 5.1 repeating headers do not work

williamdes:
Cherry-picked into 5.1 as b5a6729886f27e614fee113209aa5bd166f6699b

williamdes:
Also e84e1bda0233910e15c5d7e7bb52d3d246573437 to fix:
![image](https://user-images.githubusercontent.com/7784660/152684578-11e22c78-1ac8-4787-ac7a-080fe7eea131.png)

PR #17243, opened 2021-12-12
Refactor results.php

kamil-tekiela:
This PR is a little chaotic as I wasn't very organized when refactoring this class. The goal was to reduce complexity. I removed redundant code, checks, temporary variables and unnecessary parameters. I added type hints, type declarations, stricter conditions, early returns, etc. 

This PR contains 2 bug fixes:
- Bug #17239, which should be fixed in full. I wasn't able to determine if one of the existing unit tests should cover this. Can someone suggest which test to modify?
- Undocumented bug related to sorting functionality. When using dynamic column names (e.g. ``SELECT CONCAT(table.col1, table1.`col2`)``) the sorting wouldn't work. The logic still seems flaky, but at least that small inconsistency is fixed. 

I have removed some methods that didn't look useful at all. I have completely redesigned or changed the signatures of a few other methods. 

I decided to remove unnecessary whitespace from HTML. HTML ignores contiguous whitespace. This should reduce slightly the length of the HTML output. I had to adjust unit tests to account for the removed whitespace. A lot of the functionality involved in the generation of CSS classes was outdated and no longer required. 

While the goal wasn't to reduce the length of code, this PR reduced it by 157 lines. It also got rid of many Psalm and PHPStan issues. I tested it by running PHPUnit locally and performing UAT myself. I haven't noticed any accidental regressions. 

P.S. This will probably be the last one of the small refactorings for some time. I will try to spend some time on more meaningful fixes or refactorings. For example, the master branch contains few issues related to PHP 8.1 changes. I will try to hunt them down and fix them.

### DIFF
--- a/js/src/makegrid.js
+++ b/js/src/makegrid.js
@@ -1725,7 +1725,7 @@ var makeGrid = function (t, enableResize, enableReorder, enableVisib, enableGrid
 
             // make sure we have more than one column
             if ($firstRowCols.length > 1) {
-                var $colVisibTh = $(g.t).find('th:not(.draggable)');
+                var $colVisibTh = $(g.t).find('th:not(.draggable)').slice(0, 1);
                 Functions.tooltip(
                     $colVisibTh,
                     'th',
--- a/libraries/classes/Display/Results.php
+++ b/libraries/classes/Display/Results.php
@@ -4243,8 +4243,6 @@ class Results
         string $printLink,
         array $analyzedSqlResults
     ): array {
-        global $printview;
-
         $urlParams = [
             'db' => $this->properties['db'],
             'table' => $this->properties['table'],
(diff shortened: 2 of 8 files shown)


williamdes (approved):

MauricioFauth (approved):

mauriciofauth:
Merged. Thank you!

js/src/makegrid.js
```diff
@@ -1725,7 +1725,7 @@ var makeGrid = function (t, enableResize, enableReorder, enableVisib, enableGrid
 
             // make sure we have more than one column
             if ($firstRowCols.length > 1) {
-                var $colVisibTh = $(g.t).find('th:not(.draggable)');
+                var $colVisibTh = $(g.t).find('th:not(.draggable)').slice(0, 1);
```
williamdes, line 1728: I do not understand what does this fix change, can you help me ?

kamil-tekiela: When repeating headers are switched on the column selections buttons would appear in the wrong place. To overcome this, I decided to keep them only in the first header.

williamdes: Thanks, for the record here is a screenshot
image

PR #17246, opened 2021-12-13
Bug fix: Fix triple HTML encoding

kamil-tekiela:
Some values were double or even triple encoded in the HTML/JS. This resulted in garbled up text. This PR removed HTML escaping from places where it is not needed.

PR #17247, opened 2021-12-14
Bug fix: Fix triple HTML encoding

kamil-tekiela:
Some values were double or even triple encoded in the HTML/JS. This resulted in garbled up text. This PR removed HTML escaping from places where it is not needed.

Cherry-picked from #17246 onto QA_5_1 branch

### DIFF
--- a/libraries/classes/Controllers/Table/RelationController.php
+++ b/libraries/classes/Controllers/Table/RelationController.php
@@ -16,7 +16,6 @@ use PhpMyAdmin\Util;
 use function array_key_exists;
 use function array_keys;
 use function array_values;
-use function htmlspecialchars;
 use function mb_strtoupper;
 use function md5;
 use function strtoupper;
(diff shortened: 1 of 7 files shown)


libraries/classes/Controllers/Table/RelationController.php
```diff
@@ -322,14 +321,11 @@ public function getDropdownValueForTable(): void
...
-        foreach ($columnList as $column) {
-            $columns[] = htmlspecialchars($column);
-        }
         if ($GLOBALS['cfg']['NaturalOrder']) {
-            usort($columns, 'strnatcasecmp');
+            usort($columnList, 'strnatcasecmp');
         }
-        $this->response->addJSON('columns', $columns);
+
+        $this->response->addJSON('columns', $columnList);
```
williamdes, line 328: I can confirm this part of the fix in the column select to add a foreign key when the column has a strange name

williamdes (approved):
Could you provide ways to test all the parts of this fix please ?
I could confirm one of the changes

kamil-tekiela:
1. Create a table with varchar(255) as a primary key. Enter a record with primary key `'"<>&a`. Press `delete` button to delete that row. The SQL is garbled.
2. Create a table called `'"<>&a` with a column called `'"<>&a`. Create index on the column with the default name. Try to drop that index. The SQL is garbled.
3. Create a table `'"<>&a` with a column `'"<>&a` and `foo`. Create relationship called `'"<>&a` on the table `'"<>&a` on the column `foo` pointing to column `'"<>&a`. Try to force the value dropdown via AJAX. The column name is garbled and cannot be selected. 

There is also 1 more error when you try to delete a column called `'"<>&a` but I haven't fixed that one. 

I hope these steps can be reproduced, let me know if not and I will prepare more detailed steps.

MauricioFauth (approved):

PR #17253, opened 2021-12-21
Remove $row_number from DatabaseInterface::fetchValue

kamil-tekiela:
The parameter is not used anywhere. It is always set to 0. This doesn't sound particularly useful. It's most likely a remnant from PHP 4 era, where existed a function called `mysqli_result`. 

Removing this parameter helps with further changes I have planned.


williamdes (approved):
Very nice catch ! 💯

MauricioFauth (approved):

PR #17263, opened 2021-12-29
Major refactoring of Dbal pt.1

kamil-tekiela:
Primary goal: Abstract `mysqli_result` class by providing a new contract for accessing the query results.

Secondary goals:
- Provide reasonable type hinting for Dbal results
- Remove less useful functions
  - `fetchArray` isn't very useful and most uses can be replaced with either fetchAssoc or fetchRow
  - `freeResult` is not utilized properly. There's usually no need to free the internal mysqli result set. The only possible situation I can think of would be if there's a chance the object is leaked out of scope and the unbuffered queries are used. `mysqli_result::close()` would fetch all remaining rows from MySQL and immediately discard them, thus freeing the connection, but I haven't observed any such use in this project. `unset` can be used in places where it might have had some sense to use that function before. 
  - `fieldName` does exactly what `getFieldsMeta` does but only for a single field. Yet, all uses of it are done in a loop for all fields
  - `fieldLen` is used only once in the whole project and does the same thing as `getFieldsMeta`. For a single usage, a tiny bit of performance can be sacrificed
- Add new helper methods to abstract a little bit of common functionality related to fetching results. This is inspired by PDO
- Make the result set iterable. `mysqli_result` is natively iterable, but due to the way it was used in phpMyAdmin, one could never use `foreach`
- Reduce union types with false and null. e.g. `getFieldsMeta` was usually used with `?? []`, so why doesn't it just always return an array? `fetchAssoc` was usually used in `while` loop where an empty array is coerced into `false` just as `null` is. Many of `DbalInterface` functions took `false` as a possible argument, e.g. `numRows`. The union types made it more difficult to spot problems. Nonetheless, the goal is to cause as little disruption as possible.
- Replace `mixed` with proper types in some places, e.g. `fetchValue` or `query`
- Reveal all places where a deprecation notice might be raised in PHP 8.1. 

The PR grew significantly in size, which is I decided to split it into smaller parts. This is the main change that prepares further changes I have planned. As I have seen the project uses mostly textual API to MySQL and relies on implicit setting `MYSQLI_OPT_INT_AND_FLOAT_NATIVE` to be off. Changing this would be way too disruptive as would be switching over to prepared statements. For this reason, the values are typed as `string|null`. If prepared statements would be used we would have to think about handling `string|int|float|null`. 

Prepared statements are used in one place. I decided to use `MysqliResult` there as well, despite the problem stated above. I don't think this should be an issue though. In the future, the project should consider some ways of dealing with this. 

Mysqli is used with error reporting disabled. I will investigate the possibility of enabling error reporting as it should be relatively simple. A simple try-catch should solve the problem. The native error reporting is usually better. 

The problems faced with this PR:
- `DbiDummy` produces a message "Not supported query" in one place. I wasn't able to fix it, but it doesn't cause a test failure. I have no idea what I messed up there...
- The tests are less than ideal. I think I had the biggest problem getting `InsertEditTest` to work properly. This is testing just way too much of functionality. In general, tests are testing too much, which means that some paths are not covered or are ignored. `assertEquals` is used in many places instead of `assertSame` and `expects($this->any())` is used a lot. This makes it difficult to understand what exactly the test is meant to validate. 
- Tests provide wrong value types in many instances. `query` returned strings or ints instead of real objects. Some tests relied on implicit `null` return values. Some tests expected an int but the function produced a string. etc. 
- PHP has a weird feature that allows you to destructure null values as if they were arrays. https://3v4l.org/DMa69 This meant that achieving one of my secondary goals was challenging. `while([$someValue] = $dbi->fetchRow($result))` worked perfectly fine, but when replaced with `while([$someValue] = $result->fetchRow())` would throw an error. 
- `tryQuery` might produce `false` but the code doesn't check for it in some places. 
- I have absolutely no idea what `TrackerTest` is testing for. I made the test execute again, but I think it should be improved. 
- `Relation::queryAsControlUser` is a method that should belong to `DatabaseInterface` and should be split into two separate methods. Type hinting it for Psalm and PHPStan proved to be impossible. Next PR should solve this issue. 

I understand this PR is not perfect and I would really appreciate how it can be improved. Nonetheless, I think this is an overall improvement to the project.

### DIFF
--- a/libraries/classes/ConfigStorage/Relation.php
+++ b/libraries/classes/ConfigStorage/Relation.php
@@ -6,6 +6,7 @@ namespace PhpMyAdmin\ConfigStorage;
 
 use PhpMyAdmin\ConfigStorage\Features\PdfFeature;
 use PhpMyAdmin\DatabaseInterface;
+use PhpMyAdmin\Dbal\ResultInterface;
 use PhpMyAdmin\InternalRelations;
 use PhpMyAdmin\RecentFavoriteTable;
 use PhpMyAdmin\SqlParser\Parser;
@@ -29,7 +30,6 @@ use function htmlspecialchars;
 use function implode;
 use function in_array;
 use function is_array;
-use function is_bool;
 use function is_scalar;
 use function is_string;
 use function ksort;
@@ -69,8 +69,12 @@ class Relation
      * @param string $sql        the query to execute
      * @param bool   $show_error whether to display SQL error messages or not
      * @param int    $options    query options
+     * @psalm-param T $show_error
      *
-     * @return mixed|bool the result set, or false if no result set
+     * @return ResultInterface|false the result set, or false if no result set
+     * @psalm-return (T is true ? ResultInterface : ResultInterface|false)
+     *
+     * @template T as bool
      */
     public function queryAsControlUser($sql, $show_error = true, $options = 0)
     {
@@ -78,19 +82,11 @@ class Relation
         // is called for tracking purposes but we want to display the correct number
         // of rows affected by the original query, not by the query generated for
         // tracking.
-        $cache_affected_rows = false;
-
         if ($show_error) {
-            $result = $this->dbi->query($sql, DatabaseInterface::CONNECT_CONTROL, $options, $cache_affected_rows);
-        } else {
-            $result = @$this->dbi->tryQuery($sql, DatabaseInterface::CONNECT_CONTROL, $options, $cache_affected_rows);
+            return $this->dbi->query($sql, DatabaseInterface::CONNECT_CONTROL, $options, false);
         }
 
-        if ($result) {
-            return $result;
-        }
-
-        return false;
+        return $this->dbi->tryQuery($sql, DatabaseInterface::CONNECT_CONTROL, $options, false);
     }
 
     public function getRelationParameters(): RelationParameters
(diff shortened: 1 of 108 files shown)


libraries/classes/ConfigStorage/Relation.php
```diff
@@ -69,28 +69,24 @@ public function __construct(DatabaseInterface $dbi)
...
-            $result = @$this->dbi->tryQuery($sql, DatabaseInterface::CONNECT_CONTROL, $options, $cache_affected_rows);
-        }
-
-        if ($result) {
-            return $result;
+            return $this->dbi->query($sql, DatabaseInterface::CONNECT_CONTROL, $options, false);
         }
 
-        return false;
+        return @$this->dbi->tryQuery($sql, DatabaseInterface::CONNECT_CONTROL, $options, false);
```
williamdes, line 89: could we avoid using `@` ?

kamil-tekiela: Definitely. I strongly believe that every `@` should have a corresponding comment explaining which warning it suppresses. I can't see what possible warning it could suppress here.

MauricioFauth (approved):
Thank you very much. This is a much needed change. :100:

kamil-tekiela:
I'm happy to hear that. I rebased it against master, but if there's something I can improve in this PR, please let me know

PR #17272, opened 2022-01-06
Query as control user

kamil-tekiela:
Split the method into two methods, where one returns false instead of displaying an error message. Move the methods to DatabaseInterface. 

The new methods will always return buffered results and they will never cache the rows. 

In theory, these methods are not very useful as they are just syntactic sugar for `query()` and `tryQuery()`. I think it at least helps with static analysis and visual comprehension. Code with multiple arguments would be less understandable than a dedicated method.


MauricioFauth (approved):

williamdes (approved):
💯 
This is way more clear

PR #17274, opened 2022-01-07
Minor refactoring of Sql.php

kamil-tekiela:
Just fixing a couple of Psalm/PHPStan issues and improving ease of reading of some code.

### DIFF
--- a/libraries/classes/Sql.php
+++ b/libraries/classes/Sql.php
@@ -1610,9 +1610,7 @@ class Sql
                 $tableFromSql,
             ] = ParseAnalyze::sqlQuery($sqlQuery, $db);
 
-            if ($table != $tableFromSql && $tableFromSql) {
-                $table = $tableFromSql;
-            }
+            $table = $tableFromSql ?: $table;
         }
 
         return $this->executeQueryAndGetQueryResponse(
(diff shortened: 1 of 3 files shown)


libraries/classes/Sql.php
```diff
@@ -1336,40 +1345,41 @@ private function getHtmlForPreviousUpdateQuery(
...
+            $text = __(
+                'Current selection does not contain a unique column.'
+                . ' Grid edit, Edit, Copy and Delete features may result in'
+                . ' undesired behavior. %s'
+            );
+        } else {
+            return '';
+        }
+
+        return Message::notice(
```
williamdes, line 1374: Why the else is needed ?
I am not sure to understand how it would reach the last return

kamil-tekiela: Good catch! That is clearly my mistake. I will fix it tomorrow. I might even roll it back to what it was before as my attempt to make it more readable wasn't very successful.

kamil-tekiela: I reverted this change partially. It wasn't more readable than before.

williamdes (approved):

MauricioFauth (approved):

mauriciofauth:

PR #17280, opened 2022-01-09
fix themes cards heights

faissaloux:
Hello 👋🏻

In this PR I've fixed themes cards heights

# Before
Screenshot 2022-01-10 002218

# After
Screenshot 2022-01-10 002227


MauricioFauth (changes requested):
I don't like this change because it stretches the image. Maybe cropping the image would be a better solution.

faissaloux:
@mauriciofauth Cropped!

Screenshot 2022-02-05 191536

mauriciofauth:
Sorry, I meant cropping the image using CSS.

faissaloux:
@mauriciofauth done!

williamdes (approved):
The code looks good but I really think some bootstrap class could be used instead of the `height` attributes

mauriciofauth:
This makes the images have the same height, but the cards still could have different heights. This will help with that.

```diff
diff --git a/templates/home/themes.twig b/templates/home/themes.twig
index 3fbc4b4e40..604e8d7e90 100644
--- a/templates/home/themes.twig
+++ b/templates/home/themes.twig
@@ -3,7 +3,7 @@
   <div class="row row-cols-1 row-cols-md-2 row-cols-lg-3 row-cols-xl-4 g-4">
     {% for theme in themes %}
       <div class="col">
-        <div class="card">
+        <div class="card h-100">
           <img src="./themes/{{ theme.id }}/screen.png" class="card-img-top" alt="{{ 'Screenshot of the %s theme.'|trans|format(theme.name) }}">
           <div class="card-body">
             <h5 class="card-title">{{ theme.name }} <small>({{ theme.version }})</small></h5>
```

faissaloux:
@mauriciofauth Fixed!

themes/bootstrap/scss/_card.scss
```diff
@@ -0,0 +1,8 @@
+.card-img-container {
+    height: 140px;
+    overflow: hidden;
+    img {
+      height: 100%;
```
williamdes, line 5: `h-100` can replace this :)

themes/bootstrap/scss/_card.scss
```diff
@@ -0,0 +1,8 @@
+.card-img-container {
+    height: 140px;
```
williamdes, line 2: maybe some boostrap class could remove the need for this fixed value, I am thinking about very large screens ^^

themes/bootstrap/scss/_card.scss
faissaloux: Yeah right, I'll fix this :)

themes/bootstrap/scss/_card.scss
faissaloux: What about `vh` unit!!

themes/bootstrap/scss/_card.scss
```diff
@@ -0,0 +1,7 @@
+.card-img-container {
+  height: 16vh;
```
williamdes, line 2: https://caniuse.com/viewport-units
For the record

williamdes (approved):

themes/bootstrap/scss/_card.scss
faissaloux: Emmm yeah thank you ^^

williamdes: Saying this because we have https://github.com/phpmyadmin/phpmyadmin/blob/master/.browserslistrc to respect ;)

mauriciofauth, line 2: You should extract this value into a variable.

templates/home/themes.twig
```diff
@@ -3,8 +3,10 @@
   <div class="row row-cols-1 row-cols-md-2 row-cols-lg-3 row-cols-xl-4 g-4">
     {% for theme in themes %}
       <div class="col">
-        <div class="card">
-          <img src="./themes/{{ theme.id }}/screen.png" class="card-img-top" alt="{{ 'Screenshot of the %s theme.'|trans|format(theme.name) }}">
+        <div class="card h-100 mt-0">
```
mauriciofauth, line 6: This is redundant since cards do not have margin.
```suggestion
        <div class="card h-100">
```

MauricioFauth (changes requested):

templates/home/themes.twig
faissaloux: Actually `.card` have a `margin-top: 2rem;` 😕

themes/bootstrap/scss/_card.scss
faissaloux: Oh! alright, good to know :)

faissaloux: Done!

templates/home/themes.twig
mauriciofauth: Oh, I see what you mean. The pmahomme's card class has a top margin, however cards in general doesn't have margin. So applying the class to the HTML doesn't make sense.

faissaloux: Can I remove the margin-top from css! I think that's the option I have. Isn't it using elsewhere!!

williamdes: FYI: https://getbootstrap.com/docs/5.0/components/card/#variables

I am not sure this helps but in case here is the link where they could have defined the `margin-top` value

faissaloux: @williamdes No it's not bootstrap style, In `original` and `pmahomme` themes we have that `margin-top` for `.card`.

I wanted to remove it and it will be fixed but afraid it's used somewhere else !! 🤔

williamdes: Seems to have been added in f19f27942e315764667994c4a8f2f6c5542f12a6 by @mauriciofauth

mauriciofauth: @faissaloux You can leave the margin-top issue for another PR. The issue is that, for pmahomme and Original themes, the margin-top should only be applied if the card has a card-header.

faissaloux: @mauriciofauth Yeah exactly. But if we want to apply style to `.card` if only have `.card-header` we need to use js and there's no specific js files for specific theme (I guess) So in this case I think the `mt-0` class fixed this problem for themes cards and it have no other effects.

mauriciofauth:

williamdes:
@mauriciofauth should this be picked into 5.2 ?

mauriciofauth:
> @mauriciofauth should this be picked into 5.2 ?

I don't think it's needed. It's more like an enhancement than a fix.

PR #17303, opened 2022-01-20
Remove globals

kamil-tekiela:
Not sure if this is the best solution, but at least `$route` seems like it didn't need to be global at all.


williamdes (approved):

MauricioFauth (approved):

mauriciofauth:

PR #17310, opened 2022-01-22
Add CI workflow for PHPUnit tests

MauricioFauth:


williamdes (approved):

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,47 @@
+name: "Continuous Integration"
+
+on:
+  pull_request:
+  push:
+
+jobs:
+  phpunit:
+    name: "PHPUnit tests"
+    runs-on: "ubuntu-latest"
```
williamdes, line 10: Maybe use 18.04 because it seems apt broken for gettext

See: https://github.com/actions/virtual-environments

williamdes (approved):

.github/workflows/continuous-integration.yml
mauriciofauth: I fixed it by installing gettext before installing PHP.

williamdes:
Can you skip the selenium group to avoid the PHP errors ?

ibennetch:
This is only for 4.9, correct?

mauriciofauth:
> This is only for 4.9, correct?

Yes.

PR #17319, opened 2022-01-26
Remove code coverage for QA branches

MauricioFauth:
This will remove code coverage for QA branches (should keep it for master branch when merging), improving test time.

### DIFF
--- a/.github/workflows/lint-and-analyse-php.yml
+++ b/.github/workflows/lint-and-analyse-php.yml
@@ -43,6 +43,7 @@ jobs:
               uses: shivammathur/setup-php@v2
               with:
                   php-version: 7.1
+                  coverage: "none"
             - name: Validate composer.json and composer.lock
               run: composer validate
             - name: Get Composer Cache Directory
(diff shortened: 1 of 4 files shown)


williamdes (changes requested):
I like this change, but I will not approve it because of this argument: some code may change to fix something and make some code not covered. We may not notice this without having a coverage drop notification. The coverage drop is not really the point but more about introducing a very big issue or dead chunk of code without us noticing

mauriciofauth:
> I like this change, but I will not approve it because of this argument: some code may change to fix something and make some code not covered. We may not notice this without having a coverage drop notification. The coverage drop is not really the point but more about introducing a very big issue or dead chunk of code without us noticing 

If a change to a covered line doesn't make the tests fail means that it wasn't really covered by the tests.

What we can do is to require test for bug fixes. But this can be really hard to do for some parts of the code, especially for new contributors. Static analysis also helps with this.

I like the resource savings by removing the coverage, but we can think of better ways for future branches and leave the current ones as is.

Issue #17308, opened 2022-01-22
Navigation panel - table list is broken (5.1.2)


svivian:
### Describe the bug

The navigation panel in the latest version (5.1.2) is very broken. Using the page selector replaces the table list with a database list.

### To Reproduce

1. Open PMA.
2. Click a table name to see the list of tables.
3. Click the `>` button to move to the next page of tables.

Instead of showing the next page, it shows the list of databases instead. Example from the demo server:

![Screenshot 2022-01-22 at 13 13 12](https://user-images.githubusercontent.com/207887/150640130-6c149cd3-b952-41f3-8516-84dbf2a4e296.png)

![Screenshot 2022-01-22 at 13 13 27](https://user-images.githubusercontent.com/207887/150640135-16ff8eb9-4219-41cf-8719-97d484511047.png)

After that PMA is full of errors, for example clicking another database in the outer list says "No databases selected". Or sometimes the tables get put in a "Views" subcategory even if they're not views.

PMA 5.2.1, PHP 7.4, MySQL 5.7, latest Chrome. Also tested on demo server.

spyjo:
I confirm this bug after upgrading from 5.1.1 to 5.1.2
Selecting an other number with the select form is still functioning, but not the arrows.

ibennetch:
Git bisect tells me that 4fd7aa7a2c45e99c45d5a196e8c66cb24bd02787 is the first bad commit, "Use POST for Navigation's aPath and vPath."

mauriciofauth:
@ibennetch I opened two PRs. A fix for 4.9.x and another for 5.1.x.

PR #17323, opened 2022-01-27
[4.9] Fix broken pagination links in the navigation sidebar

MauricioFauth:
Fix for phpMyAdmin 4.9.

#17308


williamdes (approved):
Very strange 4.9 is affected

PR #17325, opened 2022-01-28
Fix broken tests

kamil-tekiela:
The first one fixes `$route` parameter. In real execution the value should have never been null, which means that it should always be present in the `action=` and the default value is `/`.

The second one is basically whitespace in the generated HTML. We can ask Twig to trim it around the `{%` but I don't think there's anyway to trim it around `<strong>`


williamdes (approved):

MauricioFauth (approved):

mauriciofauth:

PR #17332, opened 2022-02-02
Remove `db` and `table` properties from controllers

MauricioFauth:
Replaces them with their respective globals.

Since these properties have the same values of the global variables, it's better to get them from the globals or from the `PhpMyAdmin\Http\ServerRequest` object.


williamdes (commented):
💯

Issue #15582, opened 2019-11-22
Don't disable "Empty"-button after clicking


raimundmeyer:
### Is your feature request related to a problem? Please describe.
 I'm always frustrated when I'm working with data that is inserted programatically and I want to delete it periodically  in phpMyAdmin. In the browse view of a database there is a button to truncate the table, but after it is clicked, it is disabled until you refresh the page.  It is btw. not disabled  when the table is empty at page load.
![image](https://user-images.githubusercontent.com/50366929/69434131-4f21c200-0d3d-11ea-85ae-7c01f583dafe.png)

### Describe the solution you'd like
The "Empty" button in the table view should remain clickable after being clicked.

### Additional context

current master

Tithugues:
@williamdes @mauriciofauth @ibennetch , do you expect to have a fix for this please?
I agree with @raimundmeyer that it's a bit annoying.

williamdes:
> I agree with @raimundmeyer that it's a bit annoying.

I also agree, but it must be clear that the operation succeeded (eg: 0 rows in rows column, maybe it is already the case)

PR #17333, opened 2022-02-03
Remove `Search` link from tables that have no rows

faissaloux:
Hello 👋🏻
I think there is no need for a `Search` link in empty tables, so in this PR I've removed the `a` tag from `Search` in tables that have no rows.

Screenshot 2022-02-03 005531

Signed-off-by: Faissal Wahabali <fwahabali@gmail.com>

### DIFF
--- a/templates/database/structure/structure_table_row.twig
+++ b/templates/database/structure/structure_table_row.twig
@@ -49,9 +49,13 @@
         </a>
     </td>
     <td class="text-center d-print-none">
-        <a href="{{ url('/table/search', table_url_params) }}">
+        {% if may_have_rows %}
+            <a href="{{ url('/table/search', table_url_params) }}">
+        {% endif %}
           {{ may_have_rows ? get_icon('b_select', 'Search'|trans) : get_icon('bd_select', 'Search'|trans) }}
-        </a>
+        {% if may_have_rows %}
+            </a>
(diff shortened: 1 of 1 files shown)


ibennetch:
Hi,

Thanks for your suggestion. I think that in order to present a consistent
user interface it’s better if we don’t hide this in that situation. Also, a
user could insert data from another source while the empty table is open,
we could be in a situation where there is data but they can’t search for it.

That’s my opinion, any other thoughts?

williamdes:
> Hi, Thanks for your suggestion. I think that in order to present a consistent user interface it’s better if we don’t hide this in that situation. Also, a user could insert data from another source while the empty table is open, we could be in a situation where there is data but they can’t search for it. That’s my opinion, any other thoughts?

I agree about this, for the same reasons #15582 did enable at all times the empty button so that user can do something else in another tab

faissaloux:
> Hi, Thanks for your suggestion. I think that in order to present a consistent user interface it’s better if we don’t hide this in that situation. Also, a user could insert data from another source while the empty table is open, we could be in a situation where there is data but they can’t search for it. That’s my opinion, any other thoughts?

Yeah that make sense, I agree.

ibennetch:
Thanks, I appreciate your suggestion and continued interest in contributing.

PR #17334, opened 2022-02-03
remove search tab from empty tables menu

faissaloux:
Following #17333. In this PR I have removed `Search` tab from empty tables menu.

Signed-off-by: Faissal Wahabali <fwahabali@gmail.com>


williamdes (changes requested):
This is a very expensive operation, thank you for the idea. But for technical reasons I can not approve it :)

faissaloux:
@williamdes it's okay 😄

ibennetch:
As discussed in #17333, I think that in order to present a consistent user interface it’s better if we don’t hide this in that situation.

Thanks again for the suggestion.

PR #17351, opened 2022-02-07
Drop stripslashes

kamil-tekiela:
Was there any reason to strip slashes? What slashes are we stripping and why? It seems to me like some bad copy-pasta. 

I tested it with the following table and it definitely garbles up the result.
```sql
CREATE TABLE IF NOT EXISTS `slashes` (
  `\'table` int(11) NOT NULL,
  `\\Some table` int(11) NOT NULL,
  PRIMARY KEY (`\'table`)
```

### DIFF
--- a/libraries/classes/Plugins/Export/ExportCsv.php
+++ b/libraries/classes/Plugins/Export/ExportCsv.php
@@ -21,7 +21,6 @@ use function mb_strtolower;
 use function mb_substr;
 use function preg_replace;
 use function str_replace;
-use function stripslashes;
 use function trim;
 
 /**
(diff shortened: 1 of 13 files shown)


williamdes:
Looks okay but I will need to search in the git history to understand why they where added

mauriciofauth:
> Looks okay but I will need to search in the git history to understand why they where added

Looks like they are just added since from the beginning. Example: https://github.com/phpmyadmin/phpmyadmin/commit/9d6bed8d34534b97abb16899efc828a8a4d2b764#diff-f95b3fd37b383696b8386128a86960fbd5d590683b6ceb6539f37cd314e781f3R307

MauricioFauth (approved):

PR #17356, opened 2022-02-08
Fix table navigation layout if row count is unknown

lrb2:
### Description

Now that #17055 has been fixed, there are a pair of issues with the navigation buttons for system tables (and any table that has an unknown number of rows) that I have fixed. The End button was nonfunctional, so it is no longer shown, and the separator between the navigation buttons and the adjacent "Show all" checkbox was missing.

The following GIFs show the fix in action:

- Expected behavior (table with known row count):
NavLayoutKnownRowCountCurrent

- Current behavior (table with unknown row count):
NavLayoutUnknownRowCountCurrent

- Fixed behavior (table with unknown row count):
NavLayoutUnknownRowCountFixed

### DIFF
--- a/libraries/classes/Display/Results.php
+++ b/libraries/classes/Display/Results.php
@@ -933,6 +933,11 @@ class Results
             false
         );
 
+        // If the number of rows is unknown, stop here (don't add the End button)
+        if ($this->properties['unlim_num_rows'] === false) {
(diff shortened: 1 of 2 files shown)


MauricioFauth (approved):

mauriciofauth:

PR #17361, opened 2022-02-09
Add webpack bundler

MauricioFauth:
Adds webpack to bundle the JavaScript files and other assets.

The main difference from the previous attempt is that [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) wrappers are not being added to code. So we don't need to declare every global in the source files.

Related to #15580


mauriciofauth:
The generated code is almost the same as before.

webpack.config.js
```diff
@@ -0,0 +1,263 @@
...
+            'themes/metro/css/mono-theme': './themes/metro/scss/mono-theme.scss',
+            'themes/metro/css/redmond-theme': './themes/metro/scss/redmond-theme.scss',
+            'themes/metro/css/teal-theme': './themes/metro/scss/teal-theme.scss',
+            'themes/metro/css/theme': './themes/metro/scss/theme.scss',
+            'themes/original/css/theme': './themes/original/scss/theme.scss',
+            'themes/pmahomme/css/theme': './themes/pmahomme/scss/theme.scss',
+            'setup/styles': './setup/scss/styles.scss',
+        },
+        output: {
+            filename: 'build/css/[name].js',
```
mauriciofauth: webpack generates these empty .js files. I don't know how to not generate them, so I chose to output them into the build directory.

webpack.config.js
```diff
@@ -0,0 +1,263 @@
+const path = require('path');
+const webpack = require('webpack');
+const autoprefixer = require('autoprefixer');
+const CopyPlugin = require('copy-webpack-plugin');
+const WebpackConcatPlugin = require('webpack-concat-files-plugin');
+const MiniCssExtractPlugin = require('mini-css-extract-plugin');
+const WebpackRTLPlugin = require('webpack-rtl-plugin');
```
mauriciofauth: We should revisit this plugin in the future, as it's triggering some deprecation messages. And looks like it's an abandoned plugin.
https://github.com/phpmyadmin/phpmyadmin/runs/5148016820?check_suite_focus=true#step:6:14

williamdes (approved):
This really looks nice, maybe the gitattributes file and the release script cleanup will need adjustments

PR #17375, opened 2022-02-12
Fix broken escaping on Insert page

kamil-tekiela:
I have started working on fixing all escaping issues some time ago but then I got busy with other stuff and I never finished it. Here's another small fix. This time it is the Insert page that mangled up the HTML for fields with `"'<>&`.

I'll slowly chip away at the rest of the issues as I can still see plenty.

### DIFF
--- a/libraries/classes/InsertEdit.php
+++ b/libraries/classes/InsertEdit.php
@@ -363,7 +363,6 @@ class InsertEdit
         array $commentsMap,
         $timestampSeen
     ) {
-        $column['Field_html'] = htmlspecialchars($column['Field']);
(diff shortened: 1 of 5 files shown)


williamdes (approved):
Can you reset this on QA_5_1 and the cherry pick + force push your commits?

kamil-tekiela:
> Can you reset this on QA_5_1 and the cherry pick + force push your commits?

I can't do that. The bug was introduced in 5.2. I reset it to QA_5_2.

williamdes (approved):
Thank you

MauricioFauth (approved):

PR #17376, opened 2022-02-12
getDbInfo mangles table with numerical names

kamil-tekiela:
When tables have numerical keys `array_merge` will reindex the array keys and then we can't process them for example when changing table collations in bulk. Additionally, PHP arrays are ... strange, and when working with mixed numerical and string keys, they will cast numerical keys to int, so to avoid fatal error due to invalid parameter type, we simply don't use keys. The table names are available in the array structure anyway.


williamdes (approved):
If this is not a bug fix, the PR target is okay for me
Thank you for this improvement

MauricioFauth (approved):

PR #17377, opened 2022-02-12
Incorrect usage of escapeMysqlWildcards

kamil-tekiela:
I observed a bug when a column name is named `\'column` and you would like to change it, the application can't find a column with that name. 

The phpdoc clearly states that `escapeMysqlWildcards()` does not escape backslashes. However, backslashes are the default escaping mechanism in `LIKE` queries which means that this function is unsuitable to be used in `LIKE` escaping. `escapeMysqlWildcards` can only be used in statements like `GRANT` where the escaping mechanism seems to not care about a backslash. 

When I observed this bug in phpMyAdmin at first I was confused and [asked on Stack Overflow](https://stackoverflow.com/questions/71024061/show-column-escaping-mechanism). The proper escaping for LIKE should be:
1. Escape the following three characters `_%\` with a backslash `\`. 
2. Perform the usual string escaping if the value is interpolated into SQL or use parameter binding with prepared statements.

This means that in certain places, the escaping was not performed correctly and in fact, the bug I observed was fixed by introducing a proper escaping mechanism.  

This led me to investigate why `escapeMysqlWildcards` belongs to `Util` class and not `Privilege` class. I then noticed that there's another issue with incorrect escaping when creating a user with a database. In the end, I couldn't move the method to the `Privilege` class as there's one more incorrect usage in `Generator`.

### DIFF
--- a/libraries/classes/DatabaseInterface.php
+++ b/libraries/classes/DatabaseInterface.php
@@ -55,6 +55,7 @@ use function strncmp;
 use function strpos;
 use function strtolower;
 use function strtoupper;
+use function strtr;
 use function substr;
 use function syslog;
 use function trigger_error;
@@ -490,9 +491,7 @@ class DatabaseInterface implements DbalInterface
                                 ) . '\')';
                         } else {
                             $sql .= " `Name` LIKE '"
-                                . Util::escapeMysqlWildcards(
-                                    $this->escapeString($table, $link)
-                                )
+                                . $this->escapeMysqlLikeString($table, $link)
                                 . "%'";
                         }
                         $needAnd = true;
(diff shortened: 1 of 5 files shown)


williamdes:
Is 5.1 affected?

kamil-tekiela:
> Is 5.1 affected?

Yes, I just built it and I got this:
image

williamdes:
Oh okay, would you mind providing a fix for target: QA_5_1 ?

kamil-tekiela:
I cherry-picked onto 5.1 but merging up could be a little tricky.

libraries/classes/DatabaseInterface.php
```diff
@@ -2322,6 +2321,19 @@ public function escapeString(string $str, $link = self::CONNECT_USER)
         return $this->extension->escapeString($this->links[$link], $str);
     }
 
+    /**
+     * returns properly escaped string for use in MySQL LIKE clauses
+     *
+     * @param string $str  string to be escaped
+     * @param mixed  $link optional database link to use
```
williamdes, line 2328: I think this is int

williamdes (approved):

williamdes:
Thank you!
I will merge this up and you may want to rebase your other PR of it is needed

MauricioFauth (approved):

PR #17399, opened 2022-02-17
Handle connection errors with Exception instead of warnings

MauricioFauth:
Sets the [MySQLi error reporting mode](https://www.php.net/manual/en/mysqli-driver.report-mode.php) to exception for connections and disables it afterwards.

If an error occurs, catches it and then use `trigger_error` to simulate the previous behavior.

Since this is the new PHP default, the idea is to enable exceptions incrementally, also to ensure that all exceptions are being caught.

### DIFF
--- a/libraries/classes/Dbal/DbiMysqli.php
+++ b/libraries/classes/Dbal/DbiMysqli.php
@@ -8,6 +8,7 @@ declare(strict_types=1);
 namespace PhpMyAdmin\Dbal;
 
 use mysqli;
+use mysqli_sql_exception;
 use mysqli_stmt;
 use PhpMyAdmin\DatabaseInterface;
 use PhpMyAdmin\Query\Utilities;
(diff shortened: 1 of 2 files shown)


mauriciofauth:
@kamil-tekiela Could you please review this pull request?

williamdes (approved):

kamil-tekiela (approved):
I had something similar in the pipeline before but I was unhappy with my solution and I scrapped it. 
What you are proposing works. The error message that is displayed is even more user-friendly as it does not contain the name of the function anymore. 
The issue with switching to exception mode is that we barely gain anything. The phpMyAdmin project does not use exceptions often and errors are handled in a bespoke way. Enabling exceptions only means that we have to catch them and then emulate the old behaviour. 
But I think this is a good patch. It's a nice small incremental change. 

You may move `mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT);` before `mysqli_init()` as none of mysqli_init(), ssl_set(), or options can throw exceptions.

PR #17427, opened 2022-03-05
Replace `global` keyword with `$GLOBALS`

MauricioFauth:
Disallows the usage of the `global` keyword in favor of the `$GLOBALS` super global.

Using `$GLOBALS['config']->get()` is worse than using `$config->get()` for example, but at least this will make it clear when the variable is a global variable. So it's worth it as it will be easier to remove this global variable in the future. 

Note: I wrote a simple script to replace the variables, ~~so I still need to carefully check if everything went well~~.


test/classes/Controllers/Sql/EnumValuesControllerTest.php
```diff
@@ -26,7 +26,7 @@ protected function setUp(): void
 
     public function testGetEnumValuesError(): void
     {
-        global $containerBuilder, $_POST;
+         , $_POST;
```
williamdes, line 29:
```suggestion
```

libraries/classes/Common.php
```diff
@@ -81,22 +81,18 @@ final class Common
...
 
         if ($route === '/import-status') {
-            $isMinimumCommon = true;
+            $GLOBALS['isMinimumCommon'] = true;
         }
 
-        $containerBuilder = Core::getContainerBuilder();
+        $GLOBALS['containerBuilder'] = Core::getContainerBuilder();
 
         /** @var ErrorHandler $errorHandler */
```
williamdes, line 94:
```suggestion
```

PR #17444, opened 2022-03-17
Refactor import sql

kamil-tekiela:
This is a simplification of the existing Import functionality. It removes certain globals and reduces code complexity in other places.

The biggest change is the removal of SQL with comments aka `$full` query. All parameters and array storage for this was removed which allowed for significant simplification.


th4gbr0:
Thanks for this

williamdes (approved):
Looks good, thank you

MauricioFauth (approved):

PR #17456, opened 2022-03-23
Extract `console.js` objects into `console` module

MauricioFauth:
All objects in `js/src/console.js` were extracted and moved into the `js/src/modules/console.js` file and then the `Console` object was imported back into the `js/src/console.js` file.


js/src/modules/console.js
```diff
@@ -0,0 +1,1535 @@
+import $ from 'jquery';
+import CodeMirror from 'codemirror';
```
williamdes, line 2: If you can make th

js/src/modules/console.js
```diff
@@ -0,0 +1,1535 @@
+import $ from 'jquery';
```
williamdes, line 1: Can you make this a two commit operation so git shows a rename?
This will keep the file history and show a better diff :)

rkgchpc:
I don't know what to do. But I help [Accountants](https://rahulconsulting.com/home) and Bookkeepers [start and grow hyper profitable accounting firms](https://rahulconsulting.com/home) even if they don't have a license, degree, or job experience.

PR #17492, opened 2022-04-23
Disable dragging breadcrumbs links

### LINKED ISSUE #17095: Copy table name

MauricioFauth:
This makes easy to select and copy the text.

- Fixes #17095 
- Closes #17104
- Closes #17142

### DIFF
--- a/templates/menu/breadcrumbs.twig
+++ b/templates/menu/breadcrumbs.twig
@@ -3,7 +3,7 @@
   <ol class="breadcrumb">
     <li class="breadcrumb-item">
       {{ show_icons('TabsMode') ? get_image('s_host') }}
-      <a href="{{ url(server.url) }}" data-raw-text="{{ server.name }}">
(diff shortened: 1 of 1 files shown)


MauricioFauth:
@williamdes Are you okay with merging this and closing the other two PRs?

williamdes (approved):
This looks nice, I agree
Did not try it but from what I understand the text is only copyable and not anymore draggable when trying to select it for copy

williamdes:
The copy button is maybe still a good idea, I would have to deploy and try it
Not sure, what do you think about it?

PR #17509, opened 2022-05-08
Add workflow to generate daily snapshots

MauricioFauth:
Creates a GitHub Actions workflow that generates phpMyAdmin daily snapshots.

Currently it persists the snapshots only on GitHub and do not upload to the phpMyAdmin's website.

### DIFF
--- a/.github/workflows/daily-snapshots.yml
+++ b/.github/workflows/daily-snapshots.yml
@@ -1,0 +1,74 @@
+name: Daily snapshots
+
+on:
+  pull_request:
+    paths:
+      - '.github/workflows/daily-snapshots.yml'
+      - 'scripts/**'
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  generate-snapshots:
+    name: Generate ${{ matrix.version }} snapshots
+    runs-on: ubuntu-latest
+    if: (github.event_name == 'schedule' && github.repository == 'phpmyadmin/phpmyadmin') || (github.event_name != 'schedule')
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - { version: '5.1', branch: 'QA_5_1', php-version: '7.1', node-version: '12', python-version: '3.6' }
+          - { version: '5.2', branch: 'QA_5_2', php-version: '7.2', node-version: '12', python-version: '3.6' }
+          - { version: '5.3', branch: 'master', php-version: '7.2', node-version: '12', python-version: '3.6' }
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ matrix.branch }}
+
+      - name: Install Gettext
+        run: sudo apt-get install -y gettext
+
+      - name: Set up PHP ${{ matrix.php-version }}
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php-version }}
+          extensions: mbstring, iconv, mysqli, zip, gd
+          coverage: none
+
+      - name: Set up Node ${{ matrix.node-version }}
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: yarn
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v3
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install Composer dependencies
+        uses: ramsey/composer-install@v2
+        with:
+          dependency-versions: highest
+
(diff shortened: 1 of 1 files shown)


.github/workflows/daily-snapshots.yml
```diff
@@ -0,0 +1,60 @@
...
+      - '.github/workflows/daily-snapshots.yml'
+      - 'scripts/create-release.sh'
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  generate-snapshot:
+    name: Generate snapshot
+    runs-on: ubuntu-latest
```
williamdes, line 15: can a global if be added like in some other workflows to not run on forks of this repo ?

williamdes (approved):
💯

.github/workflows/daily-snapshots.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      - name: Build the release
+        run: ./scripts/create-release.sh --daily 5.1+snapshot QA_5_1
+
+      - name: Upload snapshots
+        uses: actions/upload-artifact@v3
+        with:
+          name: daily-snapshots
+          path: release/*
+          retention-days: 1
```
williamdes, line 60: can you run my check release excludes on the tar gz release bundle after the upload step ?

williamdes:
I am wondering if this will work for QA because crons are run on master branch
Maybe you need to target master and in your script change to QA_5_1

MauricioFauth:
> I am wondering if this will work for QA because crons are run on master branch Maybe you need to target master and in your script change to QA_5_1

Do you think we should only have this file on master branch and create all QA snapshots from there?

williamdes:
> > I am wondering if this will work for QA because crons are run on master branch Maybe you need to target master and in your script change to QA_5_1
> 
> Do you think we should only have this file on master branch and create all QA snapshots from there?

Yes because as I think I am sure of that fact, only crons on the default branch of the repo are run

.github/workflows/daily-snapshots.yml
```diff
@@ -0,0 +1,182 @@
...
+
+  snapshots-5-2:
+    name: Generate 5.2 snapshots
+    runs-on: ubuntu-latest
+    if: (github.event_name == 'schedule' && github.repository == 'phpmyadmin/phpmyadmin') || (github.event_name != 'schedule')
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
```
williamdes, line 76: I think you should switch branches here too

.github/workflows/daily-snapshots.yml
```diff
@@ -0,0 +1,182 @@
...
+    - cron: '0 0 * * *'
+
+jobs:
+  snapshots-5-1:
+    name: Generate 5.1 snapshots
+    runs-on: ubuntu-latest
+    if: (github.event_name == 'schedule' && github.repository == 'phpmyadmin/phpmyadmin') || (github.event_name != 'schedule')
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
```
williamdes, line 19:
```suggestion
        uses: actions/checkout@v3
        with:
            ref: QA_5_1
```
https://github.com/actions/checkout#checkout-a-different-branch

.github/workflows/daily-snapshots.yml
```diff
@@ -0,0 +1,182 @@
...
+  pull_request:
+    paths:
+      - '.github/workflows/daily-snapshots.yml'
+      - 'scripts/create-release.sh'
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  snapshots-5-1:
```
williamdes, line 13: could a matrix make this file one step ?

williamdes (approved):
Awesome!

.github/workflows/daily-snapshots.yml
```diff
@@ -0,0 +1,74 @@
...
+
+      - name: Install Sphinx for the documentation build
+        run: pip install 'Sphinx'
+
+      - name: Build the release
+        run: ./scripts/create-release.sh --daily ${{ matrix.version }}+snapshot ${{ matrix.branch }}
+
+      - name: Check release excludes
+        run: ./scripts/check-release-excludes.sh release/phpMyAdmin-${{ matrix.version }}+snapshot-all-languages.tar.gz
+        continue-on-error: true
```
MauricioFauth, line 67: This script is failing, that's why I added this. It can be removed later.

williamdes: It can be kept as it is, just warning us of failures is enough
I will do some change so it outputs errors in the GitHub format

PR #17516, opened 2022-05-11
Remove `jquery-debounce-throttle` JavaScript dependency

MauricioFauth:
This package is not supported anymore.

Removes the call to `.throttle()` when resizing the navigation sidebar.

- https://github.com/dfilatov/jquery-plugins


williamdes (approved):
Yayy!

Issue #17519, opened 2022-05-12
Can't see tab export display


kanran111:
### Describe the bug

As title, when i go into database and click export and i see nothing, then i go to web demo of phpmyadmin ( demo.phpmyadmin.net )  and it not show too . 

how can i do? 

i using both lastes chrome ,firefox, edge. 
![image](https://user-images.githubusercontent.com/19228771/168024250-d8680a52-9cd5-487e-b69a-6a53ea393e32.png)

williamdes:
The export tab seems to do nothing 🤔 
@MauricioFauth could you have a look please ?
https://demo.phpmyadmin.net/QA_5_2/index.php?route=/server/export

liviuconcioiu:
Hasn't been functioning for almost half a year, neither on master. I thought it was a intentional server config.

MauricioFauth:
This is probably something related to the routes cache file not being updated on the demo server.

toonvd:
@MauricioFauth this issue is actually present on all our servers. @kanran111 also mentioned he noticed it on his own env first.

ShiftTimoK:
We have the same issue on our live systems.
No errors show up in any logs and we also tried clearing all the caches.

williamdes:
Hi 
Could you try the latest 5.2 version in development [(phpMyAdmin 5.2+snapshot)](https://www.phpmyadmin.net/downloads/#snapshot_5.2+snapshot) as it will have a fix for the routing cache issue and let us know ?

kanran111:
@williamdes  I try download 5.2 dev version and got a new bug like this
![phpmyadmin issue](https://user-images.githubusercontent.com/19228771/169348521-4518a4c7-7d3a-4aab-849b-6a8fc1b09b0e.jpg)
Last time i installed by clone git hub and run "composer update --no-dev && yarn install --production" , The export tab not shown ( follow this document https://docs.phpmyadmin.net/en/latest/setup.html )  . 
But when i download again 5.2 Stable version, the tab export work perfectly. https://www.phpmyadmin.net/downloads/#stable

liviuconcioiu:
> @williamdes I try download 5.2 dev version and got a new bug like img below and some ajax request got error 404. 

ref #17532

williamdes:
> > @williamdes I try download 5.2 dev version and got a new bug like img below and some ajax request got error 404.
> 
> ref #17532

Thank you. I just re-built the snapshots.
Let me know @kanran111

[...]

PR #17536, opened 2022-05-19
Improve `ErrorHandler::handleException` to show errors

MauricioFauth:
It doesn't shows a blank page when not in `development` environment anymore. Now it shows a simple page with the error, and when it's an AJAX request, it shows the error instead of doing nothing.

Related to #17519.

### DIFF
--- a/js/src/ajax.js
+++ b/js/src/ajax.js
@@ -466,6 +466,14 @@ var AJAX = {
         if (typeof data === 'undefined' || data === null) {
             return;
         }
+        // Can be a string when an error occurred and only HTML was returned.
+        if (typeof data === 'string') {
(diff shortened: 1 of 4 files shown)


williamdes (approved):
Looks nice, can a test be written for that ?

js/src/ajax.js
```diff
@@ -466,6 +466,13 @@ var AJAX = {
         if (typeof data === 'undefined' || data === null) {
             return;
         }
+        if (typeof data === 'string') {
```
williamdes, line 469: I think a comment would really help understand why we have some very specific "hack" to display some kind of error
Can this create a code injection vulnerability ?

MauricioFauth: I striped the HTML tags, just in case.

Issue #17124, opened 2021-09-25
Preview SQL does not show SQL query in Add Index Dialog


shucontech:
### Describe the bug

Preview SQL does not show SQL query in Add Index Dialog.

### To Reproduce

1. Click on the button to create a new table.
2. Fill up the table name and column name
3. Select Index dropdown to any option
4. Fill up the dialog
5. Click on Preview SQL

### Expected behavior
The corresponding SQL query should be shown.

### Screenshots

![Screenshot from 2021-09-25 20-52-33](https://user-images.githubusercontent.com/85239205/134776670-f10f1822-553e-49e5-aefc-35ee61cde552.png)

### Server configuration

- phpMyAdmin version: master and QA_5_1

### Client configuration

- Browser: Firefox
- Operating system: Ubuntu

williamdes:
I just tried on https://demo.phpmyadmin.net/QA_5_2 and it seems fixed by #17542

Could you try the latest 5.2 version in development [(phpMyAdmin 5.2+snapshot)](https://www.phpmyadmin.net/downloads/#snapshot_5.2+snapshot) ?

PR #17542, opened 2022-05-22
Fix the indexes modal not opening the SQL preview

MauricioFauth:
It's not possible to open two modals at the same time, so it needs to toggle to another modal instead.

### DIFF
--- a/js/src/functions.js
+++ b/js/src/functions.js
@@ -3376,6 +3376,39 @@ AJAX.registerOnload('functions.js', function () {
 Functions.indexDialogModal = function (routeUrl, url, title, callbackSuccess, callbackFailure) {
     /* Remove the hidden dialogs if there are*/
     var modal = $('#indexDialogModal');
+
+    const indexDialogPreviewModal = document.getElementById('indexDialogPreviewModal');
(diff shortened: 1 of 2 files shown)


js/src/functions.js
```diff
@@ -3376,6 +3376,39 @@ AJAX.registerOnload('functions.js', function () {
...
     /* Remove the hidden dialogs if there are*/
     var modal = $('#indexDialogModal');
+
+    const indexDialogPreviewModal = document.getElementById('indexDialogPreviewModal');
+    indexDialogPreviewModal.addEventListener('shown.bs.modal', () => {
+        const modalBody = indexDialogPreviewModal.querySelector('.modal-body');
+        const $form = $('#index_frm');
+        const formUrl = $form.attr('action');
+        const sep = CommonParams.get('arg_separator');
+        const formData = $form.serialize() +
```
williamdes, line 3386: Is the server Id pram added?

MauricioFauth: Yes. It's included in the form. I copied from the Functions.previewSql function.

williamdes (approved):

williamdes:
Does this fix #17124 ?

PR #17555, opened 2022-05-24
Removing search columns in foreigners from twig

pokemaobr:


libraries/classes/Controllers/Table/ZoomSearchController.php
```diff
@@ -389,6 +393,7 @@ public function zoomSubmitAction($dataLabel, $goto): void
                 '',
                 $GLOBALS['cfg']['ForeignKeyMaxLimit']
             );
+
```
MauricioFauth, line 396:
```suggestion
```

MauricioFauth (changes requested):

PR #17563, opened 2022-05-27
Extract Database and Table controllers from Server\PrivilegesController class

MauricioFauth:
Creates new routes and extracts the `Database\PrivilegesController` and the `Table\PrivilegesController` classes from the `Server\PrivilegesController` class.

### DIFF
--- a/templates/database/privileges/index.twig
+++ b/templates/database/privileges/index.twig
@@ -132,7 +132,7 @@
         <a id="add_user_anchor" href="{{ url('/server/privileges', {
           'adduser': true,
           'dbname': db
-        }) }}" rel="{{ get_common({'checkprivsdb': db}) }}">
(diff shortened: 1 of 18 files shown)


templates/database/privileges/index.twig
```diff
@@ -132,7 +132,7 @@
         <a id="add_user_anchor" href="{{ url('/server/privileges', {
           'adduser': true,
           'dbname': db
-        }) }}" rel="{{ get_common({'checkprivsdb': db}) }}">
+        }) }}">
```
williamdes, line 135: This was probably to prevent opening multiple tabs for the same thing, anyway I think it is useless nowdays

williamdes (approved):
The code looks okay 👍🏻

PR #17579, opened 2022-06-02
Refactor getHtmlForDisplayedExportHeader

kamil-tekiela:
I just wanted to remove the GLOBAL that shouldn't be there, but I ended up splitting the method into 3 separate ones.

### DIFF
--- a/libraries/classes/Export.php
+++ b/libraries/classes/Export.php
@@ -219,14 +219,16 @@ class Export
     /**
      * Returns HTML containing the footer for a displayed export
      *
-     * @param string $backButton    the link for going Back
-     * @param string $refreshButton the link for refreshing page
+     * @param string $exportType the export type
+     * @param string $db         the database name
+     * @param string $table      the table name
      *
      * @return string the HTML output
      */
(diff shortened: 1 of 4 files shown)


kamil-tekiela:
Unrelated, but the resulting textarea seems to overflow my screen.

libraries/classes/Export.php
```diff
@@ -1339,4 +1278,72 @@ public function processExportSchema(?string $exportType): void
...
+            );
+        } else {
+            $backButton .= Url::getFromRoute('/table/export') . '" data-post="' . Url::getCommon(
+                ['db' => $db, 'table' => $table],
+                '',
+                false
+            );
+        }
+
+        $postParams = array_filter($this->getPostParams($exportType), static fn ($value) => ! is_array($value));
```
MauricioFauth, line 1323: Please use an anonymous function here. It should support PHP 7.2.

MauricioFauth (changes requested):

MauricioFauth (approved):

PR #17582, opened 2022-06-03
Fix errors in InsertEdit

### LINKED ISSUE #17121: password_hash function incorrectly adds single quotes to password before hashing

kamil-tekiela:
Fixes #17121

All I wanted to do was to fix the bug with `password_hash`, but I ended up rewrite a lot of the code as it was really hard to understand. I still have no idea why there are two separate methods and one claims to return an array. 

I removed many globals that were used exclusively within that single method. The rest can probably be unglobalized too, but I couldn't verify that.

### DIFF
--- a/libraries/classes/Controllers/Table/ReplaceController.php
+++ b/libraries/classes/Controllers/Table/ReplaceController.php
@@ -325,7 +325,7 @@ final class ReplaceController extends AbstractController
         );
 
         // Builds the sql query
-        if ($isInsert && count($valueSets) > 0) {
+        if ($isInsert && $valueSets !== []) {
             $GLOBALS['query'] = $this->insertEdit->buildSqlQuery($isInsertignore, $queryFields, $valueSets);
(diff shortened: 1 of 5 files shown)


williamdes (approved):
Thank you @kamil-tekiela !
The code looks okay, but can a simpler fix be done for the `QA_5_2` branch in another PR ?

MauricioFauth (approved):

PR #17594, opened 2022-06-11
Add basic TypeScript support

MauricioFauth:
Adds a basic `tsconfig.json` file to be able to use TypeScript for static analysis of the JavaScript files. TypeScript is not used for compilation yet.

This enables the type checking in some IDEs.

To check all files:
```
yarn tsc
```


williamdes (approved):
Wow, that's thin

Issue #17262, opened 2021-12-28
Changing password should respect AllowNoPassword


ibennetch:
When a user is changing their password, we should check if [`AllowNoPassword`](https://docs.phpmyadmin.net/en/latest/config.html#cfg_Servers_AllowNoPassword) is false, and if it is false then we should warn the user strongly before allowing them to set their password to a blank value. This check should apply to both the "Change password" link on the main page and actions on the "User accounts" tab.

The warning can be "You are trying to set this account to log in with no password, but the configuration directive `AllowNoPassword` is false. If you proceed, the account will not be able to log in through phpMyAdmin."

The default action should be cancel, and I think the other button should be labelled "proceed"

Ref: #17261

Ritterstahl:
Where does this happen?

ibennetch:
On the Change password link from the main page and also on the User accounts tab where a super user can change anyone's password.

conectis360:
Hi! I want to help, but **I only know vanilla PHP** is it enough? If it is, after cloning it, what's next?

Siddharth-Ashri:
I am happy to take care of this solution. I have forked / cloned the repository ,but. not sure where to go from there! 
any help would be appreciated. is there a set up guide for developer that i can refer to?

TIA

williamdes:
I think all this questions are good amd we should maybe have some easier documentation to answer them

PR #17613, opened 2022-06-19
#17262 - Changing password should respect AllowNoPassword

RamiRima:
#17262

### DIFF
--- a/js/src/functions.js
+++ b/js/src/functions.js
@@ -2451,7 +2451,10 @@ Functions.checkPassword = function ($theForm) {
     var alertMessage = false;
 
     if ($password.val() === '') {
-        alertMessage = Messages.strPasswordEmpty;
+         if(window.confirm("You are trying to set this account to log in without a password but the confihuration directive AllowNoPassword is false. If you procced, the account will not be able to log in through phpMyAdmin. Do you wish to procced ? ")){
+        }else{
+            alert_msg = PMA_messages.strPasswordEmpty;        
+        }
     } else if ($password.val() !== $passwordRepeat.val()) {
         alertMessage = Messages.strPasswordNotSame;
     }
--- a/js/src/functions.js
+++ b/js/src/functions.js
@@ -2451,7 +2451,9 @@ Functions.checkPassword = function ($theForm) {
     var alertMessage = false;
 
     if ($password.val() === '') {
-         if(window.confirm("You are trying to set this account to log in without a password but the confihuration directive AllowNoPassword is false. If you procced, the account will not be able to log in through phpMyAdmin. Do you wish to procced ? ")){
+        alertMessage = Messages.strAllowNoPassowrdIsFalse;
+         if(window.confirm(alertMessage)){
+             alertMessage=false;
         }else{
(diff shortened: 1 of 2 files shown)


williamdes (changes requested):
Please use a translation string in the controller file: https://github.com/phpmyadmin/phpmyadmin/blob/master/libraries/classes/Controllers/JavaScriptMessagesController.php#L40

By the way there is a typo in the sentence

libraries/classes/Controllers/JavaScriptMessagesController.php
```diff
@@ -41,6 +41,9 @@ private function setMessages(): array
     {
         return [
             /* For confirmations */
+            'strAllowNoPassowrdIsFalse' => __(
+                'You are trying to set this account to log in without a password but the configuration directive AllowNoPassword is false. If you procced, the account will not be able to log in through phpMyAdmin. Do you wish to procced ? '
```
liviuconcioiu, line 45: `You are trying to set this account to log in without a password but the configuration directive AllowNoPassword is false. If you proceed, the account will not be able to log in through phpMyAdmin. Do you wish to proceed?`

liviuconcioiu (changes requested):

libraries/classes/Controllers/JavaScriptMessagesController.php
```diff
@@ -42,7 +42,7 @@ private function setMessages(): array
         return [
             /* For confirmations */
             'strAllowNoPassowrdIsFalse' => __(
-                'You are trying to set this account to log in without a password but the configuration directive AllowNoPassword is false. If you procced, the account will not be able to log in through phpMyAdmin. Do you wish to procced ? '
+            'You are trying to set this account to log in without a password but the configuration directive AllowNoPassword is false. If you proceed, the account will not be able to log in through phpMyAdmin. Do you wish to proceed?'
```
liviuconcioiu, line 45: Can you remove the white space at the end?

liviuconcioiu (changes requested):

js/src/functions.js
```diff
@@ -2451,7 +2451,12 @@ Functions.checkPassword = function ($theForm) {
     var alertMessage = false;
 
     if ($password.val() === '') {
-        alertMessage = Messages.strPasswordEmpty;
+        alertMessage = Messages.strAllowNoPassowrdIsFalse;
+         if(window.confirm(alertMessage)){
+             alertMessage=false;
+        }else{
+            alertMessage = PMA_messages.strPasswordEmpty;        
+        }
```
williamdes, line 2459: I think this is not the proper way to do this, @MauricioFauth do you have some advice ?

You should run `yarn js-lint --fix` on this block of code

MauricioFauth:
Hi @RamiRima, thanks for your contribution.

As described in the issue, we should check if [`$GLOBALS['cfg']['Server']['AllowNoPassword'] === false`](https://docs.phpmyadmin.net/en/latest/config.html#cfg_Servers_AllowNoPassword) before warning the user.

Also, it's a best practice to use a different branch to open a pull request, this allows you to open more than one pull request.

PR #17625, opened 2022-07-02
Use JavaScript module to import jQuery

MauricioFauth:


williamdes (approved):
That would be cool that for such changes the genrated output diff is sent on the PR

MauricioFauth:
> That would be cool that for such changes the genrated output diff is sent on the PR

Do you mean, to avoid duplication? In this case, since the jQuery is set in [externals](https://webpack.js.org/configuration/externals/), it hasn't been really imported.

https://github.com/phpmyadmin/phpmyadmin/blob/0056236a4c52caa4f9a9013aa4786381c3b8077c/webpack.config.js#L101-L102

williamdes:
I meant that what I feel frustrating is that we can not see the actual generated JS code for such PRs

MauricioFauth:
> I meant that what I feel frustrating is that we can not see the actual generated JS code for such PRs

I know what you mean. I'm always checking the generated code. How do you think it should be compared?

williamdes:
> > I meant that what I feel frustrating is that we can not see the actual generated JS code for such PRs
> 
> I know what you mean. I'm always checking the generated code. How do you think it should be compared?

maybe a line by line diff just to be sure it is pretty much the same as before

MauricioFauth:
> maybe a line by line diff just to be sure it is pretty much the same as before

Comparing it to the previous generated code, or comparing it to the source?

williamdes:
> > maybe a line by line diff just to be sure it is pretty much the same as before
> 
> Comparing it to the previous generated code, or comparing it to the source?

I was thinking about comparing with previously generated code

PR #17657, opened 2022-07-22
Supporting the column compression option on MariaDB

### LINKED ISSUE #14956: Implement Page/Column Compression options
Make use of the new PAGE_COMPRESSED=1 and Column's COMPRESSED=1 features.

Page compression: https://mariadb.com/kb/en/library/compression/

Column compression: https://jira.mariadb.org/browse/MDEV-11371

These don't seem to be available anywhere on phpMyAdmin.

(edited post as I was also reporting the issue on #14673, which appears to have a different cause)

iifawzi:
Hi, this PR should fix #14956
Fixes #14956

image

### DIFF
--- a/libraries/classes/Types.php
+++ b/libraries/classes/Types.php
@@ -7,6 +7,8 @@ declare(strict_types=1);
 
 namespace PhpMyAdmin;
 
+use PhpMyAdmin\Query\Compatibility;
+
 use function __;
 use function _pgettext;
 use function array_diff;
@@ -724,7 +726,7 @@ class Types
             'on update CURRENT_TIMESTAMP',
         ];
 
-        if ($isMariaDB && $serverVersion >= 100100) {
+        if (Compatibility::supportsCompressedColumns($serverVersion)) {
             $attributes[] = 'COMPRESSED=zlib';
         }
 
--- a/libraries/classes/Types.php
+++ b/libraries/classes/Types.php
@@ -713,13 +713,22 @@ class Types
      */
     public function getAttributes()
     {
-        return [
+        $isMariaDB = $this->dbi->isMariaDB();
+        $serverVersion = $this->dbi->getVersion();
+
+        $attributes = [
             '',
             'BINARY',
             'UNSIGNED',
             'UNSIGNED ZEROFILL',
             'on update CURRENT_TIMESTAMP',
         ];
+
+        if ($isMariaDB && $serverVersion >= 100100) {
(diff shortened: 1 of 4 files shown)


williamdes:
@MauricioFauth could this go to 5.2 ?

libraries/classes/Types.php
```diff
@@ -713,13 +713,22 @@ public function getAllFunctions()
...
+
+        $attributes = [
             '',
             'BINARY',
             'UNSIGNED',
             'UNSIGNED ZEROFILL',
             'on update CURRENT_TIMESTAMP',
         ];
+
+        if ($isMariaDB && $serverVersion >= 100100) {
```
iifawzi, line 727: I've assumed that it's supported starting from `10.1.0`, based on this 
https://mariadb.com/kb/en/innodb-page-compression/#comment_1992

libraries/classes/Util.php
```diff
@@ -1357,6 +1360,10 @@ public static function extractColumnSpec($columnSpecification)
             $attribute = 'UNSIGNED ZEROFILL';
         }
 
+        if ($compressed) {
+            $attribute = 'COMPRESSED=zlib';
```
williamdes, line 1364: it's not always zlib ?
Can we use what was parsed ?

iifawzi: In fact, from this particular scope, no. we're indicating that the compression is used from the `compressed` word shown in the type, which's the way we also use to detect other attrs, `BINARY`, `UNSIGNED`, or`UNSIGNED ZEROFILL`. getting it dynamically from the parser would need more work

williamdes: Okay, thanks for clarifying

williamdes (approved):

linaspasv:
What is the state of this? The support for MariaDB is very needed.

MauricioFauth (approved):

williamdes:
For the record:
> With InnoDB page compression, multiple compression algorithms are supported. In contrast, with InnoDB's [COMPRESSED](https://mariadb.com/kb/en/innodb-compressed-row-format/) row format, [zlib](https://www.zlib.net/) is the only supported compression algorithm. This means that the [COMPRESSED](https://mariadb.com/kb/en/innodb-compressed-row-format/) row format has less compression options than InnoDB page compression does.

Ref: https://mariadb.com/kb/en/innodb-page-compression/#comparison-with-the-compressed-row-format

PR #17662, opened 2022-07-25
add gitHub action close stale issues

### LINKED ISSUE #17488: Probot Stale bot should be replaced with the Close Stale Issues action
Looks like [Probot Stale](https://github.com/probot/stale) bot is not working properly lately, so it should be replaced with the [Close Stale Issues](https://github.com/marketplace/actions/close-stale-issues) GitHub Action.

The Probot Stale configuration file is located at `.github/stale.yml`.

NF11:
Hi, this PR could help for https://github.com/phpmyadmin/phpmyadmin/issues/17488
Signed-off-by: Nassim FARES [nassim.fares.dz@gmail.com](mailto:nassim.fares.dz@gmail.com)
Fixes https://github.com/phpmyadmin/phpmyadmin/issues/17488


.github/workflows/stale.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+      - uses: actions/stale@v5
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          days-before-stale: 60
+          days-before-close: 7
+          stale-issue-label: "wontfix"
+          stale-pr-label: "wontfix"
+          stale-issue-message: >
+            This issue has been automatically marked as stale because it has not had
```
MauricioFauth, line 21: I don't think it's a good idea to close stale issues right now, only pull requests.

MauricioFauth (changes requested):

.github/workflows/stale.yml
williamdes: I would even argue that I do not think we wave stale "issues"
Only stale issues with label: question

NF11: I tried to follow what was on Probot Stale

.github/workflows/stale.yml
```diff
@@ -0,0 +1,25 @@
+name: "Mark or close stale PRs"
+on:
+  schedule:
+    - cron: "30 1 * * *"
+
```
williamdes, line 5: You should add the permissions section, see other workflows for this and documentation :)

.github/workflows/stale.yml
```diff
@@ -0,0 +1,25 @@
...
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          days-before-stale: 60
+          days-before-close: 7
+          stale-pr-label: "wontfix"
+          stale-pr-message: >
+            This pr has been automatically marked as stale because it has not had
+            recent activity. It will be closed if no further activity occurs. Thank you
+            for your contributions.
+          exempt-issue-labels: "type: pinned,type: security"
```
williamdes, line 23: We do not have a security label and it is not used actually
Is there other ways to only mark stale question labels without any response?

divinity76:
instead of marking as "wontfix", can we instead make a dedicated "stale" label?

williamdes (changes requested):
Unfortunately I do not see any use of this change. Pull-requests do not require stale management.
Could you rather help me with label:"question" as they often go stale and cause me much work to keep up to date and do manually the closing.

The message I use:
> _Could not reproduce, closing since no feedback or no way of reproducing was found.
Comment if you find a way to reproduce this issue, or want to re-open for a valid reason._

.github/workflows/stale.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - uses: actions/stale@v5
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          days-before-stale: 60
+          days-before-close: 7
+          only-pr-labels : "question"
```
williamdes, line 17:
```suggestion
          only-issue-labels: "question"
```

.github/workflows/stale.yml
```diff
@@ -0,0 +1,27 @@
...
+          stale-pr-message: >
+            This pr has been automatically marked as stale because it has not had
+            recent activity. It will be closed if no further activity occurs. Thank you
+            for your contributions.
```
williamdes, line 21: This can be removed

.github/workflows/stale.yml
```diff
@@ -0,0 +1,27 @@
...
+          close-issue-message: >
+            Could not reproduce, closing since no feedback or no way of reproducing was found.
+            Comment if you find a way to reproduce this issue, or want to re-open for a valid reason.
+          exempt-issue-labels: "type: pinned,type: security"
+          exempt-pr-labels: "type: pinned,type: security"
```
williamdes, line 26:
```suggestion
```
This is not needed, but if we can post a ping message before closing it could be great

williamdes (approved):

williamdes:
Merged as 0d27683e941aca2556edc225cc51f6026d3644ae

PR #17673, opened 2022-08-04
Allow empty values to be inserted into columns

### LINKED ISSUE #17658: SQL query appearing when a value is made NULL
### Describe the bug

A SQL query popup appears when a value is made NULL.

### To Reproduce

1. Create a table with a unique column with NULL as default
2. Have a few different values
3. Click on any value and make it NULL

### Expected behavior

The popup should not appear when a value is made NULL.

### Screenshots

https://user-images.githubusercontent.com/25424343/180584937-e1ceffdd-73da-413b-8214-25d99924b699.mp4

### Server configuration

- Operating system:
- Web server:
- Database version:
- PHP version:
- phpMyAdmin version: QA_5_2

### Client configuration

- Browser: Firefox
- Operating system: Windows

### Additional context

### LINKED ISSUE #17663: Updating a value in an INT column to 0 is not possible
If I edit a record and try to modify an INT column (which is not part of any index) to zero, the resulting UPDATE query does not contain this column, therefore, the update of the value to 0 is not performed, unless you manually edit the UPDATE query. Testing in 5.3.0-dev.

kamil-tekiela:
The first commit is unrelated; it's just something that was annoying me and I had to fix. 

The main commit fixes bug that was uncovered by my previous PR. I don't know why the code didn't allow empty values previously, but I assume it was just bad design that I didn't notice while refactoring. Anyway, this code still looks like a mess, so I might create another PR to refactor it more. 

Fixes: #17658 #17663

### DIFF
--- a/libraries/classes/InsertEdit.php
+++ b/libraries/classes/InsertEdit.php
@@ -1674,9 +1674,8 @@ class InsertEdit
         } elseif (
             ! (empty($multiEditFuncs[$key])
                 && isset($multiEditColumnsPrev[$key])
-                && (($currentValue === "'" . $this->dbi->escapeString($multiEditColumnsPrev[$key]) . "'")
-                    || ($currentValue === '0x' . $multiEditColumnsPrev[$key])))
-            && $currentValue
+                && $currentValue === $multiEditColumnsPrev[$key])
+            && $currentValueAsAnArray !== ''
         ) {
             // avoid setting a field to NULL when it's already NULL
             // (field had the null checkbox before the update
--- a/libraries/classes/Sql.php
+++ b/libraries/classes/Sql.php
@@ -1448,7 +1448,7 @@ class Sql
             }
         }
 
-        $hasUnique = $table && $this->resultSetContainsUniqueKey($db, $table, $fieldsMeta);
+        $hasUnique = $table !== null && $this->resultSetContainsUniqueKey($db, $table, $fieldsMeta);
 
         $editable = ($hasUnique
             || $GLOBALS['cfg']['RowActionLinksWithoutUnique']
--- a/test/classes/InsertEditTest.php
+++ b/test/classes/InsertEditTest.php
@@ -2023,6 +2023,126 @@ class InsertEditTest extends AbstractTestCase
             ],
             $result
         );
+
+        // Test to see if a zero-string is not ignored
+        $result = $this->insertEdit->getQueryValuesForInsertAndUpdateInMultipleEdit(
+            $multi_edit_columns_name,
+            [],
+            '0',
+            [],
+            [],
+            false,
+            [],
+            [],
+            "'0'",
+            [],
+            '0',
+            []
+        );
+
+        $this->assertEquals(
+            [
+                ["`fld` = '0'"],
+                [],
+            ],
+            $result
+        );
+
+        // Can only happen when table contains blob field that was left unchanged during edit
+        $result = $this->insertEdit->getQueryValuesForInsertAndUpdateInMultipleEdit(
+            $multi_edit_columns_name,
+            [],
+            '',
+            [],
+            [],
+            false,
+            [],
+            [],
+            '',
+            [],
+            '0',
+            []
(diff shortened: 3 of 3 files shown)


williamdes:
Should this target QA_5_2 ?

kamil-tekiela:
> Should this target QA_5_2 ?

You are right. Thanks!

williamdes (approved):
🚀 Thank you for all fixes !

ibennetch:
Yes, thanks for these.

MauricioFauth (approved):

kamil-tekiela:
Please don't merge yet. I finally figured out why the check was there. I will adjust it soon.

kamil-tekiela:
@williamdes @MauricioFauth Please re-review. I applied a different fixed and added more tests.

libraries/classes/InsertEdit.php
```diff
@@ -1674,9 +1674,9 @@ public function getQueryValuesForInsertAndUpdateInMultipleEdit(
         } elseif (
             ! (empty($multiEditFuncs[$key])
                 && isset($multiEditColumnsPrev[$key])
-                && (($currentValue === "'" . $this->dbi->escapeString($multiEditColumnsPrev[$key]) . "'")
-                    || ($currentValue === '0x' . $multiEditColumnsPrev[$key])))
-            && $currentValue
+                && (($currentValueAsAnArray === "'" . $this->dbi->escapeString($multiEditColumnsPrev[$key]) . "'")
```
williamdes, line 1677: Why does it also change thoes two lines?

kamil-tekiela: Excellent question! I have no idea... I changed it again just now. The short answer is that I am embarrassed that after spending so much time on this already, I still cannot understand how this code was meant to work or at what point in time it stopped working properly. 

At least I understand what it was meant to do, now. The code is used during both AJAX editing, copying, inserting, and multi-editing rows. During multi-editing (it's when you select a row and click edit) the page navigates to a normal HTML form with all fields prepopulated with current values. The frontend doesn't check which values were actually adjusted, but instead submits both the current values and the values from the form. The backend needs to check which values were actually changed and dynamically build SQL `UPDATE`. Some rows could be left unchanged during multi-edit, so the code will skip that row entirely. 

The issue is that we have two values: the one submitted to us and the one we will put in SQL (the SQL formatted representation of this value - `$currentValueAsAnArray`). If we want to check if the value was changed, we need to compare same to same. Previously, the code was comparing the submitted value to SQL-formatted previous value. It didn't work. 

The middle clause `($currentValue === '0x' . $multiEditColumnsPrev[$key])` checked for binary values... but if we compare same to same, it isn't needed anymore. 

The last clause was meant for file uploads. If a column with blob data is unchanged, the value is an empty string. That would break an SQL, so the check needs to strictly compare the value to an empty string `$currentValueAsAnArray !== ''`, but when we submit `NULL` it is submitted as an empty string too, so we need to check the SQL formatted value. Also, PHP type coercion considers zero-string to be empty. 

As a last step, the code checks if the value was `NULL` before and if is is still marked as `NULL`. That inner `if` should be merged with the outer one, but that's for another PR. I am in the process of refactoring all of this, so hopefully it will become less confusing soon. 

My second PR #17679 is also related to this, but I think it's acceptable to be fixed only on `master` rather than 5_2. The TypeError mentioned there only happens on `master`. 

I tested it with this table:

```
CREATE TABLE `a0` (
  `0` int(11) NOT NULL AUTO_INCREMENT,
  `1` varchar(11) COLLATE utf8mb4_turkish_ci DEFAULT NULL,
  `ti` timestamp NOT NULL DEFAULT current_timestamp() ON UPDATE current_timestamp(),
  `set` set('a','b','v','') COLLATE utf8mb4_turkish_ci DEFAULT NULL,
  `blob` blob DEFAULT NULL,
  `bin` binary(4) NOT NULL,
  PRIMARY KEY (`0`)
)
```

MauricioFauth (approved):

PR #17676, opened 2022-08-06
Extract `url.php` entry point into a route

MauricioFauth:
- Related to #16922


libraries/classes/Common.php
```diff
@@ -209,6 +209,10 @@ public static function run(): void
             $GLOBALS['containerBuilder']->set('theme_manager', ThemeManager::getInstance());
             Tracker::enable();
 
+            if ($route === '/url') {
```
williamdes, line 212: It can be before tracking init
Ideally this should just work without dbi and things
Or not?

williamdes (approved):
Thank you for this 💯

libraries/classes/Common.php
MauricioFauth: Probably, yes. But I added it in the same order. We can check this later when refactoring the Common class.

PR #17679, opened 2022-08-07
Fix bug not allowing insertion of set values

kamil-tekiela:
Due to the recent changes, a bug was introduced that causes a TypeError whenever set values are inserted into a table. The bug does not appear on 5.2 because of redundant code. The redundant code is removed and the logic is moved to the controller. When in ajax edit more, the values are already comma-separated by the frontend so the bug only happens during multi-edit or insertion. 

The assertion seems unnecessary... I was hoping it would make the code easier to understand, but I can remove it.


libraries/classes/Controllers/Table/ReplaceController.php
```diff
@@ -180,10 +182,17 @@ public function __invoke(ServerRequest $request): void
...
                 // Note: $key is an md5 of the fieldname. The actual fieldname is
                 // available in $multi_edit_columns_name[$key]
 
+                $current_value = $multi_edit_columns[$key];
+                if (is_array($current_value)) {
+                    // Some column types accept comma-separated values e.g. set
+                    $current_value = implode(',', $current_value);
+                }
+
+                Assert::string($current_value);
```
williamdes, line 194: Yeah it am not sure it is useful

williamdes (approved):

libraries/classes/Controllers/Table/ReplaceController.php
kamil-tekiela: Removed

williamdes (approved):

MauricioFauth (approved):

PR #17681, opened 2022-08-09
Extract `/setup/index.php` entry point to the `/setup` route

MauricioFauth:
The `/setup` entry point were removed and was moved to the `/setup` route.

For example:
```diff
- http://localhost/setup/index.php
+ http://localhost/index.php?route=/setup
```

This was made basically for two reasons. To have only one entry point (`index.php`) and to have only one call to `Common::run` method. This will make easier to improve the Common class.


MauricioFauth:
I'm not sure if this should be done in 5.3.0 or 6.0.0. What do you think about this? @ibennetch  @williamdes

williamdes (changes requested):
I think that for some packaging reasons and historical reasons we should keep the setup folder
But if the folder is removed all the setup routes should not work

We should think the setup feature as a separate feature that users should be able to remove by simple file deletes (many users do that in fact and do file guessing for that)

Besides that maybe for other users it could also be nice that the main router is able to handle setup routes (Google app platform for example)
But if it is too complicated I think we can conclude they can do the setup by other means

MauricioFauth:
> I think that for some packaging reasons and historical reasons we should keep the setup folder But if the folder is removed all the setup routes should not work
> 
> We should think the setup feature as a separate feature that users should be able to remove by simple file deletes (many users do that in fact and do file guessing for that)
> 
> Besides that maybe for other users it could also be nice that the main router is able to handle setup routes (Google app platform for example) But if it is too complicated I think we can conclude they can do the setup by other means

What about the having a file (`/setup/index.php`) that redirects to `/index.php?route=/setup` and if this file/directory is deleted, then the `/setup` route is disabled?

We could also introduce a configuration directive that disables the `/setup` route, even if the setup directory is available.

williamdes:
> What about the having a file (/setup/index.php) that redirects to /index.php?route=/setup and if this file/directory is deleted, then the /setup route is disabled?

That an interesting idea
But I would prefer the setup routes to be separated in a separate file so they can be thrown out my a simple rm command

> We could also introduce a configuration directive that disables the /setup route, even if the setup directory is available.

I am not sure this is a good thing for us, let's keep the setup feature as a separate module that file deletes can remove. We still can have a setup/index.php file and libraries/setup code files. That would mean a folder and an entry point to delete for any user or package maintainer

PR #17684, opened 2022-08-13
Refactor insert edit

kamil-tekiela:
This is a major redesign of the code that handles these four actions: in-place edit(AJAX), edit of multiple rows, copying of rows, and insertion of new rows. The goal is to make the code easier to read and more understandable. 

- I introduced a new DTO for better readability. 
- Some of the methods were made private to the model
- The controller has access to two methods for getting the value for INSERT and for UPDATE
- Unfortunately, I have introduced method envy on EditField, but I do not want to move the functionality away from InsertEdit.php
- Unit tests have not improved in readability but I added helpful comments explaining what we are testing. There's probably a way to improve it too. 

I will probably merge all commits once I decide it's ready for review.

### DIFF
--- a/libraries/classes/EditField.php
+++ b/libraries/classes/EditField.php
@@ -1,0 +1,56 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpMyAdmin;
+
+/**
+ * @psalm-immutable
+ */
+final class EditField
+{
+    /** @var string $columnName */
+    public $columnName;
+    /** @var string $value */
+    public $value;
+    /** @var string $type */
+    public $type;
+    /** @var bool $autoIncrement */
+    public $autoIncrement;
+    /** @var bool $isNull */
+    public $isNull;
+    /** @var bool $wasPreviouslyNull */
+    public $wasPreviouslyNull;
+    /** @var string $function */
+    public $function;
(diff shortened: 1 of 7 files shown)


libraries/classes/EditField.php
```diff
@@ -0,0 +1,56 @@
...
+    public $type;
+    /** @var bool $autoIncrement */
+    public $autoIncrement;
+    /** @var bool $isNull */
+    public $isNull;
+    /** @var bool $wasPreviouslyNull */
+    public $wasPreviouslyNull;
+    /** @var string $function */
+    public $function;
+    /** @var ?string $salt */
```
williamdes, line 26:
```suggestion
    /** @var string|null $salt */
```
This is coherent with the codebase and is phpdoc standard syntax valid ;)

kamil-tekiela: Thanks. I will change this. This is only a very rough sketch of what I want to do. I still have a lot of work to do here.

kamil-tekiela:
There's no test for insert scenario in ReplaceController.php but I don't have the time now to create one. I open this PR for review, but if you'd like me to make some changes, I probably won't reply for at least the next 2 weeks.

MauricioFauth (approved):

PR #17695, opened 2022-08-25
Replace `cfg` global with `Settings` class

MauricioFauth:
The idea is to replace usages of the `$GLOBALS['cfg']` with the `PhpMyAdmin\Config\Settings` class to improve type checking and reduce global variable usage.

As the `settings` property of `Config` class is changing in a few places, the instance hold by the `config` property is recreated in a few places. It should be refactored in the future to avoid or reduce that.

### DIFF
--- a/libraries/classes/Common.php
+++ b/libraries/classes/Common.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace PhpMyAdmin;
 
+use PhpMyAdmin\Config\Settings\Server;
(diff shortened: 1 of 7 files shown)


williamdes (approved):

williamdes (approved):

MauricioFauth:
I'm not confident enough about this approach. I'm afraid of it getting out of sync. I'll do it using bottom-up approach. It may end up creating more instances, but I think it's safer.

Issue #17635, opened 2022-07-11
Semicolon repeatedly suffixed onto SQL if query ends in comment


ChrisHSandN:
### Describe the bug

If the end of an SQL statement is a comment, PHPMyAdmin will repeatedly inject a semicolon onto the end each time it is edited.

### To Reproduce

1. Go to for latest demo server https://www.phpmyadmin.net/try/ and use MariaDB
2. Click on SQL
3. Enter `SELECT 1 -- foobar`
4. Press Run
5. Press "Edit" then "Run" again
6. SQL is now shown as `SELECT 1 -- foobar;;`, repeating step 5 will add another `;` on each time.

### Expected behavior

SQL should be `SELECT 1 -- foobar;`, or ideally I suppose, leaving my SQL unaltered and not adding or removing any semicolons.

### Screenshots

![image](https://user-images.githubusercontent.com/13466620/178245823-57233140-b8e5-4f87-9d40-2130dd80b2ab.png)

### Additional context

Apologies for reporting such a trivial thing!

williamdes:
> Apologies for reporting such a trivial thing!

I am happy to have such detailed reports. We need users to report things they are not happy about ;)

Arpit-Gulgulia:
Can I give it a try to fix this issue if it is not fixed?

williamdes:
> Can I give it a try to fix this issue if it is not fixed?

Sure, let me know if you need help

williamdes:
I still can confirm this bug on the demo server 498a1a3353254e878b16469b2cb895c19e7fcd44

A test should be written first before fixing the code

PR #17708, opened 2022-09-03
semicolon on comment issue solved. Issues #17635

hameddanesh:
semicolon on comment issue solved. Issues #17635

### DIFF
--- a/test/classes/UtilTest.php
+++ b/test/classes/UtilTest.php
@@ -731,6 +731,11 @@ class UtilTest extends AbstractTestCase
         ];
     }
 
+    public function testFormatByteDownWithNullValue(): void
+    {
+        $this->assertNull(Util::formatByteDown(null));
+    }
+
     /**
      * format byte test, globals are defined
      *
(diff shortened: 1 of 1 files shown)


hameddanesh:
unnecessary changes on previous commit undone.

hameddanesh:
@williamdes 
unnecessary changes on previous commit undone

libraries/classes/Controllers/Import/ImportController.php
```diff
@@ -204,6 +204,10 @@ public function __invoke(ServerRequest $request): void
             $GLOBALS['format'] = 'sql';
             $_SESSION['sql_from_query_box'] = true;
 
+            if (strpos($GLOBALS['import_text'], " -- ") != -1 && substr($GLOBALS['import_text'], -1) == ';') {
```
kamil-tekiela, line 207: `strpos` never returns `-1` and the condition should use identical operator `===`. You can also use `str_ends_with()`. 

I am not sure that this is the correct fix for this issue though.

hameddanesh: thank you for your suggestions but actually this solution works, you can compile and see it for yourself.
but also i commit that it's not the best solution to the problem at hand (i couldn't find were semicolon gets concatnate to query string so i fixed bug here, if you can guide be to the correct function i will be grateful)
and about strpos() it returns 0 to maxInt if needle is found and -1 (or false as you may mean it ) if it didn't.
anyway, thank you for your honest review, ill try to get better. 🌹

williamdes:
Hi @hameddanesh 
Please do not forget to create a pull-request from another branch than `master`, the best is to create a branch from `master` branch and not use `master` directly. You will find some usefull information on: https://github.com/phpmyadmin/phpmyadmin/wiki/GIT_MANUAL

PR #17714, opened 2022-09-08
Fix for handle large number of tables in a single database (more than 10k)

PicoI2:
Hi, this is a patch to handle large number of table in a single database. I test it with more than 300 000 tables.
It works with or without `DisableIS`.
To be able to use phpMyAdmin with this amount of table, $cfg['EnableAutocompleteForTablesAndColumns'] must be set to false.
I hesitated to automatically disable `EnableAutocompleteForTablesAndColumns` when there was more than 10 000 tables, I can add it if you think it can be helpful.

### DIFF
--- a/libraries/classes/DatabaseInterface.php
+++ b/libraries/classes/DatabaseInterface.php
@@ -387,6 +387,10 @@ class DatabaseInterface implements DbalInterface
         }
 
         $tables = [];
+		
+		if ($limitCount && is_array($table)) {
+			$table = array_slice($table, $limitOffset, $limitCount);
+		}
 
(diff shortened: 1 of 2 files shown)


williamdes:
This is an interesting fix!
Could you rebase your branch onto QA_5_2?

PicoI2:
Yes I will.
I saw I also have to signed-off my commit.
I will do it soon.

PicoI2:
I close this one, I opened a [new one](https://github.com/phpmyadmin/phpmyadmin/pull/17717) with the rebased branch.

PR #17718, opened 2022-09-10
Fix for handle large number of tables in a single database (based on QA_5_2)

### LINKED ISSUE #17702: Allowed memory size exhausted on db_sql_autocomplete.php with a large number of tables
### Describe the bug

Our phpmyadmin was (seemingly) randomly throwing at us errors code 500 so I took a look at the logs and found this:
```
PHP Fatal error:  Allowed memory size of 67108864 bytes exhausted (tried to allocate 2097152 bytes) in [...]/phpmyadmin/db_sql_autocomplete.php on line 41
PHP message: PHP Fatal error:  Allowed memory size of 67108864 bytes exhausted (tried to allocate 131072 bytes) in Unknown on line 0" while reading response header from upstream, client: [...], server: [...], request: "POST /utils-sqladmin/db_sql_autocomplete.php HTTP/1.1"
```

### To Reproduce

1. Have a very large number of tables on a database (ours is ~3000)
2. Open the SQL panel on that database
3. Start writing/adding a WHERE clause
4. See error 500 popup

### Server configuration

- Operating system: Gentoo Linux
- Web server: Nginx
- Database version: 10.2.30-MariaDB
- PHP version: 7.1.33
- phpMyAdmin version: 5.0.1

PicoI2:
Hi, this is a patch to handle large number of table in a single database. I test it with more than 300 000 tables.
It works with or without DisableIS.
To be able to use phpMyAdmin with this amount of table, [$cfg['EnableAutocompleteForTablesAndColumns']](https://docs.phpmyadmin.net/en/latest/config.html#cfg_EnableAutocompleteForTablesAndColumns) must be set to false.
I hesitated to automatically disable EnableAutocompleteForTablesAndColumns when there was more than 10 000 tables, I can add it if you think it can be helpful.

Fixes:  #17702

### DIFF
--- a/libraries/classes/DatabaseInterface.php
+++ b/libraries/classes/DatabaseInterface.php
@@ -385,6 +385,15 @@ class DatabaseInterface implements DbalInterface
         }
 
         $tables = [];
+        $paging_applied = false;
+
+        if ($limit_count && is_array($table) && $sort_by === 'Name') {
+            if ($sort_order === 'DESC') {
+                $table = array_reverse($table);
+            }
+            $table = array_slice($table, $limit_offset, $limit_count);
+            $paging_applied = true;
+        }
 
         if (! $GLOBALS['cfg']['Server']['DisableIS']) {
             $sql_where_table = QueryGenerator::getTableCondition(
@@ -412,7 +421,7 @@ class DatabaseInterface implements DbalInterface
             // Sort the tables
             $sql .= ' ORDER BY ' . $sort_by . ' ' . $sort_order;
 
-            if ($limit_count) {
+            if ($limit_count && ! $paging_applied) {
                 $sql .= ' LIMIT ' . $limit_count . ' OFFSET ' . $limit_offset;
             }
 
@@ -593,7 +602,7 @@ class DatabaseInterface implements DbalInterface
                 unset($sortValues);
             }
 
-            if ($limit_count) {
+            if ($limit_count && ! $paging_applied) {
                 $each_tables = array_slice($each_tables, $limit_offset, $limit_count, true);
             }
 
--- a/libraries/classes/Util.php
+++ b/libraries/classes/Util.php
@@ -2292,7 +2292,7 @@ class Util
 
             $tables = $groupTable + $dbi->getTablesFull(
                 $db,
-                $groupWithSeparator !== false ? $groupWithSeparator : '',
+                $groupWithSeparator !== false ? $groupWithSeparator : $tables,
                 $groupWithSeparator !== false,
                 $limitOffset,
                 $limitCount,


williamdes (approved):
This looks good
Maybe @kamil-tekiela and @MauricioFauth maybe want to review

kamil-tekiela:
Could you please explain these changes? How do they fix the issue? What exactly did you change in the program behaviour? At first sight, the changes look good, but I don't fully understand how they address the problem.

PicoI2:
Yes sure.
`getTablesFull` was taking too much time, because MySQL do not have a real table where it stores all tables information.
So when it runs a query like `select * from information_schema.TABLES ORDER BY Name limit 250 offset 0`, MySQL has to read information about all tables to build his in memory table, then sort it by name, and then return 250 first tables.
But reading information about one table takes time, and that's why the query is very slow when you have a lot of tables.
Note that the problem is the same with ' SHOW TABLE STATUS'.

So, in order to speed up this query, we have to limit it to a small list of tables, as it, MySQL has to read table information only for few tables.
Note that `SHOW TABLES` is not that long, even with a large amount of tables.

The idea here is to use the result of `SHOW TABLES` query, to apply paging on this table list, and then to run the query on `information_schema.TABLES` only on tables we want.

That makes me think that my fix may not work if sorting is not by name, or if order is not ascending. I will add an other commit to handle theses cases.

libraries/classes/DatabaseInterface.php
```diff
@@ -385,6 +385,15 @@ public function getTablesFull(
         }
 
         $tables = [];
+        $paging_applied = false;
+
+        if ($limit_count && is_array($table) && $sort_by === 'Name') {
```
williamdes, line 390: I think that this block of logic should not be there, it should be upper in the call stack
Let see what @kamil-tekiela thinks about it

kamil-tekiela:
Would you be able to provide some benchmarks or some way to test the performance improvement for ourselves?

PicoI2:
I wrote a small bash script to create lots of tables (it takes a lot of time, try it with 10k will be enough)
```
#/bin/bash
db='-h 127.0.0.1 testdatabase -u root -proot'

for i in {1..230000}
do
	table="CREATE TABLE tbl$i ( id int(10) unsigned NOT NULL AUTO_INCREMENT, k int(10) unsigned NOT NULL DEFAULT '0', c varchar(120) NOT NULL DEFAULT '', pad varchar(60) NOT NULL DEFAULT '', PRIMARY KEY (id), KEY k_1 (k) ) ENGINE=InnoDB;"
	mysql $db -e "$table"
done
```

Then 
```
SHOW TABLES FROM `testdatabase`; 
took 0.457 seconds
```
With my fix :
```
SELECT * FROM `information_schema`.`TABLES` t WHERE `TABLE_SCHEMA`  IN ('testdatabase') AND t.`TABLE_NAME`  IN ('testtable1', 'testtable2', 'testtable3', 'testtable4', 'testtable5', 'testtable6', 'testtable7', 'testtable8', 'testtable9', 'testtable10') ORDER BY Name ASC 
took 0.331 seconds
```
Without my fix :
```
SELECT * FROM `information_schema`.`TABLES` t WHERE `TABLE_SCHEMA`  IN ('testdatabase')  ORDER BY Name DESC LIMIT 10 OFFSET 0 
took 848.339 seconds 
```

kamil-tekiela:
Thanks for the script. I have reviewed it and I can't really find anything wrong with this PR. However, the description is a little misleading because it doesn't fix the issue that you linked to. It does fix at least 2 other bugs from what I observed. It also improves performance for databases with a lot of tables. 

Now, with that many tables, phpMyAdmin chokes up in a lot of areas. Perhaps, it would be a good idea to do more benchmarking and try to improve performance elsewhere too. That's some idea for future PRs. :) 

As for what this actually fixed, it seems to have fixed two issues. First, passing an integer to `array_map`. I understand why it was done, but it was definitely wrong and in PHP 8 would result in a hard error. Surprisingly, it didn't come up in tests or in real life. PHPStan caught it (this PR could also adjust the baseline). I think a proper fix is in order but that's a small priority. Secondly, this fixes an issue with natural sorting. MySQL ordered the results by name using lexical sort and then the limited results would be sorted using natural sort in PHP. It's certainly not how it was intended to work. It definitely caught me by surprise while testing this PR. Big thanks for fixing this bug. Still, if the table list is not sorted by name, the sorting might be off. However, a similar sorting problem can be observed in other places, mainly in navigation. There the items are still ordered by MySQL. While before this change, the sorting was the same in navigation and in the table list, now it's different. Probably another PR should be created to address sorting issues in other places. 

As for this PR, I think it can be merged. Please adjust PHPStan baseline and add DCO to the commits (you can merge them all into one).

williamdes:
Hi @MauricioFauth
What do you think about this PR?
I would be tempted to land it into 5.2.2, not 5.2.1

kamil-tekiela:
Can we rebase this and merge?

MauricioFauth:
> Can we rebase this and merge?

I'll rebase and merge it.

williamdes:
> > Can we rebase this and merge?
> 
> I'll rebase and merge it.

I have post poned PR merges into QA since I am still stitching up the merge into master
For you to know @MauricioFauth
One of the PRs is not that easy to fix conflits for master 😄

MauricioFauth (approved):

Issue #16419, opened 2020-10-20
Errors in 5.0.4 when removing pragmarx folder


ibennetch:
Migrated from https://github.com/phpmyadmin/phpmyadmin/issues/16396#issuecomment-712003652

In version 5.0.2, the manual removal of `vendor/pragmarx/*` would not cause an error. With 5.0.4, there are several warnings/errors displayed on the main page:

```
Warning in ./vendor/composer/ClassLoader.php#444
 include(): Failed opening '/var/www/pma-dev/rel/phpMyAdmin-5.0.4-all-languages/vendor/composer/../pragmarx/google2fa-qrcode/src/Google2FA.php' for inclusion (include_path='.:/usr/share/php')

Backtrace

ClassLoader.php#444: include()
ClassLoader.php#322: Composer\Autoload\includeFile(string '/var/www/pma-dev/rel/phpMyAdmin-5.0.4-all-languages/vendor/composer/../pragmarx/google2fa-qrcode/src/Google2FA.php')
Composer\Autoload\ClassLoader->loadClass(string 'PragmaRX\\Google2FAQRCode\\Google2FA')
TwoFactor.php#132: class_exists(string 'PragmaRX\\Google2FAQRCode\\Google2FA')
TwoFactor.php#71: PhpMyAdmin\TwoFactor->getAvailable()
AuthenticationPlugin.php#342: PhpMyAdmin\TwoFactor->__construct(string 'root')
common.inc.php#402: PhpMyAdmin\Plugins\AuthenticationPlugin->checkTwoFactor()
index.php#23: require_once(./libraries/common.inc.php)
```

I don't see anything that's obviously changed from 5.0.2 to 5.0.4, but if we can it would be helpful to fix this.

To reproduce:

1. Download the compressed file from www.phpmyadmin.net/downloads
2. Uncompress it, create your own config.inc.php, and remove the vendor/pragmarx directory

Version 5.0.2 was fine, version 5.0.4 gives the errors. Note that this is not officially supported by our list in FAQ 1.44, but it used to work and now doesn't, so we should try to fix it anyway.

After fixing this, we could add that directory (and maybe more from `vendor`) to the list at [FAQ 1.44](https://docs.phpmyadmin.net/en/latest/faq.html#how-can-i-reduce-the-installed-size-of-phpmyadmin-on-disk).

williamdes:
Fixed in #17719

PR #17719, opened 2022-09-10
Fix removing vendor folders (FAQ 1.44)

### LINKED ISSUE #16418: Have a small and empty version of phpMyAdmin
> well, it would be nice to have a "basics" tarball without as much as possible 3rd party stuff not present at all instead the "rm -rf" orgy in phpMyAdmin.spec to begin with

_Originally posted by @thelounge-zz in https://github.com/phpmyadmin/phpmyadmin/issue_comments/712133312_

### LINKED ISSUE #17661: Broken composer package in 5.2.0
### Describe the bug

Cannot install additional package with zipped file.

I know that composer packages are preinstalled, but adding new package was working just fine with version 5.1.1.

### To Reproduce

```
wget -q -c https://files.phpmyadmin.net/phpMyAdmin/5.2.0/phpMyAdmin-5.2.0-english.tar.gz -O - | tar -xz --strip-components 1
composer install / composer require some-package
```

### Expected behavior

Successful install of package.

### What happens

composer fails with:

```
composer install
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. It is recommended that you run `composer update` or `composer update <package name>`.
Your lock file does not contain a compatible set of packages. Please run composer update.

  Problem 1
    - phpmyadmin/phpmyadmin is present at version 1.0.0+no-version-set and cannot be modified by Composer
    - roave/security-advisories dev-latest conflicts with phpmyadmin/phpmyadmin <5.1.3.
    - roave/security-advisories is locked to version dev-latest and an update of this package was not requested.
```

What I think is happening - composer.json is missing explicit version declaration.

Yeap, explicitly adding `    "version": "5.2.0"` to composer.json worked.

Probably worth rethinking zip strategy?

williamdes:
### Description

Ref: https://github.com/Roave/SecurityAdvisories/issues/76
Fixes: https://github.com/phpmyadmin/phpmyadmin/issues/17661
Ref: #16418
Fixes: #16419

### DIFF
--- a/.github/workflows/other-tools.yml
+++ b/.github/workflows/other-tools.yml
@@ -70,9 +70,6 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-yarn-
 
-      - name: Install modules
-        run: yarn install --non-interactive
-
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,10 @@
         "psr-4": {
             "PhpMyAdmin\\": "libraries/classes"
         },
-        "files": ["vendor/phpmyadmin/motranslator/src/functions.php"]
+        "files": ["vendor/phpmyadmin/motranslator/src/functions.php"],
+        "exclude-from-classmap": [
+            "/test/"
+        ]
     },
     "autoload-dev": {
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,12 @@
         },
         "files": ["vendor/phpmyadmin/motranslator/src/functions.php"],
         "exclude-from-classmap": [
-            "/test/"
+            "/test/",
+            "/vendor/tecnickcom/tcpdf/tcpdf_barcodes_*.php",
+            "/vendor/tecnickcom/tcpdf/tcpdf_import.php",
+            "/vendor/tecnickcom/tcpdf/tcpdf_parser.php",
+            "/vendor/tecnickcom/tcpdf/include/tcpdf_filters.php",
+            "/vendor/tecnickcom/tcpdf/include/barcodes"
         ]
     },
     "autoload-dev": {
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -686,16 +686,20 @@ Some users have requested to be able to reduce the size of the phpMyAdmin instal
 This is not recommended and could lead to confusion over missing features, but can be done.
 A list of files and corresponding functionality which degrade gracefully when removed include:
 
-* :file:`./vendor/tecnickcom/tcpdf` folder (exporting to PDF)
 * :file:`./locale/` folder, or unused subfolders (interface translations)
-* Any unused themes in :file:`./themes/`
-* :file:`./js/vendor/jquery/src/` (included for licensing reasons)
-* :file:`./js/line_counts.php` (removed in phpMyAdmin 4.8)
+* Any unused themes in :file:`./themes/` except the default theme `pmahomme`.
+* :file:`./libraries/language_stats.inc.php` (translation statistics)
 * :file:`./doc/` (documentation)
 * :file:`./setup/` (setup script)
-* :file:`./examples/`
-* :file:`./sql/` (SQL scripts to configure advanced functionality)
-* :file:`./js/vendor/openlayers/` (GIS visualization)
+* :file:`./examples/` (configuration examples)
+* :file:`./sql/` (SQL scripts to configure advanced functionalities)
+* :file:`./js/src/` (Source files to re-build `./js/dist/`)
+* :file:`./js/config/` (Configuration files to re-build `./js/dist/`)
+* Run `rm -rv vendor/tecnickcom/tcpdf && composer dump-autoload --no-interaction --optimize --dev` (exporting to PDF)
+* Run `rm -rv vendor/williamdes/mariadb-mysql-kbs && composer dump-autoload --no-interaction --optimize --dev` (external links to MariaDB and MySQL documentations)
+* Run `rm -rv vendor/code-lts/u2f-php-server && composer dump-autoload --no-interaction --optimize --dev` (U2F second factor authentication)
+* Run `rm -rv vendor/pragmarx/* && composer dump-autoload --no-interaction --optimize --dev` (2FA second factor authentication)
+* Run `rm -rv vendor/bacon/bacon-qr-code && composer dump-autoload --no-interaction --optimize --dev` (QRcode generation for 2FA second factor authentication)
 
 .. _faq1_45:
 
--- a/scripts/create-release.sh
+++ b/scripts/create-release.sh
@@ -302,6 +302,42 @@ security_checkup() {
     fi
 }
 
+autoload_checkup() {
+    php <<'CODE'
+<?php
+
+$classMapFiles = require __DIR__ . '/vendor/composer/autoload_classmap.php';
+
+$foundFiles = 0;
+$notFoundFiles = 0;
+
+foreach ($classMapFiles as $classMapFile) {
+        if (! file_exists($classMapFile)) {
+            echo 'Does not exist: ' . $classMapFile . PHP_EOL;
+            $notFoundFiles++;
+            continue;
+        }
+        $foundFiles++;
+}
+
+echo '[autoload class map checkup] Found files: ' . $foundFiles . PHP_EOL;
+echo '[autoload class map checkup] NOT found files: ' . $notFoundFiles . PHP_EOL;
+$minFilesToHave = 1100;// An arbitrary value based on how many files the autoload has on average
+
+if ($foundFiles < $minFilesToHave) {
+    echo '[autoload class map checkup] The project expects at least ' . $minFilesToHave . ' in the autoload class map' . PHP_EOL;
+    exit(1);
+}
+
+if ($notFoundFiles > 0) {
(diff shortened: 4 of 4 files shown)


composer.json
```diff
@@ -24,7 +24,19 @@
...
-        "files": ["vendor/phpmyadmin/motranslator/src/functions.php"]
+        "files": ["vendor/phpmyadmin/motranslator/src/functions.php"],
+        "exclude-from-classmap": [
+            "/test/",
+            "/vendor/williamdes/mariadb-mysql-kbs",
+            "/vendor/tecnickcom/tcpdf/tcpdf_barcodes_*.php",
+            "/vendor/tecnickcom/tcpdf/tcpdf_import.php",
+            "/vendor/tecnickcom/tcpdf/tcpdf_parser.php",
+            "/vendor/tecnickcom/tcpdf/include/tcpdf_filters.php",
+            "/vendor/tecnickcom/tcpdf/include/barcodes",
```
williamdes, line 35: just a cleanup, we do not need the classes do be in the class map as we do not use them

composer.json
```diff
@@ -24,7 +24,19 @@
         "psr-4": {
             "PhpMyAdmin\\": "libraries/classes"
         },
-        "files": ["vendor/phpmyadmin/motranslator/src/functions.php"]
+        "files": ["vendor/phpmyadmin/motranslator/src/functions.php"],
+        "exclude-from-classmap": [
+            "/test/",
```
williamdes, line 29: The files are present when we do composer commands in the release script so so it dumps class maps for it, that's not problematic but very useless and false.
It also breaks my code I wrote to check all classmap files exist

doc/faq.rst
```diff
@@ -690,12 +690,17 @@ A list of files and corresponding functionality which degrade gracefully when re
...
-* :file:`./examples/`
-* :file:`./sql/` (SQL scripts to configure advanced functionality)
+* :file:`./examples/` (configuration examples)
+* :file:`./sql/` (SQL scripts to configure advanced functionalities)
 * :file:`./js/vendor/openlayers/` (GIS visualization)
+* :file:`./js/src/` (Source files to re-build `./js/dist/`)
+* :file:`./js/config/` (Configuration files to re-build `./js/dist/`)
+* :file:`./vendor/williamdes/mariadb-mysql-kbs/` folder (external links to MariaDB and MySQL documentations)
+* :file:`./vendor/code-lts/u2f-php-server/` folder (U2F second factor authentication)
+* :file:`./vendor/pragmarx/google2fa-qrcode/` folder (2FA second factor authentication)
```
MauricioFauth, line 703: I'm not sure about this. Instead of adding more vendor files to remove, I think we should remove the existing ones.

williamdes: Hmm, no ?
Why should we remove our vendors that make 2FA possible ?

Or do you mean saying that no vendor is removable ?

MauricioFauth: Any vendor directory. That's not the proper way to "uninstall" a dependency.

williamdes: That's true, I will change the documentation about it

doc/faq.rst
```diff
@@ -686,16 +686,19 @@ Some users have requested to be able to reduce the size of the phpMyAdmin instal
...
-* :file:`./sql/` (SQL scripts to configure advanced functionality)
-* :file:`./js/vendor/openlayers/` (GIS visualization)
+* :file:`./examples/` (configuration examples)
+* :file:`./sql/` (SQL scripts to configure advanced functionalities)
+* :file:`./js/src/` (Source files to re-build `./js/dist/`)
+* :file:`./js/config/` (Configuration files to re-build `./js/dist/`)
+* Run `composer remove --update-no-dev tecnickcom/tcpdf` (exporting to PDF)
+* Run `composer remove --update-no-dev williamdes/mariadb-mysql-kbs` (external links to MariaDB and MySQL documentations)
+* Run `composer remove --update-no-dev code-lts/u2f-php-server` (U2F second factor authentication)
+* Run `composer remove --update-no-dev pragmarx/google2fa-qrcode` (2FA second factor authentication)
```
williamdes, line 701: @MauricioFauth this is problematic because it will remove all other vendors we installed manually from the suggest space. But because we did revert the changes to composer.json composer does not know why they are here and removes them

```
phpMyAdmin-5.2.1-dev-all-languages$ composer remove williamdes/mariadb-mysql-kbs --dry-run --update-no-dev --no-interaction
williamdes/mariadb-mysql-kbs is not required in your composer.json and has not been removed
./composer.json has been updated
Running composer update williamdes/mariadb-mysql-kbs
Loading composer repositories with package information
Package "williamdes/mariadb-mysql-kbs" listed for update is not locked.
Updating dependencies
Nothing to modify in lock file
Installing dependencies from lock file
Package operations: 0 installs, 0 updates, 8 removals
  - Removing williamdes/mariadb-mysql-kbs (v1.2.13)
  - Removing tecnickcom/tcpdf (6.5.0)
  - Removing pragmarx/google2fa-qrcode (v2.1.1)
  - Removing pragmarx/google2fa (v8.0.1)
  - Removing paragonie/constant_time_encoding (v2.6.3)
  - Removing dasprid/enum (1.0.3)
  - Removing code-lts/u2f-php-server (v1.2.0)
  - Removing bacon/bacon-qr-code (2.0.7)
```

MauricioFauth: What about using `--no-install` instead of `--update-no-dev`?

composer.json
MauricioFauth: Do we really have to maintain this list of excludes?

scripts/create-release.sh
```diff
@@ -468,15 +509,27 @@ do
     PACKAGES_VERSIONS="$PACKAGES_VERSIONS $PACKAGES:$PKG_VERSION"
 done
 
-echo "Installing composer packages '$PACKAGES_VERSIONS'"
+echo "* Installing composer packages '$PACKAGES_VERSIONS'"
 
-composer require --no-interaction --optimize-autoloader --update-no-dev $PACKAGES_VERSIONS
```
MauricioFauth, line 473: Why removing `--optimize-autoloader` was necessary?

scripts/create-release.sh
```diff
@@ -284,7 +284,7 @@ delete_phpunit_sandbox() {
         echo 'No phpunit sandbox to delete'
         exit 1;
     fi
-    rm -rf "${TEMP_PHPUNIT_FOLDER}"
+    rm -r "${TEMP_PHPUNIT_FOLDER}"
```
MauricioFauth, line 287: Not related to this PR, but could we use the [PHPUnit's PHAR](https://phpunit.de/getting-started/phpunit-9.html) instead of installing it in a sandbox?

doc/faq.rst
williamdes: Well the `--no-install` will make that the package is not removed and composer will warn about it. Basically it does nothing

composer.json
williamdes: Yes, unfortunately. It's not really needed about the tcpdf lib, I can remove the lines. At some point I would like to add the files to the vendor exclude removal list. But for now each time I do this composer fails building the class map because they are hard coded files in tcpdf's composer json and if files are missing it gets un happy

composer.json
```diff
@@ -24,7 +24,19 @@
...
+            "/vendor/williamdes/mariadb-mysql-kbs",
+            "/vendor/tecnickcom/tcpdf/tcpdf_barcodes_*.php",
+            "/vendor/tecnickcom/tcpdf/tcpdf_import.php",
+            "/vendor/tecnickcom/tcpdf/tcpdf_parser.php",
+            "/vendor/tecnickcom/tcpdf/include/tcpdf_filters.php",
+            "/vendor/tecnickcom/tcpdf/include/barcodes",
+            "/vendor/pragmarx/google2fa-qrcode",
+            "/vendor/bacon/bacon-qr-code",
+            "/vendor/code-lts/u2f-php-server"
```
williamdes, line 38: Other excludes could be removed if we solve the composer remove library problem

scripts/create-release.sh
williamdes: Yeah probably if we can extract the version that we use. It's probably more complicated than the sandbox code. Or maybe not

doc/faq.rst
MauricioFauth: What about adding the optional dependencies that are included in the release to the composer.json's `require` with the release script?

williamdes: Yes, it's an option, but anyone doing packaging will then be blocked because of that. The libs will be required but they still are not. Maybe we could ask composer team some help ?

.github/workflows/other-tools.yml
```diff
@@ -70,9 +70,6 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-yarn-
 
-      - name: Install modules
-        run: yarn install --non-interactive
```
MauricioFauth, line 74: Shouldn't the previous step that caches the Yarn dependencies be removed as well?

MauricioFauth (approved):

.github/workflows/other-tools.yml
williamdes: Probably not because the install normally uses the system cache

MauricioFauth: :100:

Issue #17369, opened 2022-02-11
blowfish_secret of not exactly 32 characters causes server error


ibennetch:
QA_5_2 and master have a problem if blowfish_secret isn't exactly 32 bytes long. The main pane is blank aside from the top tabs and this is logged:

```
PHP Fatal error:  Uncaught ErrorException: The secret passphrase in configuration (blowfish_secret) is not the correct length. It should be 32 bytes long. in /var/www/pma-dev/fork/libraries/classes/Controllers/HomeController.php:330\nStack trace:\n#0 [internal function]: PhpMyAdmin\\ErrorHandler->handleError()\n#1 /var/www/pma-dev/fork/libraries/classes/Controllers/HomeController.php(330): trigger_error()\n#2 /var/www/pma-dev/fork/libraries/classes/Controllers/HomeController.php(220): PhpMyAdmin\\Controllers\\HomeController->checkRequirements()\n#3 /var/www/pma-dev/fork/libraries/classes/Routing.php(187): PhpMyAdmin\\Controllers\\HomeController->__invoke()\n#4 /var/www/pma-dev/fork/index.php(43): PhpMyAdmin\\Routing::callControllerForRoute()\n#5 {main}\n  thrown in /var/www/pma-dev/fork/libraries/classes/Controllers/HomeController.php on line 330
```

We should definitely handle this more gracefully and show the user a warning that their blowfish_secret isn't the correct length.

xkill:
Hi, I'm using a 93 chars secret, and I'm also having this problem. 5.2.0 just updated from 5.1.3.

svivian:
Is there a reason it needs to be exactly 32 characters long? Surely a longer key is more secure?

williamdes:
> 32 bytes long

Seems like this is a technical issue from the php extension: https://www.php.net/manual/en/function.sodium-crypto-secretbox.php

> If key has a length of bytes different than [SODIUM_CRYPTO_SECRETBOX_KEYBYTES](https://www.php.net/manual/en/sodium.constants.php#constant.sodium-crypto-secretbox-keybytes) (32 bytes), a [SodiumException](https://www.php.net/manual/en/class.sodiumexception.php) will be thrown.

J0riz:
We just updated phpMyAdmin on one of our staging servers to phpMyAdmin 5.2.
We noticed that after login you see the follow message in a red box:
[...]

PR #17725, opened 2022-09-14
Fix server error when `blowfish_secret` is not exactly 32 bytes long

MauricioFauth:
- [x] Back-port commit c5611531c03a5c9dfe13510e7fc5b901daa77039
- [x] Update documentation about the `blowfish_secret` setting
- [x] Update the setup page to reflect the `blowfish_secret` changes

Fixes https://github.com/phpmyadmin/phpmyadmin/issues/17369.

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -1911,6 +1911,15 @@ Cookie authentication options
         // The string is a hexadecimal representation of a 32-bytes long string of random bytes.
         $cfg['blowfish_secret'] = sodium_hex2bin('f16ce59f45714194371b48fe362072dc3b019da7861558cd4ad29e4d6fb13851');
 
+    Using a binary string is recommended. However, if all 32 bytes of the string are visible
+    characters, then a function like `sodium_bin2hex <https://www.php.net/sodium_bin2hex>`_ is not required. For
+    example:
+
+    .. code-block:: php
+
+        // A string of 32 characters.
+        $cfg['blowfish_secret'] = 'JOFw435365IScA&Q!cDugr!lSfuAz*OW';
+
     .. warning::
 
         The encryption key must be 32 bytes long. If it is longer than the length of bytes, only the first 32 bytes will
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -1896,13 +1896,23 @@ Cookie authentication options
     :type: string
     :default: ``''``
 
-    The "cookie" auth\_type uses AES algorithm to encrypt the password. If you
-    are using the "cookie" auth\_type, enter here a random passphrase of your
-    choice. It will be used internally by the AES algorithm: you won’t be
-    prompted for this passphrase.
+    The "cookie" auth\_type uses the :term:`Sodium` extension to encrypt the cookies (see :term:`Cookie`). If you are
+    using the "cookie" auth\_type, enter here a generated string of random bytes to be used as an encryption key. It
+    will be used internally by the :term:`Sodium` extension: you won't be prompted for this encryption key.
 
-    The secret should be 32 characters long. Using shorter will lead to weaker security
-    of encrypted cookies, using longer will cause no harm.
+    Since a binary string is usually not printable, it can be converted into a hexadecimal representation (using a
+    function like `sodium_bin2hex <https://www.php.net/sodium_bin2hex>`_) and then used in the configuration file. For
+    example:
+
+    .. code-block:: php
+
+        // The string is a hexadecimal representation of a 32-bytes long string of random bytes.
+        $cfg['blowfish_secret'] = sodium_hex2bin('f16ce59f45714194371b48fe362072dc3b019da7861558cd4ad29e4d6fb13851');
+
+    .. warning::
+
+        The encryption key must be 32 bytes long. If it has a different length of bytes, a new key will be automatically
+        generated for you. However it will only last for the duration of the session.
 
     .. note::
 
@@ -1910,10 +1920,18 @@ Cookie authentication options
         Blowfish algorithm was originally used to do the encryption.
 
     .. versionchanged:: 3.1.0
+
         Since version 3.1.0 phpMyAdmin can generate this on the fly, but it
         makes a bit weaker security as this generated secret is stored in
         session and furthermore it makes impossible to recall user name from
         cookie.
+
+    .. versionchanged:: 5.2.0
+
+        Since version 5.2.0, phpMyAdmin uses the
+        `sodium\_crypto\_secretbox <https://www.php.net/sodium_crypto_secretbox>`_ and
+        `sodium\_crypto\_secretbox\_open <https://www.php.net/sodium_crypto_secretbox_open>`_ PHP functions to encrypt
+        and decrypt cookies, respectively.
 
 .. config:option:: $cfg['CookieSameSite']
 
@@ -3809,8 +3827,8 @@ following example shows two of them:
 .. code-block:: php
 
     <?php
-    $cfg['blowfish_secret'] = 'multiServerExample70518';
-    // any string of your choice
+    // The string is a hexadecimal representation of a 32-bytes long string of random bytes.
+    $cfg['blowfish_secret'] = sodium_hex2bin('f16ce59f45714194371b48fe362072dc3b019da7861558cd4ad29e4d6fb13851');
     $i = 0;
 
     $i++; // server 1 :
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -1913,8 +1913,9 @@ Cookie authentication options
 
     .. warning::
 
-        The encryption key must be 32 bytes long. If it has a different length of bytes, a new key will be automatically
-        generated for you. However it will only last for the duration of the session.
+        The encryption key must be 32 bytes long. If it is longer than the length of bytes, only the first 32 bytes will
+        be used, and if it is shorter, a new temporary key will be automatically generated for you. However, this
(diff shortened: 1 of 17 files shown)


williamdes (approved):
Was there a technical reason for not using trigger error, I am not sure I remember

MauricioFauth:
> Was there a technical reason for not using trigger error, I am not sure I remember 

The trigger_error is converted to a error exception.
With this way, the error is still shown, but without the need for a error/exception handler.

I'll also update the documentation to reflect the changes and update the setup page to provide better and more secure secrets.

libraries/classes/Setup/ConfigGenerator.php
```diff
@@ -94,6 +100,12 @@ public static function getConfigFile(ConfigFile $cf)
      */
     private static function getVarExport($var_name, $var_value, $crlf)
     {
+        if ($var_name === 'blowfish_secret') {
+            $secret = self::getBlowfishSecretKey($var_value);
+
+            return sprintf('$cfg[\'blowfish_secret\'] = \sodium_hex2bin(\'%s\');%s', sodium_bin2hex($secret), $crlf);
```
MauricioFauth, line 106: I think that having a function call ([`sodium_hex2bin`](https://www.php.net/manual/en/function.sodium-hex2bin.php)) in the config file is a good trade-off for having better keys.

libraries/classes/Config/ServerConfigChecks.php
```diff
@@ -345,55 +348,21 @@ protected function performConfigChecksCookieAuthUsed(
...
-        $hasDigits = (bool) preg_match('/\d/', $blowfishSecret);
-        $hasChars = (bool) preg_match('/\S/', $blowfishSecret);
-        $hasNonword = (bool) preg_match('/\W/', $blowfishSecret);
-        if (! $hasDigits || ! $hasChars || ! $hasNonword) {
-            $blowfishWarnings[] = Sanitize::sanitizeMessage(
-                __(
-                    'Key should contain letters, numbers [em]and[/em] special characters.'
-                )
-            );
-        }
```
MauricioFauth, line 386: This was removed because it generates a new key not only when the key is not set, but when it does not have the correct length as well. And since it's a binary string, does not makes sense to check for used characters in the string.

doc/config.rst
```diff
@@ -1896,25 +1896,42 @@ Cookie authentication options
     :type: string
     :default: ``''``
 
-    The "cookie" auth\_type uses AES algorithm to encrypt the password. If you
-    are using the "cookie" auth\_type, enter here a random passphrase of your
-    choice. It will be used internally by the AES algorithm: you won’t be
-    prompted for this passphrase.
+    The "cookie" auth\_type uses the Sodium extension to encrypt the cookies. If you are using the "cookie" auth\_type,
+    enter here a generated string of random bytes to be used as an encryption key. It will be used internally by the
+    Sodium extension: you won’t be prompted for this encryption key.
```
williamdes, line 1901:
```suggestion
    Sodium extension: you won't be prompted for this encryption key.
```

libraries/classes/Setup/ConfigGenerator.php
williamdes: we could use binary strings but in the format `0xfffffabcd0123` and it would not need a call to an extension. And that would keep our config files future proof and have no "logic"

MauricioFauth: Could you provide an example? Since `0xfffffabcd0123` is treated as a number.
```
php > var_dump(hex2bin('48656c6c6f20776f726c6421'));
php shell code:1:
string(12) "Hello world!"
php > var_dump(0x48656c6c6f20776f726c6421);
php shell code:1:
double(2.2405534230754E+28)
php > var_dump((string) 0x48656c6c6f20776f726c6421);
php shell code:1:
string(19) "2.2405534230754E+28"
```

MauricioFauth: > it would not need a call to an extension

It's a bundled extension since PHP 7.2, so it's not an exotic extension. That's why I think it's safe to add this function call. And `sodium_hex2bin` is constant-time, which is important for encryption keys to avoid attacks, while `hex2bin` is not.

williamdes: > It's a bundled extension since PHP 7.2, so it's not an exotic extension.

That's a good point. Maybe at some point we should remove the need for the compat lib ?
https://github.com/phpmyadmin/phpmyadmin/blob/b9d1abb1013a4719fe8b7d2ea4cca6b2fb59374f/composer.json#L53

> And sodium_hex2bin is constant-time, which is important for encryption keys to avoid attacks, while hex2bin is not.

Can you explain a bit more how this brings more security ?
I mean most of our users will not have `sodium_hex2bin` in their config files so how does this work our for actual users vs new ones ?
Should people use `sodium_hex2bin` in their config to be more secure ?
Do we add a message to ask people to change ?

MauricioFauth:
Hi, @ibennetch. Could you please review the wording of the documentation changes?

libraries/classes/Setup/ConfigGenerator.php
MauricioFauth: > That's a good point. Maybe at some point we should remove the need for the compat lib ?

Yes, I think we can remove the `paragonie/sodium_compat` package in the future.

> Can you explain a bit more how this brings more security ?

If you converted the string to hexadecimal, then when converting back to binary, the `sodium_hex2bin` function will prevent the key being leaked by a timing attack.

> I mean most of our users will not have sodium_hex2bin in their config files so how does this work our for actual users vs new ones ?
> Should people use sodium_hex2bin in their config to be more secure ?
> Do we add a message to ask people to change ?

Since the `blowfish_secret` was already 32 chars long, there is no need to change as it's already a valid key. However, when using a binary string, it's possible to generate better keys because there is more possibilities. But since it's not printable, it should be converted to hexadecimal or base64 before pasting it in the config file.

williamdes: I just had an idea, could we let the config with the hex and in our code call the sodium function?
Like that the config would still remain logic less

MauricioFauth: And what about the users that already have a valid key? We'll have to support a 64-chars long hexadecimal string and a 32-bytes long string. The user may have this key saved in another place, for example in a file, they will have to convert to hexadecimal first?

williamdes: For valid keys we can detect that it's not a 64 hex and assume it's a 32 one
For users with files they can do what they want in their config file as before nothing changes
From a config file point of view nothing changes. Only allowing 64 hex keys
In our code we handle each format and use or not the sodium lib

Would that work?

MauricioFauth: We could support both formats. But I think this is more confusing.

> In our code we handle each format and use or not the sodium lib

We use the sodium lib regardless. For the 64 hex case, the function call will only move from the config file to the code that handles it.

williamdes: My fear is that some day it will crash in our faces and we will have to ask users to update the config because we can not do anything about it. If you think it is best to have it like you proposed that will be okay for me and we will see what is better in the end after some years :)

MauricioFauth: Oh, I see. For example, the sodium_hex2bin being deprecated or we changing the method to a new one and having to asking the users to update the config.

williamdes: > Oh, I see. For example, the sodium_hex2bin being deprecated or we changing the method to a new one and having to asking the users to update the config.

Yup, exactly. For now the config file traveled 10+ years without anything breaking and no updates needed ^^

MauricioFauth: Except when I broke it when I added a requirement for a 32 bytes string. I changed the code and now it accepts longer strings, but will only use the first 32 bytes.
The sodium_hex2bin function is not required in the config file, only a 32 bytes long string. So the current values will keep working seamlessly. My intention in suggesting using this function is to help users having better keys by using bytes, instead of only characters.
The Sodium is a bundled extension, which means that it's part of the language. So a possible deprecation will be hopefully well planned. That way, we'll have time to work on a solution if we need to.

williamdes: Okay, let's do that. Let's say to users they can do one or another while updating keys

williamdes (approved):
🚢

libraries/classes/Controllers/HomeController.php
```diff
@@ -292,39 +296,43 @@ private function checkRequirements(): void
...
-                trigger_error(
-                    __(
-                        'The configuration file now needs a secret passphrase (blowfish_secret).'
+            $encryptionKeyLength = mb_strlen($cfg['blowfish_secret'], '8bit');
+            if ($encryptionKeyLength < SODIUM_CRYPTO_SECRETBOX_KEYBYTES) {
+                $this->errors[] = [
+                    'message' => __(
+                        'The configuration file needs a valid key for cookie encryption.'
+                        . ' A temporary key was automatically generated for you.'
+                        . ' Please refer to the [doc@cfg_blowfish_secret]documentation[/doc].'
```
williamdes, line 320: I am not on my workstation but was wondering if the link target did have a citation for users to find the FAQ entry you added

PR #17732, opened 2022-09-21
Replace form-control with form-select

serjlc:
Replaced by #17733


williamdes:
I fixed the base branch issue even if you where right about the target. Like that we can see the real diff of the PR.
You can create a branch on your repository from QA_5_2 and copy your changes ontop of it. Then open a pull-request here to QA_5_2
Let me know if you need some help :)

serjlc:
Hey William, thanks for the info. Actually, I could use some help to make sure I'm doing it properly, as I'm looking into making my first contributions to the community so I'm new to this.

PR #17752, opened 2022-10-04
Remove Sodium polyfill

MauricioFauth:
Removes the [paragonie/sodium_compat](https://packagist.org/packages/paragonie/sodium_compat) package.

This polyfill is not required as the Sodium extension is bundled in PHP 7.2+.

From the [PHP documentation](https://www.php.net/manual/en/sodium.installation.php):

> As of PHP 7.2.0 this extension is bundled with PHP.

For Debian, it's a built-in extension [since Buster](https://packages.debian.org/buster/php7.3-cgi). So installing PHP will automatically install the Sodium extension.


williamdes (approved):
Yay

PR #17757, opened 2022-10-08
Add the phpstan/phpstan-strict-rules package

MauricioFauth:
Adds additional rules for PHPStan to make it more strict.

The disabled rules are reporting too much errors right now.

- https://github.com/phpstan/phpstan-strict-rules


phpstan.neon.dist
```diff
@@ -22,3 +22,10 @@ parameters:
     dynamicConstantNames:
         - ROOT_PATH
         - VERSION_SUFFIX
+    strictRules:
+        # Allow non-booleans in conditions.
+        booleansInConditions: false
+        # Allow the empty() construct.
+        disallowedConstructs: false
+        # Allow dynamic calls to static methods.
+        strictCalls: false
```
MauricioFauth, line 31: Enabling `strictCalls` causes errors like the following:
```
Dynamic call to static method PHPUnit\Framework\Assert::assertStringContainsString().
```

williamdes (approved):

Issue #17756, opened 2022-10-06
Edit/Copy/Delete row doesn't work when using GROUP BY


liviuconcioiu:
### Describe the bug

Edit/Copy/Delete row doesn't work when using GROUP BY, because of missing single quotes.

### To Reproduce

1. Import table attached
2. Run
```sql
SELECT `col` FROM `test` GROUP BY `col`
```
4. Test the Edit/Copy/Delete functions
5. See error

### Expected behavior

Edit/Copy/Delete should work.

### Screenshots

https://user-images.githubusercontent.com/25424343/194418213-31ec73d9-ca7d-4a57-b65b-8ea6eee0ba2d.mp4

### Server configuration

- Operating system: Linux
- Web server: nginx/1.14.2
- Database version: 8.0.18 - MySQL Community Server
- PHP version: 7.4.29
- phpMyAdmin version: 5.2.1-dev

### Client configuration

- Browser: Firefox
- Operating system: Windows

### Additional context

[test.zip](https://github.com/phpmyadmin/phpmyadmin/files/9728619/test.zip)

sureemo:
may i work on this issue?

kamil-tekiela:
Please do! Thank you.

Issue #11406, opened 2015-08-17
Missing quotes after calling "distinct values"


plabong:
Calling "distinct values" in table/structure ...

![17-08-2015 13-22-15](https://cloud.githubusercontent.com/assets/462559/9304032/093f82ac-44e4-11e5-9e49-8a42e1846826.png)

...gives these nice list with distinct values and its counts ...

![17-08-2015 13-22-50](https://cloud.githubusercontent.com/assets/462559/9304094/c5c33838-44e4-11e5-9712-d4472d7b2dd9.png)

Click on one of this distinct values should list datasets limited to this distinct value. This works for numeric data types but not for strings; because the value is not quoted ...

![17-08-2015 13-23-11](https://cloud.githubusercontent.com/assets/462559/9304056/58b5a8d4-44e4-11e5-8998-c07596d40d0b.png)

devenbansod:
Hi, I tried this with datatypes varchar(25), text, date, int etc.
But on both the current master as well as the latest release 4.4.13.1, I am not able to reproduce this.

Can you please confirm the phpMyAdmin version you are using?

lem9:
I am able to reproduce this with sakila.actor.last_name. In QA_4_4 I see the error and in master I get blank results.

plabong:
pma-version = 4.4.13.1,
properties of this column:
`location` VARCHAR(32) CHARACTER SET utf8 COLLATE utf8_general_ci NULL DEFAULT NULL;

madhuracj:
@lem9 I couldn't reproduce the bug with sakila.actor.last_name on QA_4_4. Anything special about the remembered sorting order or column visibility?

madhuracj:
@lem9 Just noticed that you have assigned yourself and and working on the bug :)

lem9:
@madhuracj I bet that for you, the logic goes thru _getDataCellForNonNumericColumns(). But for me, it goes thru _getDataCellForNumericColumns() !

lem9:
@devenbansod Can you confirm the scenario I just wrote to @madhuracj ?

devenbansod:
@lem9 Yes. I can confirm it.
It, indeed, goes for me through _getDataCellForNonNumericColumns()

And so, even I can not reproduce the bug with the sakila.actor.last_name column.

lem9:
@devenbansod Are you sure that you are showing Structure for sakila.actor and picking the "Distinct values" action for last_name?

[...]

PR #17758, opened 2022-10-08
Fix issue - Edit/Copy/Delete row doesn't work when using GROUP BY

sureemo:
### Description

Reference issue: https://github.com/phpmyadmin/phpmyadmin/issues/17756

### Issue Detail

Step to reproduce are explained in the [issue detail](https://github.com/phpmyadmin/phpmyadmin/issues/17756).

When query with group by and try to edit/copy/delete, following are the sample issues:

**copy row with all string value, error message occurs**

Screen Shot 2565-10-09 at 00 57 39

Screen Shot 2565-10-09 at 00 58 11

**copy row with double quote, value is not available in form**

Screen Shot 2565-10-09 at 00 59 03

Screen Shot 2565-10-09 at 00 58 40

**copy row with single quote, value is not available in form**

Screen Shot 2565-10-09 at 00 59 08

Screen Shot 2565-10-09 at 00 59 17

### DIFF
--- a/libraries/classes/FieldMetadata.php
+++ b/libraries/classes/FieldMetadata.php
@@ -240,8 +240,8 @@ final class FieldMetadata
             // reference: https://www.php.net/manual/en/mysqli-result.fetch-fields.php
             // so check field type instead of flags
             // but if no or unknown field type then check flags
-            $this->isNumeric = $this->isType(self::TYPE_INT)
-                ||  ($fieldType <= 0 && ($fieldFlags & MYSQLI_NUM_FLAG));
+            $this->isNumeric = (bool) $this->isType(self::TYPE_INT)
+                || ($fieldType <= 0 && ($fieldFlags & MYSQLI_NUM_FLAG));
 
             $this->isBlob = (bool) ($fieldFlags & MYSQLI_BLOB_FLAG);
             $this->isEnum = (bool) ($fieldFlags & MYSQLI_ENUM_FLAG);
--- a/libraries/classes/FieldMetadata.php
+++ b/libraries/classes/FieldMetadata.php
@@ -227,6 +227,8 @@ final class FieldMetadata
 
     public function __construct(int $fieldType, int $fieldFlags, object $field)
     {
+            $this->mappedType = $this->getTypeMap()[$fieldType] ?? null;
+
             $this->isMultipleKey = (bool) ($fieldFlags & MYSQLI_MULTIPLE_KEY_FLAG);
             $this->isPrimaryKey = (bool) ($fieldFlags & MYSQLI_PRI_KEY_FLAG);
             $this->isUniqueKey = (bool) ($fieldFlags & MYSQLI_UNIQUE_KEY_FLAG);
@@ -236,9 +238,8 @@ final class FieldMetadata
 
             // as flags 32768 can be NUM_FLAG or GROUP_FLAG
             // reference: https://www.php.net/manual/en/mysqli-result.fetch-fields.php
-            // so also check field type that not 253 (VARCHAR)
-            $this->isNumeric = (bool) (($fieldFlags & MYSQLI_NUM_FLAG)
-                && $this->isType(self::TYPE_INT));
+            // so also check field type instead of flags
+            $this->isNumeric = $this->isType(self::TYPE_INT);
 
             $this->isBlob = (bool) ($fieldFlags & MYSQLI_BLOB_FLAG);
             $this->isEnum = (bool) ($fieldFlags & MYSQLI_ENUM_FLAG);
@@ -249,8 +250,6 @@ final class FieldMetadata
                 MYSQLI_TIMESTAMP_FLAG => 'timestamp',
                 MYSQLI_AUTO_INCREMENT_FLAG => 'auto_increment',
             */
-
-            $this->mappedType = $this->getTypeMap()[$fieldType] ?? null;
 
             $this->isMappedTypeBit = $this->isType(self::TYPE_BIT);
             $this->isMappedTypeGeometry = $this->isType(self::TYPE_GEOMETRY);
--- a/libraries/classes/FieldMetadata.php
+++ b/libraries/classes/FieldMetadata.php
@@ -67,7 +67,6 @@ final class FieldMetadata
     public const TYPE_TIMESTAMP = 11;
     public const TYPE_DATETIME = 12;
     public const TYPE_YEAR = 13;
-    public const TYPE_VARCHAR = 253;
 
     /**
      * @var bool
@@ -234,8 +233,13 @@ final class FieldMetadata
             $this->isNotNull = (bool) ($fieldFlags & MYSQLI_NOT_NULL_FLAG);
             $this->isUnsigned = (bool) ($fieldFlags & MYSQLI_UNSIGNED_FLAG);
             $this->isZerofill = (bool) ($fieldFlags & MYSQLI_ZEROFILL_FLAG);
+
+            // as flags 32768 can be NUM_FLAG or GROUP_FLAG
+            // reference: https://www.php.net/manual/en/mysqli-result.fetch-fields.php
+            // so also check field type that not 253 (VARCHAR)
             $this->isNumeric = (bool) (($fieldFlags & MYSQLI_NUM_FLAG)
-                && $fieldType !== self::TYPE_VARCHAR);
+                && $this->isType(self::TYPE_INT));
+
             $this->isBlob = (bool) ($fieldFlags & MYSQLI_BLOB_FLAG);
             $this->isEnum = (bool) ($fieldFlags & MYSQLI_ENUM_FLAG);
             $this->isSet = (bool) ($fieldFlags & MYSQLI_SET_FLAG);
--- a/libraries/classes/FieldMetadata.php
+++ b/libraries/classes/FieldMetadata.php
@@ -239,9 +239,9 @@ final class FieldMetadata
             // as flags 32768 can be NUM_FLAG or GROUP_FLAG
             // reference: https://www.php.net/manual/en/mysqli-result.fetch-fields.php
             // so check field type instead of flags
-            // but if no field type then check flags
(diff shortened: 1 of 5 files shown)


libraries/classes/Util.php
```diff
@@ -880,7 +880,7 @@ private static function getConditionValue(
         // See commit: 049fc7fef7548c2ba603196937c6dcaf9ff9bf00
         // See bug: https://sourceforge.net/p/phpmyadmin/bugs/3064/
         if ($meta->isNumeric && ! $meta->isMappedTypeTimestamp && $meta->isNotType(FieldMetadata::TYPE_REAL)) {
-            $conditionValue = '= ' . $row;
+            $conditionValue = '= "' . $dbi->escapeString((string) $row) . '"';
```
williamdes, line 883: This does not seem right, the field is assumed to be a numeric value
Is there some flag in the meta object we could use to fall into a case where it is quoted?

kamil-tekiela:
Thanks for this fix. It's certainly a very interesting take on it and one that would solve the problem, but I am not sure that this is the right one. 

I think the bigger question is why is the `col` column considered to be numeric when `GROUP BY col`? Running mysql.exe with [`--column-type-info`](https://dev.mysql.com/doc/refman/5.7/en/mysql-command-options.html#option_mysql_column-type-info) proves that the field indeed is reported as `NUM`. 😕 😮 

According to MySQL documentation:
> NUM_FLAG indicates that a column is numeric. This includes columns with a type of MYSQL_TYPE_DECIMAL, MYSQL_TYPE_NEWDECIMAL, MYSQL_TYPE_TINY, MYSQL_TYPE_SHORT, MYSQL_TYPE_LONG, MYSQL_TYPE_FLOAT, MYSQL_TYPE_DOUBLE, MYSQL_TYPE_NULL, MYSQL_TYPE_LONGLONG, MYSQL_TYPE_INT24, and MYSQL_TYPE_YEAR.

Both MySQL and MariaDB return the same value for flags. I haven't got an explanation as to why that is yet. 

The MySQL documentation states that it offers C macro called [`IS_NUM`](https://github.com/mysql/mysql-server/blob/8d8c986e5716e38cb776b627a8eee9e92241b4ce/include/mysql.h#L115) that should help in identifying numerical fields. It begs the question what's the purpose of that NUM_FLAG then? Perhaps, the correct solution would be to stop using NUM_FLAG and use the logic from this macro instead.

sureemo:
Thank you for comments. I have done the new fix.

`GROUP_FLAG` and `NUM_FLAG`. have the same value `32768`

Reference to https://www.php.net/manual/en/mysqli-result.fetch-fields.php

libraries/classes/FieldMetadata.php
```diff
@@ -233,7 +234,8 @@ public function __construct(int $fieldType, int $fieldFlags, object $field)
             $this->isNotNull = (bool) ($fieldFlags & MYSQLI_NOT_NULL_FLAG);
             $this->isUnsigned = (bool) ($fieldFlags & MYSQLI_UNSIGNED_FLAG);
             $this->isZerofill = (bool) ($fieldFlags & MYSQLI_ZEROFILL_FLAG);
-            $this->isNumeric = (bool) ($fieldFlags & MYSQLI_NUM_FLAG);
+            $this->isNumeric = (bool) (($fieldFlags & MYSQLI_NUM_FLAG)
+                && $fieldType !== self::TYPE_VARCHAR);
```
williamdes, line 238: That's a very good fix, maybe we could use `$this->isType`

williamdes (approved):
Awesome, ~~I need to find out why this did occur~~
If you find our before me feel free to document this in the code line you changed with a comment

williamdes:
> 

@kamil-tekiela why is that?
How can we differentiate between int and group without changing the SQL query?

kamil-tekiela:
> why is that?

Ahh yes, well it's because... it's... hmmmm... uhhh... well it's because... next question, please.

Honestly, MySQL metadata is so poorly designed... but it's not so easy to modernize it even if somebody wanted to. We can't fix this in PHP because we get garbage information from MySQL API. People who originally designed mysqlnd have struggled with proper behaviour, and since then, nothing changed. 

So, I guess we can't rely on `flags` giving accurate information. At least not in regards to field being numeric. Let's follow what MySQL API does with its macro or what mysqlnd does: https://github.com/php/php-src/blob/master/ext/mysqlnd/mysqlnd_wireprotocol.c#L1248

libraries/classes/Util.php
```diff
@@ -911,7 +911,7 @@ private static function getConditionValue(
                 . self::printableBitValue((int) $row, (int) $meta->length) . "'";
         } else {
             $conditionValue = '= \''
-                . $dbi->escapeString($row) . '\'';
+                . $dbi->escapeString((string) $row) . '\'';
```
kamil-tekiela, line 914: Casts are potentially dangerous and we shouldn't be using them unless strictly necessary. I think we should not use it here.

libraries/classes/FieldMetadata.php
```diff
@@ -233,7 +233,13 @@ public function __construct(int $fieldType, int $fieldFlags, object $field)
...
             $this->isUnsigned = (bool) ($fieldFlags & MYSQLI_UNSIGNED_FLAG);
             $this->isZerofill = (bool) ($fieldFlags & MYSQLI_ZEROFILL_FLAG);
-            $this->isNumeric = (bool) ($fieldFlags & MYSQLI_NUM_FLAG);
+
+            // as flags 32768 can be NUM_FLAG or GROUP_FLAG
+            // reference: https://www.php.net/manual/en/mysqli-result.fetch-fields.php
+            // so also check field type that not 253 (VARCHAR)
+            $this->isNumeric = (bool) (($fieldFlags & MYSQLI_NUM_FLAG)
+                && $this->isType(self::TYPE_INT));
+
```
kamil-tekiela, line 242:
```suggestion
            $this->isNumeric = $this->isType(self::TYPE_INT));
```

This is "safe" definition of numeric because all integers are numeric by definition. `YEAR` type is zero-padded so I wouldn't consider it numeric. DECIMAL should be treated as a string and it seems phpMyAdmin makes no distinction between DECIMAL and FLOAT, which is probably for the best. N.B. the result of `SUM` or `AVG` is always DECIMAL

libraries/classes/Util.php
sureemo: Fixed this line for unit test as it's found possible value not match the signature.

Screen Shot 2565-10-09 at 05 01 59

kamil-tekiela:
As we change the definition of numeric, the check on this line (and the comment above it) becomes too much:

https://github.com/phpmyadmin/phpmyadmin/blob/4427470529d148a4448dcbbb0102decb2870f1ad/libraries/classes/Util.php#L882

libraries/classes/Util.php
kamil-tekiela: Ok, sorry. Disregard my comment then.

kamil-tekiela:
Related https://github.com/phpmyadmin/phpmyadmin/issues/11406

test/classes/FieldMetadataTest.php
```diff
@@ -100,7 +100,7 @@ public function testIsNumeric(): void
         $this->assertFalse($fm->isNotNull());
         $this->assertFalse($fm->isPrimaryKey());
         $this->assertFalse($fm->isMultipleKey());
-        $this->assertTrue($fm->isNumeric());
+        $this->assertFalse($fm->isNumeric());
```
kamil-tekiela, line 103: Please set the type to one of the numeric types instead.

test/classes/UtilTest.php
```diff
@@ -259,6 +261,73 @@ public function testGetUniqueConditionWithUniqueKey(): void
...
+     *
+     * @return array<string, array<int, array<int, array<string, string>|bool|int|string>>>
+     */
+    public function providerGetUniqueConditionForGroupFlag(): array
+    {
+        return [
+            'value is number' => [
+                [123],
+                [
+                    "`table`.`col` = '123'",
```
williamdes, line 299: It should not be quoted for a number, right?

libraries/classes/Util.php
```diff
@@ -875,11 +875,7 @@ private static function getConditionValue(
         $isBinaryString = $meta->isType(FieldMetadata::TYPE_STRING) && $meta->isBinary();
         // 63 is the binary charset, see: https://dev.mysql.com/doc/internals/en/charsets.html
         $isBlobAndIsBinaryCharset = $meta->isType(FieldMetadata::TYPE_BLOB) && $meta->charsetnr === 63;
-        // timestamp is numeric on some MySQL 4.1
-        // for real we use CONCAT above and it should compare to string
-        // See commit: 049fc7fef7548c2ba603196937c6dcaf9ff9bf00
-        // See bug: https://sourceforge.net/p/phpmyadmin/bugs/3064/
-        if ($meta->isNumeric && ! $meta->isMappedTypeTimestamp && $meta->isNotType(FieldMetadata::TYPE_REAL)) {
+        if ($meta->isNumeric) {
```
williamdes, line 878: Just to be sure this code was dropped because it was old compact for a version we do not support anymore?
If it is added back do the tests still pass?

williamdes (approved):
This is incredible work

test/classes/UtilTest.php
sureemo: generally it's depend on `field type` if field type is `varchar` but value is `integer`, it will treat as string.
anyway it's rare case so i've improve the test to avoid confusion. thank you for comment ^_^

libraries/classes/Util.php
sureemo: this change is suggested by @kamil-tekiela and i agree because it's no longer necessary to check `timestamp` or `real` because the `isNumeric` is now specific integer only (no longer include timestamp or real)

MauricioFauth (approved):
Great work!

MauricioFauth:
Merged, thank you for your contribution!

PR #17759, opened 2022-10-08
Remove redundant tests

kamil-tekiela:
Typecasting to object is the same as creating the object with `new`. The test just repeats the previous test. 

In the whole codebase, we never create `FieldMetadata` recursively, so that test is also unnecessary. I assume the test was just to verify that `FieldMetadata` exposes the raw properties, but IMHO this isn't needed. 

Unrelated, but what should we do about the signature of `FieldMetadata::__construct()`? `type` and `flags` are their own properties while the rest is accepted in the form of an object.

### DIFF
--- a/test/classes/FieldMetadataTest.php
+++ b/test/classes/FieldMetadataTest.php
@@ -34,46 +34,11 @@ class FieldMetadataTest extends AbstractTestCase
         $this->assertFalse($fm->isBlob());
     }
 
-    public function testIsBinaryStdClassAsObject(): void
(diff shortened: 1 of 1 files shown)


MauricioFauth (approved):

MauricioFauth:
> Unrelated, but what should we do about the signature of FieldMetadata::__construct()? type and flags are their own properties while the rest is accepted in the form of an object.

Maybe it's better to have only one parameter as object? What do you think, @williamdes?

PR #17761, opened 2022-10-09
Fix indentation

kamil-tekiela:


MauricioFauth (approved):
Thank you. I was expecting this kind of issue to be reported by PHPCS. Looks like I was wrong.

Issue #16896, opened 2021-05-11
Distinct URLs are broken on nullable text fields


williamdes:
### Describe the bug

Distinct URLs are broken on text fields because they are detected as numeric

### To Reproduce

```sql
CREATE TABLE `t1` (
  `id` int(11) NOT NULL,
  `text` text CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_ai_ci,
  `text2` text NOT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;

INSERT INTO `t1` (`id`, `text`, `text2`) VALUES
(1, 'zz', 'd'),
(2, 'bb', 'f');
```

1. Go to table structure
2. Click on distinct values for **nullable** column
3. The link is broken

### Expected behavior

Working links

### Server configuration

- phpMyAdmin version: 5.1

williamdes:
For information: the code calls `getDataCellForNumericColumns` for this column

williamdes:
```diff
diff --git a/libraries/classes/Display/Results.php b/libraries/classes/Display/Results.php
index 6ff0992919..fcfee0d0d5 100644
--- a/libraries/classes/Display/Results.php
+++ b/libraries/classes/Display/Results.php
@@ -2950,7 +2950,9 @@ class Results
             // even for a string type
             // for decimal numeric is returning 1
             // have to improve logic
-            if (($meta->numeric == 1 && $meta->type !== 'string') || $meta->type === 'real') {
+            // Nullable text fields and text fields have the blob flag
+            $isNumericAndNotBlob = $meta->numeric === 1 && $meta->blob === 0;
+            if (($isNumericAndNotBlob && $meta->type !== 'string') || $meta->type === 'real') {
                 // n u m e r i c
 
                 $display_params['data'][$row_no][$i]
```

here is the patch, to be tested
real blob fields do not work for distinct values

PR #17762, opened 2022-10-09
Fix isNumeric

kamil-tekiela:
Follow-up from #17758

I looked into it more. MySQL doesn't require quotes for integers, floats or decimals. phpMyAdmin already abstracts the MySQL types with two: TYPE_INT and TYPE_REAL. There's already a method called `isNumericType()` so why not simplify it all and get rid of `isNumeric` completely?

### DIFF
--- a/test/classes/Controllers/Export/ExportControllerTest.php
+++ b/test/classes/Controllers/Export/ExportControllerTest.php
@@ -14,6 +14,7 @@ use PhpMyAdmin\Tests\Stubs\ResponseRenderer;
 
 use function htmlspecialchars;
 
+use const ENT_COMPAT;
 use const MYSQLI_NUM_FLAG;
 use const MYSQLI_PRI_KEY_FLAG;
 use const MYSQLI_TYPE_DATETIME;
@@ -168,6 +169,6 @@ SQL;
         $exportController = new ExportController(new ResponseRenderer(), new Template(), new Export($this->dbi));
         $exportController($request);
         $output = $this->getActualOutputForAssertion();
-        $this->assertStringContainsString(htmlspecialchars($expectedOutput), $output);
+        $this->assertStringContainsString(htmlspecialchars($expectedOutput, ENT_COMPAT), $output);
     }
 }
--- a/libraries/classes/Display/Results.php
+++ b/libraries/classes/Display/Results.php
@@ -2682,16 +2682,7 @@ class Results
 
             $displayParams = $this->properties['display_params'] ?? [];
 
-            // in some situations (issue 11406), numeric returns 1
-            // even for a string type
-            // for decimal numeric is returning 1
-            // have to improve logic
-            // Nullable text fields and text fields have the blob flag (issue 16896)
-            $isNumericAndNotBlob = $meta->isNumeric && ! $meta->isBlob;
-            if (
-                ($isNumericAndNotBlob && $meta->isNotType(FieldMetadata::TYPE_STRING))
-                || $meta->isType(FieldMetadata::TYPE_REAL)
-            ) {
+            if ($meta->isNumeric) {
                 // n u m e r i c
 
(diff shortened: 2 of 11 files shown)


test/classes/Controllers/Export/ExportControllerTest.php
```diff
@@ -168,6 +168,6 @@ public function testExportController(): void
         $exportController = new ExportController(new ResponseRenderer(), new Template(), new Export($this->dbi));
         $exportController($request);
         $output = $this->getActualOutputForAssertion();
-        $this->assertStringContainsString(htmlspecialchars($expectedOutput), $output);
+        $this->assertStringContainsString(htmlspecialchars($expectedOutput, ENT_QUOTES), $output);
```
williamdes, line 171: only quotes ?
is this not displayed in a text area ?

williamdes:
> I looked into it more. MySQL doesn't require quotes for integers, floats or decimals. phpMyAdmin already abstracts the MySQL types with two: TYPE_INT and TYPE_REAL. There's already a method called isNumericType() so why not simplify it all and get rid of isNumeric completely?

Okay but be aware that in the export where `isNumericType` is used you may considerably change the performance.
Use a pre-evaluated property stored in a property. Calling getters costs a lot. That's why I created the `isXyz` properties

test/classes/Controllers/Export/ExportControllerTest.php
kamil-tekiela: Regardless of where it will be displayed, we should do a thorough review of all usages of `htmlspecialchars` and `htmlentities` in the whole codebase. PHP 8.1 changed the default value. Previously it was `ENT_COMPAT` and now it's `ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401`. 

For this PR, I wanted to make the change as small as possible to ensure that the tests pass. As ENT_QUOTES can be considered a superset of ENT_COMPAT, that is it encodes everything ENT_COMPAT does and more, there should be no backwards incompatible changes. All it changes is that from now on single quotes will also be encoded.

kamil-tekiela:
> Okay but be aware that in the export where isNumericType is used you may considerably change the performance.
Use a pre-evaluated property stored in a property. Calling getters costs a lot. That's why I created the isXyz properties

Thanks for the heads up. I will benchmark the change now.

test/classes/Controllers/Export/ExportControllerTest.php
williamdes: Oh okay, its not great that `ENT_QUOTES` also encodes simple quotes `'`. But I understand the change. Why not `ENT_COMPAT` ?
here htmlentities is just some code to protect html properties I think

> All it changes is that from now on single quotes will also be encoded.

Yeah but I see not good reason to encode them

kamil-tekiela: Makes sense. Changed.

williamdes:
> > Okay but be aware that in the export where isNumericType is used you may considerably change the performance.
> > Use a pre-evaluated property stored in a property. Calling getters costs a lot. That's why I created the isXyz properties
> 
> Thanks for the heads up. I will benchmark the change now.

Let me know :)

I used xhprof: https://github.com/phpmyadmin/phpmyadmin/wiki/How-to-use-xhprof
Having a working xhprof profile file viewer is a bit complicated but I can write a wiki article for it

kamil-tekiela:
@williamdes The performance is much worse with the additional function call. I reverted back to readonly property. But I removed the `isNumericType` method. I did it to reduce confusion as now we would have two methods that produce the same result. I kept the getter for `isNumeric` but it's not used anywhere except two tests. Is it acceptable?

MauricioFauth (approved):

williamdes (approved):
Did you check there is not regressions regarding #16896 ?

kamil-tekiela:
I cannot find any refressions.

PR #17781, opened 2022-10-13
Fix phpstan `return type` errors

ghost:
Hello,

ich checked out the phpmyadmin baseline and fixed some return type errors, for example

```
			message: "#^Method PhpMyAdmin\\\\Config\\\\ServerConfigChecks\\:\\:performConfigChecksServersSetBlowfishSecret\\(\\) return type has no value type specified in iterable type array\\.$#"
			count: 1
			path: libraries/classes/Config/ServerConfigChecks.php
```

This is my first PR in the phpymadmin repo. Please feel free to comment, if i have to change things :)

Greetings

Daniel

#hacktoberfest

PR #17782, opened 2022-10-13
Fix phpstan `return type` errors - add more specific array types

ghost:
Hello,

this PR fixes some phpstan `type errors` from the baseline, for example:

```
			message: "#^Method PhpMyAdmin\\\\Config\\\\ServerConfigChecks\\:\\:performConfigChecksServersSetBlowfishSecret\\(\\) return type has no value type specified in iterable type array\\.$#"
			count: 1
			path: libraries/classes/Config/ServerConfigChecks.php
```

This is my first PR in the phpymadmin repo. Please feel free to comment, if i have to change things :)

Greetings

Daniel

#hacktoberfest

Refs https://github.com/phpmyadmin/phpmyadmin/pull/17781 -> now with signed commits. Sorry :D

Signed-off-by: Daniel Platt <82732185+standan-hulk@users.noreply.github.com>

### DIFF
--- a/examples/signon-script.php
+++ b/examples/signon-script.php
@@ -18,7 +18,7 @@ declare(strict_types=1);
  *
  * @param string $user User name
  *
- * @return array
+ * @return array<int,string>
  */
 function get_login_credentials($user)
 {
--- a/libraries/classes/Config/ServerConfigChecks.php
+++ b/libraries/classes/Config/ServerConfigChecks.php
@@ -240,7 +240,7 @@ class ServerConfigChecks
      * @param bool        $cookieAuthServer  Cookie auth is used
      * @param bool        $blowfishSecretSet Blowfish secret set
      *
-     * @return array
+     * @return array<int, bool|string|null>
      */
     protected function performConfigChecksServersSetBlowfishSecret(
         $blowfishSecret,
--- a/libraries/classes/BrowseForeigners.php
+++ b/libraries/classes/BrowseForeigners.php
@@ -271,7 +271,7 @@ class BrowseForeigners
      *
      * @param string $description the key name's description
      *
-     * @return array the new description and title
+     * @return array<int,string> the new description and title
      */
     private function getDescriptionAndTitle(string $description): array
     {
--- a/libraries/classes/CheckUserPrivileges.php
+++ b/libraries/classes/CheckUserPrivileges.php
@@ -37,7 +37,7 @@ class CheckUserPrivileges
      *
      * @param string $row grant row
      *
-     * @return array
+     * @return array<int,string>
      */
     public function getItemsFromShowGrantsRow(string $row): array
     {


libraries/classes/Config/ServerConfigChecks.php
```diff
@@ -240,7 +240,7 @@ protected function performConfigChecksServers(
      * @param bool        $cookieAuthServer  Cookie auth is used
      * @param bool        $blowfishSecretSet Blowfish secret set
      *
-     * @return array
+     * @return array<int, bool|string|null>
```
standan-hulk, line 243:
```suggestion
     * @return list<bool|string|null>
```

standan-hulk: maybe this is a much more precise way for the return type.

williamdes: list does not exist
But you can use an array shape like you did elsewhere `array{foo: int,...}`

standan-hulk: You can have a look here for the list type: https://phpstan.org/writing-php-code/phpdoc-types#lists

I would prefer the ` @return array<int, bool|string|null>` doctype instead of using a specific declaration, because we do not know the key values of the array. they are just incremental integer values.

kamil-tekiela: @williamdes  I believe a list is something that would be valid in [`array_is_list()`](https://www.php.net/manual/en/function.array-is-list.php)

examples/signon-script.php
```diff
@@ -18,7 +18,7 @@
  *
  * @param string $user User name
  *
- * @return array
+ * @return array<int,string>
  */
 function get_login_credentials($user)
```
kamil-tekiela, line 23: I would recommend setting the proper return type and parameter type while you are at it so that the change has a little bit more meat on the bone. :) But that's just a suggestion.

libraries/classes/Config/ServerConfigChecks.php
kamil-tekiela: Made obsolete by #17792
No offence to anyone, but this was a rubbish method. All it does is set a variable to true if the condition matches. `$blowfishSecret` doesn't need to be returned as it's never modified.

williamdes: > @williamdes I believe a list is something that would be valid in [`array_is_list()`](https://www.php.net/manual/en/function.array-is-list.php)

Thank you, in 8 php years I never used it 😄maybe lists are not as much a thing, more arrays?

MauricioFauth: I think you haven't heard of it because it is PHP 8.1 + only. ^^

standan-hulk: if the PR is obsolete, you may close it :) no problem

kamil-tekiela: Not the whole PR. Only this change. The rest is really good.

MauricioFauth:

Issue #17534, opened 2022-05-18
Index separation border


nunoperalta:
Hey

On 5.2.0, it's hard to differentiate each index, when they have multiple columns.

Before, the separations used to be clearer (more contrast).

Now, it just seems like a big table, like if all was the same index.

![image](https://user-images.githubusercontent.com/487079/169168277-8f0b3fb6-4853-493d-a094-25fc59f51673.png)

williamdes:
Hi @nunoperalta 
Some changes have been done to tables for  #17455 but I confirm that a stripped table would be easier to read

williamdes:
Hi @nunoperalta 
After applying #17534 locally here is what it looks like
![image](https://user-images.githubusercontent.com/7784660/195959443-339267f9-82e2-45cb-ac98-4ee7e6108914.png)
I am not sure if it is right that the second line of the composite index is another color

nunoperalta:
Hey @williamdes 

I use the classic UI, but what I meant when I raised this issue is that before this update, each index/row had a proper border, to clearly separate each index.

Even yesterday, actually, I was confused whether one of the composite columns belongs to which index, because of the lack of borders.

I believe that having different background colors can also help, yes.
The columns of the index probably shouldn't alternate columns, as you said above.

Thank you very much.

williamdes:
> Hey @williamdes
> 
> I use the classic UI, but what I meant when I raised this issue is that before this update, each index/row had a proper border, to clearly separate each index.
> 
> Even yesterday, actually, I was confused whether one of the composite columns belongs to which index, because of the lack of borders.
> 
> I believe that having different background colors can also help, yes. The columns of the index probably shouldn't alternate columns, as you said above.
> 
> Thank you very much.

Thank you for the feedback, that will be a left over to fix then. The row coloring is now fixed

nunoperalta:
> that will be a left over to fix then

If the colors alternate, I think that's good enough.
That's probably how it was before anyway.

Yeah, I confirm that my issue at the moment is that all rows have the same color, and that's what confuses me a lot.
Alternating colors is the fix for this issue I reported :-)

Thank you!

PR #17783, opened 2022-10-13
Table index list

clxswalther:
### Description

Move tbody out of for-loop in table-index list to make rows more distinguishable.
This still looks a little weird, because now there are some big grey blobs, but that is another issue.

Addresses https://github.com/phpmyadmin/phpmyadmin/issues/17534


williamdes:
You probably missed to base your branch on QA_5_2
You can reset hard to QA_5_2 add commits and then force push it
You can re open the pull request, no need for a new one :)

Issue #16276, opened 2020-07-22
Move HTTP request superglobals access to controllers


mauriciofauth:
[Superglobals](https://www.php.net/manual/en/language.variables.superglobals.php) like $_POST, $_GET and $_REQUEST should be moved to controllers.

Related to https://github.com/phpmyadmin/phpmyadmin/issues/15507.

williamdes:
Could you provide an example?

mauriciofauth:
For example, currently we directly access `$_POST` variables when needed. With PSR-7, we can get these variables from the Request object. So it is better to read these superglobals only on controllers and when other methods need these values, we can pass them as parameters.

PR #17795, opened 2022-10-14
Ref #16276 - Move HTTP request superglobals access to controllers

MrMadClown:
# Description

As described in [#16276](https://github.com/phpmyadmin/phpmyadmin/issues/16276)

This PR aims to reduce the usage of SuperGlobals outside of Controllers.

### DIFF
--- a/libraries/classes/Controllers/Server/ReplicationController.php
+++ b/libraries/classes/Controllers/Server/ReplicationController.php
@@ -75,7 +75,15 @@ class ReplicationController extends AbstractController
                 isset($_POST['replica_changeprimary']),
                 isset($_POST['sr_replica_server_control']),
                 $_POST['sr_replica_action'] ?? null,
-                isset($_POST['sr_replica_skip_error'])
+                isset($_POST['sr_replica_skip_error']),
+                isset($_POST['sr_skip_errors_count']) ? (int) $_POST['sr_skip_errors_count'] : 1,
+                $_POST['sr_replica_control_param'] ?? null,
+                [
+                    'username' => $GLOBALS['dbi']->escapeString($_POST['username']),
+                    'pma_pw' => $GLOBALS['dbi']->escapeString($_POST['pma_pw']),
+                    'hostname' => $GLOBALS['dbi']->escapeString($_POST['hostname']),
+                    'port' => (int) $GLOBALS['dbi']->escapeString($_POST['text_port']),
+                ]
             );
         }
 
@@ -85,7 +93,9 @@ class ReplicationController extends AbstractController
             $primaryReplicationHtml = $this->replicationGui->getHtmlForPrimaryReplication(
                 $_POST['primary_connection'] ?? null,
                 $params['repl_clear_scr'],
-                $_POST['primary_add_user'] ?? null
+                $_POST['primary_add_user'] ?? null,
+                $_POST['username'] ?? null,
+                $_POST['hostname'] ?? null
             );
         }
 
@@ -96,7 +106,8 @@ class ReplicationController extends AbstractController
                 $replicaConfigurationHtml = $this->replicationGui->getHtmlForReplicaConfiguration(
                     $_POST['primary_connection'] ?? null,
                     $replicaInfo['status'],
-                    $replicationInfo->getReplicaStatus()
+                    $replicationInfo->getReplicaStatus(),
+                    isset($_POST['replica_configure'])
                 );
             }
 
--- a/libraries/classes/ReplicationGui.php
+++ b/libraries/classes/ReplicationGui.php
@@ -17,7 +17,6 @@ use function mb_strtolower;
 use function mb_substr;
 use function sprintf;
 use function str_replace;
-use function strlen;
 use function strtok;
 use function time;
 
@@ -73,7 +72,9 @@ class ReplicationGui
     public function getHtmlForPrimaryReplication(
         ?string $connection,
         ?bool $replClearScr,
-        ?string $primaryAddUser
+        ?string $primaryAddUser,
+        ?string $username,
+        ?string $hostname
     ): string {
         if ($replClearScr === null) {
             $primaryStatusTable = $this->getHtmlForReplicationStatusTable($connection, 'primary', true, false);
@@ -85,7 +86,7 @@ class ReplicationGui
         }
 
         if ($primaryAddUser !== null) {
-            $primaryAddReplicaUser = $this->getHtmlForReplicationPrimaryAddReplicaUser();
+            $primaryAddReplicaUser = $this->getHtmlForReplicationPrimaryAddReplicaUser($username, $hostname);
         }
 
         return $this->template->render('server/replication/primary_replication', [
@@ -116,16 +117,18 @@ class ReplicationGui
     /**
      * returns HTML for replica replication configuration
      *
-     * @param ?string $connection               Primary connection
-     * @param bool    $serverReplicaStatus      Whether it is Primary or Replica
-     * @param array   $serverReplicaReplication Replica replication
+     * @param string|null $connection               Primary connection
+     * @param bool        $serverReplicaStatus      Whether it is Primary or Replica
+     * @param array       $serverReplicaReplication Replica replication
+     * @param bool        $replicaConfigure         Replica configure
      *
      * @return string HTML code
      */
     public function getHtmlForReplicaConfiguration(
         ?string $connection,
         $serverReplicaStatus,
-        array $serverReplicaReplication
+        array $serverReplicaReplication,
+        bool $replicaConfigure
     ): string {
         $serverReplicaMultiReplication = $GLOBALS['dbi']->fetchResult('SHOW ALL SLAVES STATUS');
         if ($serverReplicaStatus) {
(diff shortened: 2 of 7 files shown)


libraries/classes/UrlRedirector.php
```diff
@@ -30,9 +30,9 @@ public static function redirect(): void
         $response->disable();
 
         if (
-            ! isset($_GET['url']) || ! is_scalar($_GET['url']) || strlen((string) $_GET['url']) === 0
-            || ! preg_match('/^https:\/\/[^\n\r]*$/', (string) $_GET['url'])
-            || ! Core::isAllowedDomain((string) $_GET['url'])
+            ! is_scalar($url) || strlen((string) $url) === 0
```
kamil-tekiela, line 33:
```suggestion
            $url === null || $url === ''
```

libraries/classes/UrlRedirector.php
```diff
@@ -17,7 +17,7 @@ final class UrlRedirector
     /**
      * @psalm-return never
      */
-    public static function redirect(): void
+    public static function redirect(?string $url): void
```
kamil-tekiela, line 20:
```suggestion
    public static function redirect(string $url): void
```
It doesn't need to be nullable. Why not just provide an empty string instead? Then all the `(string)` can be removed and the check for `is_scalar` or `=== null` be removed too.

libraries/classes/ReplicationGui.php
```diff
@@ -356,20 +361,20 @@ public function getUsernameHostnameLength(): array
...
 
-        if (isset($_POST['username']) && strlen($_POST['username']) === 0) {
+        if ($postUsername !== null && strlen($postUsername) === 0) {
             $GLOBALS['pred_username'] = 'any';
         }
 
         $username = '';
-        if (! empty($_POST['username'])) {
-            $username = $GLOBALS['new_username'] ?? $_POST['username'];
+        if (! empty($postUsername)) {
```
kamil-tekiela, line 376:
```suggestion
        if ($postUsername !== null && $postUsername !== '' && $postUsername !== '0') {
```
`empty` is bad because it hides typos. Please use more explicit check. In my experience the check for 0-string can be ignored, but please check the logic to make sure it's redundant.

libraries/classes/ReplicationGui.php
```diff
@@ -356,20 +361,20 @@ public function getUsernameHostnameLength(): array
...
-    public function getHtmlForReplicationPrimaryAddReplicaUser(): string
+    public function getHtmlForReplicationPrimaryAddReplicaUser(?string $postUsername, ?string $hostname): string
     {
         [
             $usernameLength,
             $hostnameLength,
         ] = $this->getUsernameHostnameLength();
 
-        if (isset($_POST['username']) && strlen($_POST['username']) === 0) {
+        if ($postUsername !== null && strlen($postUsername) === 0) {
```
kamil-tekiela, line 371:
```suggestion
        if ($postUsername !== null && $postUsername === '') {
```

kamil-tekiela (changes requested):

williamdes (approved):

libraries/classes/ReplicationGui.php
```diff
@@ -116,16 +119,18 @@ public function getHtmlForPrimaryConfiguration(): string
     /**
      * returns HTML for replica replication configuration
      *
-     * @param ?string $connection               Primary connection
-     * @param bool    $serverReplicaStatus      Whether it is Primary or Replica
-     * @param array   $serverReplicaReplication Replica replication
+     * @param null|string $connection               Primary connection
```
kamil-tekiela, line 122:
```suggestion
     * @param string|null $connection               Primary connection
```

libraries/classes/ReplicationGui.php
```diff
@@ -245,10 +250,10 @@ public function getHtmlForReplicationChangePrimary($submitName): string
     /**
      * This function returns html code for table with replication status.
      *
-     * @param ?string $connection primary connection
-     * @param string  $type       either primary or replica
-     * @param bool    $isHidden   if true, then default style is set to hidden, default value false
-     * @param bool    $hasTitle   if true, then title is displayed, default true
+     * @param null|string $connection primary connection
```
kamil-tekiela, line 253:
```suggestion
     * @param string|null $connection primary connection
```

libraries/classes/ReplicationGui.php
```diff
@@ -356,20 +361,18 @@ public function getUsernameHostnameLength(): array
...
-        if (isset($_POST['username']) && strlen($_POST['username']) === 0) {
-            $GLOBALS['pred_username'] = 'any';
-        }
-
         $username = '';
-        if (! empty($_POST['username'])) {
-            $username = $GLOBALS['new_username'] ?? $_POST['username'];
+        if ($postUsername === '') {
+            $GLOBALS['pred_username'] = 'any';
+        }else if ($postUsername !== null && $postUsername !== '0') {
```
kamil-tekiela, line 374:
```suggestion
        } elseif ($postUsername !== null && $postUsername !== '0') {
```

libraries/classes/ReplicationGui.php
```diff
@@ -73,8 +73,11 @@ public function getHtmlForErrorMessage(): string
     public function getHtmlForPrimaryReplication(
         ?string $connection,
         ?bool $replClearScr,
-        ?string $primaryAddUser
-    ): string {
+        ?string $primaryAddUser,
+        ?string $username,
+        ?string $hostname
+    ): string
+    {
```
kamil-tekiela, line 80:
```suggestion
    ): string {
```

libraries/classes/ReplicationGui.php
```diff
@@ -423,7 +426,10 @@ public function handleControlRequest(
         bool $replicaChangePrimary,
         bool $srReplicaServerControl,
         ?string $srReplicaAction,
-        bool $srReplicaSkipError
+        bool $srReplicaSkipError,
+        $srSkipErrorsCount,
```
kamil-tekiela, line 430: I think the type should be nullable int, but if you are not sure then specifying the type in the phpdoc should be enough.

libraries/classes/ReplicationGui.php
```diff
@@ -585,12 +581,11 @@ public function handleRequestForReplicaServerControl(?string $srReplicaAction):
         return $result;
     }
 
-    public function handleRequestForReplicaSkipError(): bool
+    public function handleRequestForReplicaSkipError($srSkipErrorsCount): bool
```
kamil-tekiela, line 584: Same here.

kamil-tekiela (changes requested):

libraries/classes/ReplicationGui.php
MrMadClown: Putting it in the phpdoc, should be a nullable int but the `* 1`  part threw me off, looks like it could also be a string

kamil-tekiela: then maybe a cast to int is in order in the controller...

libraries/classes/Controllers/Server/ReplicationController.php
```diff
@@ -75,7 +75,15 @@ public function __invoke(ServerRequest $request): void
                 isset($_POST['replica_changeprimary']),
                 isset($_POST['sr_replica_server_control']),
                 $_POST['sr_replica_action'] ?? null,
-                isset($_POST['sr_replica_skip_error'])
+                isset($_POST['sr_replica_skip_error']),
+                isset($_POST['sr_skip_errors_count']) ? (int)$_POST['sr_skip_errors_count'] : null,
```
kamil-tekiela, line 79: Even better, use `1` as default then we don't need nullable type nor the ternary inside that method. Much simpler :).

kamil-tekiela:
Please also fix the issues from coding standard check https://github.com/phpmyadmin/phpmyadmin/actions/runs/3255197954/jobs/5344264915 and after that I think the PR will be ready to merge. Thanks.

williamdes (approved):

MrMadClown:
Why does the DCO check fail? is the `Signed-off-by...` not correct?

kamil-tekiela:
The signoff should be in the commit body, not in the commit message. i.e. use two newlines to separate commit message from signoff.

williamdes:
> The signoff should be in the commit body, not in the commit message. i.e. use two newlines to separate commit message from signoff.

You can squash your commits and add the signoff :)

kamil-tekiela:
@williamdes  The DCO check still failed. Do you know why?

MrMadClown:
@kamil-tekiela my default git config is work related. The signed-off-by didn't match.

kamil-tekiela (approved):
Good job!

williamdes:
> @williamdes The DCO check still failed. Do you know why?

Sometimes it expects some email but honestly we do not care about the email, just have some working one. The important part is having a real name before the email

MrMadClown:
But now it's fine right?

MrMadClown:
What's needed to get this merged? @williamdes Anything I can do?

williamdes:
> What's needed to get this merged? @williamdes Anything I can do?

I asked for another review if it passes it should get merged right after

MauricioFauth (approved):

MauricioFauth:

PR #17797, opened 2022-10-15
Fix #17793 - insert UUID

### LINKED ISSUE #17793: Inserting a row with UUID should insert the user value when not empty
By default, when inserting a row, the uuid type column has UUID as function set by default. I think it should be empty, because when the column is set to NULL, it still inserts a uuid, instead of NULL.

The same case applies when a user enters his own UUID, it will still insert the one generated by MariaDB and not what user has entered.

What can be done, is to have UUID (uuid() function) added to the Default select (if MariaDB and column is uuid), like it is with CURRENT_TIMESTAMP.

https://user-images.githubusercontent.com/25424343/195730197-5f711440-2ed0-4e2a-af7d-a0a16531db80.mp4

sureemo:
### Description

- Add/Change table schema - Add UUID in default options
  Screen Shot 2565-10-15 at 04 13 22
- Insert/Edit - Generate uuid only when not defined value and not check null option

Reference issue: https://github.com/phpmyadmin/phpmyadmin/issues/17793

### DIFF
--- a/libraries/classes/Html/Generator.php
+++ b/libraries/classes/Html/Generator.php
@@ -311,6 +311,11 @@ class Generator
             $defaultFunction = $cfg['DefaultFunctions']['first_timestamp'];
         }
 
+        // For uuid field, no default function
+        if ($field['True_Type'] === 'uuid') {
+            $defaultFunction = '';
+        }
+
         // For primary keys of type char(36) or varchar(36) UUID if the default
         // function
         // Only applies to insert mode, as it would silently trash data on updates.
--- a/libraries/classes/Html/Generator.php
+++ b/libraries/classes/Html/Generator.php
@@ -313,7 +313,7 @@ class Generator
 
         // For uuid field, no default function
         if ($field['True_Type'] === 'uuid') {
-            $defaultFunction = '';
+            return '';
         }
 
         // For primary keys of type char(36) or varchar(36) UUID if the default
--- a/libraries/classes/InsertEdit.php
+++ b/libraries/classes/InsertEdit.php
@@ -870,7 +870,7 @@ class InsertEdit
                     . $columnNameAppendix . '" value="' . $type . '">';
             }
 
-            if (in_array($column['True_Type'], ['bit', 'uuid'])) {
+            if (in_array($column['True_Type'], ['bit', 'uuid'], true)) {
                 $htmlOutput .= '<input type="hidden" name="fields_type'
                     . $columnNameAppendix . '" value="' . $column['True_Type'] . '">';
             }
--- a/libraries/classes/InsertEdit.php
+++ b/libraries/classes/InsertEdit.php
@@ -870,9 +870,9 @@ class InsertEdit
                     . $columnNameAppendix . '" value="' . $type . '">';
             }
 
-            if ($column['True_Type'] === 'bit') {
+            if (in_array($column['True_Type'], ['bit', 'uuid'])) {
                 $htmlOutput .= '<input type="hidden" name="fields_type'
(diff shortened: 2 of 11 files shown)


libraries/classes/InsertEdit.php
```diff
@@ -870,9 +870,9 @@ private function getValueColumnForOtherDatatypes(
                     . $columnNameAppendix . '" value="' . $type . '">';
             }
 
-            if ($column['True_Type'] === 'bit') {
+            if (in_array($column['True_Type'], ['bit', 'uuid'])) {
```
MauricioFauth, line 873:
```suggestion
            if (in_array($column['True_Type'], ['bit', 'uuid'], true)) {
```

libraries/classes/Util.php
```diff
@@ -1756,6 +1757,18 @@ public static function unsupportedDatatypes(): array
...
+    /**
+     * This function is to check whether database support UUID
+     *
+     * @return bool
+     */
+    public static function isUUIDSupported()
+    {
+        global $dbi;
+
+        return Compatibility::isUUIDSupported($dbi);
```
MauricioFauth, line 1769: We're using `$GLOBALS` instead of `global` for new code.
```suggestion
        return Compatibility::isUUIDSupported($GLOBALS['dbi']);
```

MauricioFauth (changes requested):

libraries/classes/InsertEdit.php
sureemo: updated. thank you. ^_^

libraries/classes/Util.php
sureemo: updated. thank you. ^_^

libraries/classes/Html/Generator.php
```diff
@@ -311,6 +311,11 @@ public static function getDefaultFunctionForField(array $field, $insertMode): st
             $defaultFunction = $cfg['DefaultFunctions']['first_timestamp'];
         }
 
+        // For uuid field, no default function
+        if ($field['True_Type'] === 'uuid') {
+            $defaultFunction = '';
```
williamdes, line 316:
```suggestion
            return '';
```

williamdes: Or is there some logic we want to enter below?

libraries/classes/Util.php
```diff
@@ -1756,6 +1757,16 @@ public static function unsupportedDatatypes(): array
         return [];
     }
 
+    /**
+     * This function is to check whether database support UUID
+     *
+     * @return bool
+     */
+    public static function isUUIDSupported()
```
williamdes, line 1765:
```suggestion
    public static function isUUIDSupported(): bool
```
phpdoc return can be removed

williamdes (approved):
It looks good, I will ask for another review
Maybe there is different ways for some changes

libraries/classes/Html/Generator.php
sureemo: updated. thank you ^_^

libraries/classes/Util.php
sureemo: updated. thank you ^_^

MauricioFauth (approved):

MauricioFauth:

test/classes/InsertEditTest.php
```diff
@@ -2472,6 +2542,46 @@ public function testGetCurrentValueForDifferentTypes(): void
         );
 
         $this->assertEquals("''", $result);
+
+        // case 10
```
williamdes, line 2546: this turned out to be very difficult to know what was actually tested while merging into master

PR #17799, opened 2022-10-15
Refactor JS escaping mechanism

kamil-tekiela:
Here's something that bothered me for some time. There is logic for JS escaping that doesn't really need to be there. The usual way to output a variable into JS context is to use `json_encode`. We also have Twig which implements JS string escaping. Why does phpMyAdmin need custom JS escaping logic? 

My solution is to replace all custom JS escaping with either Twig's escaping (for when a string is needed only) or `json_encode` for when an unknown type is being formatted in JS context. 

I admit this was a little risky because the custom escaping wasn't the same. I also don't know if everything is covered by tests. I can't find any regressions though. 

Overall, I think this is an improvement, but some areas should be looked at more in detail at some later point in time. Perhaps, some code could be moved into templates.

### DIFF
--- a/libraries/classes/InsertEdit.php
+++ b/libraries/classes/InsertEdit.php
@@ -28,6 +28,7 @@ use function in_array;
 use function is_array;
 use function is_file;
 use function is_string;
+use function json_encode;
 use function max;
 use function mb_stripos;
 use function mb_strlen;
@@ -595,7 +596,7 @@ class InsertEdit
             . ' cols="' . $textareaCols . '"'
             . ' dir="' . $textDir . '"'
             . ' id="field_' . $idindex . '_3"'
-            . ($onChangeClause ? ' ' . $onChangeClause : '')
+            . ($onChangeClause ? ' onchange="' . htmlspecialchars($onChangeClause, ENT_COMPAT) . '"' : '')
             . ' tabindex="' . ($tabindex + $tabindexForValue) . '"'
             . ' data-type="' . $dataType . '">'
             . $specialCharsEncoded
@@ -715,7 +716,7 @@ class InsertEdit
             . ($readOnly ? ' readonly="readonly"' : '')
             . ($inputMinMax ? ' ' . $inputMinMax : '')
             . ' data-type="' . $dataType . '"'
-            . ' class="' . $theClass . '" ' . $onChangeClause
+            . ' class="' . $theClass . '" onchange="' . htmlspecialchars($onChangeClause, ENT_COMPAT) . '"'
             . ' tabindex="' . ($tabindex + $tabindexForValue) . '"'
             . ($isInteger ? ' inputmode="numeric"' : '')
             . ' id="field_' . $idindex . '_3">';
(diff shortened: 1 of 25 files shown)


libraries/classes/InsertEdit.php
```diff
@@ -2035,10 +2035,9 @@ private function getHtmlForInsertEditFormColumn(
         }
 
         //Call validation when the form submitted...
-        $onChangeClause = $chgEvtHandler
-            . "=\"return verificationsAfterFieldChange('"
-            . Sanitize::escapeJsString($fieldHashMd5) . "', '"
-            . Sanitize::escapeJsString($jsvkey) . "','" . $column['pma_type'] . "')\"";
+        $onChangeClause = 'return verificationsAfterFieldChange('
+            . json_encode($fieldHashMd5, JSON_HEX_TAG) . ', '
```
williamdes, line 2039: I am not sure there is a point using hex tag here
But if it is to standardize maybe we should have an util function that does
Hex tag and JSON_HEX_QUOT

libraries/classes/Plugins/Transformations/Abs/DateFormatTransformationsPlugin.php
```diff
@@ -140,7 +143,7 @@ public function applyTransformation($buffer, array $options = [], ?FieldMetadata
                 $text = 'INVALID DATE TYPE';
             }
 
-            return '<dfn onclick="alert(\'' . Sanitize::jsFormat($source, false) . '\');" title="'
+            return '<dfn onclick="alert(' . htmlspecialchars(json_encode($source, JSON_HEX_TAG), ENT_COMPAT) . ');" title="'
```
williamdes, line 146: Using JSON_HEX_QUOT should remove the need for another function call

templates/javascript/redirect.twig
```diff
@@ -1,5 +1,5 @@
 <script type='text/javascript'>
     window.onload = function () {
-        window.location = '{{ url }}';
+        window.location = '{{ url|e('js') }}';
```
williamdes, line 3: could we use a full word function and not just `e`

libraries/classes/Plugins/Transformations/Abs/DateFormatTransformationsPlugin.php
kamil-tekiela: Removing `JSON_HEX_TAG` feels like a better choice to me, but you are absolutely correct. Just in case, in the future we would want to change this to a template it seems easier to stay with `htmlspecialchars` anyway.

templates/javascript/redirect.twig
kamil-tekiela: We certainly can, but I was trying to stay consistent with the rest of the project. Mauricio used the shorter form elsewhere so I also used it. I will use longer form now.

MauricioFauth (approved):

MauricioFauth:
Merged, thank you for your contribution!

> Why does phpMyAdmin need custom JS escaping logic?

That was the way to escape JS before introducing Twig. No one took the effort to remove this function before.

PR #17805, opened 2022-10-16
Fix static analysis issues

kamil-tekiela:


libraries/classes/Export.php
```diff
@@ -1129,7 +1129,7 @@ public function mergeAliases(array $aliases1, array $aliases2): array
                 $val1 = $db['alias'][0];
                 $val2 = $db['alias'][1];
                 // Use aliases2 alias if non empty
-                $aliases[$dbName]['alias'] = empty($val2) ? $val1 : $val2;
+                $aliases[$dbName]['alias'] = $val2 ?: $val1;
```
williamdes, line 1132: I do not like this syntax, it's not very explicit. Can you use another one please ?

kamil-tekiela: Sure, I will think of something tomorrow.

kamil-tekiela: It seems the values can either be string or null, so I improved the condition.

MauricioFauth (approved):

PR #14076, opened 2018-03-13
Enable columns names by default for CSV in issue #14007 solved

rijaspk:
Enable columns names by default for CSV
Added a export  type check for CSV 
IF true added csv columns 
fixes #14007

Signed-off-by: Rijas Ahammed PK Rijaspk5195@gmail.com

Before submitting pull request, please check that every commit:

- [x] Has proper Signed-Off-By
- [x] Has commit message which describes it
- [ ] Is needed on it's own, if you have just minor fixes to previous commits, you can squash them
- [ ] Any new functionality is covered by tests

Issue #14007, opened 2018-02-14
Enable columns names by default for CSV exports


OlafvdSpek:
The CSV Excel exporter has this enabled but the normal one doesn't.

psaxena130:
It will be appreciated if the problem is explained a bit more. Regards.

ibennetch:
@psaxena130 When exporting, the CSV Excel type enables exporting column names on the first line as a default. For the CSV type, this is not the default.

rijaspk:
I would like to work on this issue and am a newbie.

PR #17814, opened 2022-10-19
Fix column names option for CSV Export

scottmcgowan24:
Previously work was done to enable column names in CSV exports by default. This change actually ended up overwriting the column name value so that column names were included in all exports regardless of whether or not the property was disabled. This commit changes the default for column names to enabled as the original change intended, and removes the property overwrite.

Signed-off-by: Scott McGowan <mcgowanscott16@gmail.com>

### Description

Undoes the change made by https://github.com/phpmyadmin/phpmyadmin/pull/14076 and changes the default value of `csv_columns` to true. Also changes usage of the property to `true`/`false` instead of `"on"`/`"off"` as the value is a boolean. 


libraries/config.default.php
```diff
@@ -1677,7 +1677,7 @@
 /**
  * @global boolean $cfg['Export']['csv_columns']
  */
-$cfg['Export']['csv_columns'] = false;
+$cfg['Export']['csv_columns'] = true;
```
williamdes, line 1680: This does not seem right, we should not change the default config
What do you think @MauricioFauth?

scottmcgowan24: @williamdes I think this behavior was previously intended when https://github.com/phpmyadmin/phpmyadmin/pull/14076 was created to address https://github.com/phpmyadmin/phpmyadmin/issues/14007 but to my understanding it was not done properly.

williamdes: Indeed, it looks like a hack

williamdes: I think the rst documentation should be updated to explain that

scottmcgowan24: @williamdes not only a hack, but it removes the utility of the property entirely. If you attempt to deselect `Put columns names in the first row` for exporting a table as a CSV it adds the column names anyways since the value is always overwritten.

scottmcgowan24: @williamdes If you point me to the rst docs I can update as needed

scottmcgowan24:
Any update on this? @MauricioFauth can you confirm that this is a desired fix?

williamdes (approved):
~~I will get the documentation done~~ and merge this right now

Edit: this is not documented 😢

PR #17825, opened 2022-10-21
Improve `index.php` and `setup/index.php`

MauricioFauth:
Simplifies the `index.php` and `setup/index.php` entry points with some refactoring to the `Common` class.


setup/index.php
```diff
@@ -3,41 +3,32 @@
...
+    );
 }
 
-Core::fatalError(sprintf(
-    __('Error 404! The page %s was not found.'),
-    '[code]' . htmlspecialchars($route) . '[/code]'
-));
+require AUTOLOAD_FILE;
+
+chdir(ROOT_PATH);
```
williamdes, line 32: I remember working on #16187 
Is changing dir still needed ?

williamdes (approved):
Neat !

setup/index.php
MauricioFauth: You're right. This is not needed anymore.

williamdes (approved):
🚢

Issue #17727, opened 2022-09-17
`DatabaseName` and `TableName` should be used for `$db` and `$table`


MauricioFauth:
The `PhpMyAdmin\Dbal\DatabaseName` and `PhpMyAdmin\Dbal\TableName` classes should be used as a type for the database and table names instead of using plain strings.

```diff
- public function foo(string $db, string $table)
+ public function foo(DatabaseName $db, TableName $table)
```

Nithin1728:
Hi, I'm new to open source, Can you assign this issue to me.

williamdes:
> Hi, I'm new to open source, Can you assign this issue to me.

Feel free to contribute on this one when you want and make pull-requests to the `master` branch :)

Adams-Ijachi:
Hello also new to open source , i can find a doc on how i would start contributing to this sorry if this sounds dumb

williamdes:
> Hello also new to open source , i can find a doc on how i would start contributing to this sorry if this sounds dumb

That's a really good question, I keep this on my todo list until now
We should have a wiki page to help people contribute

Adams-Ijachi:
> > Hello also new to open source , i can find a doc on how i would start contributing to this sorry if this sounds dumb
> 
> That's a really good question, I keep this on my todo list until now We should have a wiki page to help people contribute

That would be nice let me know how I can help

williamdes:
> > > Hello also new to open source , i can find a doc on how i would start contributing to this sorry if this sounds dumb
> > 
> > 
> > That's a really good question, I keep this on my todo list until now We should have a wiki page to help people contribute
> 
> That would be nice let me know how I can help

Basically if you have a web server, node, yarn and composer installed you are ready to code
Clone the repository, run:
```
yarn install
composer update
```
And create a config.inc.php file

You will also need a MariaDB or MySQL server running somewhere (or locally) that the workstation can access

Once you access the cloned folder from the web server you will see your phpmyadmin

sivadasrajan:
Is help still needed?. I'm tryin to do some contribution as part of hacktober fest

SubaruSama:
Hello. I am reading the contents of DatabaseName and TableName to make myself familiar.

As I can see, there is a lot of places that is being used string $db and string $table as args. My doubt is for every update of string $db and string $table to DatabaseName $db and TableName $table I should open:

* individual PRs for every instance;
* try to group by functionalities;
* following the structure of folders.

[...]

PR #17829, opened 2022-10-23
Abstract & CheckRelations Controllers - Refactored

Maria-Papa:
Signed-off-by: Maria Papadopoulou <1maria.papa@gmail.com>

### Description

Please describe your pull request: This is a partial fix of the issue #17727.
Controllers `AbstractController` & `CheckRelationsController` were refactored.

Fixes #

- [X] Any new functionality is covered by tests.

### DIFF
--- a/libraries/classes/Controllers/AbstractController.php
+++ b/libraries/classes/Controllers/AbstractController.php
@@ -59,8 +59,10 @@ abstract class AbstractController
         }
 
         $GLOBALS['is_db'] = false;
-        $db = DatabaseName::fromValue($GLOBALS['db']);
-        if (strlen($db->getName()) > 0) {
+        $db = DatabaseName::tryFromValue($GLOBALS['db']);
+        $dbName = isset($db) ? $db->getName() : '';
+
+        if (!empty($dbName)) {
             $GLOBALS['is_db'] = $GLOBALS['dbi']->selectDb($db->getName());
             // This "Command out of sync" 2014 error may happen, for example
             // after calling a MySQL procedure; at this point we can't select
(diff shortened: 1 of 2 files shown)


williamdes (approved):

libraries/classes/Controllers/AbstractController.php
```diff
@@ -58,8 +59,9 @@ protected function hasDatabase(): bool
         }
 
         $GLOBALS['is_db'] = false;
-        if (strlen($GLOBALS['db']) > 0) {
-            $GLOBALS['is_db'] = $GLOBALS['dbi']->selectDb($GLOBALS['db']);
+        $db = DatabaseName::fromValue($GLOBALS['db']);
+        if (strlen($db->getName()) > 0) {
+            $GLOBALS['is_db'] = $GLOBALS['dbi']->selectDb($db->getName());
```
kamil-tekiela, line 64: That's not right. The `DatabaseName` value object ensures that the database name is not an empty string. It will throw `InvalidDatabaseName` if it's empty. If there's a requirement to handle this then you either should check that it's not empty string before creating the value object or catch the exception and recover from it. 

P.S. This whole method should be refactored as it currently both has side-effects and returns value.

williamdes (changes requested):
Thank you @kamil-tekiela for the reviews, I missed the strlen part

libraries/classes/Controllers/AbstractController.php
```diff
@@ -58,8 +59,11 @@ protected function hasDatabase(): bool
         }
 
         $GLOBALS['is_db'] = false;
-        if (strlen($GLOBALS['db']) > 0) {
-            $GLOBALS['is_db'] = $GLOBALS['dbi']->selectDb($GLOBALS['db']);
+        $db = DatabaseName::tryFromValue($GLOBALS['db']);
+        $dbName = isset($db) ? $db->getName() : '';
+
+        if (!empty($dbName)) {
```
kamil-tekiela, line 65: I'm sorry but that's worse than what it was before. You need to either check if the `$GLOBALS['db']` is not en empty string before calling `DatabaseName::tryFromValue` or use try-catch.

PR #17837, opened 2022-10-26
Replace `PhpMyAdmin\Core::fatalError()` method calls

MauricioFauth:
The `PhpMyAdmin\Core::fatalError()` method terminates the execution with `exit` when called.
This PR replaces the calls to `fatalError` with other options, like throwing an exception for example.

### DIFF
--- a/libraries/classes/Common.php
+++ b/libraries/classes/Common.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace PhpMyAdmin;
 
+use Exception;
 use PhpMyAdmin\Config\ConfigFile;
 use PhpMyAdmin\ConfigStorage\Relation;
 use PhpMyAdmin\Dbal\DatabaseName;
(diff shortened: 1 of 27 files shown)


libraries/classes/Common.php
```diff
@@ -225,13 +239,17 @@ public static function run(bool $isSetupPage = false): void
...
+                    'dir' => $GLOBALS['text_dir'] ?? 'ltr',
+                    'error_message' => sprintf(
+                        __('You should upgrade to %s %s or later.'),
                         'MySQL',
-                        $GLOBALS['cfg']['MysqlMinVersion']['human'],
-                    ]
-                );
+                        (string) $GLOBALS['cfg']['MysqlMinVersion']['human']
+                    ),
+                ]);
```
williamdes, line 250: the render pattern is duplicated, maybe a method to render an error would be better ?

libraries/classes/Common.php
```diff
@@ -535,11 +551,9 @@ private static function checkServerConfiguration(): void
             return;
         }
 
-        Core::fatalError(
-            __(
-                'The ini_get and/or ini_set functions are disabled in php.ini. phpMyAdmin requires these functions!'
-            )
-        );
+        throw new Exception(__(
```
williamdes, line 554: could we have an internal exception class to avoid catching the entire world afterwards ?

libraries/classes/Common.php
MauricioFauth: That's true. However I'll handle it later when introducing the PSR-7 Response class. Something like an ErrorResponse class.

PR #17844, opened 2022-10-29
Fix #16042 - gzip encoding in Firefox for 5.2.x

### LINKED ISSUE #16042: Exported gzip file of database has first ~73 kB uncompressed and rest is gzip compressed
### Describe the bug

I'm exporting 280 MB database with Export to file with gzip compression.
Resulting file `database.sql.gz` has first 73 kB written in plain SQL and then the rest of file is gzip compressed.

### To Reproduce

1. Export database to file with gzip compression checked.
2. Leave all other export settings default

### Expected behavior

Resulting export file should be fully gzip compresed.

### Server configuration

- Operating system: Debian 8
- Web server: nginx/1.6.2 
- Database version: 8.0.13 - MySQL Community Server - GPL 
- PHP version: 7.4.4 
- phpMyAdmin version: 4.9.5

### Client configuration

- Browser: Firefox
- Operating system: Win10

kamil-tekiela:
Fixes: #16042

### DIFF
--- a/libraries/classes/Core.php
+++ b/libraries/classes/Core.php
@@ -42,7 +42,6 @@ use function preg_match;
 use function preg_replace;
 use function session_write_close;
 use function sprintf;
-use function str_contains;
 use function str_replace;
 use function strlen;
 use function strpos;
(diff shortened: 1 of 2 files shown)


libraries/classes/Core.php
```diff
@@ -488,24 +487,10 @@ public static function downloadHeader(
...
-        } else {
-            // The default output in PMA uses gzip,
-            // so if we want to output uncompressed file, we should reset the encoding.
-            // See PHP bug https://github.com/php/php-src/issues/8218
-            header_remove('Content-Encoding');
-        }
+        // The default output in PMA uses gzip,
+        // so if we want to output uncompressed file, we should reset the encoding.
+        // See PHP bug https://github.com/php/php-src/issues/8218
+        header_remove('Content-Encoding');
```
williamdes, line 493: This was only fixed in php 8.0.19+ and other 8.1.x+...
Can we make the workaround conditional?

kamil-tekiela: I don't think that's necessary. If the header is there, we want to remove it, if it's not there we have nothing to remove. As far as I know `header_remove` doesn't generate any warnings when the header you want to remove is not there. IMHO it can be completely unconditional.

williamdes (approved):
💯

MauricioFauth (approved):

Issue #17683, opened 2022-08-12
Count subquery in Sql::countQueryResults fails on JOIN of tables with same column names


Loopia:
### Describe the bug
Otherwise valid join query will cause pagination not to render, wrong number of total rows to be shown, and `#1060 - Duplicate column name X` error to be shown at the bottom of the page if some of the tables have columns with the same name.

The problem is caused by a query in  PhpMyAdmin\Sql::countQueryResults.
For example, for query in demo.phpmyadmin.net db world `SELECT * FROM City join Country on City.CountryCode=Country.Code`  `PhpMyAdmin\Sql::countQueryResults` will run query  ` SELECT COUNT(*) FROM (SELECT * FROM City join Country on City.CountryCode=Country.Code) as cnt `. This mysql query will fail with message `Error: #1060 Duplicate column name 'Name' ` because City.Name and Country.Name exist.
The problem will only happen if the total number of rows selected is larger than the current "Number of rows" selected in Query box.

Stack trace from local install running release 5.20
```
SELECT COUNT(*) FROM ( ...) as cnt
#1060 - Duplicate column name column_name;
PhpMyAdmin\DatabaseInterface->tryQuery(...)./libraries/classes/DatabaseInterface.php#1229
PhpMyAdmin\DatabaseInterface->fetchValue(...)./libraries/classes/Sql.php#757
PhpMyAdmin\Sql->countQueryResults(...)./libraries/classes/Sql.php#855
PhpMyAdmin\Sql->executeTheQuery(...)./libraries/classes/Sql.php#1718
PhpMyAdmin\Sql->executeQueryAndGetQueryResponse(...)./libraries/classes/Sql.php#1609
PhpMyAdmin\Sql->executeQueryAndSendQueryResponse(...)./libraries/classes/Controllers/Sql/SqlController.php#211
PhpMyAdmin\Controllers\Sql\SqlController->__invoke(...)./libraries/classes/Routing.php#192
PhpMyAdmin\Routing::callControllerForRoute(...)./index.php#43
```

### To Reproduce

1. Go to https://demo.phpmyadmin.net/master-config/index.php?route=/database/sql&db=world
2. Run query `SELECT * FROM City join Country on City.CountryCode=Country.Code`
3. Scroll down to the bottom of the page
4. See error Error:  1060 Duplicate column name 'Name' 
5. Pagination is missing

### Expected behavior
 - Pagination
 - Correct count of total rows

### Server configuration
- Any

[...]

PR #17851, opened 2022-10-31
Fix #17028, #17683 - Re build query count management

### LINKED ISSUE #17028: This solution is not a direct replacement of found_rows() as one would expect. It generated random results even on such queries:
This solution is not a direct replacement of found_rows() as one would expect. It generated random results even on such queries:

 Showing rows 0 - 49 (164056 total, 0 in query, Query took 0.1498 seconds.)
```select * from game_auth_logs l join ( select al.user_id, max(al.id) as id from game_auth_logs al where al.successfull = 1 group by al.user_id ) last_log on last_log.id = l.id;```

Showing rows 0 - 24 (267 total, Query took 0.1533 seconds.)
```select l.* from game_auth_logs l join ( select al.user_id, max(al.id) as id from game_auth_logs al where al.successfull = 1 group by al.user_id ) last_log on last_log.id = l.id;```

Notice that only change was star or alias select.

I think it should be configurable to allow end user to decide how total rows is count.

Mysql as a replacement proposes running two queries:

```
The SQL_CALC_FOUND_ROWS query modifier and accompanying FOUND_ROWS() function are deprecated as of MySQL 8.0.17; expect them to be removed in a future version of MySQL. As a replacement, considering executing your query with LIMIT, and then a second query with COUNT(*) and without LIMIT to determine whether there are additional rows. For example, instead of these queries:
```

_Originally posted by @fliespl in https://github.com/phpmyadmin/phpmyadmin/issues/16470#issuecomment-884844311_

Version: 5.1.1

williamdes:
### Description

This should fix #17028

Fixes: #17028
Fixes: #17683

It currently lacks test cases for ORDER BY and LIMIT

I think this should not get into 5.2.1 but in 5.2.**2** because it could have bad side effects

### DIFF
--- a/libraries/classes/Sql.php
+++ b/libraries/classes/Sql.php
@@ -13,6 +13,7 @@ use PhpMyAdmin\Html\Generator;
 use PhpMyAdmin\Html\MySQLDocumentation;
 use PhpMyAdmin\Query\Generator as QueryGenerator;
 use PhpMyAdmin\Query\Utilities;
+use PhpMyAdmin\SqlParser\Components\Expression;
 use PhpMyAdmin\SqlParser\Statements\AlterStatement;
 use PhpMyAdmin\SqlParser\Statements\DropStatement;
 use PhpMyAdmin\SqlParser\Statements\SelectStatement;
@@ -745,6 +746,11 @@ class Sql
                 // Removes LIMIT clause that might have been added
                 if ($analyzedSqlResults['limit'] !== null) {
                     $statement->limit = null;
+                }
+
+                if ($analyzedSqlResults['is_group'] === false && count($statement->expr) === 1) {
+                    $statement->expr[0] = new Expression();
+                    $statement->expr[0]->expr = '1';
                 }
 
                 $countQuery = 'SELECT COUNT(*) FROM (' . $statement->build() . ' ) as cnt';
--- a/libraries/classes/Sql.php
+++ b/libraries/classes/Sql.php
@@ -736,24 +736,19 @@ class Sql
                 }
             } else {
                 $statement = $analyzedSqlResults['statement'];
-                $tokenList = $analyzedSqlResults['parser']->list;
-                $replaces = [
-                    // Remove ORDER BY to decrease unnecessary sorting time
-                    [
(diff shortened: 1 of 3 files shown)


MauricioFauth (approved):

fliespl:
Shouldn't test dataset actually be included for the test to achieve a goal?

Currently I don't see how it can decide if row count is correct.

fliespl:
Thanks for explanation then :) I was under impression there is SQL dataset that actually checks it with real database.


 Fix works a charm!

williamdes:
> Thanks for explanation then :) I was under impression there is SQL dataset that actually checks it with real database.
> 
> Fix works a charm!

Perfect, I think this will be sent into 5.2.2 because it may have some aspects we do not know for now
I prefer to not distribute a version that breaks user workflows. We will have some nice testing time between 5.2.1 and 5.2.2 to find bugs in the implementation (or the parser that rebuilds the query in my new implementation)

fliespl:
That's fine by me. 

We are not using stable versions verbatim internally, but building custom versions based on stable releases either way (with different patches applied & some custom coding stuff). We have already incorporated before my proposed solution, but will switch to your suggested one :)

kamil-tekiela (approved):

fliespl:
A few things we do:
- multi-user support with permissions from oauth2 (while database info is stored internally),
- per user (from oauth2) history - without different mysql accounts,
- file based mysql history for audit with time, user, and additional details,
- hiding different aspects,
- coloring based on environment used (red on highly critical systems) :)

It's more of internal stuff rather than improvements that would suit for community.

williamdes:
> A few things we do:
> 
> * multi-user support with permissions from oauth2 (while database info is stored internally),
> * per user (from oauth2) history - without different mysql accounts,
> * file based mysql history for audit with time, user, and additional details,
> * hiding different aspects,
> * coloring based on environment used (red on highly critical systems) :)
> 
> It's more of internal stuff rather than improvements that would suit for community.

Very cool !

> file based mysql history for audit with time, user, and additional details

This one could be nice to have here. I think the current Tracking solution is not that awesome

Issue #17849, opened 2022-10-31
TypeError: Argument 1 passed to PhpMyAdmin\DatabaseInterface::escapeString() must be of the type string, null given


liviuconcioiu:
### Describe the bug

TypeError: Argument 1 passed to PhpMyAdmin\DatabaseInterface::escapeString() must be of the type string, null given

Only master is affected.

### To Reproduce

1. Go to [https://demo.phpmyadmin.net/master/index.php?route=/server/replication](https://demo.phpmyadmin.net/master/index.php?route=/server/replication)
3. See error

### Expected behavior

No error should happen

### Screenshots

![1](https://user-images.githubusercontent.com/25424343/198986332-e9e5ec3a-7d15-4183-82dd-547f7b906568.png)

![2](https://user-images.githubusercontent.com/25424343/198986356-edb1c7e2-7d8f-444f-b653-99b514b84f96.png)

### Server configuration

- phpMyAdmin version:  5.3.0-dev

williamdes:
Broken by #17819

PR #17856, opened 2022-11-01
restrict request body parameter parsing to post requests on /server/replication page

DAcodedBEAT:
### Description

the `/server/replication` page was not loading because the same controller function is used on `GET` and `POST` requests.

This change now only checks if the `POST` parameters are provided only when a POST request is received, which alleviates the type mismatch in the function signature.

Fixes #17849

Before submitting pull request, please review the following checklist:

- [x] Make sure you have read our [CONTRIBUTING.md](https://github.com/phpmyadmin/phpmyadmin/blob/master/CONTRIBUTING.md) document.
- [x] Make sure you are making a pull request against the correct branch. For example, for bug fixes in a released version use the corresponding QA branch and for new features use the _master_ branch. If you have a doubt, you can ask as a comment in the bug report or on the mailing list.
- [x] Every commit has proper `Signed-off-by` line as described in our [DCO](https://github.com/phpmyadmin/phpmyadmin/blob/master/DCO). This ensures that the work you're submitting is your own creation.
- [x] Every commit has a descriptive commit message.
- [x] Every commit is needed on its own, if you have just minor fixes to previous commits, you can squash them.
- [x] Any new functionality is covered by tests.

PR #17859, opened 2022-11-02
Fix Replica bugs

kamil-tekiela:
This PR fixes recently refactored code. The escaping for SQL should ALWAYS be done right when it's concatenated into SQL. Otherwise, we introduce random escaping bugs. The values also don't need to be passed as an array, only to be split and merged into an array again. I opted for splitting them in controller and passing them individually to other methods. The values also cannot be null, so I provided empty string as default.


williamdes (approved):

williamdes:
Fixes #17849

Replaces #17856

MauricioFauth (approved):

PR #17863, opened 2022-11-02
Fix import/export escaping

kamil-tekiela:
- A bit string consists of only 1s and 0s so there's no need to escape it. 
- Escaping is not needed for numeric strings because they are treated as a number and not a string
- We can optimize export for tables with a lot of empty columns by not escaping them at all.


williamdes (approved):
I think the import optimisation should go to QA

MauricioFauth (approved):

PR #17860, opened 2022-11-02
Fix #17793 - insert record - fix null not selected if nullable UUID & insert error

sureemo:
### Description

- Fix issue - null is not selected for uuid field with nullable setting
- Fix issue - default select option not select `UUID` when alter column 

Related PR: https://github.com/phpmyadmin/phpmyadmin/pull/17797
Reference issue: https://github.com/phpmyadmin/phpmyadmin/issues/17793

Signed-off-by: Mo Sureerat <sureemo@gmail.com>

PR #17864, opened 2022-11-02
Fix #17793 UUID - Fix insert error for 5.3

### LINKED ISSUE #17793: Inserting a row with UUID should insert the user value when not empty

sureemo:
### Description

Same fix as commits based on `QA_5_2` branch
- Fix issue - null is not selected for uuid field with nullable setting
- Fix issue - default select option not select `UUID` when alter column 

But add new commit (required change on `master` because effected code not existed in `QA_5_2`)
- Fix issue - insert error

Related PR: https://github.com/phpmyadmin/phpmyadmin/pull/17860
Reference issue: https://github.com/phpmyadmin/phpmyadmin/issues/17793

### DIFF
--- a/libraries/classes/InsertEdit.php
+++ b/libraries/classes/InsertEdit.php
@@ -1746,7 +1746,7 @@ class InsertEdit
         if (
             $editField->type === 'uuid'
                 && ! $editField->isNull
-                && in_array($editField->value, ["''", '', "'uuid()'"], true)
+                && in_array($editField->value, ["''", '', "'uuid()'", 'uuid()'], true)
         ) {
             return 'uuid()';
         }
--- a/libraries/classes/InsertEdit.php
+++ b/libraries/classes/InsertEdit.php
@@ -1109,7 +1109,14 @@ class InsertEdit
     private function getSpecialCharsAndBackupFieldForInsertingMode(
         array $column
     ) {
-        if (! isset($column['Default'])) {
+        $isNullableUUID = ! empty($column['True_Type'])
+            && ! empty($column['Default'])
+            && ! empty($column['Null'])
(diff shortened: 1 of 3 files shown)


williamdes (approved):
Thank you!

liviuconcioiu:
Thank you @sureemo! Is perfect! @williamdes I've tested the code and it fixes all the issues.

williamdes:
Then is #17860 ready for review and merge ?

liviuconcioiu:
> Then is #17860 ready for review and merge ?

Yes. Tested that too and it works! Everything is fixed.

sureemo:
> > Then is #17860 ready for review and merge ?
> 
> Yes. Tested that too and it works! Everything is fixed.

@liviuconcioiu Thank you for help verify and feedback.

williamdes:
This is what got merged into 5.3 after 5.2 got merged into 5.3: c00a90482717764334bcccb47e91da3e85a463e3

williamdes:
Seems like you have pushed a merge of master into your branch (I prefer a rebase ;))
And I was about to push the merge, so this pull-request will be marked as closed but is merged for sure ;)

PR #17866, opened 2022-11-03
Introducing quoteString()

kamil-tekiela:
I have started reviewing the usages of `DatabaseInterface::escapeString()` method. I found out that it's usually used correctly. However, it leads to somewhat messy code and several instances are undiscovered bugs. It seems to me that phpMyAdmin has followed the outdated `mysql_*` API technique and it got a little out of hand. 

I am not saying that **proper** string escaping is wrong. I think when it comes to a tool like phpMyAdmin, it makes a lot of sense that we should be able to build SQL queries with string literals properly formatted. Of course, when the SQL doesn't need to be presented to a user in a final form, we can use prepared statements with parameter binding. But I looked into it last year and this project is not yet ready for this. (for the curious: the biggest challenge would be to solve the issue of int/float types with PS). What's more important is that phpMyAdmin doesn't need to switch to prepared statements any time soon. What we have works, but we must ensure that we do it properly. String escaping is flawed because it's very easy to get wrong and break the resulting SQL. The important points are:
- Only string literals need to be escaped
- String literals must be enclosed with single quotes `'` and never double quotes `"` (see `NO_BACKSLASH_ESCAPES` mode)
- Not every SQL statement uses the same escaping. Some require double escaping. See [How should I escape characters inside this LIKE query?](https://stackoverflow.com/a/24590043/1839439). This makes escaping very context-dependent. 
- Escaping needs to be done by the client API that will execute the query. This is something that we can't control when it comes to generated SQL queries though. But phpMyAdmin uses two mysqli connections internally, so ideally the escaping should be done by the connection that will execute the SQL query. 
- Escaping should be done only when putting the string into SQL. It cannot be done earlier as this can lead to accidental reuse of the escaped variable in another context e.g. HTML output. 

There might be even more considerations I can't think of right now. But to the point now. PDO got things right. To avoid SQL injection due to `NO_BACKSLASH_ESCAPES` it doesn't expose any method for string escaping, but it does provide the [`PDO::quote()`](https://www.php.net/manual/en/pdo.quote.php) method. This solves points number 1 and 2 above. 

I propose to add a new method called `DatabaseInterface::quoteString()` that will do what PDO does. It will escape the string and put it in single quotes. I would also like to remove `DatabaseInterface::escapeString()` to prevent accidental usages, but it seems to me that this isn't so straightforward, for example, bookmarks use it to interpolate user input into bookmarked queries. This method is used ~750 times and sometimes it's not used correctly, which will make the review a rather long process. I would like to ask you what you think about this proposal and what do you see as the best way forward. 

This PR contains a demo code of what I am proposing.

@MauricioFauth @williamdes

### DIFF
--- a/libraries/classes/DatabaseInterface.php
+++ b/libraries/classes/DatabaseInterface.php
@@ -485,18 +485,18 @@ class DatabaseInterface implements DbalInterface
                 $needAnd = false;
                 if ($table || ($tableIsGroup === true)) {
                     if (is_array($table)) {
-                        $sql .= ' `Name` IN (\''
+                        $sql .= ' `Name` IN ('
                             . implode(
-                                '\', \'',
+                                ', ',
                                 array_map(
                                     [
                                         $this,
-                                        'escapeString',
+                                        'quoteString',
                                     ],
                                     $table,
                                     $link
                                 )
-                            ) . '\')';
+                            ) . ')';
                     } else {
                         $sql .= " `Name` LIKE '"
                             . $this->escapeMysqlLikeString($table, $link)
@@ -832,9 +832,9 @@ class DatabaseInterface implements DbalInterface
     ): array {
         if (! $GLOBALS['cfg']['Server']['DisableIS']) {
             $sql = QueryGenerator::getInformationSchemaColumnsFullRequest(
-                $database !== null ? $this->escapeString($database, $link) : null,
-                $table !== null ? $this->escapeString($table, $link) : null,
-                $column !== null ? $this->escapeString($column, $link) : null
+                $database !== null ? $this->quoteString($database, $link) : null,
+                $table !== null ? $this->quoteString($table, $link) : null,
+                $column !== null ? $this->quoteString($column, $link) : null
             );
             $arrayKeys = QueryGenerator::getInformationSchemaColumns($database, $table, $column);
 
@@ -1122,9 +1122,7 @@ class DatabaseInterface implements DbalInterface
         // Set timezone for the session, if required.
         if ($GLOBALS['cfg']['Server']['SessionTimeZone'] != '') {
             $sqlQueryTz = 'SET ' . Util::backquote('time_zone') . ' = '
-                . '\''
-                . $this->escapeString($GLOBALS['cfg']['Server']['SessionTimeZone'])
-                . '\'';
+                . $this->quoteString($GLOBALS['cfg']['Server']['SessionTimeZone']);
 
             if (! $this->tryQuery($sqlQueryTz)) {
                 $errorMessageTz = sprintf(
@@ -1167,9 +1165,9 @@ class DatabaseInterface implements DbalInterface
         }
 
         $result = $this->tryQuery(
-            "SET collation_connection = '"
-            . $this->escapeString($collation)
-            . "';"
+            'SET collation_connection = '
+            . $this->quoteString($collation)
+            . ';'
         );
 
         if ($result === false) {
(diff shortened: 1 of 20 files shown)


williamdes:
Hello,

This took me two times to really understand this request, from what I understand I am okay with the proposal. 
I am okay with moving the escaping at the last moment as it will ensure things get right and will be ready for some conversions to prepared statements. 

It's confusing that it needs to be different for `LIKE`.

So I can say it looks right to fix the escaping 
Maybe this can get merged with some tests for the news methods. And following PRs can change the queries to the new way 

> But phpMyAdmin uses two mysqli connections internally, so ideally the escaping should be done by the connection that will execute the SQL query.

The second one is for phpMyAdmin configuration storage. It should be only one open if it is not configured.

libraries/classes/DatabaseInterface.php
```diff
@@ -1893,6 +1890,19 @@ public function getFieldsMeta(ResultInterface $result): array
         return $fields;
     }
 
+    /**
+     * returns properly quoted string for use in MySQL queries
```
MauricioFauth, line 1894:
```suggestion
     * Returns properly quoted string for use in MySQL queries.
```

libraries/classes/DatabaseInterface.php
```diff
@@ -1893,6 +1890,19 @@ public function getFieldsMeta(ResultInterface $result): array
...
+     *
+     * @param string $str  string to be quoted
+     * @param mixed  $link optional database link to use
+     *
+     * @return string a MySQL quoted string
```
MauricioFauth, line 1899:
```suggestion
     * @psalm-return non-empty-string
```

MauricioFauth (approved):
I really like this method.
What about marking `escapeString` as `@deprecated` to avoid adding new usages of it, and opening an issue like "`escapeString` should be replaced with `quoteString`" and adding the "help wanted" tag.

kamil-tekiela:
I made it deprecated. Once this PR is merged an issue can be created. I decided to leave `LIKE` out of it for the moment because I want to do it in another PR.

libraries/classes/Dbal/DbalInterface.php
```diff
@@ -577,9 +577,21 @@ public function affectedRows($link = DatabaseInterface::CONNECT_USER, bool $getF
      */
     public function getFieldsMeta(ResultInterface $result): array;
 
+    /**
+     * Returns properly quoted string for use in MySQL queries.
+     *
+     * @param string $str  string to be quoted
+     * @param mixed  $link optional database link to use
+     *
+     * @psalm-return non-empty-string
```
williamdes, line 586:
```suggestion
     * @psalm-return non-empty-string
     * @psalm-taint-escape sql
```
I think it may work 
https://psalm.dev/docs/security_analysis/avoiding_false_positives/

libraries/classes/DatabaseInterface.php
```diff
@@ -1888,9 +1886,26 @@ public function getFieldsMeta(ResultInterface $result): array
...
+     */
+    public function quoteString(string $str, $link = self::CONNECT_USER): string
+    {
+        return "'" . $this->extension->escapeString($this->links[$link], $str) . "'";
+    }
+
     /**
      * returns properly escaped string for use in MySQL queries
      *
+     * @deprecated Use quoteString() instead.
```
MauricioFauth, line 1907:
```suggestion
     * @deprecated Use {@see quoteString()} instead.
```

libraries/classes/Dbal/DbalInterface.php
```diff
@@ -577,9 +577,23 @@ public function affectedRows($link = DatabaseInterface::CONNECT_USER, bool $getF
...
+     *
+     * @psalm-taint-escape sql
+     */
+    public function quoteString(string $str, $link = DatabaseInterface::CONNECT_USER): string;
+
     /**
      * returns properly escaped string for use in MySQL queries
      *
+     * @deprecated Use quoteString() instead.
```
MauricioFauth, line 595:
```suggestion
     * @deprecated Use {@see quoteString()} instead.
```

MauricioFauth (changes requested):

MauricioFauth (approved):

PR #17870, opened 2022-11-04
Introduce ServerRequest::hasBodyParam()

kamil-tekiela:
This is done in an effort to reduce usage of the infamous `isset`. The `isset` is moved to `ServerRequest`. The idea for this is [borrowed from Laravel](https://github.com/illuminate/http/blob/50ded11e21531c3e00a88928819bb511c4d45398/Concerns/InteractsWithInput.php#L85). 
While implementing this method I accidentally made some other refactorings.

### DIFF
--- a/libraries/classes/Controllers/Sql/ColumnPreferencesController.php
+++ b/libraries/classes/Controllers/Sql/ColumnPreferencesController.php
@@ -11,13 +11,15 @@ use PhpMyAdmin\Http\ServerRequest;
 use PhpMyAdmin\Message;
 use PhpMyAdmin\ResponseRenderer;
 use PhpMyAdmin\Sql;
+use PhpMyAdmin\Table;
 use PhpMyAdmin\Template;
+
+use function array_map;
+use function explode;
+use function is_string;
 
 final class ColumnPreferencesController extends AbstractController
 {
-    /** @var Sql */
-    private $sql;
-
     /** @var CheckUserPrivileges */
     private $checkUserPrivileges;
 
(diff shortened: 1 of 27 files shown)


libraries/classes/Controllers/Sql/ColumnPreferencesController.php
```diff
@@ -44,16 +45,21 @@ public function __invoke(ServerRequest $request): void
...
         $status = false;
 
+        /** @var string|null $tableCreateTime */
+        $tableCreateTime = $request->getParsedBodyParam('table_create_time');
+
         // set column order
         $colorder = $request->getParsedBodyParam('col_order');
-        if ($colorder !== null) {
-            $status = $this->sql->setColumnProperty($tableObject, 'col_order');
+        if (is_string($colorder)) {
```
williamdes, line 53: Why not checking that it is not null?

williamdes (approved):
Very good work!

libraries/classes/Controllers/Sql/ColumnPreferencesController.php
kamil-tekiela: I changed it because we know that we expect a string and nothing else. `explode()` accepts only a string. To help out with static analysis, I decided to restrict type to string.

williamdes: Makes sense, thank you

MauricioFauth (approved):
Those are very useful methods. Thank you!

PR #17873, opened 2022-11-06
Refactor BinlogController.php

kamil-tekiela:
I could not find `is_full_query` anywhere in the code so I have no idea what values we are actually expecting.

### DIFF
--- a/libraries/classes/Controllers/Server/BinlogController.php
+++ b/libraries/classes/Controllers/Server/BinlogController.php
@@ -44,33 +44,27 @@ class BinlogController extends AbstractController
 
     public function __invoke(ServerRequest $request): void
     {
-        $GLOBALS['errorUrl'] = $GLOBALS['errorUrl'] ?? null;
+        $log = $request->getParsedBodyParam('log');
+        $position = (int) $request->getParsedBodyParam('pos', 0);
 
-        $params = [
-            'log' => $_POST['log'] ?? null,
-            'pos' => $_POST['pos'] ?? null,
-            'is_full_query' => $_POST['is_full_query'] ?? null,
-        ];
         $GLOBALS['errorUrl'] = Url::getFromRoute('/');
 
         if ($this->dbi->isSuperUser()) {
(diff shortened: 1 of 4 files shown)


williamdes:
> I could not find `is_full_query` anywhere in the code so I have no idea what values we are actually expecting.

I think I will have to search in the history
It's probably an old parameter or a bug
https://github.com/phpmyadmin/phpmyadmin/search?q=is_full_query&type=code

libraries/classes/Controllers/Server/BinlogController.php
```diff
@@ -44,33 +44,27 @@ public function __construct(ResponseRenderer $response, Template $template, Data
...
         $urlParams = [];
-        if (isset($params['log']) && array_key_exists($params['log'], $this->binaryLogs)) {
-            $urlParams['log'] = $params['log'];
+        if (array_key_exists($log, $this->binaryLogs)) {
+            $urlParams['log'] = $log;
         }
 
         $isFullQuery = false;
-        if (! empty($params['is_full_query'])) {
+        if (! empty($request->getParsedBodyParam('is_full_query'))) {
```
MauricioFauth, line 62:
```suggestion
        if ($request->getParsedBodyParam('is_full_query')) {
```

The `is_full_query` parameter is a flag and always has the value '`1`' when present. So, `if ($request->hasBodyParam('is_full_query')) {` will also work here.

You can see it been set here:
https://github.com/phpmyadmin/phpmyadmin/blob/97aac39a40cef8f0ec63f380be79e14c4633bd67/libraries/classes/Controllers/Server/BinlogController.php#L67-L71

And been unset here:
https://github.com/phpmyadmin/phpmyadmin/blob/97aac39a40cef8f0ec63f380be79e14c4633bd67/libraries/classes/Controllers/Server/BinlogController.php#L88-L91

MauricioFauth (changes requested):

MauricioFauth (approved):

Issue #17523, opened 2022-05-13
Error running phpmyadmin on Ubuntu 22.04 server


arcanisgk:
### Describe the bug

I tried to open phpmyadmin and got a 500 server error:

![image](https://user-images.githubusercontent.com/1207306/168309909-88aff5d3-3b3a-46d8-ae21-a8e7a783bc7d.png)

When trying to find out what happened to the server I see this error in the log:

> sudo tail -n 2 /var/log/apache2/error.log
> [Fri May 13 14:44:11.929538 2022] [php7:error] [pid 1334] [client 192.168.160.1:30449] PHP Parse error:  syntax error, unexpected 'static' (T_STATIC) in /usr/share/php/Symfony/Component/DependencyInjection/ContainerBuilder.php on line 272
> [Fri May 13 14:44:13.206921 2022] [php7:error] [pid 1331] [client 192.168.160.1:30459] PHP Parse error:  syntax error, unexpected 'static' (T_STATIC) in /usr/share/php/Symfony/Component/DependencyInjection/ContainerBuilder.php on line 272

### To Reproduce

having installed Apache, MySQL and php 7.4

I have installed phpmyadmin using the following command:

```
sudo add-apt-repository ppa:phpmyadmin/ppa
sudo apt-get update
sudo apt -y install phpmyadmin
```

when the installation is done it has added a new version of php, to be more specific system add version 8.1
at the end of the installation and past the configuration dialog:

![image](https://user-images.githubusercontent.com/1207306/168309611-c9e12955-6be3-4828-b1ab-1b5867238cb8.png)

![image](https://user-images.githubusercontent.com/1207306/168309632-8a359a70-8651-4236-ae09-fad2ba339415.png)r

I'm going to try the phpmyadmin portal. and get the error/bug described above.

### Expected behavior

that the phpmyadmin open normally without server problem error 500 related to dependencies.
How can we solve the problem of the package that is installed with apt?

### Server configuration

- Operating system: Ubuntu 22.04 Server
- Web server: Apache
- Database version: MySQL
- PHP version: 7.4/8.1
- phpMyAdmin version: 

> apt show phpmyadmin
> Package: phpmyadmin
> Version: 4:5.1.1+dfsg1-5ubuntu1
> Priority: optional
> Section: universe/web
> Origin: Ubuntu
> Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
[...]

PR #17883, opened 2022-11-08
Separate the minimum phpmyadmin requirement

wnunezc:
Separate the minimum phpmyadmin requirement for the download version and the package installation.

### DIFF
--- a/doc/require.rst
+++ b/doc/require.rst
@@ -12,7 +12,11 @@ web server (such as Apache, nginx, :term:`IIS`) to install phpMyAdmin's files in
 PHP
 ---
 
-* You need PHP 7.1.3 or newer, with ``session`` support, the Standard PHP Library
+* Download version: You need PHP 7.2.5 or newer.
+
+* Package installation: for the latest version of phpMyAdmin 5.X it is recommended the version of php 8.1 or higher (normally this version will be installed automatically like a dependency OS distribution).
+
+* Also need ``session`` support, the Standard PHP Library
   (SPL) extension, hash, ctype, and JSON support.
 
 * The ``mbstring`` extension (see :term:`mbstring`) is strongly recommended


ibennetch (changes requested):
I do not particularly care for this change.

I feel like it is less clear in general and puts on us some of the work of tracking the status of various distribution packages. Users installing through a package manager should trust the  version recommendations made by the maintainers as they take precedence over our distribution, just like file paths and configuration can be changed by packagers.

arcanisgk:
@ibennetch It is what is understood [here ](https://github.com/phpmyadmin/phpmyadmin/issues/17523#issuecomment-1307217243)and [here](https://github.com/phpmyadmin/phpmyadmin/wiki/DebianUbuntu#related-issues-and-comments-on-install-methods), but not visible, I don't know if you can improve it, my English is rubbish, XD.

williamdes:
This is an Ubuntu specific bug, so the documentation you wrote is wrong. I am closing this subject as this is up to Ubuntu to fix their sloppy patches. This is a sad situation that still can be fixed by someone submitting a bug fix to Ubuntu.

doc/require.rst
```diff
@@ -12,7 +12,11 @@ web server (such as Apache, nginx, :term:`IIS`) to install phpMyAdmin's files in
 PHP
 ---
 
-* You need PHP 7.1.3 or newer, with ``session`` support, the Standard PHP Library
+* Download version: You need PHP 7.2.5 or newer.
+
+* Package installation: for the latest version of phpMyAdmin 5.X it is recommended the version of php 8.1 or higher (normally this version will be installed automatically like a dependency OS distribution).
```
williamdes, line 17:
```suggestion
* Package installation on Ubuntu: due to an Ubuntu mistake in packaging symfony you will be required to use PHP 8 or above.
```
Something like that

PR #17886, opened 2022-11-09
Fix #17404 - Handling and Redirecting to the database list page after dropping the current Database

### LINKED ISSUE #17404: No database selected error after dropping a database
No database selected error after dropping a database.
It should probably just redirect to the home page.

5.2.0-dev+20220217.fdd17ddf0b

image

vimalMK:
Signed-off-by: Vimal K <vimalinfo10@gmail.com>

### Description

Handling and Redirecting to the database list page after dropping the current database.

After dropping the DB and redirecting, a gentle message is shown saying 'No database selected'. No error message pop up.

image

[drop_db_pma.webm](https://user-images.githubusercontent.com/35750792/200754647-7ac93f12-ac13-4a1a-9490-abb32e346944.webm)

Fixes #17404 

### DIFF
--- a/libraries/classes/Controllers/AbstractController.php
+++ b/libraries/classes/Controllers/AbstractController.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace PhpMyAdmin\Controllers;
 
 use PhpMyAdmin\Core;
-use PhpMyAdmin\Message;
 use PhpMyAdmin\ResponseRenderer;
 use PhpMyAdmin\Template;
 use PhpMyAdmin\Url;
@@ -67,11 +66,9 @@ abstract class AbstractController
 
         if (strlen($db) === 0 || ! $is_db) {
             if ($this->response->isAjax()) {
-                $this->response->setRequestStatus(false);
-                $this->response->addJSON(
-                    'message',
-                    Message::error(__('No databases selected.'))
-                );
+                //Redirecting to the server page and exiting the AJAX control
+                $param = ['ajax_request' => 'false', 'message' => __('No databases selected.')];
+                $this->redirect('/server/databases', $param);
 
                 return false;
             }


williamdes (approved):
Looks good!

OlafvdSpek (commented):
Is this the right place for the redirect?

libraries/classes/Controllers/AbstractController.php
```diff
@@ -67,11 +66,9 @@ protected function hasDatabase(): bool
 
         if (strlen($db) === 0 || ! $is_db) {
             if ($this->response->isAjax()) {
-                $this->response->setRequestStatus(false);
-                $this->response->addJSON(
-                    'message',
-                    Message::error(__('No databases selected.'))
-                );
+                //Redirecting to the server page and exiting the AJAX control
+                $param = ['ajax_request' => 'false', 'message' => __('No databases selected.')];
```
williamdes, line 70: after viewing the video it's maybe not the proper message to display 🤔

vimalMK: can you suggest a message you think might fit well . I can update it.

vimalMK:
@OlafvdSpek thank you for the review. what page you think might make more sense for redirect ?

libraries/classes/Controllers/AbstractController.php
vimalMK: something like "Please select a database" instead of "Database not selected"

OlafvdSpek:
> @OlafvdSpek thank you for the review. what page you think might make more sense for redirect ?

If the query is executed in the context of `?route=/server/sql` rather than `?route=/database/sql&db=xyz` there's no need for special handling.

vimalMK:
> > @OlafvdSpek thank you for the review. what page you think might make more sense for redirect ?
> 
> If the query is executed in the context of `?route=/server/sql` rather than `?route=/database/sql&db=xyz` there's no need for special handling. But that'd only work if the operation is performed via `/database/operations&db=xyz`

The PR will reflect the behavior of redirecting only when the query is executed from '/database/sql&db=xyz' or  `/database/operations&db=xyz` ->Query section 

Anything performed via  `?route=/server/sql`  will not have any change and will remain as it used to be.

MauricioFauth (approved):
I also not sure if this is the right change. But I have no better idea. I'd rather add this to the `master` branch instead. What do you think @williamdes?

MauricioFauth:
@vimalMK Unrelated to this PR. Looks like you are using the wrong video codec for the webm video. That's why it doesn't work on Firefox.

vimalMK:
> @vimalMK Unrelated to this PR. Looks like you are using the wrong video codec for the webm video. That's why it doesn't work on Firefox.

Thanks for letting me know. I had no idea. :)

williamdes:
> I also not sure if this is the right change. But I have no better idea. I'd rather add this to the `master` branch instead. What do you think @williamdes?

Yes, I agree. It's not fixing any real bug. I updated the labels on the issue

PR #17888, opened 2022-11-11
Extract schema export from schema plugins

MauricioFauth:
Moves the response handling to the `SchemaExportController`.


williamdes (approved):
💯

PR #17889, opened 2022-11-11
Refactor Index controllers

kamil-tekiela:
I'm not entirely happy with this PR but I think this may still be some kind of an improvement. I am open to suggestions. I'll keep working on it in later PRs too.

### DIFF
--- a/libraries/classes/Table/Indexes.php
+++ b/libraries/classes/Table/Indexes.php
@@ -8,6 +8,7 @@ use PhpMyAdmin\Common;
 use PhpMyAdmin\Controllers\Table\StructureController;
 use PhpMyAdmin\Core;
 use PhpMyAdmin\DatabaseInterface;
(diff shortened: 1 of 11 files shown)


libraries/classes/Table/Indexes.php
```diff
@@ -114,4 +118,29 @@ public function doSaveData(Index $index, bool $renameMode, string $db, string $t
             $this->response->addJSON('message', $error);
         }
     }
+
+    /**
+     * @param string[] $selectedColumns
+     */
+    public function getAddIndexSql(string $indexType, string $table, array $selectedColumns): string
```
williamdes, line 125: this should go to `libraries/classes/Query/Generator.php`

williamdes (approved):

MauricioFauth (approved):

PR #17904, opened 2022-11-18
Simplify gis polygon handling

MoonE:
No need to handle inner and outer rings separately


williamdes (approved):
How can I check that this will have no impact :)?
Said differently: how did you check it did not break something?

MoonE:
Export the geometries below as png / pdf and compare the results:

```sh
cmp '/tmp/polygons-master.png' '/tmp/polygons-new.png' && echo 'png same' || echo 'png different'
cmp '/tmp/polygons-master.png' '/tmp/polygons-new.png' && echo 'pdf same' || echo 'pdf different'
```

```sql
CREATE TABLE `polygons` (
  `id` int NOT NULL   AUTO_INCREMENT,
  `polygons` geometry DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB;

INSERT INTO `polygons` (`id`, `polygons`) VALUES
(1, 0xe610000001060000000200000001030000000200000005000000000000000000000000000000000000000000000000004940000000000000000000000000000049400000000000004940000000000000000000000000000049400000000000000000000000000000000005000000000000000000244000000000000024400000000000004440000000000000244000000000000044400000000000004440000000000000244000000000000044400000000000002440000000000000244001030000000100000005000000000000000000344000000000000034400000000000003e4000000000000034400000000000003e400000000000003e4000000000000034400000000000003e4000000000000034400000000000003440),
(2, 0xe6100000010300000001000000040000006666666666661e40000000000000484033333333333321400000000000004840333333333333214000000000008047406666666666661e400000000000004840),
(3, 0x0000000001030000000200000005000000000000000000f03f000000000000f03f0000000000002440000000000000f03f00000000000024400000000000002440000000000000f03f0000000000002440000000000000f03f000000000000f03f050000000000000000000040000000000000004000000000000022400000000000000040000000000000224000000000000022400000000000000040000000000000224000000000000000400000000000000040),
(4, 0xe6100000010300000001000000040000000000000000001c40000000000000484000000000000020400000000000004840000000000000204000000000008047400000000000001c400000000000004840);
```

Logically `explode('),(', $wkt_rings)`  on a string that does not container `'),('` will return an array with the original string as the only element. So it only complicates things when it is treated differently.

MauricioFauth (changes requested):
Hi, thanks for your contribution.
Could you please add tests that cover these changes?

MoonE:
Now tests compare the exported png and pdf. For the pdf it was necessary to set the creation and modifcation time to a fixed timestamp and to always use the same internal file id. Otherwise the exported pdf does not match.

Also fixes the pdf tests did not draw anything because no page was added to the document beforehand.

test/classes/Gis/GisGeomTestCase.php
```diff
@@ -62,4 +66,28 @@ public function testScaleRow(string $spatial, array $min_max): void
...
+    /**
+     * Create a new pdf document with predictable timestamps and ids.
+     *
+     * @param string $id Used as a seed for the internal file_id
+     *
+     * @return TCPDF A pdf document with an empty page
+     */
+    protected function createEmptyPdf(string $id): TCPDF
+    {
+        $pdf = new TCPDF();
```
williamdes, line 79: since tcpdf is optional, we should add a 
```suggestion

        if (! class_exists(TCPDF::class)) {
            $this->markTestSkipped('TCPDF is missing');
        }
        $pdf = new TCPDF();
```

test/classes/Gis/GisMultiLineStringTest.php
```diff
@@ -282,7 +298,109 @@ public function testPrepareRowAsPdf(
         TCPDF $pdf
     ): void {
         $return = $this->object->prepareRowAsPdf($spatial, $label, $color, $scale_data, $pdf);
-        $this->assertInstanceOf(TCPDF::class, $return);
+        /** @var string $file */
+        $file = $return->getPDFData();
+
+        $this->assertEquals(
+            'JVBERi0xLjcKJeLjz9MKNSAwIG9iago8PCAvVHlwZSAvUGFnZSAvUGFyZW50IDEgMCBSIC9MYXN0TW9kaWZpZWQgKEQ6MjAyMDA5MTM'
```
williamdes, line 305: maybe such large blobs should be files in `test/test_data/` ?

test/classes/Gis/GisGeomTestCase.php
MoonE: TCPDF is listed under `require-dev`. Wouldn't that always make it available for tests?
https://github.com/phpmyadmin/phpmyadmin/blob/22ab240b3316b0dae04db346e654b975266af8bb/composer.json#L97-L113

MoonE:
The i386 build job produces a slightly different pdf for point and multipoint, The point-circles are drawn with slightly different coordinates. I added a different test pdf just for that build job.

MauricioFauth (approved):
Thank you for adding the test commits before the change. That helped a lot.

PR #17906, opened 2022-11-18
Fix #17482 - default to "Full texts" when running explain statements 

### LINKED ISSUE #17482: Default to "Full texts" when running ANALYZE
Hey

When the query begins with `ANALYZE` (MariaDB) or `EXPLAIN ANALYZE` (MySQL),
could you please default to "Full texts", so we don't have to run the query again?

`ANALYZE` is used to understand why a query is slow, so having to run it twice is a "waste of time", if you know what I mean :)

image

Example query:

`ANALYZE FORMAT=JSON SELECT * FROM test_table`

Thank you!

iifawzi:
### Description

Hi, this should fix #17482.
Fixes #17482

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -61,7 +61,7 @@
         "paragonie/sodium_compat": "^1.17",
         "phpmyadmin/motranslator": "^5.0",
         "phpmyadmin/shapefile": "^3.0.1",
-        "phpmyadmin/sql-parser": "dev-master#a9cd167be77829ec2dd4496ce1010ed6abde92f8",
+        "phpmyadmin/sql-parser": "dev-master#8fddb4becdfb657db3b9ade52e952ef3c822a26b",
         "phpmyadmin/twig-i18n-extension": "^4.0",
         "psr/http-factory": "^1.0",
         "psr/http-message": "^1.0",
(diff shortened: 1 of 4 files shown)


iifawzi:
I'm not very sure that this's the correct way of testing this.

williamdes:
> I think it's not testing against the SQL-Parser#dev-master. I've hard-coded the SQL-Parser changes on my machine and it passed.

yeah, you should update in this PR the hash I put in the `composer.json` file ;)

iifawzi:
what do you mean by updating the hash @williamdes?

williamdes:
> what do you mean by updating the hash @williamdes?

https://github.com/phpmyadmin/phpmyadmin/blob/9a8452735c653ac0c35059ab8c93d5b90561fd8c/composer.json#L64

williamdes (approved):
Looks good, will test it
Thank you for your work on this subject :)

nunoperalta:
Thank you very much for this, guys. You're amazing! 🤩🥳🎉

williamdes:
The [(phpMyAdmin 5.2+snapshot)](https://www.phpmyadmin.net/downloads/#snapshot_5.2+snapshot) also available as a [non official docker image](https://github.com/sudo-bot/docker-phpmyadmin-snapshots) is updated with this PR 🚀

PR #17917, opened 2022-11-22
Remove unnecessary Export JS object

MauricioFauth:


williamdes (changes requested):
I really prefer when functions are in an object and not free in the wild. Like we do in other files

MauricioFauth:
> I really prefer when functions are in an object and not free in the wild. Like we do in other files

I like that as well. What bother you the most when functions aren't in a object? For example, those are all "private" functions, they aren't exposed to the global space, because Webpack wraps this file in an IIFE.

MauricioFauth:
Okay. Makes sense. I'll look for better ways to improve the code.

PR #17919, opened 2022-11-24
Create a `shared.js` file for common JS dependencies

MauricioFauth:
Instead of using the Webpack's `dependOn` setting, uses the `splitChunks` setting to automatically generate the `shared.js` file.

As the contents of these shared dependencies are being moved to the `shared.js` file, the generated file for each dependency is now empty.

Is probably safe to not generate them, but some adjustments are needed for that first. For example, the AJAX's onload and teardown functions expects that some of these files exists.

For now, only dependencies that are loaded for all pages should be added in this `shared.js` file.

This will facilitate the refactoring of the JS files, but the trade off, at least for now, is that will be a single large file instead of a few medium ones.

The Webpack's runtime seems to work like a container, so this allows us to stop relying on the Window global object.

- https://webpack.js.org/plugins/split-chunks-plugin/
- https://webpack.js.org/guides/code-splitting/#splitchunksplugin


libraries/classes/Header.php
```diff
@@ -139,6 +139,7 @@ private function addDefaultScripts(): void
         $this->scripts->addFile('vendor/jquery/jquery.validate.min.js');
         $this->scripts->addFile('vendor/jquery/jquery-ui-timepicker-addon.js');
         $this->scripts->addFile('index.php', ['route' => '/messages', 'l' => $GLOBALS['lang']]);
+        $this->scripts->addFile('shared.js');
         $this->scripts->addFile('ajax.js');
```
williamdes, line 143: We don't remove the adding of small ones?

williamdes (approved):

libraries/classes/Header.php
MauricioFauth: This is also used to trigger the event handlers of these respective files, so an extra care is needed when removing them.

williamdes: Indeed !

williamdes (approved):

PR #17921, opened 2022-11-25
Clean up LIKE escaping

kamil-tekiela:
This is follow-up from #17377 and #17866

What I did:
- Removed escape_mysql_wildcards Twig extension
- Moved `escapeMysqlWildcards()` and `unescapeMysqlWildcards()` to Privileges.php where they belong. These methods should only be used for escaping `LIKE` in `GRANT`, but only in the global database mask part, not in tables or routines. This whole code is a little bit of a mess, so I might refactor it later. 
- Dropped `DatabaseInterface::escapeMysqlLikeString()` which I have added earlier. I did it because I saw no other way to allow deliberate wildcards while escaping `_&\`. e.g. `$dbi->quoteString($dbi->escapeMysqlWildcards('db_100%') . '%')`
- Refactored `Generator::getTableCondition()`. I spent way too much time thinking how to best design it and I only came up with this. 
- Fixed escaping issue in `UserPreferences` and `ListDatabases`. While this is technically a bug fix, it's not worth backporting to 5.2. 
- I also fixed a bug (?) in HTML/Generator. I think it's only a visual bug and I doubt it affects any functionality. It's fixed anyway now.


libraries/classes/DatabaseInterface.php
```diff
@@ -1931,16 +1939,16 @@ public function escapeString(string $str, int $link = self::CONNECT_USER): strin
...
-     * @param int    $link optional database link to use
+     * @param string $str string to be escaped
      *
      * @return string a MySQL escaped LIKE string
      */
-    public function escapeMysqlLikeString(string $str, int $link = self::CONNECT_USER)
+    public function escapeMysqlWildcards(string $str): string
     {
-        return $this->escapeString(strtr($str, ['\\' => '\\\\', '_' => '\\_', '%' => '\\%']), $link);
+        return strtr($str, ['\\' => '\\\\', '_' => '\\_', '%' => '\\%']);
```
williamdes, line 1951: This should fix the performance issue I had detected, can this line be picked to 5.2 ?
What did escapeString do exactly ?

libraries/classes/ListDatabase.php
```diff
@@ -133,7 +134,7 @@ protected function checkOnlyDatabase(): bool
             // thus containing not escaped _ or %
             if (! preg_match('/(^|[^\\\\])(_|%)/', $each_only_db)) {
                 // ... not contains wildcard
-                $items[] = Util::unescapeMysqlWildcards($each_only_db);
+                $items[] = strtr($each_only_db, ['\\\\' => '\\', '\\_' => '_', '\\%' => '%']);
```
williamdes, line 137: this feels like it should be near the escaping method

libraries/classes/DatabaseInterface.php
kamil-tekiela: I doubt it will. I haven't changed anything related to performance. This is a brand new method anyway. 
`escapeString` is a method that calls `mysqli::real_escape_string()`. It formats a string to be used as a string literal in SQL. Any time you inject any string directly into SQL, you must use this function to format it.

libraries/classes/ListDatabase.php
kamil-tekiela: Which escaping method? This is for unescaping manual input in the config file. There is no corresponding escape method unless you meant the one I added in `DbalInterface.php`. But this sounds like an overkill to create a method that will be used only in one place. I could, of course, add such a method to `DbalInterface.php` but this will mean I have to use another global here. I don't know if it's such a good idea.

MauricioFauth (approved):

PR #17924, opened 2022-11-25
Extract inline JS for import's upload progress logic

MauricioFauth:


js/src/import.js
```diff
@@ -58,6 +62,7 @@ AJAX.registerTeardown('import.js', function () {
     $('#input_import_file').off('change').off('focus');
     $('#select_local_import_file').off('focus');
     $('#text_csv_enclosed').add('#text_csv_escaped').off('keyup');
+    $('#buttonGo').off('click');
```
williamdes, line 65: maybe we can make this more specific to the import form to avoid binding other buttonGo ?

PR #17926, opened 2022-11-26
Change text direction on selects for RTL

liviuconcioiu:
Signed-off-by: Liviu-Mihail Concioiu <liviu.concioiu@gmail.com>

### Description

This PR fixes text direction on selects for RTL languages.

Before:

before

After:

after

### DIFF
--- a/libraries/classes/Language.php
+++ b/libraries/classes/Language.php
@@ -166,7 +166,7 @@ class Language
      */
     public function isRTL(): bool
     {
-        return in_array($this->code, ['ar', 'fa', 'he', 'ur']);
+        return in_array($this->code, ['ar', 'ar_LY', 'fa', 'he', 'ur']);
     }
 
     /**
--- a/templates/columns_definitions/column_attributes.twig
+++ b/templates/columns_definitions/column_attributes.twig
@@ -61,7 +61,7 @@
 </td>
 <td class="text-center">
   {# column collation #}
-  <select lang="en" dir="ltr" name="field_collation[{{ column_number }}]" id="field_{{ column_number }}_{{ ci - ci_offset }}">
+  <select lang="{{ lang }}" dir="{{ text_dir }}" name="field_collation[{{ column_number }}]" id="field_{{ column_number }}_{{ ci - ci_offset }}">
     <option value=""></option>
     {% for charset in charsets %}
       <optgroup label="{{ charset.name }}" title="{{ charset.description }}">
--- a/templates/login/form.twig
+++ b/templates/login/form.twig
@@ -31,7 +31,7 @@
       <div class="card-body">
         <form method="get" action="{{ url('/') }}" class="disableAjax">
           {{ get_hidden_inputs(form_params) }}
-          <select name="lang" class="form-select autosubmit" lang="{{ lang }}" dir="{{ text_dir }}" id="languageSelect" aria-labelledby="languageSelectLabel">
+          <select name="lang" class="form-select autosubmit" lang="en" dir="ltr" id="languageSelect" aria-labelledby="languageSelectLabel">
             {% for language in available_languages %}
               <option value="{{ language.getCode()|lower }}"{{ language.isActive() ? ' selected' }}>
                 {{- language.getName()|raw -}}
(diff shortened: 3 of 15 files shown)


williamdes (approved):

templates/login/form.twig
```diff
@@ -31,7 +31,7 @@
       <div class="card-body">
         <form method="get" action="{{ url('/') }}" class="disableAjax">
           {{ get_hidden_inputs(form_params) }}
-          <select name="lang" class="form-select autosubmit" lang="en" dir="ltr" id="languageSelect" aria-labelledby="languageSelectLabel">
+          <select name="lang" class="form-select autosubmit" lang="{{ lang }}" dir="{{ text_dir }}" id="languageSelect" aria-labelledby="languageSelectLabel">
```
williamdes, line 34: I am not sure for this one, I think the language list is english only. Not sure there is a text direction

williamdes (approved):
The diff looks good, is there more commits to be pushed ?

liviuconcioiu:
> The diff looks good, is there more commits to be pushed ?

No.

libraries/classes/Language.php
```diff
@@ -166,7 +166,7 @@ public function matchesUserAgent($header): bool
      */
     public function isRTL(): bool
     {
-        return in_array($this->code, ['ar', 'fa', 'he', 'ur']);
+        return in_array($this->code, ['ar', 'ar_LY', 'fa', 'he', 'ur']);
```
iifawzi, line 169: Do we really need to have a specific country locale here? `_LY`?
Isn't `ar` the code already? if it's not, I think it would be better to substr only the language `ar`, `en`.. from the code strings, otherwise, we will need to have a list of all the countries' dialects for all the languages  ( `ly`, `eg`, `sa`, .... ) in this array

liviuconcioiu: > Do we really need to have a specific country locale here? `_LY`? Isn't `ar` the code already? if it's not, I think it would be better to substr only the language `ar`, `en`.. from the code strings, otherwise, we will need to have a list of all the countries' dialects for all the languages ( `ly`, `eg`, `sa`, .... ) in this array

`ar_LY` is present only on the demo server [https://demo.phpmyadmin.net/QA_5_2/index.php?route=/](https://demo.phpmyadmin.net/QA_5_2/index.php?route=/) and not other dialects.

I'll use substr.

iifawzi: > ar_LY is present only on the demo server https://demo.phpmyadmin.net/QA_5_2/index.php?route=/ and not other dialects.

Ahh, ok, I thought that the locale code is being detected from the browser, not from a pre-defined list. 
I think it's fine having it in the array then, what do you think?

liviuconcioiu: > Ahh, ok, I thought that the locale code is being detected from the browser, not from a pre-defined list. I think it's fine having it in the array then, what do you think?

By looking at [https://hosted.weblate.org/projects/phpmyadmin/#languages](https://hosted.weblate.org/projects/phpmyadmin/#languages) only ar_LY is missing from RTL languages. There are no other translation, so it will be useless to use substr since it will slower the speed. I think the best way is to leave it in array, since no other translation are there. Either way, `ar_LY` is not shipped in the downloaded version (is only on demo).

MauricioFauth (changes requested):
I think this change is semantically wrong. The options in these selects are in English and read left to right.
This is [an issue with Bootstrap](https://github.com/twbs/bootstrap/issues/37567) and should be fixed using CSS.

templates/columns_definitions/column_attributes.twig
```diff
@@ -61,7 +61,7 @@
 </td>
 <td class="text-center">
   {# column collation #}
-  <select lang="en" dir="ltr" name="field_collation[{{ column_number }}]" id="field_{{ column_number }}_{{ ci - ci_offset }}">
+  <select lang="{{ lang }}" dir="{{ text_dir }}" name="field_collation[{{ column_number }}]" id="field_{{ column_number }}_{{ ci - ci_offset }}">
```
MauricioFauth, line 64: This is redundant, as the `<html>` tag already sets the `lang` and `dir` attributes with the same values.

https://github.com/phpmyadmin/phpmyadmin/blob/44a0844f4083f09f78330b9e0e3568d43ba859ca/templates/header.twig#L2

PR #17927, opened 2022-11-26
Fix key navigation between select inputs

MoonE:
This compatibility fix is no longer necessary and actually causing issues with current Firefox versions.
Not sure since what version it is no longer necessary, but was probably fixed a long time ago in Firefox.

Test by going to the insert form, focus a select input and press Ctrl + Down or Ctrl + Up. It should not change the value in the  select but  focus the one below or above.

As a side note:
It checked versions from 3 to 24 but actually also matched all versions >= 30 as no word boundary check was made on the version number

See #1100

### DIFF
--- a/js/src/keyhandler.js
+++ b/js/src/keyhandler.js
@@ -70,9 +70,6 @@ function onKeyDownArrowsHandler (event) {
         return;
     }
 
-    // eslint-disable-next-line compat/compat
-    var isFirefox = navigator.userAgent.toLowerCase().indexOf('firefox/') > -1;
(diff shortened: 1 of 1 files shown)


williamdes (approved):
Thank you!

MauricioFauth (approved):
Thank you!
These version of Firefox are not supported anymore.
https://github.com/phpmyadmin/phpmyadmin/blob/e649dada4969b9514ac93e53d8cff8cea309bdea/.browserslistrc#L7-L8

PR #17946, opened 2022-12-02
Replace escapeString with quoteString

kamil-tekiela:


libraries/classes/CreateAddField.php
```diff
@@ -399,7 +399,7 @@ public function getTableCreationQuery(string $db, string $table): string
 
         // Adds table type, character set, comments and partition definition
         if (! empty($_POST['tbl_storage_engine']) && ($_POST['tbl_storage_engine'] !== 'Default')) {
-            $sqlQuery .= ' ENGINE = ' . $this->dbi->escapeString($_POST['tbl_storage_engine']);
+            $sqlQuery .= ' ENGINE = ' . $_POST['tbl_storage_engine'];
```
williamdes, line 402: I think we should secure this one to avoid injections
I agree that it uses POST

kamil-tekiela: Yes, I saw that too. I was thinking of looking into it later, but let me see if I can muster something up now.

williamdes (approved):

MauricioFauth (approved):

PR #17969, opened 2022-12-16
Remove dead code in Navigation

kamil-tekiela:
Here's the commit that introduced this method https://github.com/phpmyadmin/phpmyadmin/commit/dfe8dc283a5caa631ae1a21380e188bf5b4489de
Since then, the images have been deleted (or renamed) and all other usages of this method are gone too. 

It's worth mentioning that these buttons (or links actually) do not show up on the UI without an image. There is JS code for them, but I have no idea what it does. The "collapse all" button might have been useful. Perhaps the buttons should get fixed again? @williamdes Any idea?

### DIFF
--- a/libraries/classes/Navigation/NavigationTree.php
+++ b/libraries/classes/Navigation/NavigationTree.php
@@ -1330,15 +1330,9 @@ class NavigationTree
      */
     private function controls(): string
     {
-        // always iconic
-        $showIcon = true;
-        $showText = false;
-
         $collapseAll = Generator::getNavigationLink(
             '#',
-            $showText,
             __('Collapse all'),
-            $showIcon,
             's_collapseall',
             'pma_navigation_collapse'
         );
@@ -1349,7 +1343,7 @@ class NavigationTree
             $title = __('Unlink from main panel');
         }
 
-        $unlink = Generator::getNavigationLink('#', $showText, $title, $showIcon, $syncImage, 'pma_navigation_sync');
+        $unlink = Generator::getNavigationLink('#', $title, $syncImage, 'pma_navigation_sync');
 
         return $this->template->render('navigation/tree/controls', [
             'collapse_all' => $collapseAll,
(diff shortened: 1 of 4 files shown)


williamdes:
It's fixed now by 90f9e708cc6faca16a27b0d87f49e70231b3766c
Right ?

liviuconcioiu:
@williamdes Collapse all is fixed.

collapse

libraries/classes/Navigation/NavigationTree.php
```diff
@@ -1335,26 +1334,18 @@ private function fastFilterHtml(Node $node): string
...
-        $showText = false;
-
-        $collapseAll = Generator::getNavigationLink(
+        $collapseAll = sprintf(
+            '<a href="%s" id="%s" title="%s"></a>',
             '#',
-            $showText,
-            __('Collapse all'),
-            $showIcon,
-            's_collapseall',
```
MauricioFauth, line 1347: This missing icon was fixed by 90f9e708cc6faca16a27b0d87f49e70231b3766c.

MauricioFauth (changes requested):

libraries/classes/Navigation/NavigationTree.php
kamil-tekiela: Sorry, but I got lost a little bit. Do you want me to add `ic_s_collapseall` to this button? What about the other button?
Even if I add the new class to this button it still won't show up. It has no content therefore it has no dimensions and no background. I'm not sure how these buttons were designed, can you suggest how to proceed here?

MauricioFauth: This button wasn't working, but it's now fixed. When you click on a database, it expands the tree. You can use this button to collapse all items.

collapse

kamil-tekiela: Ahh ok. I got confused there. I reverted some of my changes.

MauricioFauth (approved):

Issue #17229, opened 2021-12-07
2FA U2F FIDO - U2F API deprecating soon 


MordiSacks:
### Describe the bug
 U2F API will be deprecated by the end of February 2022 by chrome (and all chrome based browsers) ,
Currently when attempting to authenticate I get a message 

“This site won’t be able to use the U2F API after February 2022. If you own this site, you should change it to use the Web Authentication API”. 

### To Reproduce
1. Set up PMA
2. Create PMA storage DB
3. Setup FIDO U2F as 2FA (home > settings > Two-factor authentication > FIDO) (I use a yubikey)
4. logout
5. login

### Screenshots
<img src="https://image.prntscr.com/image/k4UvkMezRUGDdO-8JOTQQg.png" />

### Server configuration

- Operating system: Debian 11
- Web server: Nginx 1.18
- Database version: 10.6.5-MariaDB-1:10.6.5
- PHP version: 8.0.13
- phpMyAdmin version: 5.1.1

### Client configuration

- Browser: Chrome 96.0.4664.45 x64
- Operating system: Windows 10

williamdes:
Thank you so much for the report @MordiSacks !

MordiSacks:
Just updating that as of chrome 98.0.4758.80 2FA via U2F is broken :(
Thank you for taking this issue in high priority

MordiSacks:
If anyone is looking for an easy temporary solution, latest firefox still supports U2F.

mauriciofauth:
- https://blog.chromium.org/2021/09/chrome-95-beta-secure-payment.html
> Chrome's legacy U2F API for interacting with security keys is deprecated and beginning a deprecation trial in Chrome 95 wherein the API remains enabled by default, but the trial token will disable the key for participating sites. U2F security keys themselves are not deprecated and will continue to work.
>
> Affected sites should migrate to the [Web Authentication API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Authentication_API). Credentials that were originally registered via the U2F API can be challenged via web authentication. USB security keys that are supported by the U2F API are also supported by the Web Authentication API.

MauricioFauth:
I'm working on adding support for WebAuthn.

williamdes:
> I'm working on adding support for WebAuthn.

Awesome !
Do not hesitate to PR to https://github.com/code-lts/U2F-php-server any back end code

williamdes:
[...]

PR #17972, opened 2022-12-17
Add support for Web Authentication API

MauricioFauth:
Adds a two factor authentication plugin that supports FIDO2/WebAuthn security keys.

The automatic migration from FIDO U2F to FIDO2 WebAuthn is not implemented yet.

- Fixes https://github.com/phpmyadmin/phpmyadmin/issues/17229

Some of the functions in `webauthn.js` file are copied from the `@web-auth/webauthn-helper` Yarn package. I'll probably rewrite them to avoid installing this package or I'll install another package. It's basically for encoding and decoding base64url strings.

@williamdes, Do you have more than one security key? If you do, could you please test using a different key?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -79,7 +79,8 @@
     "conflict": {
         "bacon/bacon-qr-code": "<2.0",
         "pragmarx/google2fa-qrcode": "<2.1",
-        "tecnickcom/tcpdf": "<6.4.4"
+        "tecnickcom/tcpdf": "<6.4.4",
+        "web-auth/webauthn-lib": "<3.3"
     },
     "suggest": {
         "ext-curl": "Updates checking",
@@ -93,7 +94,11 @@
         "tecnickcom/tcpdf": "For PDF support",
         "pragmarx/google2fa-qrcode": "^2.1 - For 2FA authentication",
         "bacon/bacon-qr-code": "^2.0 - For 2FA authentication",
-        "code-lts/u2f-php-server": "For FIDO U2F authentication"
+        "code-lts/u2f-php-server": "For FIDO U2F authentication",
+        "web-auth/webauthn-lib": "^3.3 - For FIDO2/WebAuthn authentication",
+        "ext-gmp": "For FIDO2/WebAuthn authentication",
+        "ext-bcmath": "For FIDO2/WebAuthn authentication",
+        "phpseclib/bcmath_compat": "For FIDO2/WebAuthn authentication"
     },
     "require-dev": {
         "bacon/bacon-qr-code": "^2.0",
@@ -111,7 +116,8 @@
         "squizlabs/php_codesniffer": "~3.6.0",
         "symfony/console": "^5.2.3",
         "tecnickcom/tcpdf": "^6.4.4",
-        "vimeo/psalm": "^4.22"
+        "vimeo/psalm": "^4.22",
+        "web-auth/webauthn-lib": "^3.3"
     },
     "extra": {
         "branch-alias": {
--- a/libraries/classes/TwoFactor.php
+++ b/libraries/classes/TwoFactor.php
@@ -13,13 +13,16 @@ use PhpMyAdmin\ConfigStorage\Relation;
 use PhpMyAdmin\Plugins\TwoFactor\Application;
 use PhpMyAdmin\Plugins\TwoFactor\Invalid;
 use PhpMyAdmin\Plugins\TwoFactor\Key;
+use PhpMyAdmin\Plugins\TwoFactor\WebAuthn;
 use PhpMyAdmin\Plugins\TwoFactorPlugin;
 use PragmaRX\Google2FAQRCode\Google2FA;
+use Webauthn\Server;
 use XMLWriter;
 
 use function array_merge;
 use function class_exists;
 use function extension_loaded;
+use function function_exists;
 use function in_array;
 use function ucfirst;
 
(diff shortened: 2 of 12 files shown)


williamdes:
> @williamdes, Do you have more than one security key? If you do, could you please test using a different key?

Yes, it's a recommendation. I will test using my two keys :)

composer.json
```diff
@@ -93,7 +94,11 @@
         "tecnickcom/tcpdf": "For PDF support",
         "pragmarx/google2fa-qrcode": "^2.1 - For 2FA authentication",
         "bacon/bacon-qr-code": "^2.0 - For 2FA authentication",
-        "code-lts/u2f-php-server": "For FIDO U2F authentication"
+        "code-lts/u2f-php-server": "For FIDO U2F authentication",
+        "web-auth/webauthn-lib": "^3.3 - For FIDO2/WebAuthn authentication",
```
williamdes, line 98: As it adds one more library and it costs a lot of maintenance for me, maybe we can pick the code or add the support on the actual library

libraries/classes/TwoFactor.php
```diff
@@ -135,6 +138,10 @@ class_exists(Google2FA::class)
             $result[] = 'application';
         }
 
+        if (class_exists(Server::class) && (function_exists('gmp_add') || function_exists('bcadd'))) {
```
williamdes, line 141: Should recommend ext bc and ext gmp

composer.json
MauricioFauth: What do you mean by pick the code?

libraries/classes/TwoFactor.php
MauricioFauth: The package uses GMP, and if it's not available, uses BCMath. There is no need to have both installed.

composer.json
MauricioFauth: I could try to write a simple implementation myself, to simplify the packaging, but we should add that package for master branch.

libraries/classes/TwoFactor.php
williamdes: Yes but composer.json should suggest them

composer.json
williamdes: This should also be for QA
We can maybe cherry pick or adapt the current library to handle webauthn

The cost of a new library to package in Debian is so high

If we can have only one library it would be awesome

MauricioFauth: > This should also be for QA

I meant writing a simple implementation for QA to avoid needing to package, but requiring the third party package in master.

> We can maybe cherry pick or adapt the current library to handle webauthn
> 
> The cost of a new library to package in Debian is so high
> 
> If we can have only one library it would be awesome 

I'm not sure if I'm following you here. What do you mean by having only one library?

libraries/classes/TwoFactor.php
MauricioFauth: I added them to suggest section of composer.json.

composer.json
williamdes: > I meant writing a simple implementation for QA to avoid needing to package, but requiring the third party package in master.

Okay, that can be a good alternative

> I'm not sure if I'm following you here. What do you mean by having only one library?

By changing the code in the u2f library 
As it is not old method specific, it can support webauthn too

Reducing all u2f code to one library on the back end

MauricioFauth: I was reading about how to implement a WebAuthn server and it will be hard to do using only dependencies that has Debian packages. I'll try it anyway. 
I think we could work to remove the u2f lib and use only WebAuthn lib, because WebAuthn has compatibility with U2F, but I think the opposite is not true.

williamdes: Thank you, let me know
Reducing it to only one library is a good idea, whatever the lib is

That's okay if I packaged a lib for nothing
It's better if I know it now before debian freezes

MauricioFauth: When it will freeze?

williamdes: Start of 2023, I found this article https://www.linuxadictos.com/en/they-already-released-the-debian-12-bookworm-freeze-date.html
It was twitter or mailing list announced if you want to see the official dates

I am currently pending acceptation for slim psr7 to be able to package 5.2
Yeah here we are ^^

MauricioFauth: Is there another PSR-7 debian package? Use can use those instead.

williamdes: yes, niholm's but we depend on some cool things from slim 

dont worry it should be accepted soon

williamdes:
I configured my key using Firefox and tried to use it on Brave

```
TypeError: Cannot read properties of undefined (reading 'get')
```

This is because you need to do like the other implementation, prevent if HTTPS is not active. It does not work without as far as i know

williamdes:
> @williamdes, Do you have more than one security key? If you do, could you please test using a different key?

There is no way to add a second one :/ So I can not test it for now ;)

MauricioFauth:
> > @williamdes, Do you have more than one security key? If you do, could you please test using a different key?
> 
> There is no way to add a second one in the preferences page :/ So I can not test it for now ;)

This is a phpMyAdmin's limitation. I meant test if using a different key to authenticate shows an error. Could you also test the custom server in https://github.com/phpmyadmin/phpmyadmin/pull/17989?

MauricioFauth:
Closing this in favor of https://github.com/phpmyadmin/phpmyadmin/pull/17989.

PR #17974, opened 2022-12-18
NodeDatabase memoization

kamil-tekiela:
This simple memoization improved the execution time by 38%. I didn't test all variants though. 

The getPresence method is called from at least 3 places. We should cache the result for the same arguments. 

Stalled by #17973 and WIP


williamdes (approved):
This seems pretty straightforward, should it go to QA @MauricioFauth?

MauricioFauth (approved):

PR #17989, opened 2022-12-22
Add support for Web Authentication API

MauricioFauth:
Adds a two factor authentication plugin that supports FIDO2/WebAuthn security keys.

- Related to https://github.com/phpmyadmin/phpmyadmin/issues/17229
- Related to https://github.com/phpmyadmin/phpmyadmin/pull/17972

This pull request implements a custom WebAuthn server. 
The WebAuthn server doesn't support [attestation](https://www.w3.org/TR/webauthn-3/#sctn-attestation) and some checks are still incomplete.
The CBOR decoder is not specification complete.

References:
- https://webauthn.guide/
- https://www.w3.org/TR/webauthn-3/
- https://www.rfc-editor.org/rfc/rfc7049
- https://developer.mozilla.org/en-US/docs/Web/API/Web_Authentication_API

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -93,7 +93,8 @@
         "tecnickcom/tcpdf": "For PDF support",
         "pragmarx/google2fa-qrcode": "^2.1 - For 2FA authentication",
         "bacon/bacon-qr-code": "^2.0 - For 2FA authentication",
-        "code-lts/u2f-php-server": "For FIDO U2F authentication"
+        "code-lts/u2f-php-server": "For FIDO U2F authentication",
+        "web-auth/webauthn-lib": "For better WebAuthn/FIDO2 authentication support"
     },
     "require-dev": {
         "bacon/bacon-qr-code": "^2.0",
@@ -111,7 +112,8 @@
         "squizlabs/php_codesniffer": "~3.6.0",
         "symfony/console": "^5.2.3",
         "tecnickcom/tcpdf": "^6.4.4",
-        "vimeo/psalm": "^4.22"
+        "vimeo/psalm": "^4.22",
+        "web-auth/webauthn-lib": "^3.3"
     },
     "extra": {
         "branch-alias": {
--- a/js/src/webauthn.js
+++ b/js/src/webauthn.js
@@ -1,0 +1,133 @@
+/**
+ * @param {ArrayBuffer} buffer
+ *
+ * @return {string}
+ */
+const arrayBufferToBase64 = buffer => {
+    const bytes = new Uint8Array(buffer);
+    let string = '';
+    for (const byte of bytes) {
+        string += String.fromCharCode(byte);
+    }
+
+    return window.btoa(string);
+};
+
+/**
+ * @param {string} string
+ *
+ * @return {Uint8Array}
+ */
+const base64ToUint8Array = string => {
+    return Uint8Array.from(window.atob(string), char => char.charCodeAt(0));
+};
+
+/**
+ * @param {JQuery<HTMLElement>} $input
+ *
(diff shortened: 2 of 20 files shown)


MauricioFauth:
@williamdes This PR does not have any third-party dependency. However, I'm thinking about using this custom server only as a fallback and adding `web-auth/webauthn-lib` as a suggested package for better WebAuthn support and using it if it's available. What do you think?

williamdes (approved):

williamdes:
https://github.com/phpmyadmin/phpmyadmin/actions/runs/3797400536/jobs/6458304595#step:11:783

williamdes (approved):
Can I test it once more ?

MauricioFauth:
@williamdes Could you test it?

There are some things still missing that can be done in other pull requests:
- The possibility to add more than one key.
- Automatic migration of U2F keys to use the new plugin. For now, users that had register with the old API can use Firefox to login, deactivate the U2F plugin and enable the WebAuthn plugin to use with other browsers.

composer.json
```diff
@@ -93,7 +93,8 @@
         "tecnickcom/tcpdf": "For PDF support",
         "pragmarx/google2fa-qrcode": "^2.1 - For 2FA authentication",
         "bacon/bacon-qr-code": "^2.0 - For 2FA authentication",
-        "code-lts/u2f-php-server": "For FIDO U2F authentication"
+        "code-lts/u2f-php-server": "For FIDO U2F authentication",
+        "web-auth/webauthn-lib": "For better WebAuthn/FIDO2 authentication support"
```
williamdes, line 97: You should add this to PACKAGE_LIST on `create-release.sh`

williamdes:
> Two-factor authentication failed: Expected a different value than "".

Without HTTPS on http://8.2.local/@phpmyadmin/phpMyAdmin-QA_5_2/index.php?route=/preferences/two-factor

On click on the button below "Please connect your WebAuthn/FIDO2 device. Then confirm registration on the device."

williamdes:
image
Same on Firefox

williamdes:
> > image Same on Firefox
> 
> I can't replicate this. Looks like there is something wrong with your JavaScript. Did you compile it and updated the cache?

Ah, sorry
I forgot to do this
Will test it soon

williamdes:
This error still exists when you persist on clicking the enable button.
That said the feature works fine, awesome job ! 🎉 


image

I think you can merge this PR after fixing this error so testing can be done on a larger scale

MauricioFauth:
> 



> This error still exists when you persist on clicking the enable button. That said the feature works fine, awesome job ! tada
> 
> image
> 
> I think you can merge this PR after fixing this error so testing can be done on a larger scale

I'm not sure why this button is appearing to you, since it's getting hidden with JS, but I fixed the error message anyway.

williamdes:
> I'm not sure why this button is appearing to you, since it's getting hidden with JS, but I fixed the error message anyway.

Now it is fixed (the error) but the button only appears when I am using HTTP now. So I can click it but at least it gives an consistent error each time.

One last thing is to add the optional package to PACKAGE_LIST on create-release.sh and use the check release excludes script but nothing should show up

williamdes:
The snapshot was re-built right now, thank you for your awesome work on this !

If possible an error should be show on chrome and not nothing when the user tries to login using the old U2F method ?

MauricioFauth:
> The snapshot was re-built right now, thank you for your awesome work on this !

Thank you!

> If possible an error should be show on chrome and not nothing when the user tries to login using the old U2F method ?

This will done in another PR.

PR #18007, opened 2023-01-07
Fix #18006 - UUID columns can't be moved

### LINKED ISSUE #18006: Moving columns causes the default uuid() value to get quotes around it
### Describe the bug

When I set a default value for a column that is `UUID` to `uuid()`, and use PHPMyAdmin to move it in the list of columns, and then use `show create table X`, I see that the `uuid()` would become `'uuid()'` and if I move it one more time, it becomes `'''uuid()'''`.

### To Reproduce

- Create a table with at least 3 columns
- Set two columns data types to be UUID
- Set default values for those two columns to `uuid()`
- Then use `Move column` feature to change the order of columns
  - Select the table
  - Go to `Structure` tab
  - Use `Move column` link (if you can't find it, search it inside the page)
- Then use `show create table X`
- You should see quotes around the default value
- If you move columns multiple times, then those quotes will increase in number

### Expected behavior

The default value `uuid()` should remain as is when we move columns (change the order of columns)

### Screenshots

From left hand (default state) we get to the right hand after moving columns around

Screenshot from 2023-01-07 16-19-15

### Server configuration

- Operating system:
- Web server:
- Database version:
- PHP version: MariaDB
- phpMyAdmin version: latest

### Client configuration

- Browser: Chrome
- Operating system:

### Additional context

liviuconcioiu:
Signed-off-by: Liviu-Mihail Concioiu <liviu.concioiu@gmail.com>

### Description

This PR fixes the issue when uuid columns can't be moved.

Fixes #18006

### DIFF
--- a/libraries/classes/Controllers/Table/Structure/MoveColumnsController.php
+++ b/libraries/classes/Controllers/Table/Structure/MoveColumnsController.php
@@ -81,11 +81,17 @@ final class MoveColumnsController extends AbstractController
             $timeDefault = $data['Default'] === 'CURRENT_TIMESTAMP' || $data['Default'] === 'current_timestamp()';
             $current_timestamp = $timeType && $timeDefault;
 
+            $uuidType = $data['Type'] === 'uuid';
+            $uuidDefault = $data['Default'] === 'UUID' || $data['Default'] === 'uuid()';
+            $uuid = $uuidType && $uuidDefault;
+
             // @see https://mariadb.com/kb/en/library/information-schema-columns-table/#examples
             if ($data['Null'] === 'YES' && in_array($data['Default'], [$defaultNullValue, null])) {
                 $default_type = 'NULL';
             } elseif ($current_timestamp) {
                 $default_type = 'CURRENT_TIMESTAMP';
+            } elseif ($uuid) {
+                $default_type = 'UUID';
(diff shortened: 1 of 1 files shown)


libraries/classes/Controllers/Table/Structure/MoveColumnsController.php
```diff
@@ -81,11 +81,17 @@ public function __invoke(): void
             $timeDefault = $data['Default'] === 'CURRENT_TIMESTAMP' || $data['Default'] === 'current_timestamp()';
             $current_timestamp = $timeType && $timeDefault;
 
+            $uuidType = $data['Type'] === 'uuid';
+            $uuidDefault = $data['Default'] === 'UUID' || $data['Default'] === 'uuid()';
+            $uuid = $uuidType && $uuidDefault;
```
kamil-tekiela, line 86: Could these 3 lines be extracted as a new method?

williamdes (approved):

libraries/classes/Controllers/Table/Structure/MoveColumnsController.php
williamdes: I am merging this as it is, the above code looks alike
On master branch this could be improved

PR #18009, opened 2023-01-11
Bugfix for changing $cfg['RepeatCells'] no effect

still-dreaming-1:
Fixes changing $cfg['RepeatCells'] in configuration has no effect.

I'm not sure if this is a good way to fix this or not, but it does fix this problem for me in my local environment. I'm not sure what the purpose of checking `empty($query['repeat_cells'])` was in the original code.

Signed-off-by: still-dreaming-1 <jessedickey@gmail.com>

### DIFF
--- a/libraries/classes/Display/Results.php
+++ b/libraries/classes/Display/Results.php
@@ -3460,9 +3460,7 @@ class Results
 
         $query['sql'] = $this->properties['sql_query'];
 
-        if (empty($query['repeat_cells'])) {
-            $query['repeat_cells'] = $GLOBALS['cfg']['RepeatCells'];
-        }
+        $query['repeat_cells'] = $GLOBALS['cfg']['RepeatCells'];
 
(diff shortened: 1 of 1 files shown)


still-dreaming-1:
Actually, I thought I wanted this setting to work, but now that it does, I don't like it. I just realized the header floats at the top as I scroll down the list of results anyways, so I don't need this setting. I guess I would like to be able to disable it if it is not already disabled. I'm not sure if the original code had it disabled or stuck at the default setting of 100.

williamdes:
Thank you for this patch, we need to have a deeper look before merging this

still-dreaming-1:
Yeah, I'm not sure it makes sense anymore. Maybe the setting should be removed entirely in favor of the floating header. I think it doesn't work with some of the other PhpMyAdmin themes though.

williamdes:
Let's keep it open until it's debugged :)

PR #18023, opened 2023-01-20
Create `Dbal/Statement` interface

MauricioFauth:
Wraps `mysqli_stmt` class with the `Dbal\MysqliStatement` class.

It's not feature complete and only covers what's being used now.

### DIFF
--- a/libraries/classes/Server/Privileges.php
+++ b/libraries/classes/Server/Privileges.php
@@ -7,7 +7,6 @@ declare(strict_types=1);
 
 namespace PhpMyAdmin\Server;
 
-use mysqli_stmt;
 use PhpMyAdmin\ConfigStorage\Features\ConfigurableMenusFeature;
 use PhpMyAdmin\ConfigStorage\Relation;
 use PhpMyAdmin\ConfigStorage\RelationCleanup;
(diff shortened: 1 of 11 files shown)


williamdes (approved):

kamil-tekiela:
I think it would be a good idea to call it `PreparedStatement` instead of `Statement` to avoid confusion.

libraries/classes/Server/Privileges.php
```diff
@@ -3789,13 +3785,12 @@ private function getUserPrivileges(string $user, string $host, bool $hasAccountL
         $userPrivileges['account_locked'] = 'N';
 
         $query = 'SELECT * FROM `mysql`.`global_priv` WHERE `User` = ? AND `Host` = ?;';
-        /** @var mysqli_stmt|false $statement */
         $statement = $this->dbi->prepare($query);
-        if ($statement === false || ! $statement->bind_param('ss', $user, $host) || ! $statement->execute()) {
+        if ($statement === null || ! $statement->execute([$user, $host])) {
```
kamil-tekiela, line 3789: I wonder when would these two statements ever fail? The syntax looks correct and isn't something that I would ever expect to work conditionally.

MauricioFauth: You are right. I tried to not change too much.
When `$query` has an error, `$statement` will be `null`, and, when passing a different number of parameters, `execute` will fail, for example. But these should be treated as exceptions instead.

PR #18029, opened 2023-01-23
Refactor config forms

kamil-tekiela:
Improves SA for these classes. 
This is working towards the goal of running Psalm 5.6 someday with as few false positives as possible.


MauricioFauth (approved):

MauricioFauth:
:100:

PR #18046, opened 2023-01-28
Remove globals from create controller

kamil-tekiela:
Remove local globals.

### DIFF
--- a/libraries/classes/Controllers/View/CreateController.php
+++ b/libraries/classes/Controllers/View/CreateController.php
@@ -65,7 +65,6 @@ class CreateController extends AbstractController
         $GLOBALS['urlParams'] = $GLOBALS['urlParams'] ?? null;
 
         $GLOBALS['message'] = $GLOBALS['message'] ?? null;
-        $GLOBALS['systemDb'] = $GLOBALS['systemDb'] ?? null;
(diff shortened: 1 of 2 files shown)


libraries/classes/Controllers/View/CreateController.php
```diff
@@ -36,6 +36,22 @@ class CreateController extends AbstractController
     /** @var DatabaseInterface */
     private $dbi;
 
+    private const VIEW_SECURITY_OPTIONS = [
```
williamdes, line 39: This like some stuff that should be on the parser repository
Can you add some todo phpdocs?
All this weird code that rebuilds view definitions should be on the parser. And it probably already is :)

kamil-tekiela: Makes sense, but is a TODO comment good idea? They are rarely revisited. Maybe a GH issue would be a better idea?

MauricioFauth (approved):

libraries/classes/Controllers/View/CreateController.php
williamdes: I think a todo would be better than nothing, an issue is a great idea for new contributors

PR #18047, opened 2023-01-29
Move getColumnMapFromSql() to SystemDatabase

kamil-tekiela:


libraries/classes/SystemColumn.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpMyAdmin;
+
+class SystemColumn
```
williamdes, line 7: It could have psalm sealed or readonly phpdoc tags that other classes have?

williamdes (approved):

libraries/classes/SystemColumn.php
kamil-tekiela: Good idea. I added `@psalm-immutable`

MauricioFauth (approved):

libraries/classes/SystemColumn.php
```diff
@@ -0,0 +1,25 @@
...
+
+namespace PhpMyAdmin;
+
+/**
+ * @psalm-immutable
+ */
+class SystemColumn
+{
+    /** @var string */
+    public $tableName;
```
MauricioFauth, line 13:
```suggestion
    public readonly string $tableName;
```

MauricioFauth (changes requested):

libraries/classes/SystemColumn.php
kamil-tekiela: Yeah, this is completely unnecessary now 😄

MauricioFauth (approved):

Issue #16229, opened 2020-06-25
Create a public directory for public files


MauricioFauth:
By moving public files to a public directory, we improve the security of phpMyAdmin because users will not have access to files that are outside the public directory.

Related to #14814.

williamdes:
Fixes [Debian #583588](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=583588) reported 11 years ago by @OlafvdSpek

OlafvdSpek:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=583588 is marked as fixed-upstream now?

OlafvdSpek:
Ah..

> Since phpMyAdmin 5.0.0 all public files have been moved to the

Did this actually ship with 5.0? I can't find the public dir in my 5.1 snapshot.

williamdes:
This is in progress and should be done soon but I do not know when.
That said we are close to be able to ship it

And I emailed control to remove the wrong tag

williamdes:
@MauricioFauth what is the current state of this?
There is not much entrypoints left. Two will be kept: index.php in root and setup directories. Only `show_config_errors.php` is the left over to disappear before this issue is near to happen?

PR #18078, opened 2023-02-10
[RFC] Change phpMyAdmin version from 5.3 to 6.0 and bump min PHP version

### LINKED ISSUE #15607: Bump minimum version to PHP 7.4
We should start to discuss when we want to drop the support for PHP 7.1, PHP 7.2 and PHP 7.3, because PHP 7.1 is no longer supported and PHP 7.2 is supported for critical security issues only.

PHP 7.4 will be the last 7.x release before the [PHP 8](https://news-web.php.net/php.internals/103862).

Another reason to chose PHP 7.4, besides the new features, is the fix for the MySQL authentication issue.

https://www.php.net/supported-versions.php

### LINKED ISSUE #17953: Plans to drop support for PHP 7.2 and 7.3
Now that PHP 8.2 is released, does phpMyAdmin have any plans to drop support for ancient PHP versions such as 7.2 and 7.3? I understand supporting 7.4 for the time being as the jump from 7.4 to 8.0 might be too big for certain users (even though phpMyAdmin is not coupled with any other applications they may have, they might not be ready to install a new PHP version yet just for PMA). 

From the development perspective, supporting these old versions is holding us back. For example, we cannot use property types. We also cannot use new features such as `match`, Enums, attributes, etc. We have to constantly think of what limitations these old PHPs have. We are also limited to old dependencies, some of which may no longer receive PHP 7.2 updates. 

Dropping support for very old PHP versions is also good for PMA users. These old PHP versions aren't patched anymore and new security vulnerabilities are still found. By dropping the support, we give users another push to upgrade to more secure environments.

MauricioFauth:
This pull request changes the phpMyAdmin version in the `master` branch from `5.3.0-dev` to `6.0.0-dev` and bumps the minimum PHP version from `7.2.5` to `8.1.0`.

If this PR got merged, that means that phpMyAdmin 6.0.0 will replace the phpMyAdmin 5.3.0. The reason is that I feel that phpMyAdmin 5.2 should be supported for a longer time and that the development version in the `master` branch (`5.3.0-dev`) should have a longer development time. So, as there are some features planned for phpMyAdmin 6.0, makes sense to merge 5.3 and 6.0 and drop the 5.3 version.

With phpMyAdmin 6.0 been a new major version, I think we should drop support for old PHP versions and require at least PHP 8.1.0. The reason is that PHP 8.0 is not actively supported anymore and security support will end in November, and PHP 8.1 has a lot of benefits like enumerations and read-only properties, and the new features from PHP 7.3, 7.4 and 8.0 as well.

- Fixes https://github.com/phpmyadmin/phpmyadmin/issues/17953
- Fixes https://github.com/phpmyadmin/phpmyadmin/issues/15607

### DIFF
--- a/.github/workflows/daily-snapshots.yml
+++ b/.github/workflows/daily-snapshots.yml
@@ -22,7 +22,7 @@ jobs:
       matrix:
         include:
           - { version: '5.2', branch: 'QA_5_2', php-version: '7.2', node-version: '12', python-version: '3.7' }
-          - { version: '5.3', branch: 'master', php-version: '7.2', node-version: '14', python-version: '3.7' }
+          - { version: '6.0', branch: 'master', php-version: '8.1', node-version: '14', python-version: '3.7' }
     steps:
       - name: Check out code
         uses: actions/checkout@v3
--- a/.github/workflows/lint-and-analyse-php.yml
+++ b/.github/workflows/lint-and-analyse-php.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        php-version: ["7.2"]
+        php-version: ["8.1"]
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
@@ -65,7 +65,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        php-version: ["7.2"]
+        php-version: ["8.1"]
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
--- a/.github/workflows/other-tools.yml
+++ b/.github/workflows/other-tools.yml
@@ -33,7 +33,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        php-version: ["7.2"]
+        php-version: ["8.1"]
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,13 +20,13 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - { php-version: '8', experimental: false, arch: 'amd64', exclude-phpunit-groups: 'extension-iconv' }
-          - { php-version: '8', experimental: false, arch: 'arm64v8', exclude-phpunit-groups: 'extension-iconv' }
-          - { php-version: '8', experimental: false, arch: 'arm32v7', exclude-phpunit-groups: 'extension-iconv,32bit-incompatible' }
-          - { php-version: '8', experimental: false, arch: 'arm32v6', exclude-phpunit-groups: 'extension-iconv,32bit-incompatible' }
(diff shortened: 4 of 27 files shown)


williamdes (approved):
I agree with that

and with the routing change

williamdes (approved):

ibennetch (approved):
This is a thorough pull request, nicely done.

I guess this makes sense, but part of me is slightly unhappy about jumping to version 6 with few internal changes. I think on some level, many people still expect a new major software version to have some major changes itself (as opposed to just dropping old PHP support that isn't even generally active).

It's hard to decide on version numbers when chasing major PHP, MySQL, and MariaDB versions because they're all independently adding and removing aspects all the time.

So I think this is the right time to do this, but k do hesitate a little bit saying that.

williamdes:
> I guess this makes sense, but part of me is slightly unhappy about jumping to version 6 with few internal changes. I think on some level, many people still expect a new major software version to have some major changes itself (as opposed to just dropping old PHP support that isn't even generally active).

We will have the routing change that will also be a major change ;)
See: #16229

MauricioFauth:
> I guess this makes sense, but part of me is slightly unhappy about jumping to version 6 with few internal changes. I think on some level, many people still expect a new major software version to have some major changes itself (as opposed to just dropping old PHP support that isn't even generally active).

I did only the minimal required changes for this pull request. We will do more changes later.
Regarding major changes, the main one, as William mentioned, is the directory structure change that will introduce the public directory (#16229).

williamdes:
could we also rename the default branch to 6.x or something else?
It should be transparent if done from the GitHub UI as they will handle the transition for users

MauricioFauth:
> could we also rename the default branch to 6.x or something else?

Do you mean rename `master` to `6.x`? I don't like that for phpMyAdmin, as it doesn't follow SemVer. However, I agree with renaming `master` to `main` and `QA_5_2` to `5.2.x`, but we have to be careful with the tooling.

ibennetch:
I have been thinking about this. 

I prefer to keep QA_5_2 until we have released the next version, because that's mostly what we have done and there are times when we want to make a big fox release from the QA branch even before 6.0 would be ready. _Historically speaking_, people preferred to vet and test new minor versions (the `y` of x.y.z) which is why we often did bug fixes only on patch releases (the `z`) and any new development on the minor branches. That's all a little more blurry because (1) people are understanding of software to be moving a little more quickly and (2) our code is more robust (with many tools to help check quality and more modernized processes).

So this is an interesting thought about whether it's time to close QA_5_2 and my vote is no, I'd rather keep it going for a bit longer, but to continue pressing on for a alpha release of 6.0 maybe sometime this spring.

I know it's mostly just rambling, but those are my thoughts.

williamdes:
> So this is an interesting thought about whether it's time to close QA_5_2 and my vote is no, I'd rather keep it going for a bit longer, but to continue pressing on for a alpha release of 6.0 maybe sometime this spring.

Yup, let's keep 5.2 living for a bit more time since 6.0 will drop quite a lot of PHP versions.
I think we can not shutdown 5.2 any time soon for that reason

PR #18100, opened 2023-02-15
Refactor gis scale data set calculation

kamil-tekiela:
One bug less. 
I don't understand why this was written in such a complicated way, but hopefully it's much clearer after adding a value object. The VO is immutable, but I had to indicate a lack of value so I used null union (although the actual result should always be a VO and never null). 
This started with me trying to add an array shape, but I uncovered a bug and decided to fix it. I don't think it's worth the effort to backport it to 5.2 as this is only a minor visual bug. I added a regression test for it.

### DIFF
--- a/libraries/classes/Gis/GisGeometry.php
+++ b/libraries/classes/Gis/GisGeometry.php
@@ -99,9 +99,9 @@ abstract class GisGeometry
      *
      * @param string $spatial spatial data of a row
      *
-     * @return array array containing the min, max values for x and y coordinates
+     * @return ScaleData|null min, max values for x and y coordinates
      */
-    abstract public function scaleRow($spatial): array;
+    abstract public function scaleRow(string $spatial): ?ScaleData;
 
     /**
      * Generates the WKT with the set of parameters passed by the GIS editor.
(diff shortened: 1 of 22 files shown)


MauricioFauth (approved):

williamdes:
>  I don't think it's worth the effort to backport it to 5.2 as this is only a minor visual bug.

It's probably worth it is its noticeable, can you provide a screenshot?

kamil-tekiela:
I would have to record it but I don't think Windows has any built-in tools to record a video. 

I can provide you with the test table and steps:
```
CREATE TABLE IF NOT EXISTS `gis` (
  `a` polygon NOT NULL,
  `b` multipolygon NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=5 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_turkish_ci;

INSERT INTO `gis` (`id`, `a`, `b`) VALUES
(2, 0x00000000010300000002000000040000000000000000000000000000000000000000000000000000000000000000000840000000000000084000000000000000000000000000000000000000000000000004000000000000000000f03f000000000000f03f000000000000f03f00000000000000400000000000000040000000000000f03f000000000000f03f000000000000f03f, 0x0000000001060000000100000001030000000200000005000000000000000000000000000000000000000000000000000000000000000000084000000000000008400000000000000840000000000000084000000000000000000000000000000000000000000000000005000000000000000000f03f000000000000f03f000000000000f03f0000000000000040000000000000004000000000000000400000000000000040000000000000f03f000000000000f03f000000000000f03f),
(3, 0x00000000010300000002000000040000000000000000000000000000000000000000000000000000000000000000000840000000000000084000000000000000000000000000000000000000000000000004000000000000000000f03f000000000000f03f000000000000f03f00000000000000400000000000000040000000000000f03f000000000000f03f000000000000f03f, 0x00000000010600000001000000010300000002000000050000000000000000002440000000000000244000000000000024400000000000002a400000000000002a400000000000002a400000000000002a40000000000000244000000000000024400000000000002440050000000000000000002640000000000000264000000000000026400000000000002840000000000000284000000000000028400000000000002840000000000000264000000000000026400000000000002640);
```

Steps:
1. Browse this table
2. Click "visualize GIS data"
3. Select `b` column as spatial column
4. Observe that only one square is visible. The second one is hidden off-screen due to invalid zoom.

williamdes:
Okay, that might be noticeable then
Can you make a QA fix for it please?

PS: greenshot is an awesome screenshot tool, but not for videos unfortunately

PR #18101, opened 2023-02-16
Stop using PHP_EOL

kamil-tekiela:
On Windows, these cause a lot of tests to fail. The failures are bogus IMHO but either the tests need to be adapted for cross-OS compatibility or we should stop using `PHP_EOL` and stick with `"\n"`. I assume it is safe to do this but please correct me if I am wrong. For example, one import class actually used it (in a stupid way though). I don't know if the Git class needs it or not. The Config class used it to generate a config file which I think should be created with only `"\n"`.

### DIFF
--- a/libraries/classes/Git.php
+++ b/libraries/classes/Git.php
@@ -42,7 +42,6 @@ use function trim;
 use function unpack;
 
 use const DIRECTORY_SEPARATOR;
-use const PHP_EOL;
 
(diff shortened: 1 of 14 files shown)


williamdes (approved):
I think it's okay if this suite passes on Windows

MauricioFauth (approved):

Issue #18105, opened 2023-02-16
Rename database has multiple bugs


kamil-tekiela:
During some testing, I observed the following problem when renaming one database to another:
- Tables whose names are made of only digits do not work properly
- Database containing views break during renaming and refuse to move the view. 
- If the database or table name is a zero-string (i.e. `'0'`) then renaming will fail. 
- Renaming a database with a lot of tables is excruciatingly slow. I lost patience and stopped it so I don't know if or when it would finish. 
- When navigating from another tab into the operations tab, the confirmation modal for renaming is shown off-screen. You must refresh the page and try renaming again to see the modal. 

I am using this issue as a TODO list because I am already starting to forget what some of the bugs were. I will try to fix them one by one, but if anyone wants to help, please submit a PR.

PR #18106, opened 2023-02-16
Fix renaming db with a view

kamil-tekiela:
Re point number 2 in #18105
The fix is definitely working, but I cannot get the test to work.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -16554,7 +16554,8 @@
     </UndefinedMethod>
   </file>
   <file src="test/classes/TableTest.php">
-    <InvalidArrayOffset occurrences="1">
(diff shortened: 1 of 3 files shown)


psalm-baseline.xml
```diff
@@ -1,5 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <files psalm-version="4.30.0@d0bc6e25d89f649e4f36a534f330f8bb4643dd69">
+  <file src="config.inc.php">
```
williamdes, line 3: can you remove this block?

williamdes (approved):

MauricioFauth (approved):

PR #18116, opened 2023-02-17
Remove strlen when used in a condition

kamil-tekiela:


williamdes (approved):

libraries/classes/SqlQueryForm.php
```diff
@@ -145,7 +144,7 @@ public function init($query): array
                     : $GLOBALS['cfg']['Servers'][$GLOBALS['server']]['host']
                 )
             );
-        } elseif (strlen($GLOBALS['table']) === 0) {
```
williamdes, line 148: maybe this was also a null check safety, I would be cautious about that with GLOBALS

williamdes: Ha but strlen null would type fail, nevermind then

williamdes (approved):

MauricioFauth (approved):

PR #18126, opened 2023-02-20
Add native return type hints

kamil-tekiela:
1/3. Return types are first. Hopefully, type hints were mostly correct. This was done fully automatically. I will test as much as I can tomorrow and fix any CI issues if there are any.


MauricioFauth (approved):
:100:

PR #18144, opened 2023-02-23
Fix bug preventing opening information_schema

kamil-tekiela:


templates/table/structure/display_table_stats.twig
```diff
@@ -117,7 +117,7 @@
                 {% if not is_innodb and showtable['Rows'] is defined %}
                     <tr>
                     <th class="name">{% trans 'Rows' %}</th>
-                    <td class="value">{{ format_number(showtable['Rows'], 0) }}</td>
+                    <td class="value">{{ format_number(showtable['Rows'] ? 0, 0) }}</td>
```
williamdes, line 120:
```suggestion
                    <td class="value">{{ format_number(showtable['Rows'] ?? 0, 0) }}</td>
```
Does twig support ?? rather than ?
Weiiirrd syntax 😄

kamil-tekiela: Ohh yeah, you are right. I used the wrong operator. Twig has some weird syntax https://twig.symfony.com/doc/3.x/templates.html#other-operators

williamdes (approved):

MauricioFauth (approved):

williamdes:
So this bug does not exist on QA?

kamil-tekiela:
I don't think so. I think this was related to one of the refactorings on 6.0. I can't reproduce it on 5.2.

PR #18149, opened 2023-02-24
Remove $GLOBALS['PMA_PHP_SELF']

kamil-tekiela:
PMA does not need this to be a global. Only a few places actually used it and two of them don't actually need it IMHO. I moved the method to Config and made it return a string. 
Please review this carefully as I am not entirely sure that I fully understood previous logic.
I also need your help here as the tests are failing and I do not know why we even used `parse_url` there. 
The URL in format `public/index.php/foo/bar` doesn't work, but it never worked so that's not an issue.


MauricioFauth:
> PMA does not need this to be a global. Only a few places actually used it and two of them don't actually need it IMHO. I moved the method to Config and made it return a string. Please review this carefully as I am not entirely sure that I fully understood previous logic. I also need your help here as the tests are failing and I do not know why we even used `parse_url` there. The URL in format `public/index.php/foo/bar` doesn't work, but it never worked so that's not an issue.

I don't know about this variable too much. If I remember correctly, this was used for the various entry point files that were used before the route system was introduced. I'm not entirely sure, but I think it's always `'index.php'` now and it can probably be completely removed.

kamil-tekiela:
Yeah, that's my thinking too. But the two usages I left are still useful. For example, if you have the demo server path containing `public/` now then getting the root path makes sense. Also, I assume many people would have PMA installed locally in a subfolder, so this logic needs to stay.

libraries/classes/Config.php
```diff
@@ -863,6 +863,56 @@ public function getRootPath(): string
         return implode('/', $parts);
     }
 
+    /**
+     * PATH_INFO could be compromised if set, so remove it from PHP_SELF
+     * and provide a clean PHP_SELF here
+     */
+    private function cleanupPathInfo(): string
```
williamdes, line 870: i'd say this should be in the Routing class

kamil-tekiela: I moved it to Routing class.

williamdes (approved):

libraries/classes/Config.php
```diff
@@ -856,7 +856,10 @@ public function getRootPath(): string
             $parts = array_slice($parts, 0, count($parts) - 1);
         }
 
-        $parts[] = '';
+        // Add one more part to make the path end in slash unless it already ends with slash
+        if (count($parts) < 2 || $parts[array_key_last($parts)] !== '') {
+            $parts[] = '';
+        }
```
williamdes, line 862: can you add test cases for that ?

kamil-tekiela: Done.

williamdes (approved):

MauricioFauth (approved):

PR #18151, opened 2023-02-25
Refactor ol code generation

MoonE:
This makes the generated OpenLayers js code and the php code a lot smaller.


libraries/classes/Gis/GisGeometry.php
```diff
@@ -247,150 +220,114 @@ protected function extractPoints($point_set, $scale_data, $linear = false): arra
...
-    protected function getLineForOpenLayers(
-        array $points_arr,
-        int $srid,
-        $is_line_string = true,
-    ): string {
-        return 'new ol.geom.'
-        . ($is_line_string ? 'LineString' : 'LinearRing') . '('
-        . $this->getPointsArrayForOpenLayers($points_arr, $srid)
-        . ')';
+    protected function extractPoints3(string $wktCoords, array|null $scale_data): array
```
williamdes, line 301: Can't we use a better name ?

williamdes (approved):

libraries/classes/Gis/GisGeometry.php
MoonE: Well, naming things is hard. The numbers represent the different dimensions of the in- / output. I made a small improvement to make it more clear.

MauricioFauth (approved):

Issue #18117, opened 2023-02-17
Max length is ignored for input values


liviuconcioiu:
### Describe the bug

Max length is ignored for input values. (I remember it worked a few months ago).

### Expected behavior

The length of a value should not exceed the max length.

### Screenshots

https://user-images.githubusercontent.com/25424343/219812485-58139cd2-cfac-4001-9f63-a4dde2105906.mp4

### Server configuration

- phpMyAdmin version: 5.2.2-dev, 6.0.0-dev

OlafvdSpek:
IMO the HTML max length property is broken by design.. input should NOT be silently ignored / truncated.

liviuconcioiu:
> IMO the HTML max length property is broken by design.. input should NOT be silently ignored / truncated.

Then, there's another bug, because this message doesn't appear.

![missing](https://user-images.githubusercontent.com/25424343/219856154-d0d05d9a-068b-4d77-9f16-18d5436994f0.png)

nyalex:
Looks like this comment was added 6 years ago on an issue number that I can't find, but it is relevant: https://github.com/nyalex/phpmyadmin/blob/master/libraries/classes/InsertEdit.php#L2082

Anyone know?

kamil-tekiela:
HTML's `maxlength` property is stupidly badly designed as it counts the length in UTF-16 code units. When using it with a tool like PMA where the charset can be a lot of different things, it would either limit the input too much or not enough.

liviuconcioiu:
> Looks like this comment was added 6 years ago on an issue number that I can't find, but it is relevant: https://github.com/nyalex/phpmyadmin/blob/master/libraries/classes/InsertEdit.php#L2082
> 
> Anyone know?

https://sourceforge.net/p/phpmyadmin/bugs/2367/

nyalex:
@kamil-tekiela Thanks for the clarification - makes sense why now. I'm a newcomer here and didn't know about that older system.

I believe I have fixed the bug @liviuconcioiu mentioned for varchars and vars. PR is here: https://github.com/phpmyadmin/phpmyadmin/pull/18163

PR #18163, opened 2023-02-26
Ref #18117 - Fix error message for some field times when inserting

nyalex:
Refactored change.js so that varchar ("CHAR") fields show an error message immediately on the insert screen when going over the character limit. Also created logic to check for valid numeric fields for int, decimal, and float fields.

Ref: #18117

### DIFF
--- a/js/src/table/change.js
+++ b/js/src/table/change.js
@@ -294,6 +294,7 @@ function verificationsAfterFieldChange (urlField, multiEdit, theType) {
     }
 
     // Remove possible blocking rules if the user changed functions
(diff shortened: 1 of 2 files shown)


OlafvdSpek:
Does this take into account input transformations?

nyalex:
> Does this take into account input transformations?

My understanding is that transformations are rendered on the browse page only. Insert page still validates against the field type/length. This JS change should be similar to the backend validation upon submit. Happy to be corrected!

OlafvdSpek:
That sounds like output transforms, not input transforms.

PR #18177, opened 2023-02-27
Fix "IS NULL" is shown for non-nullable columns on search page

kamil-tekiela:
Furthermore, this is a bug preventing adding typed params.

### DIFF
--- a/libraries/classes/Controllers/Table/SearchController.php
+++ b/libraries/classes/Controllers/Table/SearchController.php
@@ -157,7 +157,7 @@ class SearchController extends AbstractController
             }
 
             $this->columnTypes[] = $type;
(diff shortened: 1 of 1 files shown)


williamdes (approved):

williamdes:
> Furthermore, this is a bug preventing adding typed params.

Can you explain a bit more or was it related to another PR?

kamil-tekiela:
> > Furthermore, this is a bug preventing adding typed params.
> 
> Can you explain a bit more or was it related to another PR?

No PR yet. I am working on adding native type to params. However, many existing type hints are wrong or inaccurate. Before we add the native types, we should tidy up certain things.

PR #18179, opened 2023-02-28
QA Fix gis extent calculation

MoonE:
https://github.com/phpmyadmin/phpmyadmin/pull/18100#issuecomment-1449019663

PR #18181, opened 2023-03-01
Backport: Refactor gis scale data set calculation refactoring #18100 

MoonE:
Maybe replaces #18179


williamdes:
I feel this version is heavy for QA, but maybe it's the best version to kill the bugs that where found

For me #18179 is the right one to merge into QA, and this one should go into 6.x

What do you thing @MauricioFauth ?

MauricioFauth:
> I feel this version is heavy for QA, but maybe it's the best version to kill the bugs that where found
> 
> For me #18179 is the right one to merge into QA, and this one should go into 6.x

I agree. This is too much refactoring for QA.

PR #18182, opened 2023-03-01
Always use svg display in gis editor

MoonE:
Editing geometries with SRID should display with OpenLayers visualization. As that doesn't work this now always uses the svg viewer.

Fixing the OpenLayers visualization here would be rather complicated.

Will provide a complete fix for master branch shortly ...

### DIFF
--- a/templates/gis_data_editor_form.twig
+++ b/templates/gis_data_editor_form.twig
@@ -11,16 +11,14 @@
         {{ get_hidden_inputs() }}
 
         {# Visualization section #}
-        <!-- Always uses svg-view here added `and false` as a quick-fix -->
-        <div id="placeholder"{{ srid != 0 and false ? ' class="hide"' }}>
+        <div id="placeholder">
             {{ visualization|raw }}
         </div>
 
         <div id="openlayersmap" style="width: {{width}}px; height: {{height}}px;"{{ srid == 0 ? ' class="hide"' }}></div>
 
         <div class="choice float-end">
-            <!-- Always uses svg-view here added `and false` as a quick-fix -->
-            <input type="checkbox" id="choice" value="useBaseLayer"{{ srid != 0 and false ? ' checked="checked"' }}>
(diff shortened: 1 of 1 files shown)


williamdes (approved):
This is an acceptable workaround but feels weird
Is there no way it could be fixed in a proper way, is it too broken?

MoonE:
I'll have a look.

templates/gis_data_editor_form.twig
```diff
@@ -11,14 +11,16 @@
         {{ get_hidden_inputs() }}
 
         {# Visualization section #}
-        <div id="placeholder"{{ srid != 0 ? ' class="hide"' }}>
+        <!-- Always uses svg-view here added `and false` as a quick-fix -->
+        <div id="placeholder"{{ srid != 0 and false ? ' class="hide"' }}>
```
MauricioFauth, line 15:
```suggestion
        <div id="placeholder">
```

templates/gis_data_editor_form.twig
```diff
@@ -11,14 +11,16 @@
...
             {{ visualization|raw }}
         </div>
 
         <div id="openlayersmap" style="width: {{width}}px; height: {{height}}px;"{{ srid == 0 ? ' class="hide"' }}></div>
 
         <div class="choice float-end">
-            <input type="checkbox" id="choice" value="useBaseLayer"{{ srid != 0 ? ' checked="checked"' }}>
+            <!-- Always uses svg-view here added `and false` as a quick-fix -->
+            <input type="checkbox" id="choice" value="useBaseLayer"{{ srid != 0 and false ? ' checked="checked"' }}>
```
MauricioFauth, line 23:
```suggestion
            <input type="checkbox" id="choice" value="useBaseLayer">
```

MauricioFauth (changes requested):

MauricioFauth (approved):

PR #18188, opened 2023-03-01
Fix inserting gis-data via insert/edit form

MoonE:
Inserting or updating gis data via the form  does not work currently because the  single quotes in the input control are escaped by call to `escapeString`.

This PR correctly escapes the input and makes the single quotes in the input optional as they are not needed for correct parsing of the wkt + srid data.


williamdes (approved):
Thanks !

MauricioFauth (approved):

PR #18189, opened 2023-03-01
Fix editing GEOMETRYCOLLECTION EMPTY

MoonE:
An empty geometry collection has two valid representations in wkt and MariaDB uses the latter when fetching data, which breaks editing.
`GEOMETRYCOLLECTION()` and `GEOMETRYCOLLECTION EMPTY`

With this change both formats can be successfully edited.

### DIFF
--- a/libraries/classes/Controllers/GisDataEditorController.php
+++ b/libraries/classes/Controllers/GisDataEditorController.php
@@ -26,6 +26,16 @@ use function trim;
  */
 class GisDataEditorController extends AbstractController
 {
+    private const GIS_TYPES = [
+        'POINT',
+        'MULTIPOINT',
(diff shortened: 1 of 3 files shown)


kamil-tekiela:
No test for this?

MoonE:
I thought about it, but currently GisDataEditorController is just one long function that outputs a lot of html.
Makes it hard to write a test.

williamdes:
> I thought about it, but currently GisDataEditorController is just one long function that outputs a lot of html. Makes it hard to write a test.

Maybe you can add a method validateFooBar and put the changed line into it and then you can test it easily

williamdes (approved):

MauricioFauth (approved):

PR #18213, opened 2023-03-04
Drop getDbnameAndOptions method and add named argument

kamil-tekiela:
This reduces strange polymorphism in plugins and redundant code. Psalm complains about invalid types but I tested this and I cannot see any bug. I raised a bug report with Vimeo.


kamil-tekiela:
@MauricioFauth Do you know why the tests are now failing? It looks like it has something to do with locale,

MauricioFauth (approved):

PR #18218, opened 2023-03-05
Refactor GisVisualization settings

MoonE:
This is a refactoring to  get better type safety in GisVisualization and GisVisualizationController
This gets rid of the `settings` and `userSpecifiedSettings` properties and instead stores the concrete values in class properties.

### DIFF
--- a/libraries/classes/Controllers/Table/GisVisualizationController.php
+++ b/libraries/classes/Controllers/Table/GisVisualizationController.php
@@ -130,13 +130,7 @@ final class GisVisualizationController extends AbstractController
         $this->addScriptFiles(['vendor/openlayers/OpenLayers.js', 'table/gis_visualization.js']);
 
         // If all the rows contain SRID, use OpenStreetMaps on the initial loading.
-        if (! isset($_POST['displayVisualization'])) {
-            if ($visualization->hasSrid()) {
-                $visualizationSettings['choice'] = 'useBaseLayer';
-            } else {
-                unset($visualizationSettings['choice']);
-            }
-        }
+        $useBaseLayer = isset($_POST['redraw']) ? isset($_POST['useBaseLayer']) : $visualization->hasSrid();
 
         /**
          * Displays the page
@@ -165,6 +159,7 @@ final class GisVisualizationController extends AbstractController
             'spatial_candidates' => $spatialCandidates,
             'visualization_settings' => $visualizationSettings,
             'start_and_number_of_rows_fieldset' => $startAndNumberOfRowsFieldset,
+            'useBaseLayer' => $useBaseLayer,
             'visualization' => $visualization->toImage('svg'),
             'draw_ol' => $visualization->asOl(),
         ]);
--- a/libraries/classes/Controllers/Table/GisVisualizationController.php
+++ b/libraries/classes/Controllers/Table/GisVisualizationController.php
@@ -119,6 +119,9 @@ final class GisVisualizationController extends AbstractController
             $visualizationSettings['spatialColumn'] = $spatialCandidates[0];
         }
 
+        $visualizationSettings['width'] = 600;
+        $visualizationSettings['height'] = 450;
+
         // Download as PNG/SVG/PDF use _GET and the normal form uses _POST
         // Convert geometric columns from bytes to text.
         $pos = (int) ($_POST['pos'] ?? $_GET['pos'] ?? $_SESSION['tmpval']['pos']);
--- a/libraries/classes/Controllers/Table/GisVisualizationController.php
+++ b/libraries/classes/Controllers/Table/GisVisualizationController.php
@@ -20,6 +20,7 @@ use function __;
 use function array_merge;
 use function in_array;
 use function is_array;
+use function is_string;
 
 /**
  * Handles creation of the GIS visualizations.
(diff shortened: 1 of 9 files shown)


libraries/classes/Gis/GisVisualization.php
```diff
@@ -35,76 +33,109 @@
...
 
     /**
      * Factory
      *
-     * @param string $sql_query SQL to fetch raw data for visualization
-     * @param array  $options   Users specified options
-     * @param int    $row       number of rows
-     * @param int    $pos       start position
+     * @param string                   $sql_query SQL to fetch raw data for visualization
+     * @param array<string,string|int> $options   Users specified options
```
kamil-tekiela, line 108: Shouldn't this be `array<string,string|int|null>` if `labelColumn` is nullable?

psalm-baseline.xml
```diff
@@ -7627,95 +7599,26 @@
...
     <PossiblyInvalidArgument>
       <code>$results</code>
     </PossiblyInvalidArgument>
     <PossiblyInvalidOperand>
       <code>$results</code>
       <code>$results</code>
     </PossiblyInvalidOperand>
+    <PossiblyUnusedMethod>
+      <code>toImage</code>
+    </PossiblyUnusedMethod>
```
kamil-tekiela, line 7617: Can we not remove this method?

kamil-tekiela (approved):
Awesome work. Thank you. I was going to fix these params, but now I don't have to.

libraries/classes/Gis/GisVisualization.php
MoonE: Currently only GisVisualizationController uses the labelColumn option and if no label is needed the option is not set. 
Though it does makes sense to allow null, the `@psalm-param` annotation below also allows null.

psalm-baseline.xml
MoonE: Yeah, the toImage and then also the toPng functions are not used.
Png was originally used for browsers without svg support. b2e498cbaf67b5ede9d2c198432e54a5e4340e3d

libraries/classes/Controllers/Table/GisVisualizationController.php
```diff
@@ -68,129 +59,180 @@ public function __invoke(ServerRequest $request): void
...
-        if ($result !== false) {
-            $meta = $this->dbi->getFieldsMeta($result);
-        }
+        $meta = $this->getColumnMeta($sqlQuery);
 
         // Find the candidate fields for label column and spatial column
         $labelCandidates = [];
         $spatialCandidates = [];
         foreach ($meta as $column_meta) {
+            if ($column_meta->name === '') {
```
williamdes, line 68: can you add a comment on when this occurs ?

test/classes/Gis/GisVisualizationTest.php
```diff
@@ -120,19 +118,12 @@ public function testModifyQueryOld(): void
      */
     public function testModifyQuery(): void
     {
+        $this->dbi?->setVersion(['@@version' => '8.0.0']);
```
williamdes, line 121:
```suggestion
        $this->dbi->setVersion(['@@version' => '8.0.0']);
```
I do not see why we should allow it to soft fail, the test needs it ;)

libraries/classes/Controllers/Table/GisVisualizationController.php
MoonE: It does not, but it fixes a psalm warning.

In FieldMetadata, the name is default-initialized with empty string currently. I suspect this is just for tests and mysql will never return empty string. So it may be possible to assert non-empty string there.

williamdes: Then it's better to silence psalm with a comment rather than having dead code that will stay here for decades ^^

williamdes (approved):
Very good work !

And cool GPG signature (I probably already said it 😄)

test/classes/Gis/GisVisualizationTest.php
MoonE: This is also just a convenient way to ignore analyzer warnings. If you prefer, an `assert($this->dbi !== null)` would also work.

test/classes/Gis/GisVisualizationTest.php
```diff
@@ -4,28 +4,31 @@
...
 namespace PhpMyAdmin\Tests\Gis;
 
+use PhpMyAdmin\DatabaseInterface;
 use PhpMyAdmin\Gis\GisVisualization;
 use PhpMyAdmin\Tests\AbstractTestCase;
 
 /** @covers \PhpMyAdmin\Gis\GisVisualization */
 class GisVisualizationTest extends AbstractTestCase
 {
+    private DatabaseInterface|null $dbi = null;
```
MauricioFauth, line 14: Don't need `null` here. The `setUp()` method is treated as a constructor.

```suggestion
    private DatabaseInterface $dbi;
```

MauricioFauth (changes requested):

libraries/classes/Controllers/Table/GisVisualizationController.php
MoonE: I'll make a new PR and annotate FieldMetadata::name as `non-empty-string`. Then psalm will complain and this check can be removed.

MauricioFauth (approved):

PR #18225, opened 2023-03-06
Make database and web server info separately configurable

erkia:
### Description

While showing database server information in phpMyAdmin is mostly relevant, showing web server information may not always be. 

Removing web server information may be desired just to clean up the frontpage or to follow security best? practices by giving up as little information about the server software as minimally required.

This pull request makes `ShowServerInfo` configuration variable more granular by allowing to select 'dbserver' or 'webserver' (just for completeness :) ) information only.


doc/config.rst
```diff
@@ -2427,13 +2427,23 @@ Main panel
...
+    :type: boolean|string
     :default: true
 
     Defines whether to display detailed server information on main page.
+    Possible values are:
+
+    * ``true`` to show all server information
+    * ``false`` to hide server information
+    * ``'dbserver'`` to show only database server information
+    * ``'webserver'`` to show only web server information
```
ibennetch, line 2439: I might personally prefer the words "database" and "web" instead of "dbserver" here, but that's very minor and I wonder what others think.

ibennetch (approved):
This is not a change I think most people would need or find particularly desirable myself, but your solution is quite elegant and well done. Your work fits in nicely with the style of the project and I really like that you added more options to the existing variable rather than just adding more new variables. 

Looks good to me, thanks for the contribution!

doc/config.rst
williamdes: I also agree
But maybe web-server and database-server

So people do not expect things about the database itself

erkia: Let's try 'database-server' and 'web-server'. Changed and rebased.

williamdes (approved):
💯

MauricioFauth (approved):

PR #18233, opened 2023-03-08
Fix #18212 - Query Builder doesn't replace a table name with it's alias in the `WHERE` block

### LINKED ISSUE #18212: Query Builder doesn't replace a table name with alias In WHERE block.
### Describe the bug
When I make a query with query builder and use a alias for a table,
in WHERE block, it does not replace the table name with the alias.
It seems wrong to me that it is not replaced.

### To Reproduce

1. Go to Query.
2. Select table and input table alias
3. Check "criteria"
4. Input criteria as free text
5. Click "Update query"
6. See the query made by builder

### Expected behavior

A table name is replaced by the alias if alias field is entered.

### Screenshots

image

### Server configuration

- Operating system:
- Web server: nginx/1.14.2
- Database version: 8.0.18 - MySQL
- PHP version: 8.2.3
- phpMyAdmin version: 6.0.0-dev, 5.2.1

### Client configuration

- Browser: Google Chrome

faissaloux:
Fixes #18212 

image

### DIFF
--- a/js/src/database/query_generator.js
+++ b/js/src/database/query_generator.js
@@ -33,7 +33,10 @@ function getFormatsText () {
 }
 
 function generateCondition (criteriaDiv, table) {
-    var query = '`' + Functions.escapeBacktick(table.val()) + '`.';
+    const tableName = table.val();
+    const tableAlias = table.siblings('.table_alias').val();
+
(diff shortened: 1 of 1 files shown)


williamdes (approved):
:100:

williamdes:
Thank you @faissaloux for this contribution and for the patience

PR #15585, opened 2019-11-23
Fix #15582 Don't disable "Empty"-button after clicking

Tithugues:
Fixes: #15582

PR #18239, opened 2023-03-10
Disable unnecessary tables actions when empty

faissaloux:
Hello 👋🏻 

In this PR I have disabled `Search` and `Empty` actions on empty tables since they have no use in this case.

## Before
You can click on `Search` and `Empty` actions on all tables.

image

## After
You can't click on `Search` and `Empty` actions on empty tables.

image

### DIFF
--- a/js/src/functions.js
+++ b/js/src/functions.js
@@ -516,7 +516,7 @@ Functions.checkPasswordStrength = function (value, meterObject, meterObjectLabel
         'my',
         'admin',
     ];
-    if (username !== null) {
+    if (username) {
         customDict.push(username);
     }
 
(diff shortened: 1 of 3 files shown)


williamdes:
Hi @faissaloux 
I can not accept this request because it goes against #15582 and it's PR #15585

kamil-tekiela:
On top of this, I would like to add that having disabled UI buttons is not user-friendly. A button should either return an unsuccessful response or be idempotent.

faissaloux:
@williamdes on https://github.com/phpmyadmin/phpmyadmin/issues/15582 I think the problem was unable to click `empty` after clicking it in this PR I don't allow to `empty` empty tables from the start! Or should it be enabled anyway?

What about the `search`?

kamil-tekiela:
I would leave it enabled. PMA doesn't necessarily represent the current state. It can only know the state at the time the page was opened. If the data changes in the meantime, the user is left with the only option of refreshing the page. That is quite annoying. Let the buttons be always active.

williamdes:
> I would leave it enabled. PMA doesn't necessarily represent the current state. It can only know the state at the time the page was opened. If the data changes in the meantime, the user is left with the only option of refreshing the page. That is quite annoying. Let the buttons be always active.

That's exactly the reason

williamdes:
> What about the search?

You can transform this PR to make search always clickable if it's not already like that. Basically all of the buttons should be available. 

> It can only know the state at the time the page was opened. If the data changes in the meantime, the user is left with the only option of refreshing the page. That is quite annoying.

Because of this reason

faissaloux:
Ow yeah you are right @kamil-tekiela  👌🏻 Good point

@williamdes Yeah, it makes sense, thank you.

PR #18244, opened 2023-03-11
Refactor getItemUnhideDialog()

kamil-tekiela:
This does not look right. When used with native types, the test were failing so I wanted to understand why we have these params here. I could not find any use for them so I removed them. But I have no idea what this feature is or how to test it properly. I don't know if it was a bug or left over from earlier refactoring. 

Does anyone know how to test this feature?

### DIFF
--- a/libraries/classes/Controllers/NavigationController.php
+++ b/libraries/classes/Controllers/NavigationController.php
@@ -60,15 +60,13 @@ class NavigationController extends AbstractController
             $itemName = $request->getParsedBodyParam('itemName', '');
             $itemType = $request->getParsedBodyParam('itemType', '');
             $dbName = $request->getParsedBodyParam('dbName', '');
-            $tableName = $request->getParsedBodyParam('tableName', '');
 
             if ($request->getParsedBodyParam('hideNavItem') !== null) {
-                if (! empty($itemName) && ! empty($itemType) && ! empty($dbName)) {
+                if ($itemName !== '' && $itemType !== '' && $dbName !== '') {
                     $this->navigation->hideNavigationItem(
                         $itemName,
                         $itemType,
                         $dbName,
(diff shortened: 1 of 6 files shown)


MauricioFauth:
When you hide some tables, an eye icon will appear next to the database name, clicking on it will open a dialog modal that will call this method.

Screenshot from 2023-03-13 11-10-44

Screenshot from 2023-03-13 11-12-15

Screenshot from 2023-03-13 11-13-17

kamil-tekiela:
@MauricioFauth How do I get the eye icon? Do I need to enable something in the configuration?

MauricioFauth:
> @MauricioFauth How do I get the eye icon? Do I need to enable something in the configuration?

In the navigation tree, when you hover a table's name, an eye icon will appear. When you click on it, the table will be hidden from the list and then another eye icon will appear next to the database's name, so you can click on it to show the hidden tables again.

kamil-tekiela:
Thanks. No idea why it was not showing up for me until now. I found it and I was able to test it. It works, but the column `table_name` is unused. I removed the parameter but I think the SQL table should get adjusted too.

williamdes (approved):

MauricioFauth:
> Thanks. No idea why it was not showing up for me until now. I found it and I was able to test it. It works, but the column `table_name` is unused. I removed the parameter but I think the SQL table should get adjusted too. 
> 
> P.S. I also fixed the bug that did not allow me to hide a table called `0`.

I think the idea of the table_name column is to be able hide itens from a table, like indexes or columns. But looks like it's not working right now, or was never fully implemented.

MauricioFauth (approved):

MauricioFauth:
I merged it. I'll work on adding/fixing the feature of hiding the table items later. So the table_name column will not be removed from the table structure.

PR #18247, opened 2023-03-11
Native param types

kamil-tekiela:
3/3. Adding native param types was the most challenging of native types. I caught and fixed a lot of issues but I am sure I missed many more. 

Prerequisities:
- #18244
- #18218

### DIFF
--- a/libraries/classes/Advisory/Advisor.php
+++ b/libraries/classes/Advisory/Advisor.php
@@ -136,7 +136,7 @@ class Advisor
      * @param string|int $variable Variable to set
      * @param mixed      $value    Value to set
      */
-    public function setVariable(string|int $variable, $value): void
+    public function setVariable(string|int $variable, mixed $value): void
     {
         $this->variables[$variable] = $value;
     }
--- a/libraries/classes/Cache.php
+++ b/libraries/classes/Cache.php
@@ -20,7 +20,7 @@ class Cache
      * @param string $cacheKey The key to use
      * @param mixed  $value    The value to cache
      */
-    public static function set(string $cacheKey, $value): void
+    public static function set(string $cacheKey, mixed $value): void
     {
         self::$cacheData[$cacheKey] = $value;
     }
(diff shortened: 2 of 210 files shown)


williamdes (approved):
🥇 🥇 🥇

kamil-tekiela:
The transformation tests are failing on ARM but I don't know why.

MauricioFauth:
> The transformation tests are failing on ARM but I don't know why. 

Looks like a 32 bit issue.

kamil-tekiela:
> > The transformation tests are failing on ARM but I don't know why.
> 
> Looks like a 32 bit issue.

Is it because of my PR? How can I fix this?

libraries/classes/Plugins/Transformations/Abs/LongToIPv4TransformationsPlugin.php
```diff
@@ -37,7 +37,7 @@ public static function getInfo(): string
      * @param array              $options transformation options
      * @param FieldMetadata|null $meta    meta information
      */
-    public function applyTransformation($buffer, array $options = [], FieldMetadata|null $meta = null): string
+    public function applyTransformation(string $buffer, array $options = [], FieldMetadata|null $meta = null): string
```
MauricioFauth, line 40: Not sure, but I think is this line that is causing the test to fail.

MauricioFauth: Looks like ReflectionMethod::invokeArgs does not raise a type error. 

https://3v4l.org/MBVNG

kamil-tekiela: I fixed it but it still fails. I don't think that was it. I think I will take this change out of this PR.

MauricioFauth (approved):

MauricioFauth:
:100:

PR #18253, opened 2023-03-13
Fix #15898 - escape tbl_storage_engine argument

### LINKED ISSUE #15898: sql injection in  /phpmyadmin/libraries/classesCreateAddField.php
### Describe the bug

- sql injection in  /phpmyadmin/libraries/classesCreateAddField.php
- function getTableCreationQuery() line 457
sqli_phpmyadmin
### To Reproduce

1. Go to http://localhost/phpmyadmin/tbl_create.php?db=anydb
2. Click on save
3. sniff resquest and edit param `tbl_storage_engine` or `tbl_collation`

### Expected behavior

https://www.slideshare.net/OWASPEEE/russia-mysql-oob-injections

### Screenshots

-adding `'` in param `tbl_storage_engine`
error_phpmyadmin
- adding comment `-- -`
not_error

### Server configuration

- Operating system:windows 10
- Web server:localhost TCP/ IP
- Database version:10.4.8-MariaDB
- PHP version:PHP Version 7.3.10
- phpMyAdmin version:5.0.1

### Client configuration

- Browser:firefox
- Operating system:windows 10

izsob:
Signed-off-by: William Desportes <williamdes@wdes.fr>
(cherry picked from commit ca42395ee4b2936d3702524f8fb8bec1e9502bc7)

### Description

The SQL injection was discovered and fixed a long time ago. Can you include the fix in a subsequent 4.9 release? I just cherry-picked the fixing commit, and I think the original plan was to include it in the 4.9 branch, if applicable (#15898).

### DIFF
--- a/libraries/classes/CreateAddField.php
+++ b/libraries/classes/CreateAddField.php
@@ -474,7 +474,7 @@ class CreateAddField
         if (!empty($_POST['tbl_storage_engine'])
             && ($_POST['tbl_storage_engine'] != 'Default')
         ) {
-            $sqlQuery .= ' ENGINE = ' . $_POST['tbl_storage_engine'];
+            $sqlQuery .= ' ENGINE = ' . $this->dbi->escapeString($_POST['tbl_storage_engine']);
         }
         if (!empty($_POST['tbl_collation'])) {
             $sqlQuery .= Util::getCharsetQueryPart($_POST['tbl_collation']);


kamil-tekiela:
This doesn't look like a correct fix. The ENGINE value is not a string literal and should not be quoted. The value should be validated against a list of known ENGINE types. The use of `escapeString` and by extension `quoteString` is invalid here.

williamdes:
I am closing this one since no non security fix will enter 4.9
Why do you need it in 4.9 @izsob ?

williamdes:
> This doesn't look like a correct fix. The ENGINE value is not a string literal and should not be quoted. The value should be validated against a list of known ENGINE types. The use of `escapeString` and by extension `quoteString` is invalid here.

Can you make a PR to fix this in 5.2 ?

izsob:
Hi @williamdes ! Thank you for the quick reply! Version 4.9 is still in use, and an SQL injection seems like a critical security vulnerability: https://nvd.nist.gov/vuln/detail/CVE-2020-22452
Can you include the fix in 4.9, or is it not affected? (Sorry, if I mistakenly referred to a wrong commit.)

ibennetch:
On a quick look at the history of that other issue, it appears to me like we meant to target 4.9.x with the fix as well, but because it was handled outside of our usual security process I may have missed backporting it when performing the release. Good catch by @izsob. 

I'll have to dig more in the commit history; I'm on mobile right now and it isn't easy to see all that information on one screen.

If that's true, we should fix this for QA_4_9, but we should also make sure the fix that's already in 5.x was correct. I agree with Kamil about the escaping here and appreciate the input.

kamil-tekiela:
I made a PR to backport it to 5.2. I don't mind it if you want to merge it into 4.9 too, but I don't really see any reason for it. While this is SQL injection, the nature of PMA is that it's a tool that allows users to execute arbitrary SQL statements. SQL injection would only be a problem if it allowed gaining elevated privileges, which this bug doesn't.

williamdes:
> it appears to me like we meant to target 4.9.x with the fix as well

I have re-open this PR 

Maybe you should try to adapt #18255 for 4.9 rather than the "wrong" fix I made ?

williamdes:
Ref: https://security-tracker.debian.org/tracker/CVE-2020-22452

Introduced by https://github.com/phpmyadmin/phpmyadmin/commit/4d988516bffc743daa5a124176d32e83f7246917
All versions before 4.8.4 and after 4.1.0 are vulnerable since it was `$_REQUEST`: https://github.com/phpmyadmin/phpmyadmin/commit/d6e04ca09b205cbc1e00f26da9d1f3690287a4af

/cc @carnil 
/cc @ibennetch 

please read https://github.com/phpmyadmin/phpmyadmin/issues/15898#issuecomment-1422850217

PR #18256, opened 2023-03-13
Check whether ob_gzhandler is already in use

kamil-tekiela:
When an error happens during the handling of another error, the code would try to start output compression again. This should only happen during development or on severely misconfigured ecosystems but this change should help to identify actual problems when this happens.

### DIFF
--- a/libraries/classes/OutputBuffering.php
+++ b/libraries/classes/OutputBuffering.php
@@ -8,6 +8,7 @@ use function defined;
 use function flush;
 use function function_exists;
 use function header;
+use function in_array;
 use function ini_get;
 use function ob_end_clean;
 use function ob_flush;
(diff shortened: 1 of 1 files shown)


williamdes (approved):
Should this go to QA ?

kamil-tekiela:
> Should this go to QA ?

Do we need it in QA? Normal users should not need this fix. It's mostly for developers, I think. I can change the target branch if you'd like.

williamdes:
> You can reproduce it in 6.0 by going to `ResponseRenderer` class and changing the line:
> 
> ```
> $this->HTML = '';
> ```
> 
> to
> 
> ```
> $this->HTML = 234;
> ```

Okay, so there is no real possibility that 5.2 walks into this bug somehow ?
Sorry I ask many questions to kill as much bugs as I can on 5.2

kamil-tekiela:
It should not happen with a properly installed build. At least, I can't think of a way to reproduce it without changing code.

MauricioFauth (approved):

Issue #18249, opened 2023-03-12
Speed improvements when browsing a database with multiple tables


liviuconcioiu:
### Describe the bug

Right now if you browse a database with 250 tables/page, the average page load is 3.5-4 seconds from which 2.4 seconds is taken by getTrackingIcon and 1.6 seconds is taken by sql-parser.

### To Reproduce

1. Create a database with 500-1000 tables.
2. Browse database
3. See page load

### Expected behavior

Page load needs improvement.

### Screenshots

Using xhprof:

![1](https://user-images.githubusercontent.com/25424343/224524712-aa9827df-8034-432c-8d78-5b02ea48e822.png)

![2](https://user-images.githubusercontent.com/25424343/224525182-193e33ec-2edb-448d-a2cb-59ec88143158.png)

### Server configuration

- phpMyAdmin version: 5.2.2-dev, 6.0.0-dev

williamdes:
Thank you!
Is this xhprof?

liviuconcioiu:
> Thank you! Is this xhprof?

Yes.

kamil-tekiela:
Did you try with PR #18250

liviuconcioiu:
> Did you try with PR #18250

I've tested right now. The timings are the same, no improvement when browsing a database.

I've removed the checks from this function (return empty string) and it reduces the times with more than 70%.

https://github.com/phpmyadmin/phpmyadmin/blob/869ecea9cc12efc170b4fe0e0586f756e55b01d1/libraries/classes/Controllers/Database/StructureController.php#L528-L543

I think this function needs a redesign, because it makes too many checks just to get an icon (more than 250 queries), while a single query that can return all the tables with tracking enabled can be made and those tables put in an array and checked if the table is in that array or not.

```SQL
SELECT * FROM `pma__tracking` WHERE `db_name` = 'sakila' AND `table_name` IN (SELECT `TABLE_NAME` FROM `information_schema`.`TABLES` WHERE `TABLE_SCHEMA` = 'sakila');
```

kamil-tekiela:
Yeah, this looks weird. I'll try to refactor this when I have some time.

kamil-tekiela:
[...]

PR #18259, opened 2023-03-14
Refactor and optimize Tracking

kamil-tekiela:
This is an implementation of the suggestion from #18249. The previous code was unoptimized and messy, which I assume was because of some old removed functionality. This PR optimizes page load of `/database/tracking/` and `/database/structure`. With a small number of tables, performance improvement is about 30%. With 10k tables, the performance difference is huge. 

There are still performance bottlenecks that need to be optimized and the Tracking functionality needs a proper cleanup. Further refactorings will be done in the future. 

Memory usage should remain the same, but the more tracked tables there are, the higher it will be. Nonetheless, the new code requires very little memory and is extremely efficient. 

In terms of bringing this to QA, I need to think if some temporary shim could be created. It should be possible to minimize code differences by only introducing the new SQL and changing nothing else. 

Ref: https://github.com/phpmyadmin/phpmyadmin/issues/16005


libraries/classes/Tracking/TrackedTable.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpMyAdmin\Tracking;
+
```
williamdes, line 6: You should maybe add some kind of read only phpdoc flag like other classes
Waiting for some day where we bump the PHP version and mark it readonly class ?

libraries/classes/Util.php
```diff
@@ -222,123 +219,6 @@ public static function getDocuURL(bool $isMariaDB = false): string
...
-    }
-
-    /**
-     * returns array with tables of given db with extended information and grouped
-     *
-     * @param string $db
-     *
-     * @return array (recursive) grouped table list
-     */
-    public static function getTableList($db): array
```
williamdes, line 266: Where all this config useless ?
I can not se where the logic did go, maybe dead code ?
NavigationTreeTableSeparator NavigationTreeEnableGrouping NavigationTreeTableLevel

williamdes (approved):

libraries/classes/Tracking/TrackedTable.php
kamil-tekiela: I thought it would not be necessary. All params are already readonly. Do you want it only as a human comment or do you see it necessary from SA perspective?

williamdes: It's more for SA and since it's not final maybe it could be misused ?

libraries/classes/Util.php
kamil-tekiela: Good question. I have absolutely no idea. I enabled this in config but I see no grouping happening anywhere.

libraries/classes/Tracking/TrackedTable.php
williamdes: Perfect, final seals it

williamdes (approved):

libraries/classes/Tracking/TrackedTable.php
```diff
@@ -0,0 +1,12 @@
...
+final class TrackedTable
+{
+    public function __construct(public readonly string $name, public readonly bool $active)
```
MauricioFauth, line 9: It is better to use TableName here instead of string, isn't it?

williamdes: it's possible that it's a `db.table` string

MauricioFauth: So, at least annotate it as a non-empty-string.

kamil-tekiela: wait, why would it be `db.table`? I know it can be an empty string which is why it is excluded in the query, but why db would be part of it?

MauricioFauth:
> Does anyone know why Selenium test is failing?

https://github.com/phpmyadmin/phpmyadmin/actions/runs/4418492664/jobs/7745696367#step:13:35
```
Failed asserting that 'CREATE TABLE test_table ( id int(11) NOT NULL, val int(11) NOT NULL ) ENGINE=InnoDB DEFAULT CHARSET=latin1;' contains "CREATE TABLE `test_table` (".
```

MauricioFauth:
> Does anyone know why Selenium test is failing?

You can also see the screenshot of the moment of failure:

- https://github.com/phpmyadmin/phpmyadmin/suites/11557829548/artifacts/598441798

- https://github.com/phpmyadmin/phpmyadmin/actions/runs/4418492664#artifacts

MauricioFauth (approved):

PR #18262, opened 2023-03-14
Optimize export of many tables

### LINKED ISSUE #18258: Speed improvements when exporting a database
### Describe the bug

Right now 250 empty tables are exported in ~43 seconds. This is because when exporting, the table schema is fetched each time for each table, instead for the entire database. Even if the user chooses to export only some tables, the fastest way is to fetch the schema for the entire database. If this is done, for 250 tables will only take ~24 seconds.

Ref: #16005

### To Reproduce

1. Create a database with 250 tables
2. Export the entire database

### Expected behavior

Exporting needs improvement.

### Screenshots

Untitled

### Server configuration

### Additional context

https://github.com/phpmyadmin/phpmyadmin/blob/869ecea9cc12efc170b4fe0e0586f756e55b01d1/libraries/classes/Table.php#L329

kamil-tekiela:
The method `getTablesFull()` accepts an array of tables as a second parameter, but this was not used until now. In fact, it could not have been used because it was severely broken. Let's make use of it to try and optimize database export. 

Fixes #18258 and several smaller related issues. Sorry for rather large diff but I had to fix the tests which as it turns out were incorrect. Unless they were meant to be wrong?...


liviuconcioiu:
Can you please test with this PR and let me know your test results?

Before

2 random tables - 207 ms
5000 tables db - 2 minutes 28 seconds

After

2 random tables - 189 ms
5000 tables db - 1 minute 53 seconds

kamil-tekiela:
> > Can you please test with this PR and let me know your test results?
> 
> Before
> 
> 2 random tables - 207 ms 5000 tables db - 2 minutes 28 seconds
> 
> After
> 
> 2 random tables - 189 ms 5000 tables db - 1 minute 53 seconds

These results don't match the ones you listed in the bug report. Are you testing it differently now? It seems like there is some performance improvement, but not a lot. Is this what you were expecting?

liviuconcioiu:
> > > Can you please test with this PR and let me know your test results?
> > 
> > 
> > Before
> > 2 random tables - 207 ms 5000 tables db - 2 minutes 28 seconds
> > After
> > 2 random tables - 189 ms 5000 tables db - 1 minute 53 seconds
> 
> These results don't match the ones you listed in the bug report. Are you testing it differently now? It seems like there is some performance improvement, but not a lot. Is this what you were expecting?

The results for the bug report were for 6.0.0-dev. Probably something else is the issue?

The new results for 5.2.2-dev are great!

libraries/classes/Export.php
```diff
@@ -750,6 +750,13 @@ public function exportDatabase(
 
         $views = [];
 
+        if ($tables !== []){
```
williamdes, line 753:
```suggestion
        if ($tables !== []) {
```

williamdes (approved):

libraries/classes/Export.php
kamil-tekiela: yeah, of course :D This is a very rough PR as I only wanted to test performance results. I'll clean it up.

williamdes: You rock, thank you so much for contributing !

MauricioFauth (approved):

PR #18263, opened 2023-03-15
Optimize export of many tables 6.0

kamil-tekiela:
cc @liviuconcioiu


williamdes:
Since it's the same diff as #18262 I will do a mere octopus to merge QA into master and merge your PR into it too
Or you can close it

PR #18264, opened 2023-03-16
Fix #17635 - Semicolon repeatedly suffixed onto SQL if query ends in comment

### LINKED ISSUE #17635: Semicolon repeatedly suffixed onto SQL if query ends in comment

ghost:
### Description

Added check for addition of semicolon only when it is missing while editing and/or running an SQL query.

Fixes #17635 

### DIFF
--- a/src/Import/Import.php
+++ b/src/Import/Import.php
@@ -176,8 +176,13 @@ final class Import
     {
         ImportSettings::$readMultiply = 1;
         if ($this->importRunBuffer === null) {
+            // Add semicolon only if it is missing
+            if ($sql !== '' && ! str_ends_with($sql, ';')) {
+                $sql .= ';';
+            }
+
             // Do we have something to push into buffer?
-            $this->importRunBuffer = $sql !== '' ? $sql . ';' : null;
+            $this->importRunBuffer = $sql !== '' ? $sql : null;
 
             return;
         }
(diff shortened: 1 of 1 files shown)


williamdes (approved):

williamdes:
@kamil-tekiela You did some refactoring to this code, how can this be merged without needing to use that very costy str_contains function ?

kamil-tekiela:
`str_contains` is not very costly. I guess this just needs a rebase.

williamdes:
Hi @kamil-tekiela 
Is this still needed ?
No tests seem to be unhappy, this is odd

kamil-tekiela:
I don't think this should be merged because the semicolon is still added in the wrong place. This is really a job for the sql-parser and it can already do it (although there is a bug), but ImportSql is not the only place where this method is called from, so this needs a bit more refactoring.

williamdes:
Okay, if we can have a regression test first that would be good. Because the current code seems fine

PR #18270, opened 2023-03-17
Fix failing transformation tests on 32 bit

kamil-tekiela:
Let's see if this fixes tests on ARM. They fail because the value for invalid IP is undefined and depends on the platform. I guess it's because of signed integers, but I have no way of testing this on ARM myself.


williamdes:
> but I have no way of testing this on ARM myself.

If you have Docker locally you can do like in our tests:
https://github.com/phpmyadmin/phpmyadmin/blob/3f125939c657bd74bb1897118e577f5d9c088178/.github/workflows/tests.yml#L54-L61

You can launch yourself a container running on another arch

MauricioFauth (approved):

PR #18272, opened 2023-03-18
Fix #18248 - Change password modal doesn't open 

### LINKED ISSUE #18248: Change password modal doesn't open
### Describe the bug

`Change password` modal doesn't open. It only opens on page refresh.

### To Reproduce

1. Go to https://demo.phpmyadmin.net/master/public/index.php?route=/server/status
2. Go to https://demo.phpmyadmin.net/master/public/index.php?route=/
3. Click on `Change password`

### Expected behavior

`Change password` modal should open.

### Screenshots

https://user-images.githubusercontent.com/25424343/224518539-3a47b4be-961a-4747-b487-2cac77beac9e.mp4

### Server configuration

Odachannnn:
## Description
 
This pull request fixes change password modal doesn't open.

Fixes  #18248

### DIFF
--- a/libraries/classes/Controllers/HomeController.php
+++ b/libraries/classes/Controllers/HomeController.php
@@ -213,6 +213,9 @@ class HomeController extends AbstractController
 
         $git = new Git($this->config->get('ShowGitRevision') ?? true);
 
+        $hasChangePasswordLink = ($GLOBALS['cfg']['Server']['auth_type'] ?? '') !== 'config'
+            && $GLOBALS['cfg']['ShowChgPassword'];
+
         $this->render('home/index', [
             'db' => $GLOBALS['db'],
             'table' => $GLOBALS['table'],
@@ -225,8 +228,8 @@ class HomeController extends AbstractController
             'is_demo' => $GLOBALS['cfg']['DBG']['demo'],
             'has_server_selection' => $hasServerSelection ?? false,
             'server_selection' => $serverSelection ?? '',
-            'has_change_password_link' => ($GLOBALS['cfg']['Server']['auth_type'] ?? '') !== 'config'
-                && $GLOBALS['cfg']['ShowChgPassword'],
+            'has_change_password_link' => $hasChangePasswordLink,
+            'has_change_password_modal' => $hasChangePasswordLink && !$this->response->isAjax(),
             'charsets' => $charsetsList ?? [],
             'available_languages' => $availableLanguages,
             'database_server' => $databaseServer,
--- a/templates/header.twig
+++ b/templates/header.twig
@@ -58,5 +58,4 @@
     {{ include('modals/preview_sql_modal.twig') }}
     {{ include('modals/enum_set_editor.twig') }}
     {{ include('modals/create_view.twig') }}
-    {{ include('modals/change_password.twig') }}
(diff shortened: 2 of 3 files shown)


templates/header.twig
```diff
@@ -58,5 +58,4 @@
     {{ include('modals/preview_sql_modal.twig') }}
     {{ include('modals/enum_set_editor.twig') }}
     {{ include('modals/create_view.twig') }}
-    {{ include('modals/change_password.twig') }}
```
williamdes, line 61: You should probably keep this line, because in some cases the modal will not be part of the content

I am unsure about my comment, but it's worth checking if there is other ways to trigger this modal to open or not

Odachannnn: @williamdes 
> I am unsure about my comment, but it's worth checking if there is other ways to trigger this modal to open or not

I checked it, and I found that I must not remove this line.
When I change the theme, the modal doesn't not open.


https://user-images.githubusercontent.com/63225126/226154197-67e35df1-834f-4ac6-8bb6-a79620b54e4e.mov

Odachannnn:
I'm sorry if force push is  not good after PR.
I did out of habit.

williamdes:
> I'm sorry if force push is not good after PR. I did out of habit.

No no, it's totally okay I prefer that than users that create me more work by closing an opening a new one because for some reason they can not force push

templates/home/index.twig
```diff
@@ -6,6 +6,10 @@
 
 {{ partial_logout|raw }}
 
+{%- if has_change_password_modal == false %}
```
williamdes, line 9:
```suggestion
{%- if not has_change_password_modal %}
```

libraries/classes/Controllers/HomeController.php
```diff
@@ -213,6 +213,9 @@ public function __invoke(ServerRequest $request): void
 
         $git = new Git($this->config->get('ShowGitRevision') ?? true);
 
+        $hasChangePasswordLink = ($GLOBALS['cfg']['Server']['auth_type'] ?? '') !== 'config'
```
williamdes, line 216: Maybe suggest to #18274 to pick this change in their batch so it also does not add the modal globally

williamdes (approved):
💯

williamdes (approved):

libraries/classes/Controllers/HomeController.php
Odachannnn: I'm not good at English, and I'm not used to contributing and contributing flow,
so I'd like to get a better understanding of your comment.

You mean that I should suggest to #18274 to change codes like this line, and write a comment?
I checked comments of the PR, you have already commented so.

templates/header.twig
williamdes: @MauricioFauth 
Because you pushed 01cf59bf8d1281264e32f3639ce8d8fd2dfeac90

MauricioFauth: I did not see this discussion when I was reviewing this PR, otherwise I would have accepted it.

MauricioFauth: The change I did doesn't have this issue.

williamdes: okay, thanks for checking then

PR #18280, opened 2023-03-20
Don't parse SQL if no tracking is in progress

kamil-tekiela:
Surprisingly, performing an SQL query is better for performance than parsing the query unnecessarily. Tested with 250 tables for about 50% better performance. 

Ref. #18249

### DIFF
--- a/libraries/classes/Tracker.php
+++ b/libraries/classes/Tracker.php
@@ -7,6 +7,7 @@ declare(strict_types=1);
 
 namespace PhpMyAdmin;
 
+use PhpMyAdmin\ConfigStorage\Features\TrackingFeature;
 use PhpMyAdmin\ConfigStorage\Relation;
 use PhpMyAdmin\Plugins\Export\ExportSql;
 use PhpMyAdmin\SqlParser\Parser;
@@ -836,8 +837,6 @@ class Tracker
     {
         global $dbi;
 
-        $relation = new Relation($dbi);
-
         // If query is marked as untouchable, leave
         if (mb_strstr($query, '/*NOTRACK*/')) {
             return;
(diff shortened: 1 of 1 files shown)


williamdes (approved):
Awesome !

liviuconcioiu:
@kamil-tekiela Thank you for the great work! BTW, this PR will also fix the issues with import of large tables/databases.

In my case, opening a database with 250 tables (50% performance increase) and importing a table (25 MB in size and 25000 rows), before this PR it took 4 minutes and 30 seconds. Now it takes 30 seconds!

libraries/classes/Tracker.php
```diff
@@ -954,4 +958,19 @@ public static function handleQuery($query): void
...
+        string $dbname
+    ): bool {
+        $sqlQuery = sprintf(
+            '/*NOTRACK*/ SELECT 1 FROM %s.%s WHERE tracking_active = 1 AND db_name = %s LIMIT 1',
+            Util::backquote($trackingFeature->database),
+            Util::backquote($trackingFeature->tracking),
+            "'" . $dbi->escapeString($dbname, DatabaseInterface::CONNECT_CONTROL) . "'"
+        );
+
+        return $dbi->queryAsControlUser($sqlQuery)->fetchValue() !== false;
```
williamdes, line 974: Is this kept in cache or is it intentional that it is not ?

kamil-tekiela: It's intentional. I thought about caching it but I was unclear if it's a good idea so I intentiontionally left it as is. If someones wants to evaluate the possibility of caching this in the future, they can do it in another PR.

PR #18281, opened 2023-03-20
Optimize getTrackingIcon

kamil-tekiela:
This is #18259 but with minimal changes for QA. No new classes and no refactoring. The /database/tracking/ optimization is also not included.

### DIFF
--- a/libraries/classes/Tracker.php
+++ b/libraries/classes/Tracker.php
@@ -973,4 +973,46 @@ class Tracker
 
         return $dbi->queryAsControlUser($sqlQuery)->fetchValue() !== false;
     }
+
+    /**
+     * THIS IS TEMPORARY FIX for performance issues in QA 5.2. Do not merge into 6.0!
+     */
+    public static function getTrackedTables(string $dbName): array
+    {
+        global $dbi;
+
+        $trackingEnabled = Cache::get(self::TRACKER_ENABLED_CACHE_KEY, false);
+        if (! $trackingEnabled) {
+            return [];
+        }
+
+        $relation = new Relation($dbi);
+        $trackingFeature = $relation->getRelationParameters()->trackingFeature;
+        if ($trackingFeature === null) {
+            return [];
+        }
+
+        $sqlQuery = sprintf(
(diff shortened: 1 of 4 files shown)


williamdes (approved):
🥇

libraries/classes/Tracker.php
```diff
@@ -954,4 +954,46 @@ public static function handleQuery($query): void
 
         $dbi->queryAsControlUser($sqlQuery);
     }
+
+    /**
+     * THIS IS TEMPORARY FIX for performance issues in QA 5.2. Do not merge into 6.0!
+     */
```
williamdes, line 960: Okay, if I am the one to merge, I will redact this part
And do the merge into 6.0
A merge that will be empty since 6.0 has the changes IIRC

kamil-tekiela: So I actually have no idea what will Git do when you try to merge 5.2 into master. This is why I added the note. The merge should be empty, but if it's not then you need to manually adjust it.

williamdes: Git will try to merge all changes, and then I can edit the file and amend the merge
It will maybe complain before so I fix conflicts

What git GUI are you using ?

kamil-tekiela: I just tried `git merge getTrackingIcon`, and it just appended the new method. So you need to manually delete the method.

psalm-baseline.xml
```diff
@@ -1,5 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <files psalm-version="4.30.0@d0bc6e25d89f649e4f36a534f330f8bb4643dd69">
+  <file src="config.inc.php">
```
MauricioFauth, line 3: This shouldn't have to be included.

MauricioFauth (changes requested):

libraries/classes/Tracker.php
williamdes: Yup, delete git add && git commit --amend  and push the merge

MauricioFauth (approved):

PR #18282, opened 2023-03-21
Merge config.values.php with ConfigFile class

MauricioFauth:
This file is only used by the ConfigFile class, so makes sense to this information to be inside the class.


ibennetch (approved):
I believe the reason config.values.php was kept was to give users an easy place to see the possible values for their config.inc.php. That doesn't mean we need to keep it, but it was a known place for users to go and quickly scroll through some of the available options.

Really, that's what the documentation is for anyway, but despite our best efforts things there aren't always as clear as looking directly in the source.

Still, I support this change, I just want to vocalize the thought that some users might still refer to this.

williamdes (approved):

PR #18293, opened 2023-03-28
Warm up Twig cache when creating a release

MauricioFauth:
Ref: #14604

### DIFF
--- a/scripts/create-release.sh
+++ b/scripts/create-release.sh
@@ -513,6 +513,8 @@ if [ -f ./scripts/console ]; then
     composer update --no-interaction
     # Warm up the routing cache for 5.1+ releases
     ./scripts/console cache:warmup --routing
+    # Warm up the Twig cache
+    ./scripts/console cache:warmup --twig
 fi
 
 PHP_REQ=$(sed -n '/"php"/ s/.*"\^\([0-9]\.[0-9]\.[0-9]\).*/\1/p' composer.json)
@@ -656,6 +658,9 @@ for kit in $KITS ; do
 
     # Remove possible tmp folder
     rm -rf tmp
+
+    mkdir tmp
+    mv twig-templates tmp/twig
 
     cd ..
 


scripts/create-release.sh
```diff
@@ -657,6 +659,9 @@ for kit in $KITS ; do
     # Remove possible tmp folder
     rm -rf tmp
 
+    mkdir tmp
+    mv twig-templates tmp/twig
```
williamdes, line 663: You should not do that, it will crash into peoples existing tmp folders and overwrite the permissions
The code needs to use a specific folder in `libraries/cache`

scripts/create-release.sh
```diff
@@ -513,6 +513,8 @@ if [ -f ./scripts/console ]; then
     composer update --no-interaction
     # Warm up the routing cache for 5.1+ releases
     ./scripts/console cache:warmup --routing
+    # Warm up the Twig cache
+    ./scripts/console cache:warmup --twig
```
williamdes, line 517: `security_checkup` should check for that first

MauricioFauth: What do you mean by that? What it should check exactly?

scripts/create-release.sh
MauricioFauth: I'm going to do that on master. I tried to change as little as possible on 5.2. Why it will crash peoples folders?

williamdes: Because people already have folders named tmp and tmp/twig
extracting to them will overwrite permissions
And users that do not have it, will have wrong permissions

We need to force another folder that is only used by us. The user can use another folder than `tmp` so this change would not work in this case

scripts/create-release.sh
williamdes: Just checking there is files in the folder would be enougth

scripts/create-release.sh
MauricioFauth: I think is not a good idea to change the directory for 5.2. And it is a tmp directory, it should be cleaned before doing upgrades.

scripts/create-release.sh
MauricioFauth: This command creates a new directory.

scripts/create-release.sh
williamdes: I agree, let's do this for 6.0
And use our libraries/cache folder
You know people just unzip and have poor knowledge of what's to be done for upgrades
a lot of users are non technical

scripts/create-release.sh
williamdes: And creates files in it
We should check there is files in the folder to ensure warmup has been done

MauricioFauth:
I'm closing this and I'll do it on `master` branch.

PR #18300, opened 2023-03-30
Convert JavaScript files to TypeScript

MauricioFauth:
TypeScript 5 introduced a new config option ([`--moduleResolution bundler`](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#moduleresolution-bundler)) that made the integration with Webpack easier.

The output is almost the same. There are only minor differences.

There are a lot of errors being reported by TypeScript, however it still compiles the JavaScript as before, since the config option `noEmitOnError` defaults to `false`.

- [x] Those errors must be fixed or ignored before merging this pull request.
- [x] Fix ESLint error
- [x] Fix Jest error


williamdes (approved):
Nice !
Thank you for the output diff

PR #18315, opened 2023-04-02
Fix #18266 - Disable tracking feature does not work

sfurqan2:
### Description

Added an additional strict equality check to detect if config value is empty or set to false.

Fixes #18266 

Before submitting pull request, please review the following checklist:

- [ ] Make sure you have read our [CONTRIBUTING.md](https://github.com/phpmyadmin/phpmyadmin/blob/master/CONTRIBUTING.md) document.
- [ ] Make sure you are making a pull request against the correct branch. For example, for bug fixes in a released version use the corresponding QA branch and for new features use the _master_ branch. If you have a doubt, you can ask as a comment in the bug report or on the mailing list.
- [ ] Every commit has proper `Signed-off-by` line as described in our [DCO](https://github.com/phpmyadmin/phpmyadmin/blob/master/DCO). This ensures that the work you're submitting is your own creation.
- [ ] Every commit has a descriptive commit message.
- [ ] Every commit is needed on its own, if you have just minor fixes to previous commits, you can squash them.
- [ ] Any new functionality is covered by tests.

PR #18318, opened 2023-04-02
Refactor relation

kamil-tekiela:
This should also fix #18315 on master because we are now explicitly checking for an empty string. We use the default table name only if the global is an empty string, which means the user didn't override it in the config. 
The rest is just making the code easier to read.

### DIFF
--- a/libraries/classes/ConfigStorage/Relation.php
+++ b/libraries/classes/ConfigStorage/Relation.php
@@ -1572,29 +1572,28 @@ class Relation
             // use the possible replaced name first and fallback on the table name
             // if no replacement exists
             if (! in_array($tableNameReplacements[$table] ?? $table, $existingTables)) {
-                if ($create) {
-                    $this->dbi->tryQuery($createQueries[$table], Connection::TYPE_CONTROL);
-
-                    $error = $this->dbi->getError(Connection::TYPE_CONTROL);
-                    if ($error) {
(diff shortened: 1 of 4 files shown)


libraries/classes/ConfigStorage/Relation.php
```diff
@@ -1554,64 +1551,47 @@ public function fixPmaTables(string $db, bool $create = true): void
...
-                    // Do not override a user defined value, only fill if empty
-                    $GLOBALS['cfg']['Server'][$feature] = $table;
+                    return;
                 }
             }
+
+            $foundOne = true;
+
+            // Do not override a user defined value, only fill if empty
+            if (isset($GLOBALS['cfg']['Server'][$feature]) && $GLOBALS['cfg']['Server'][$feature] !== '') {
```
sfurqan2, line 1589: Hey, just a thought, won't this ignore if feature is set to `false` (disabled)?

[https://docs.phpmyadmin.net/en/latest/config.html#cfg_Servers_tracking](url)

I think one more additional check is needed

```suggestion
            if (isset($GLOBALS['cfg']['Server'][$feature]) && $GLOBALS['cfg']['Server'][$feature] !== '' && $GLOBALS['cfg']['Server'][$feature] !== false) {
```

What do you think?

MauricioFauth (approved):

PR #18319, opened 2023-04-03
Fix #18266 - bug preventing disabling tracking feature

### LINKED ISSUE #18266: Disable tracking feature doesn't work
### Describe the bug

phpMyAdmin configuration storage database is setup, a `pma__tracking` table exists with no tables being tracked at all, yet there's an issue.

1. If a `pma__tracking` table exists, there's no way to disable tracking from `config.inc.php`. I've tried setting `$cfg['Servers'][$i]['tracking']` to `false`, empty string, wrong table name and it doesn't work. (https://docs.phpmyadmin.net/en/latest/config.html#cfg_Servers_tracking)

At least this is how I understand the setting and I hope I'm not wrong. This is why the Import is slow, because it doesn't respect the default value (in the config).

### Expected behavior

1. Even if `pma__tracking` table exists, it should respect `$cfg['Servers'][$i]['tracking']` setting.
2. Even if `pma__tracking` table exists and there's no `$cfg['Servers'][$i]['tracking']` setting (empty by default), queries (in this case when using the Import feature) shouldn't be tracked.
3. Even if `$cfg['Servers'][$i]['pmadb']` is set to another database name, this setting is ignored.

### Server configuration

kamil-tekiela:
This is an alternative to #18315. 

It's based on William's proposal. I am wondering if skipping the table creation makes sense. Technically, what we only want to do is prevent filling the `$GLOBALS['cfg']['Server'][$feature]` if it equals `false`. In this PR I added a new if at the start of the loop, but on master, we can just stick with the one at the end of the loop like I showed in the other PR #18318. 

Fixes: #18266

### DIFF
--- a/libraries/classes/ConfigStorage/Relation.php
+++ b/libraries/classes/ConfigStorage/Relation.php
@@ -1626,6 +1626,11 @@ class Relation
         $createQueries = null;
         $foundOne = false;
         foreach ($tablesToFeatures as $table => $feature) {
+            if (($GLOBALS['cfg']['Server'][$feature] ?? null) === false) {
+                // The feature is disabled by the user in config
+                continue;
+            }
+
             // Check if the table already exists
             // use the possible replaced name first and fallback on the table name
             // if no replacement exists


MauricioFauth (approved):

williamdes:
Thank you !

I pushed my tests as e0dc51563870273a5ab3dcf97a39d9098f0f6603

PR #18324, opened 2023-04-05
[UI] Remove float-end class on tracking versions table

kamil-tekiela:
This will remove the visual defect with this column.


williamdes (approved):
Before
Sélection_2023040710:09:42001
After
Sélection_2023040710:10:09001

PR #18326, opened 2023-04-06
Replace jqPlot with Chart.js in /server/status/queries

MauricioFauth:
Build the pie chart in `/server/status/queries` page with [Chart.js](https://www.chartjs.org/) (MIT licensed) instead of [jqPlot](http://www.jqplot.com/).

The idea is replace jqPlot with Chart.js, as looks like jqPlot is not maintained anymore.


williamdes (approved):
💯
Let's remove old jq stuff for charts
~~morris.js is maybe a lib that could help too~~ Edit: it's dead too and uses jquery 😄

PR #18327, opened 2023-04-06
More tracking refactoring

kamil-tekiela:
This significantly simplifies the code, fixes some bugs, and improves some functionality. The fixed bugs should not be present in 5.2, so no need to backport. The changed functionality adds a time zone to SQL dump file and retains original date when deleting tracking data. Frankly, I think it was a bug as I do not find the current behaviour useful. 

BTW if anyone has an idea how to avoid `Assert::notNull($trackingFeature);` I am all ears.

### DIFF
--- a/libraries/classes/Tracking/TrackedData.php
+++ b/libraries/classes/Tracking/TrackedData.php
@@ -1,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpMyAdmin\Tracking;
+
+final class TrackedData
+{
+    /**
+     * @psalm-param list<array{date: string, username: string, statement: string}> $ddlog
+     * @psalm-param list<array{date: string, username: string, statement: string}> $dmlog
+     */
+    public function __construct(
+        public readonly string $dateFrom,
+        public readonly string $dateTo,
+        public readonly array $ddlog,
+        public readonly array $dmlog,
+        public readonly string $tracking,
+        public readonly string $schemaSnapshot,
+    ) {
+    }
+}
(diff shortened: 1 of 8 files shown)


libraries/classes/Tracking/TrackedData.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpMyAdmin\Tracking;
+
```
williamdes, line 6: same comment as for other PRs, can we seal this class as psalm immutable ?

williamdes (approved):

libraries/classes/Tracking/TrackedData.php
kamil-tekiela: DTOs should be final. I forgot about it. Thanks.

williamdes (approved):

libraries/classes/Tracking/Tracking.php
```diff
@@ -651,23 +633,11 @@ public function getHtmlForSchemaSnapshot(string $db, string $table, string $vers
...
-     * }|array<never, never>
      */
-    public function getTrackedData(string $dbname, string $tablename, string $version): array
+    public function getTrackedData(string $dbname, string $tablename, string $version): TrackedData
     {
         $trackingFeature = $this->relation->getRelationParameters()->trackingFeature;
-        if ($trackingFeature === null) {
-            return [];
-        }
+        Assert::notNull($trackingFeature);
```
MauricioFauth, line 640: The best option is to check for `$trackingFeature !== null` in the controllers, as having the tracking feature enabled is required for everything, then pass `$trackingFeature` as a parameter.

MauricioFauth (approved):

PR #18328, opened 2023-04-07
Refactor DbiDummy and improve FieldMetadata

MoonE:
- addResult was called with `[true]` or `[[]]` in some tests, should be just `true`
- fetch_assoc may return array with int key when the column name is an int
- the result handling is now done in DummyResult, it gets constructed with column, result and fieldlmetadata
- instead of marking results as used they are now deleted
- when a Dummy result is created, the result values are now always converted to `string|null`, it's fine to use int/float when calling addResult in tests, this was previouly done only in `fetchValue`
- addResult now expects the row values always as a list, if fetchAssoc is called a separate column config is require. it was used like this most times, but not always

And it fixes a bug in ChartController when a column has an integer name by casting the name to string.
Found due to the now wider return type of fetchAssoc `array<string|null>` instead of `array<string, string|null>`

### DIFF
--- a/libraries/classes/Controllers/Table/GisVisualizationController.php
+++ b/libraries/classes/Controllers/Table/GisVisualizationController.php
@@ -65,10 +65,6 @@ final class GisVisualizationController extends AbstractController
         $labelCandidates = [];
         $spatialCandidates = [];
         foreach ($meta as $columnMeta) {
-            if ($columnMeta->name === '') {
-                continue;
-            }
-
             if ($columnMeta->isMappedTypeGeometry) {
                 $spatialCandidates[] = $columnMeta->name;
             } else {
--- a/libraries/classes/Server/Status/Processes.php
+++ b/libraries/classes/Server/Status/Processes.php
@@ -9,11 +9,11 @@ use PhpMyAdmin\Html\Generator;
 use PhpMyAdmin\Util;
 
 use function __;
-use function array_keys;
+use function array_change_key_case;
 use function count;
-use function mb_strtolower;
 use function strlen;
-use function ucfirst;
+
+use const CASE_LOWER;
 
 final class Processes
 {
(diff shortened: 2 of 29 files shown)


libraries/classes/Server/Status/Processes.php
```diff
@@ -9,11 +9,11 @@
 use PhpMyAdmin\Util;
 
 use function __;
-use function array_keys;
+use function array_change_key_case;
```
kamil-tekiela, line 12: I am pretty sure this does not work with unicode letters, only with ASCII. The previous logic applies to full Unicode.

test/classes/FieldHelper.php
```diff
@@ -0,0 +1,47 @@
...
+        return new FieldMetadata((object) ($metadata + [
+            'name' => 'c',
+            'orgname' => '',
+            'table' => '',
+            'orgtable' => '',
+            'max_length' => 0,
+            'length' => 0,
+            'charsetnr' => -1,
+            'flags' => 0,
+            // 'type' => MYSQLI_TYPE_STRING,
```
kamil-tekiela, line 40: Why is this line commented out?

kamil-tekiela (commented):
I love it. There are a lot of changes, but I reviewed most of it, and it looks good.

libraries/classes/Server/Status/Processes.php
MoonE: So I changed this to get rid of the warning that the property name could be an int which `array_change_key_case` handles without warning (and it's less verbose). I did not consider unicode handling, but I don't think it matters here. This only deals with the column names of the `information_schema.PROCESSLIST` table / output of `SHOW PROCESSLIST` with a defined set of columns.

test/classes/FieldHelper.php
MoonE: I decided to make `type` the only required parameter in the `$metadata` array, so no default is required. I guess the line could be removed ....

libraries/classes/Controllers/Table/GisVisualizationController.php
```diff
@@ -65,10 +65,6 @@ public function __invoke(ServerRequest $request): void
         $labelCandidates = [];
         $spatialCandidates = [];
         foreach ($meta as $columnMeta) {
-            if ($columnMeta->name === '') {
```
williamdes, line 68: I am pretty sure this is one of my bug fixes, not sure it should be removed

MoonE: This is from https://github.com/phpmyadmin/phpmyadmin/pull/18218#discussion_r1131794879 and can be removed.

williamdes: Oh yes, true
I had forgot about it

MauricioFauth (approved):

PR #18340, opened 2023-04-11
Psalm ignore config.inc.iphp

kamil-tekiela:
This is more of a personal request as I want to avoid accidentally adding the baselines for my config file to commits. Unless there is an easier way to avoid it.

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -23,6 +23,9 @@
             <directory name="vendor"/>
             <directory name="libraries/cache"/>
         </ignoreFiles>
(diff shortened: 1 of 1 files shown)


williamdes (approved):
Cool, when I tried it complained not having a config.inc.php file 🤔

kamil-tekiela:
> Cool, when I tried it complained not having a config.inc.php file 🤔

That's why I added `allowMissingFiles="true"`

MauricioFauth (approved):

PR #18351, opened 2023-04-15
Create a value object for ShowGrants

kamil-tekiela:

PR #18357, opened 2023-04-20
Disable JIT for Selenium tests

MauricioFauth:
Related to https://github.com/phpmyadmin/phpmyadmin/pull/18351#issuecomment-1516094785

### DIFF
--- a/test/php.ini
+++ b/test/php.ini
@@ -1929,6 +1929,8 @@ opcache.restrict_api=/disabled/
 ; Absolute path used to store shared lockfiles (for *nix only).
 ;opcache.lockfile_path=/tmp
 
+opcache.jit=disable
+
 [curl]
 ; A default value for the CURLOPT_CAINFO option. This is required to be an
 ; absolute path.


test/php.ini
```diff
@@ -1929,6 +1929,8 @@ opcache.restrict_api=/disabled/
 ; Absolute path used to store shared lockfiles (for *nix only).
 ;opcache.lockfile_path=/tmp
 
+opcache.jit=disable
```
williamdes, line 1932: https://php.watch/versions/8.0/JIT#jit-opcache.jit

off would be a bit better
But, can we debug the stituation and not ignore it ?
Because it would mean that real users can walk into this issue ?
~~Since this php.ini config is a sync of the Debian one~~
Edit: I just searched into git and it looks like my changes never did get into git. Well, maybe it's worth checking that are the default configs on other systems

MauricioFauth: Yeah, I opened it to see if the error go away. Looks like the type errors didn't happen. As @kamil-tekiela said, looks like a JIT bug.

williamdes: Okay, let's try to not hide bugs in our testing tools. If users can have this bug using a default config, it's best to not hide or ignore it

PR #18359, opened 2023-04-20
InsertEdit refactoring - part 1

kamil-tekiela:
This PR removes a lot of unnecessary code and refactors some of the methods to make it easier to understand. The PR is already getting large, so I will split it into chunks for easier reviewing.


libraries/classes/InsertEdit.php
```diff
@@ -1170,29 +1008,24 @@ public function getErrorUrl(array $urlParams): string
...
-
-        return [
-            $insertCommand . 'INTO '
-            . Util::backquote($GLOBALS['table'])
+    public function buildInsertSqlQuery(
+        string $table,
+        bool $isInsertIgnore,
+        array $queryFields,
+        array $valueSets,
+    ): string {
```
williamdes, line 1024: query building functions should go into the query generator class

kamil-tekiela: I agree, which is why I left a todo comment. This is only part 1 and I didn't want to make this change yet.

MauricioFauth (approved):

PR #18375, opened 2023-04-30
Add the `Config::getCurrentServer()` method 

MauricioFauth:
The `Config::getCurrentServer()` method returns the current server configuration.
It's a value object that is equivalent to `$GLOBALS['cfg']['Server']` setting.

### DIFF
--- a/libraries/classes/Config.php
+++ b/libraries/classes/Config.php
@@ -86,9 +86,6 @@ class Config
 
     public bool $errorConfigFile = false;
 
-    /** @var mixed[] */
-    public array $defaultServer;
-
     private bool $isHttps = false;
 
     private Settings $config;
@@ -96,7 +93,6 @@ class Config
     public function __construct()
     {
         $this->config = new Settings([]);
-        $this->defaultServer = $this->config->Servers[1]->asArray();
         $config = $this->config->asArray();
         $this->default = $config;
         $this->settings = $config;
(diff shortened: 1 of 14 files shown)


libraries/classes/Config.php
```diff
@@ -86,17 +86,14 @@ class Config
 
     public bool $errorConfigFile = false;
 
-    /** @var mixed[] */
-    public array $defaultServer;
-
     private bool $isHttps = false;
 
     private Settings $config;
+    private Settings\Server|null $currentServer = null;
```
kamil-tekiela, line 92: Could this be imported?

MauricioFauth: @kamil-tekiela I was not happy with that property, so I removed it. Instead, I created a `$server` property that is equivalent to `$GLOBALS['server']` and updated the `getCurrentServer()` method.

libraries/classes/Config.php
```diff
@@ -1252,4 +1241,9 @@ public function getSettings(): Settings
     {
         return $this->config;
     }
+
+    public function getCurrentServer(): Settings\Server|null
```
kamil-tekiela, line 1245: And what about this one?

MauricioFauth: Do you mean like injecting a value for testing?
Something like that:
```php
$config = new Config();
$config->config = new Settings(['Servers' => [2 => ['host' => 'testhost']]]);
$config->server = 2;
$currentServer = $config->getCurrentServer();
```

kamil-tekiela: No, sorry. I meant can we replace `Settings\Server` with just `Server`?

MauricioFauth: Haha of course. That makes sense. Done!

PR #18386, opened 2023-05-06
Refactor Config::getConnectionParams() method 

MauricioFauth:
- Add null as default for Server control_ properties
- Removes the $server array param and use the Server class instead.
- Uses the user and pass Server properties instead of returning an array.
- Add Server class as a param for DatabaseInterface::postConnect()

### DIFF
--- a/libraries/classes/Common.php
+++ b/libraries/classes/Common.php
@@ -244,6 +244,7 @@ final class Common
             }
 
             $authPlugin->authenticate();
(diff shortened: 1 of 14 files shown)


kamil-tekiela (approved):
Looks good. Why are the Selenium tests failing again?

MauricioFauth:
> Looks good. Why are the Selenium tests failing again?

Because the cookie auth plugin changes `$GLOBALS['cfg']['Server']` and the `Server` object wasn't update.

Issue #17769, opened 2022-10-11
Request superglobals in controllers should be replaced with `ServerRequest` object


MauricioFauth:
Instead of getting the request parameters from superglobals (`$_POST`, `$_GET` and `$_REQUEST`) in controllers, they should be got from the [`PhpMyAdmin\Http\ServerRequest`](https://github.com/phpmyadmin/phpmyadmin/blob/6785c97d22e764bb87cfa86eae199d5f7909297a/libraries/classes/Http/ServerRequest.php) object that is available for all controllers.

Controllers have a method called `__invoke` that has a signature like the following:
```php
public function __invoke(ServerRequest $request): void
```

So if the controller gets a parameter from the `$_POST` superglobal, it can be replaced to get from the `$request` variable. For example:
```diff
- $param = $_POST['param'] ?? null;
+ $param = $request->getParsedBodyParam('param');

- $value = $_POST['value'] ?? 'default';
+ $value = $request->getParsedBodyParam('value', 'default');
```

Or from the `$_GET` superglobal:
```diff
- $param = $_GET['param'] ?? null;
+ $param = $request->getQueryParam('param');

- $value = $_GET['value'] ?? 'default';
+ $value = $request->getQueryParam('value', 'default');
```

If you want to get from `$_POST` or from `$_GET`:
```diff
- $param = $_POST['param'] ?? $_GET['param'] ?? null;
+ $param = $request->getParam('param');

- $value = $_POST['param'] ?? $_GET['value'] ?? 'default';
+ $value = $request->getParam('value', 'default');
```

You can also get all the request parameters. For example:
```diff
- $getParams = $_GET;
+ $getParams = $this->getQueryParams();

- $postParams = $_POST;
+ $postParams = $this->getParsedBody();
```

> **Note**
> Please be careful that while the `$_POST` and `$_GET` are global variables, the `ServerRequest` is a immutable object.
> So, if you encounter something like this in code: 
> ```php
> $_POST['param'] = 'new value';
> ```
> You should first find all places in the code that uses this parameter and refactor the code to not write on the global variable.

- Related to https://github.com/phpmyadmin/phpmyadmin/issues/16276
- Related to https://github.com/phpmyadmin/phpmyadmin/issues/15507

clxswalther:
I'm currently working my way through this and I'm wondering, if `isset()`-statements should be refactored to `!isnull()` or if this will be addressed in a later stage/other issue.

[...]

PR #18390, opened 2023-05-07
Replace superglobals with serverrequest in GetFieldController

ru-asdx:
### Description
Ref https://github.com/phpmyadmin/phpmyadmin/issues/17769

Replace superglobals with ServerRequest in 

      libraries/classes/Controllers/Table/GetFieldController.php

### DIFF
--- a/libraries/classes/Controllers/Table/GetFieldController.php
+++ b/libraries/classes/Controllers/Table/GetFieldController.php
@@ -54,10 +54,11 @@ class GetFieldController extends AbstractController
             Generator::mysqlDie(__('Invalid table name'));
         }
 
+        $whereClause = (string) $request->getQueryParam('where_clause', '');
+        $whereClauseSign = (string) $request->getQueryParam('where_clause_sign', '');
         if (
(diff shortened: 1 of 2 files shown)


libraries/classes/Controllers/Table/GetFieldController.php
```diff
@@ -54,10 +54,12 @@ public function __invoke(ServerRequest $request): void
             Generator::mysqlDie(__('Invalid table name'));
         }
 
+        $whereClause = $request->getQueryParam('where_clause');
+        $whereClauseSign = $request->getQueryParam('where_clause_sign');
         if (
-            ! isset($_GET['where_clause'])
-            || ! isset($_GET['where_clause_sign'])
-            || ! Core::checkSqlQuerySignature($_GET['where_clause'], $_GET['where_clause_sign'])
+            ! isset($whereClause)
```
williamdes, line 60: you should check for not null because it is always is set now

MauricioFauth (approved):

PR #18393, opened 2023-05-08
Remove session cache in relation

kamil-tekiela:
So I took another look at this session cache feature and it turns out that I made it obsolete when I implemented the static cache. The session was only used by tests, which didn't need to use it. In fact, using session led to some confusing tests. The session cache needs to be removed from all tests. 

@williamdes In regards to the new test you recently added, I made it pass, but the test isn't testing anything. I tried to understand what the test was meant to do, but I gave up after some time. I left a TODO note to fix the test in the future. I think the problem for me is that I don't fully understand how `$GLOBALS['cfg']['Server']` works. If you know how to fix the test, I would appreciate.

### DIFF
--- a/libraries/classes/ConfigStorage/Relation.php
+++ b/libraries/classes/ConfigStorage/Relation.php
@@ -57,8 +57,7 @@ use const SQL_DIR;
  */
 class Relation
 {
-    /** @var RelationParameters[] $cache */
-    private static array $cache = [];
+    private static RelationParameters|null $cache = null;
 
     public function __construct(public DatabaseInterface $dbi)
     {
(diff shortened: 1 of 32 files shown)


williamdes:
Thanks for this PR, I need to check this once more

But TLDR RelationParameters seems not to be cached at all, this needs to persist in the session because it costs a lot to know if the relation features are working. Not sure if this PR is related or not

Will have another look asap ;)

kamil-tekiela:
@williamdes Can we merge it as it is and we will revisit the TODO later?

williamdes:
> @williamdes Can we merge it as it is and we will revisit the TODO later?

Sure, I made some work on this subject recently on QA and master

williamdes (approved):

williamdes:
Can you rebase it so we are sure it still passes?

williamdes:
> I don't fully understand how $GLOBALS['cfg']['Server'] works. If you know how to fix the test, I would appreciate

From what I understood it represents the state of [$cfg['Servers'][$i]](https://docs.phpmyadmin.net/en/latest/config.html#cfg_Servers_pmadb) for the selected server

MauricioFauth:
> > I don't fully understand how $GLOBALS['cfg']['Server'] works. If you know how to fix the test, I would appreciate
> 
> From what I understood it represents the state of [$cfg['Servers'][$i]](https://docs.phpmyadmin.net/en/latest/config.html#cfg_Servers_pmadb) for the selected server 

To be more precise, it holds the server information of the current connection.

MauricioFauth (approved):

PR #18430, opened 2023-05-13
Feature request: Remove QBE

### LINKED ISSUE #18041: Does Qbe actually work?
I wanted to clean up Qbe.php but not long after I started doing some refactoring I realized that there's a lot of dead code in this class, I cannot get some features to work, and the UI is very unintuitive. I don't want to list every issue I found, because I keep finding more and more, but the main ones are:
- Visual builder doesn't work at all. Whenever I select any column I get a JS error. I cannot reproduce it on STABLE so it might be something to do with my setup?
- The performance is absolutely atrocious whenever I select two columns from two different tables, but...
- What does that actually mean to select columns from two tables? After a while, I got a response back but the result is just the old style SQL join e.g. ``FROM `bar`, `foo` ``. How do I join them?
- The WHERE clauses are very brittle. It's QBE but I have to type the condition out. If I don't use the proper SQL syntax I get garbage output. 
- By default, the page contains two fields for Criteria. Why? If the user is not careful it may result in a PHP error `mb_strtoupper(): Argument #1 ($string) must be of type string, null given` or `Undefined array key ""`
- The second box allows me to "accidentally" select the wrong table. Let's say I have spent 15 minutes building my WHERE conditions and listing out all the columns but before I press "update query" I accidentally select the wrong table. All is gone! What exactly does this selection do? The only thing I noticed is that it restricts the column dropdown. Why does it have a dedicated button?
- The "ins" and "del" checkboxes are nightmarishly confusing. I think they are meant to indicate which column should be added/removed and which condition should be expanded. But then there are two dropdowns beside the "update query" button. What's the difference? Why does the dropdown say "Delete" if it doesn't seem to matter anyway?

A quick look at the code reveals more issues:
- Dead code... like a lot of it is never executed because it's unreachable. Psalm/PHPStan would complain except they can't because all is array and arrays are magic.  But not only that. There is code just because... (technically all legacy PHP codebases have this problem). Some parts of this class have no clear purpose and despite being executed, don't change the outcome. It's a problem stemming from old code that has been refactored many times and language changes in PHP, but the problem is that it's not self-explanatory so it's hard to say if it should be fixed or just removed. 
- The code is full of things that make it difficult to read and understand: superglobals, globals, variable variables, arrays, isset/empty and inline HTML. It's not specific to only this class though. 

I like refactoring old PHP code, but I need to understand what it does and how it's supposed to work. This feature has no clear purpose and doesn't work fully for me. I can take small steps and make the code clearer, but I don't know what is a bug that needs fixing and what can simply be removed. Is there any technical specification for this feature? Any documentation or guide? Do people actually use it? 

----

At a theoretical level, QBE is meant to ease the use of RDMBS by abstracting SQL through buttons, checkboxes and dropdowns. The software should offer UI that **replaces** user's knowledge of SQL. The SQL is generated and executed behind the scenes. This is not what phpMyAdmin does. In PMA I need to know SQL to use this feature. Not only that, I need to know how to use this terrible UI. PMA can and will generate broken SQL, e.g. don't select any column but select a sorting direction.

The following paper describes the QBE language: https://pages.cs.wisc.edu/~dbbook/openAccess/thirdEdition/qbe.pdf Please note that what PMA offers is not the same thing. I use a tool called "DB browser for SQLite". It offers QBE-like feature. It works great and is very intuitive. PMA's offering doesn't come close to that. 

What can I do to fix this situation? Should we try to salvage what we have or try to design something new?

kamil-tekiela:
As [explained previously](https://github.com/phpmyadmin/phpmyadmin/issues/18041), the feature of Query By Example is very flawed and has been mostly replaced by the new multi-query. Modernizing QBE is too big of a task for little to no gain. Let's remove this functionality altogether. 

Open point: Designer is currently coupled with QBE and without it, it cannot execute queries. It should be relatively simple to substitute it with a dedicated controller. This can be done in another PR. 

- Closes https://github.com/phpmyadmin/phpmyadmin/issues/18041


williamdes (approved):
I think dropping it in 6.0 is a nice idea
We will have some non happy users, but I agree the feature should be re done after fully deleting it

MauricioFauth (approved):

PR #18468, opened 2023-05-30
Refactor Template class to lazy load the Twig Environment

MauricioFauth:
Loads the Twig Environment class only when rendering a template for the first time, not when constructing the Template object.

### DIFF
--- a/libraries/classes/Template.php
+++ b/libraries/classes/Template.php
@@ -40,20 +40,16 @@ class Template
     /**
      * Twig environment
      */
-    protected static Environment $twig;
+    protected static Environment|null $twig = null;
 
     public const TEMPLATES_FOLDER = ROOT_PATH . 'templates';
 
+    private Config|null $config;
+
     public function __construct(Config|null $config = null)
     {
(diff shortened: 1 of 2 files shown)


libraries/classes/Template.php
```diff
@@ -104,6 +100,10 @@ public static function getTwigEnvironment(string|null $cacheDir): Environment
      */
     private function load(string $templateName): TemplateWrapper
     {
+        if (static::$twig === null) {
+            static::$twig = self::getTwigEnvironment($this->config?->getTempDir('twig'));
```
williamdes, line 104:
```suggestion
            static::$twig = self::getTwigEnvironment($this->config->getTempDir('twig'));
```
Looks wrong to allow a non existing config

williamdes (approved):

libraries/classes/Template.php
MauricioFauth: Yeah, I know. Moving the getTempDir call out of the constructor is the first step. Now I need to check if it's really possible that config will ever be null.

PR #18217, opened 2023-03-05
Fix `Invalid opcode 23/1/0`

faissaloux:
Hello 👋🏻 

Got this error using php8.1.1
` Fatal error: Invalid opcode 23/1/0. in C:\laragon\etc\apps\phpMyAdmin\libraries\classes\Common.php on line 82`

It's a php bug which was fixed on php8.1.2
https://bugs.php.net/bug.php?id=81684

So in this PR I have forced requirement to minimum php8.1.2 which fixed it.

PR #18474, opened 2023-05-31
Require min php 80102 (8.1.2)

faissaloux:
On #18217 I have required minimum php version to 8.1.2 because of a php bug.
In this PR I'm updating its check from `index.php`

### DIFF
--- a/public/index.php
+++ b/public/index.php
@@ -12,8 +12,8 @@ if (! defined('ROOT_PATH')) {
 define('PHPMYADMIN', true);
 // phpcs:enable
 
-if (PHP_VERSION_ID < 80100) {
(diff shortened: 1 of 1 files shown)


williamdes (approved):
Can you find the link to the PHP bug ?

williamdes:
I pushed a bunch more stuff into bf9519444a018647a47e439d300fa9d82a768d4d

williamdes:
> > Can you find the link to the PHP bug ?
> 
> https://bugs.php.net/bug.php?id=81684

https://github.com/php/php-src/pull/7717

Thank you, this will be easier to find in some years

PR #18477, opened 2023-06-01
Create the `Identifiers\Identifier` interface

MauricioFauth:
Changes `Dbal\DatabaseName`, `Dbal\TableName` and `Triggers\TriggerName` to implements the new `Identifiers\Identifier` interface.
Moves `Dbal\DatabaseName`, `Dbal\TableName` and `Triggers\TriggerName` to the
`Identifiers` namespace.
Renames the `tryFromValue()` and `fromValue()` methods to just `tryFrom()` and `from()` respectively.


MauricioFauth:
cc @kamil-tekiela

kamil-tekiela (approved):
Looks good. I only have one question. Does it have to be in Dbal namespace? Maybe a dedicated namespace would be better?

MauricioFauth:
> Looks good. I only have one question. Does it have to be in Dbal namespace? Maybe a dedicated namespace would be better?

What namespace you suggest?

MauricioFauth:
> Maybe `Identifiers`? For me, this interface doesn't have anything to do with Dbal logic.

It really makes more sense than Dbal. I also rename IdentifierName to just Identifier, as it seemed redundant.

kamil-tekiela (approved):
This looks sensible.

PR #18478, opened 2023-06-01
Remove empty option from enum select

faissaloux:
Hello 👋🏻 

In this PR I have removed an unnecessary empty option (Not included in ENUM) from enum select.

### Before
image

### After merging this PR
image

### DIFF
--- a/templates/sql/enum_column_dropdown.twig
+++ b/templates/sql/enum_column_dropdown.twig
@@ -1,5 +1,4 @@
 <select>
-  <option value="">&nbsp;</option>
   {% for value in values %}
     <option value="{{ value|raw }}"{{ value in selected_values ? " selected" }}>{{ value|raw }}</option>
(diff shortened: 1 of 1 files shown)


williamdes:
Are you sure this will have no impacts on other forms or for example editing columns ?

faissaloux:
> Are you sure this will have no impacts on other forms or for example editing columns ?

I don't think it has any impact on other forms since it's only related to enum's dropdown.

williamdes:
As this is not really a bug or annoyance, it should go to master branch I think

faissaloux:
> As this is not really a bug or annoyance, it should go to master branch I think

Done!

williamdes (approved):

ibennetch:
> As this is not really a bug

This seems like a regression to me, I wouldn’t mind seeing this fixed in
QA_5_2.

PR #18484, opened 2023-06-06
fix NULL not applied after clearing nullable field

### LINKED ISSUE #18473: `NULL` not applied after clearing nullable field
### Describe the bug

`NULL` not applied after clear nullable field, the field updated to '' (empty value) instead of `NULL` even NULL checkbox is checked automatically.
```sql
UPDATE `Country` SET `HeadOfState` = '' WHERE `Country`.`Code` = 'AZE';
```

### To Reproduce

1. Go to https://demo.phpmyadmin.net/QA_5_2/index.php?route=/sql&db=world&table=Country&pos=0
2. Double click on some nullable field to update it (eg: `HeadOfState`).
3. Clear its value. (NULL is checked automatically)
4. Click somewhere outside.
5. NULL not applied.

### Expected behavior

When clearing a nullable field, set to `NULL` instead of '' (empty value).
```sql
UPDATE `Country` SET `HeadOfState` = NULL WHERE `Country`.`Code` = 'AZE';
```

### Screenshots

https://github.com/phpmyadmin/phpmyadmin/assets/60013703/b5608868-6353-4e1a-849d-39658b977749

### Server configuration

### Additional context

This bug appears only after updating the field and clicking outside. If click `ENTER` after update it's working as expected.

faissaloux:
Fixes #18473

After @SachinBahukhandi #18483 PR that fixes the issue on `v6.0.0`
This PR fixes the same issue on `v5.2.2`

### DIFF
--- a/js/src/makegrid.js
+++ b/js/src/makegrid.js
@@ -845,7 +845,9 @@ var makeGrid = function (t, enableResize, enableReorder, enableVisib, enableGrid
                         });
                     } else {
                         $(g.cEdit).on('keypress change paste', '.edit_box', function () {
-                            $checkbox.prop('checked', false);
+                            if ($(this).val() !== '') {
+                                $checkbox.prop('checked', false);
+                            }
                         });
                         // Capture ctrl+v (on IE and Chrome)
                         $(g.cEdit).on('keydown', '.edit_box', function (e) {


williamdes:
Merged as e17e32e747ad8cd9bdd0acdbfa9695f3ca62bd02

PR #18485, opened 2023-06-07
Add a wrapper around the `exit` language construct

MauricioFauth:
Creates the `PhpMyAdmin\ResponseRenderer::callExit()` method as a wrapper to the `exit` language construct. That makes the tests more correct as the method will always stop the code execution.


williamdes (approved):
Ref: #15247

PR #18503, opened 2023-06-16
fix: empty html options need to have a label

### LINKED ISSUE #12747: Errors/Warnings from HTML checker
Using HTML markup checker such as:
https://validator.w3.org/#validate_by_uri
return many errors/warnings on PMA output, for example:
https://validator.w3.org/nu/?doc=https%3A%2F%2Fdemo.phpmyadmin.net%2Fmaster-config%2F
such as:
```
Error: The clear attribute on the br element is obsolete. Use CSS instead.
Error: The width attribute on the table element is obsolete. Use CSS instead.
Error: The cellspacing attribute on the table element is obsolete. Use CSS instead.
Error: Element option without attribute label must not be empty.
```

ThanasisMpalatsoukas:
empty options are errors in html validators and they need to have at least a label to be correct. By adding a label we solve the problem with the validator while keeping the code functionallity the same. Related to #12747

### Description

Related to #12747

Fixes part of 12747 problem that relates to options not having a label. Fixed it by checking
if the value is empty, and in case it is adding a 'None' label which is translated. 


templates/config/form_display/input.twig
```diff
@@ -40,7 +40,7 @@
       <select name="{{ path }}" id="{{ path }}" class="w-75{{ not value_is_default ? (has_errors ? ' custom field-error' : ' custom') }}">
         {% for key, val in select_values %}
           {% if val is same as(true) %}{% set val = 'Yes'|trans %}{% elseif val is same as(false) %}{% set val = 'No'|trans %}{% endif %}
-          <option value="{{ key }}"{{ key is same as(value) or (value is same as(true) and key is same as(1)) or (value is same as(false) and key is same as(0)) ? ' selected' }}{{ key in select_values_disabled ? ' disabled' }}>{{ val }}</option>
+          <option value="{{ key }}" {{ val is empty ? 'label=' ~ 'None'|trans : ""}} {{ key is same as(value) or (value is same as(true) and key is same as(1)) or (value is same as(false) and key is same as(0)) ? ' selected' }}{{ key in select_values_disabled ? ' disabled' }}>{{ val }}</option>
```
MauricioFauth, line 43: This will reduce the unnecessary white spaces. The empty string is not necessary for the else side of the ternary in Twig.

```suggestion
          <option value="{{ key }}"{{ val is empty ? ' label="' ~ 'None'|trans ~ '"'}}{{ key is same as(value) or (value is same as(true) and key is same as(1)) or (value is same as(false) and key is same as(0)) ? ' selected' }}{{ key in select_values_disabled ? ' disabled' }}>{{ val }}</option>
```

MauricioFauth (approved):

MauricioFauth:
Merged. Thank you for your contribution!

Issue #18505, opened 2023-06-17
UI bug with bootstrap theme on settings > Page related settings 


ThanasisMpalatsoukas:
### Describe the bug

The UI is buggy on settings > Page related settings 

### To Reproduce

1. Go to http://localhost:5000/public/index.php
2. Click on The cog icon

### Expected behavior

The list style should probably have a list-style: none so the dots are not visible

### Screenshots

![image](https://github.com/phpmyadmin/phpmyadmin/assets/43223812/be5fd8ea-7afc-4fdc-a862-771f712a10eb)

### Server configuration

- Operating system: Ubuntu 22.04.2 LTS
- Web server: 
- Database version: MySQL 8.0.33-0ubuntu0.22.04.2 
- PHP version: 8.2.7
- phpMyAdmin version: 6.0.0-dev

### Client configuration

- Browser: Chrome Version 112.0.5615.121 (Official Build) (64-bit)
- Operating system: 

### Additional context

Add any other context about the bug here.

ThanasisMpalatsoukas:
I will try to fix the problem if the issue is confirmed

liviuconcioiu:
Duplicate of #18162 (task 3)

> I will try to fix the problem if the issue is confirmed

@ThanasisMpalatsoukas feel free to open a PR. Make sure to fix the issue on all themes.

heidygallardo:
Hi! I'm a student in the CodePath AI301 program and I'd like to work on this issue. I'll start by reproducing the problem locally this week and investigating the root cause. Any pointers on where to start in the codebase would be appreciated. If possible, could I be assigned to this issue?

liviuconcioiu:
Feel free to open a PR. The previous one has been closed. You can see the discussion around it here - https://github.com/phpmyadmin/phpmyadmin/pull/18507

heidygallardo:
> Feel free to open a PR. The previous one has been closed. You can see the discussion around it here - [#18507](https://github.com/phpmyadmin/phpmyadmin/pull/18507)

Got it thank you!

liviuconcioiu:
@williamdes this no longer happens. Probably it was a Bootstrap bug? Issue can be closed.

PR #18507, opened 2023-06-17
fix: fixed the dots besides the navigation bar

ThanasisMpalatsoukas:
configFormDisplayTab seems to be a good selector to add the list-style: none. i checked all themes and the problem seems to dissapear.

fixes part of #18505 

### DIFF
--- a/public/themes/bootstrap/scss/_forms.scss
+++ b/public/themes/bootstrap/scss/_forms.scss
@@ -2,3 +2,7 @@ select#fieldsSelect,
 textarea#sqlquery {
   height: 20rem;
 }
+
+#configFormDisplayTab li {
+  list-style: none;
+}
--- a/public/themes/metro/scss/_forms.scss
+++ b/public/themes/metro/scss/_forms.scss
@@ -7,3 +7,7 @@ textarea#sqlquery {
 select.form-control {
   padding: 4px;
 }
+
+#configFormDisplayTab li {
+  list-style: none;
+}
--- a/public/themes/original/scss/_forms.scss
+++ b/public/themes/original/scss/_forms.scss
@@ -2,6 +2,7 @@ select#fieldsSelect,
 textarea#sqlquery {
   height: 20rem;
 }
+
 #configFormDisplayTab li {
   list-style: none;
 }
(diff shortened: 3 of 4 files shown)


liviuconcioiu:
@ThanasisMpalatsoukas while this fixes the issue, I think it can be fixed on another part of the code and avoid adding additional CSS styles.

The code that creates the modals should be changed, because the modals are created at the same time in two places on the page. The right place is at the bottom:

two modals

Let's wait for @williamdes review.

ThanasisMpalatsoukas:
@liviuconcioiu weird. There seems to be only one instance of a pageSettingsModal modal in html in what im am seeing while having the modal opened.

image

williamdes (changes requested):
> @ThanasisMpalatsoukas while this fixes the issue, I think it can be fixed on another part of the code and avoid adding additional CSS styles.

Thanks for the first review, indeed we are trying to remove all custom non bootstrap styles. Adding one more will not help
Let's see if there is another solution around that
Maybe bootstrap has some class to do the same job ?

williamdes:
Most probably the code should use tabs instead: https://getbootstrap.com/docs/5.2/components/navs-tabs/

liviuconcioiu:
Inside `maintcontainer` there is `pageSettingsModal`, but at the same time there is `pageSettingsModal` right under it. The same for `functionConfirmModal` and `addIndexModal` (don't know about those two where should be as I never encountered issues).

https://github.com/phpmyadmin/phpmyadmin/assets/25424343/253daa96-1f74-4a8c-8cdd-ffc048cd1274

ThanasisMpalatsoukas:
@liviuconcioiu it does seem that you are right that the second modal creates the problem with the css. You can see that when the modal is at the bottom the right css applies. I will try to see why the modals are generated twice.

MauricioFauth (changes requested):
I think removing this entire block from common.scss will be better:

```css
#maincontainer {
  ul {
    list-style-type: disc;
    vertical-align: middle;
  }

  li {
    margin-bottom: 0.3em;
  }
}
```

williamdes:
I am closing this one as no updates where pushed, feel free to ask for re-open and push updates

PR #18508, opened 2023-06-18
Replace jqPlot with Chart.js for the profiling chart

MauricioFauth:


williamdes (approved):

liviuconcioiu:
Awesome! Can data labels be added to show the percentage over the pie like it used with the jqPlot? Also seconds precision seems to be a bit lost, at least for the operations that take most time.

MauricioFauth:
> Awesome! Can data labels be added to show the percentage over the pie like it used with the jqPlot?

I don't know. Maybe. I'm still learning how to use Chart.js. 

> Also seconds precision seems to be a bit lost, at least for the operations that take most time.

The seconds in the chart tooltip? Or in the table?

liviuconcioiu:
> The seconds in the chart tooltip? Or in the table?

In the chart tooltip.

MauricioFauth:
> In the chart tooltip.

All values that I tested are the same, only with a different notation. However, is possible that Intl.NumberFormat is rounding it in a weird way. Can you show me an example? Is it from a sum of more calls, because that's the only time it's done math, otherwise is same value that the db server provides.

liviuconcioiu:
> > In the chart tooltip.
> 
> All values that I tested are the same, only with a different notation. However, is possible that Intl.NumberFormat is rounding it in a weird way. Can you show me an example? Is it from a sum of more calls, because that's the only time it's done math, otherwise is same value that the db server provides.

The values are okay, but I think is a limitation from ChartJS.

timing

https://www.chartjs.org/docs/latest/axes/cartesian/time.html#time-units

PR #18525, opened 2023-07-08
Fix Mutation test for all files

MauricioFauth:
This is an attempt to understand why the mutation tests for all files are failing.

### DIFF
--- a/.github/workflows/mutation-tests.yml
+++ b/.github/workflows/mutation-tests.yml
@@ -13,6 +13,7 @@ on:
       - master
     paths:
       - '*/classes/**.php'
+  workflow_dispatch:
   schedule:
     - cron: '0 0 * * *'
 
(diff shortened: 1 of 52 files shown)


MauricioFauth:
Looks like it's hitting some runner limit, like a CPU limit or memory limit. Most probably the memory limit.

williamdes:
> Looks like it's hitting some runner limit, like a CPU limit or memory limit. Most probably the memory limit.

Looks like it https://github.com/actions/runner-images/issues/6680#issuecomment-1334848113

Exit Code 143: Graceful Termination (SIGTERM)

MauricioFauth:
Yeah, I saw that issue.
I'll look more into it if there's something we could do about it without using a hosted runner or giving up running Infection for all files.

MauricioFauth:
@williamdes it was not canceled this time.

I added an annotation to ignore most of the line that caused timeouts or errors. Most of them are infinite loops.
Looks like one of those or the sum of them are causing the CPU or memory limit.

Next step is to add tests to kill these mutants.

williamdes:
Coool, let's kill them

PR #18534, opened 2023-07-16
Improve copy to clipboard code

### LINKED ISSUE #18529: add copy to export like text
In the export section, there is an option that indicates if I want this in a text box, will it be possible to add a copy button to the result view??? currently it says "update" and "back"

image

kamioon:
### Description

This is a PR for adding a **copy** to the clipboard on the export page that was requested (#18529). 

phpmyadmin_copy_export

To keep the text format copied to the clipboard temporary `input` has changed to the `textarea`.

### DIFF
--- a/js/src/modules/functions.ts
+++ b/js/src/modules/functions.ts
@@ -1519,7 +1519,7 @@ function dismissNotifications () {
             var copyText = $(this).attr('data-text');
             var copyTarget = $(this).attr('data-target');
             if (copyTarget) {
-                copyText = String($(copyTarget).val());
+                copyText = $(copyTarget).val();
             }
 
             if (Functions.copyToClipboard(copyText)) {
--- a/js/src/modules/functions.ts
+++ b/js/src/modules/functions.ts
@@ -1516,22 +1516,28 @@ function dismissNotifications () {
 
         $(document).on('click', 'a.copyQueryBtn', function (event) {
             event.preventDefault();
-            var copyText = $(this).attr('data-text');
-            var copyTarget = $(this).attr('data-target');
-            if (copyTarget) {
-                copyText = $(copyTarget).val();
-            }
+            var copyStatus = Functions.copyToClipboard($(this).attr('data-text'));
+            displayCopyStatus(this, copyStatus);
+        });
 
-            if (Functions.copyToClipboard(copyText)) {
-                $(this).after('<span id=\'copyStatus\'> (' + window.Messages.strCopyQueryButtonSuccess + ')</span>');
+        $(document).on('click', 'a.copyExportBtn', function (event) {
+            event.preventDefault();
+            var copyTarget = $(this).attr('data-target');
+            var copyStatus = Functions.copyToClipboard($(copyTarget).val());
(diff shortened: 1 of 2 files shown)


williamdes (approved):
Very nice

js/src/modules/functions.ts
```diff
@@ -1516,8 +1516,13 @@ function dismissNotifications () {
 
         $(document).on('click', 'a.copyQueryBtn', function (event) {
             event.preventDefault();
-            var res = Functions.copyToClipboard($(this).attr('data-text'));
-            if (res) {
+            var copyText = $(this).attr('data-text');
+            var copyTarget = $(this).attr('data-target');
+            if (copyTarget) {
+                copyText = String($(copyTarget).val());
```
williamdes, line 1522:
```suggestion
                copyText = $(copyTarget).val();
```

williamdes (approved):
💯

williamdes:
Can you check why tests are not passing?

williamdes (changes requested):
what was wrong with the previous version?

kamioon:
`[tsl] ERROR in /home/runner/work/phpmyadmin/phpmyadmin/js/src/modules/functions.ts(1522,17)
      TS2322: Type 'string | number | string[]' is not assignable to type 'string'.`
@williamdes the `.val()` returns a mixed type and that was the reason I had added the `String()`. I've changed a little bit more. separated them into two different functions and moved the copy message to a new function. I think it's better now. otherwise, reverting the two latest commits could work fine but your commit suggestion wouldn't be applied there.

kamioon:
> what was wrong with the previous version?

Sorry, I just saw that you had approved the previous one. I explained here the reason for failing the test and I think the new changes are cleaner than using the `copyQueryBtn` for two purposes. (it should handle two different text sources.)

> `[tsl] ERROR in /home/runner/work/phpmyadmin/phpmyadmin/js/src/modules/functions.ts(1522,17) TS2322: Type 'string | number | string[]' is not assignable to type 'string'.` @williamdes the `.val()` returns a mixed type and that was the reason I had added the `String()`. I've changed a little bit more. separated them into two different functions and moved the copy message to a new function. I think it's better now. otherwise, reverting the two latest commits could work fine but your commit suggestion wouldn't be applied there.

js/src/modules/functions.ts
```diff
@@ -1516,17 +1516,28 @@ function dismissNotifications () {
...
+            displayCopyStatus(this, copyStatus);
+        });
+
+        $(document).on('click', 'a.copyExportBtn', function (event) {
+            event.preventDefault();
+            var copyTarget = $(this).attr('data-target');
+            var copyStatus = Functions.copyToClipboard($(copyTarget).val());
+            displayCopyStatus(this, copyStatus);
+        });
+
```
williamdes, line 1529: could you add a block of documentation like other functions to document the input and output
or much better: add typescript types

williamdes (approved):
Thank for explaining, looks okay
I left one minor comment about typing

williamdes (approved):
This really looks good

Issue #18533, opened 2023-07-16
Simulate query matched rows wrong count


liviuconcioiu:
### Describe the bug

When simulating a UPDATE query, the matched rows count is wrong sometimes. This happens if the column contains NULL values.

The generated query for count is ```SELECT `b` FROM `test` WHERE `a` = '2' AND (`b` <> 'test')``` but it probably should be ```SELECT `b` FROM `test` WHERE `a` = '2' AND (`b` <> 'test' OR `b` IS NULL)```

### To Reproduce

1. Create table
```SQL
CREATE TABLE `test` (
  `id` int(10) UNSIGNED NOT NULL,
  `a` varchar(255) DEFAULT NULL,
  `b` text DEFAULT NULL
) ENGINE=MyISAM DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;

INSERT INTO `test` (`id`, `a`, `b`) VALUES
(1, '1', 'test'),
(2, '1', NULL),
(3, '1', NULL),
(4, '1', NULL),
(5, '1', 'test'),
(6, '2', NULL),
(7, '2', NULL),
(8, '2', 'test'),
(9, '2', NULL),
(10, '2', 'test');

ALTER TABLE `test`
  ADD PRIMARY KEY (`id`);

ALTER TABLE `test`
  MODIFY `id` int(10) UNSIGNED NOT NULL AUTO_INCREMENT, AUTO_INCREMENT=11;
COMMIT;
```
2. Go to SQL tab and simulate this query
```UPDATE `test` SET `b` = 'test' WHERE `a` = '2'```
3. See the count 0 which is wrong. It should be 3.

### Expected behavior

The count should be accurate.

### Screenshots

https://github.com/phpmyadmin/phpmyadmin/assets/25424343/8c46380d-1ac3-471e-8e88-64ac4a168532

### Server configuration

- phpMyAdmin version: 5.2.2-dev, 6.0.0-dev

williamdes:
Did it appear before 5.2.2-dev ?

liviuconcioiu:
> Did it appear before 5.2.2-dev ?

I really don't remember. Probably didn't catch my eye, but I've seen it first a few months ago. I'll test some previous versions and let you know.

liviuconcioiu:
> Did it appear before 5.2.2-dev ?

It happens even on QA_4_9.

kamioon:
[...]

PR #18536, opened 2023-07-16
fix: simulate query matched rows wrong count

kamioon:
This is a bug fix for #18533 
The `<>` operator excludes `NULL` values. [read more](https://stackoverflow.com/a/38375786)

Fixes: #18533

### DIFF
--- a/libraries/classes/Import/SimulateDml.php
+++ b/libraries/classes/Import/SimulateDml.php
@@ -135,7 +135,7 @@ final class SimulateDml
         }
 
         if (! empty($diff)) {
-            $where .= ' AND (' . implode(' OR ', $diff) . ')';
+            $where = '(' . $where . ') AND (' . implode(' OR ', $diff) . ')';
         }
 
         $orderAndLimit = '';
--- a/libraries/classes/Import/SimulateDml.php
+++ b/libraries/classes/Import/SimulateDml.php
@@ -132,12 +132,11 @@ final class SimulateDml
         $diff = [];
         foreach ($statement->set as $set) {
             $columns[] = $set->column;
-            $notEqualOperator = ' <> ';
             if (strtoupper($set->value) === 'NULL') {
-                $notEqualOperator = ' IS NOT ';
+                $diff[] = $set->column . ' IS NOT ' . $set->value;
+            }else{
+                $diff[] = '(' . $set->column . ' <> ' . $set->value . ' OR ' . $set->column . ' IS NULL)';
             }
-
-            $diff[] = $set->column . $notEqualOperator . $set->value;
         }
 
         if (! empty($diff)) {
--- a/libraries/classes/Import/SimulateDml.php
+++ b/libraries/classes/Import/SimulateDml.php
@@ -15,7 +15,6 @@ use PhpMyAdmin\SqlParser\Utils\Query;
 use PhpMyAdmin\Url;
 
 use function implode;
-use function strtoupper;
 
 final class SimulateDml
 {
@@ -132,11 +131,7 @@ final class SimulateDml
         $diff = [];
         foreach ($statement->set as $set) {
             $columns[] = $set->column;
-            if (strtoupper($set->value) === 'NULL') {
-                $diff[] = $set->column . ' IS NOT ' . $set->value;
-            } else {
-                $diff[] = $set->column . ' <> ' . $set->value . ' OR ' . $set->column . ' IS NULL';
-            }
+            $diff[] = 'NOT ' . $set->column . ' <=> (' . $set->value . ')';
         }
 
         if (! empty($diff)) {
--- a/libraries/classes/Import/SimulateDml.php
+++ b/libraries/classes/Import/SimulateDml.php
@@ -134,8 +134,8 @@ final class SimulateDml
             $columns[] = $set->column;
             if (strtoupper($set->value) === 'NULL') {
                 $diff[] = $set->column . ' IS NOT ' . $set->value;
-            }else{
-                $diff[] = '(' . $set->column . ' <> ' . $set->value . ' OR ' . $set->column . ' IS NULL)';
+            } else {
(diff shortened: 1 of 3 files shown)


libraries/classes/Import/SimulateDml.php
```diff
@@ -132,12 +132,11 @@ private function getSimulatedUpdateQuery(Parser $parser, UpdateStatement $statem
         $diff = [];
         foreach ($statement->set as $set) {
             $columns[] = $set->column;
-            $notEqualOperator = ' <> ';
             if (strtoupper($set->value) === 'NULL') {
-                $notEqualOperator = ' IS NOT ';
+                $diff[] = $set->column . ' IS NOT ' . $set->value;
+            }else{
+                $diff[] = '(' . $set->column . ' <> ' . $set->value . ' OR ' . $set->column . ' IS NULL)';
```
liviuconcioiu, line 138: I would get rid of both parentheses here, since aren't need. Examples of generated queries with and without parentheses.

```SQL
SELECT `b` FROM `test` WHERE `a` = 2 AND ((`b` <> 'test' OR `b` IS NULL))
SELECT `b` FROM `test` WHERE `a` = 2 AND (`b` <> 'test' OR `b` IS NULL)

SELECT `b`, `c` FROM `test` WHERE `a` = 1 AND ((`b` <> 'test' OR `b` IS NULL) OR (`c` <> 'test' OR `c` IS NULL))
SELECT `b`, `c` FROM `test` WHERE `a` = 1 AND (`b` <> 'test' OR `b` IS NULL OR `c` <> 'test' OR `c` IS NULL)

SELECT `b`, `c` FROM `test` WHERE `a` = 2 AND (`b` IS NOT NULL OR (`c` <> 'test' OR `c` IS NULL))
SELECT `b`, `c` FROM `test` WHERE `a` = 2 AND (`b` IS NOT NULL OR `c` <> 'test' OR `c` IS NULL)
```

libraries/classes/Import/SimulateDml.php
```diff
@@ -132,12 +132,11 @@ private function getSimulatedUpdateQuery(Parser $parser, UpdateStatement $statem
         $diff = [];
         foreach ($statement->set as $set) {
             $columns[] = $set->column;
-            $notEqualOperator = ' <> ';
             if (strtoupper($set->value) === 'NULL') {
-                $notEqualOperator = ' IS NOT ';
+                $diff[] = $set->column . ' IS NOT ' . $set->value;
+            }else{
```
liviuconcioiu, line 137: `} else {`

libraries/classes/Import/SimulateDml.php
kamioon: It makes sense, considering that they are all `OR`, it is not really necessary. I've removed them.

libraries/classes/Import/SimulateDml.php
```diff
@@ -132,12 +132,11 @@ private function getSimulatedUpdateQuery(Parser $parser, UpdateStatement $statem
         $diff = [];
         foreach ($statement->set as $set) {
             $columns[] = $set->column;
-            $notEqualOperator = ' <> ';
             if (strtoupper($set->value) === 'NULL') {
-                $notEqualOperator = ' IS NOT ';
+                $diff[] = $set->column . ' IS NOT ' . $set->value;
+            } else {
+                $diff[] = $set->column . ' <> ' . $set->value . ' OR ' . $set->column . ' IS NULL';
             }
```
MoonE, line 139: Some love  for the [MySQL `<=>` operator](https://dev.mysql.com/doc/refman/8.0/en/comparison-operators.html#operator_equal-to)?
```suggestion
            $diff[] = 'NOT ' . $set->column . ' <=> ' . $set->value;
```

williamdes: It broke my brain 😂
Thanks for sharing it

kamioon: Thank you @MoonE ! I kept the additional parentheses that you have added it seems to keep the `value` safe. but are there any special conditions without them that ruin the query?

williamdes (approved):
This looks good

libraries/classes/Import/SimulateDml.php
MoonE: This can happen with an update query like 
```sql
UPDATE a SET b = c OR d WHERE e OR f
```
It would generate `NOT b <=> c OR d` instead of  `NOT b <=> (c OR d)`

With that in mind the original condition should also be  set in parenthesis:
```patch
        if (! empty($diff)) {
-            $where .= ' AND (' . implode(' OR ', $diff) . ')';
+            $where = '(' . $where . ') AND (' . implode(' OR ', $diff) . ')';
        }
``` 
to generate the condition
`WHERE (e OR f) AND (NOT b <=> (c OR d))` instead of 
`WHERE e OR f AND (NOT b <=> (c OR d))`

MoonE:
@kamioon You also need  to adjust the test here
https://github.com/phpmyadmin/phpmyadmin/blob/dd5d3406f1df56d23fc6e2391d857ee0bc8577fc/test/classes/Import/SimulateDmlTest.php#L51-L54 
and also modify the same query here https://github.com/phpmyadmin/phpmyadmin/blob/dd5d3406f1df56d23fc6e2391d857ee0bc8577fc/test/classes/Stubs/DbiDummy.php#L2236-L2240

@williamdes @MauricioFauth Unit tests still pass because the test is marked incomplete here
https://github.com/phpmyadmin/phpmyadmin/blob/dd5d3406f1df56d23fc6e2391d857ee0bc8577fc/test/classes/Stubs/DbiDummy.php#L135
and PHPUnit is not run with  `--fail-on-incomplete` / `failOnIncomplete="true"`

MoonE (approved):
For what it's woth, I think this is ready.

MauricioFauth (approved):

PR #18550, opened 2023-07-19
add create user query with exporting data

### LINKED ISSUE #17996: Exporting user accounts does not contain `CREATE USER`
### Describe the bug

The function to export a specific mysql user account does only contain the `GRANT` queries but a `CREATE USER` query in the begining is missing. This is a problem when importing the users to a MySQL 8 server whic requieres the `CREATE USER` query.

### To Reproduce

1. Go to 'User accounts' in the top menu
2. Click on 'Export' on a specific user
3. See only `GRANT` queries

Gemui:
### Description

Fixes #17996 

### DIFF
--- a/libraries/classes/Server/Privileges.php
+++ b/libraries/classes/Server/Privileges.php
@@ -735,6 +735,39 @@ class Privileges
     }
 
     /**
+     * Returns create query for user on a certain host
+     * Used in the export privileges for all users section
+     *
+     * @param string $user User name
+     * @param string $host Host name
+     *
+     * @return string containing all the grants text
+     */
+    public function getUser(string $user, string $host): string
+    {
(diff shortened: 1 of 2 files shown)


williamdes:
Can this one be rebased, and some screenshots posted ?
I think that could be a nice addition

williamdes:
@kamil-tekiela what do you think about this one?
can we avoid the select in the MySQL DB?

kamil-tekiela:
I'm sorry but I don't think we should be doing this. There should be no `CREATE USER` in the export. This PR demonstrates quite well why that is not even possible. To create a user properly you need to specify a password. We can't know the user's password because it's not saved anywhere in the database. The PR uses an authentication_string instead of a password which is wrong because it would only create unusable user accounts. 

You can't export user accounts into SQL in any other way than using `SELECT` and `INSERT` (e.g. exporting the `mysql.users` table) which in itself is already a bad idea. The best is to create new users in the new DB and only export privileges.

PR #18581, opened 2023-07-26
Add a isAjax() method to ServerRequest class

MauricioFauth:
This method belongs to the ServerRequest class instead of the ResponseRenderer class, as it's a request property.

### DIFF
--- a/libraries/classes/Http/ServerRequest.php
+++ b/libraries/classes/Http/ServerRequest.php
@@ -321,4 +321,10 @@ class ServerRequest implements ServerRequestInterface
         return is_array($postParams) && isset($postParams[$param])
             || is_object($postParams) && property_exists($postParams, $param);
     }
+
+    public function isAjax(): bool
+    {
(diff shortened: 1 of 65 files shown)


libraries/classes/Http/ServerRequest.php
```diff
@@ -321,4 +321,10 @@ public function hasBodyParam(string $param): bool
         return is_array($postParams) && isset($postParams[$param])
             || is_object($postParams) && property_exists($postParams, $param);
     }
+
+    public function isAjax(): bool
+    {
+        return $this->serverRequest->getHeaderLine('X-Requested-With') === 'XMLHttpRequest'
+            || $this->has('ajax_request');
```
kamil-tekiela, line 328: Why is this necessary?

MauricioFauth: It's currently used as a way to tell if the response should be HTML or JSON.

PR #18583, opened 2023-07-26
Fix simulate query count

### LINKED ISSUE #18533: Simulate query matched rows wrong count

MoonE:
Fixes #18533, closes #18536

### DIFF
--- a/libraries/classes/Import/SimulateDml.php
+++ b/libraries/classes/Import/SimulateDml.php
@@ -72,21 +72,13 @@ final class SimulateDml
      * Executes the matched_row_query and returns the resultant row count.
      *
      * @param string $matchedRowQuery SQL query
-     *
-     * @return int|string
-     * @psalm-return int|numeric-string
      */
-    private function executeMatchedRowQuery(string $matchedRowQuery)
+    private function executeMatchedRowQuery(string $matchedRowQuery): int
     {
         $this->dbi->selectDb($GLOBALS['db']);
-        // Execute the query.
-        $result = $this->dbi->tryQuery($matchedRowQuery);
-        if (! $result) {
-            return 0;
-        }
 
-        // Count the number of rows in the result set.
-        return $result->numRows();
+        // Execute the query.
+        return (int) $this->dbi->fetchValue($matchedRowQuery);
     }
 
     /**
@@ -107,7 +99,9 @@ final class SimulateDml
             : ' ORDER BY ' . Query::getClause($statement, $parser->list, 'ORDER BY');
         $limit = $statement->limit === null ? '' : ' LIMIT ' . Query::getClause($statement, $parser->list, 'LIMIT');
 
-        return 'SELECT 1 FROM ' . $tableReferences[0] . $where . $order . $limit;
+        return 'SELECT COUNT(*) FROM (' .
+            'SELECT 1 FROM ' . $tableReferences[0] . $where . $order . $limit .
+            ') AS `pma_tmp`';
     }
 
     /**
@@ -136,6 +130,8 @@ final class SimulateDml
             : ' ORDER BY ' . Query::getClause($statement, $parser->list, 'ORDER BY');
         $limit = $statement->limit === null ? '' : ' LIMIT ' . Query::getClause($statement, $parser->list, 'LIMIT');
 
-        return 'SELECT 1 FROM ' . $tableReferences[0] . $where . $order . $limit;
+        return 'SELECT COUNT(*) FROM (' .
+            'SELECT 1 FROM ' . $tableReferences[0] . $where . $order . $limit .
+            ') AS `pma_tmp`';
     }
 }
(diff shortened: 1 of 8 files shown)


MauricioFauth:
@MoonE Could you please rebase the PR it?

MauricioFauth (approved):

libraries/classes/Import/SimulateDml.php
```diff
@@ -122,33 +115,38 @@ private function getSimulatedDeleteQuery(Parser $parser, DeleteStatement $statem
...
+            ? ''
+            : ' ORDER BY ' . Query::getClause($statement, $parser->list, 'ORDER BY');
+        $limit = $statement->limit === null ? '' : ' LIMIT ' . Query::getClause($statement, $parser->list, 'LIMIT');
+
+        return 'SELECT COUNT(*)' .
+            ' FROM (SELECT ' . implode(',', $newValues) . ') AS `pma_new`' .
+            ' JOIN (' .
+            'SELECT ' . implode(', ', $oldValues) . ' FROM ' . $tableReferences[0] . $where . $order . $limit .
+            ') AS `pma_old`' .
+            ' WHERE NOT (' . implode(', ', $newColumns) . ') <=> (' . implode(', ', $oldColumns) . ')';
```
kamil-tekiela, line 150: I'd love if there was some comment explaining how this works because I can't comprehend it myself. :)

MoonE: `pma_new` selects the `set` values into a single row which is then joined with the current table as `pma_old`. No `ON` condition needed, as `pma_new` is always just a single row.
In the `WHERE` confition a tuple comparison is made that checks if the old data tuple is different to the new data tuple.
The outer select counts all rows that are different.

```sql
UPDATE `table_1` SET `id` = 1, `name` = 'asdf' WHERE `id` = 10
```

```sql
SELECT COUNT(*)
FROM (SELECT 1 AS `n0`, 'asdf' AS `n1`) AS `pma_new`
JOIN (SELECT `id` AS `o0`, `name` AS `o1` FROM `table_1` WHERE `id` = 1) AS `pma_old`
WHERE NOT (`n0`, `n1`) <=> (`o0`, `o1`)
```

Thinking about it ... this should be equivalent to this, placing the values directly in the `WHERE` condition:
```sql
SELECT COUNT(*)
FROM (SELECT `id` AS `o0`, `name` AS `o1` FROM `table_1` WHERE `id` = 1) AS `pma_old`
WHERE NOT (1, 'asdf') <=> (`o0`, `o1`)
```

liviuconcioiu: @MoonE I'm not sure about `COUNT(*)`. Previously, when the number was clicked it would have shown the rows that will be updated, but now is only the count number.

MoonE: The count is the number of updated/affected rows. The `pma_old` subquery will give the matched rows, and the where condition reduces this to  the affected rows

liviuconcioiu: > The count is the number of updated/affected rows. The `pma_old` subquery will give the matched rows, and the where condition reduces this to the affected rows

Yeah, but when the number is clicked, previously the rows where shown, something like:

count

MoonE: I did not know about that ...

Though I'm not sure how useful this feature is. Just listing the updated values without a full row context?

liviuconcioiu: > I did not know about that ...
> 
> Though I'm not sure how useful this feature is. Just listing the updated values without a full row context?

Well, the feature was there and probably someone might use this.

Issue #18542, opened 2023-07-18
Search doesn't work with `IN (...)`, `NOT IN (...)` operators on `INT` column types with no foreign key


liviuconcioiu:
### Describe the bug

Search doesn't work with `IN (...)`, `NOT IN (...)` operators on `INT` column types with no foreign key.

### To Reproduce

1. Go to https://demo.phpmyadmin.net/master/public/index.php?route=/table/search&db=sakila&table=actor
2. For `actor_id` select `IN (...)` operator
3. Enter `1,2` as value
4. Click `Go`
5. Search isn't working

### Expected behavior

Search should work with both operators.

### Screenshots

https://github.com/phpmyadmin/phpmyadmin/assets/25424343/4074d0c9-20c9-48ce-816d-66441fa25b0f

### Server configuration

- phpMyAdmin version: 6.0.0-dev

williamdes:
I did some recent changes to this rules with adding a dedicated regex
I assume my fix has some kind of flaw

MoonE:
On 6.0.0 the field is a number field for integer columns, which only allows single numbers regardless of the validation function.

PR #18584, opened 2023-07-26
Search validate ints

MoonE:
Fixes partially #18542, need more changes in 6.0.0

I converted the tests from https://regexr.com/7h1ci and https://regexr.com/7h1eq to node/js tests. 

Not quite sure if I'm doing the exporting/importing of the functions right, I tried to match how it's done in functions.js. It works...


williamdes (approved):
That looks great, @MauricioFauth what do you think?

PR #18585, opened 2023-07-26
Redesign Triggers::getDetails()

kamil-tekiela:
Next step is to make it non-static.


templates/triggers/row.twig
```diff
@@ -3,27 +3,27 @@
     <input type="checkbox" class="checkall" name="item_name[]" value="{{ trigger.name }}">
   </td>
   <td>
-    <span class='drop_sql hide'>{{ trigger.drop }}</span>
+    <span class='drop_sql hide'>{{ trigger.getDropSql }}</span>
```
MauricioFauth, line 6: I think it's better to have the parenthesis to avoid the Twig trying to access a `getDropSql` property first.
```suggestion
    <span class='drop_sql hide'>{{ trigger.getDropSql() }}</span>
```

MauricioFauth (approved):

MauricioFauth (approved):

PR #18587, opened 2023-07-28
Remove the public/show_config_errors.php file

MauricioFauth:
A better and safer way of showing the configuration errors should be developed.


williamdes (approved):
🎉

PR #18589, opened 2023-07-29
Add support for PSR-15 middleware

### LINKED ISSUE #15507: HTTP handling should implement PSR-7

MauricioFauth:
Uses the `Laminas\HttpHandlerRunner\RequestHandlerRunner` class to handle the server request and the response emission.

Creates the `Http\Handler\QueueRequestHandler` class to handle the middleware queue.

- Fixes https://github.com/phpmyadmin/phpmyadmin/issues/15507

Follow up to:
- https://github.com/phpmyadmin/phpmyadmin/pull/18576
- https://github.com/phpmyadmin/phpmyadmin/pull/18609
- https://github.com/phpmyadmin/phpmyadmin/pull/18603


williamdes (approved):

libraries/classes/Http/Handler/ApplicationHandler.php
```diff
@@ -0,0 +1,34 @@
...
+    public function __construct(private readonly Application $application)
+    {
+    }
+
+    public function handle(ServerRequestInterface $request): ResponseInterface
+    {
+        try {
+            $response = $this->application->handle(new ServerRequest($request));
+            if ($response === null) {
+                throw new ExitException();
```
MauricioFauth, line 26: This is bad. Exceptions shouldn't be used to change the code path. We should remove that in the future.

libraries/classes/Http/Handler/ApplicationHandler.php
```diff
@@ -0,0 +1,34 @@
...
+final class ApplicationHandler implements RequestHandlerInterface
+{
+    public function __construct(private readonly Application $application)
+    {
+    }
+
+    public function handle(ServerRequestInterface $request): ResponseInterface
+    {
+        try {
+            $response = $this->application->handle(new ServerRequest($request));
```
MauricioFauth, line 24: I would like to use [`assert`](https://www.php.net/manual/en/function.assert.php) here instead of creating a new instance of `ServerRequest`, as this will always be a `ServerRequest` instance.
```diff
-            $response = $this->application->handle(new ServerRequest($request));
+            \assert($request instanceof ServerRequest);
+            $response = $this->application->handle($request);
```

MauricioFauth: @williamdes

libraries/classes/Http/Handler/ApplicationHandler.php
kamil-tekiela: Yes please. I don't know why but this PR breaks login page for me on my setup with xDebug.

MauricioFauth: Yeah, it's fully working yet. I'm adjusting it.

libraries/classes/Http/Handler/ApplicationHandler.php
williamdes: Yeah okay, that looks good
Also add a comment or a second argument if supported to say why it's impossible
just in case it does fire for some user

MauricioFauth: > just in case it does fire for some user 

That's not gonna happen for two reasons. First, ServerRequestInterface implementations are never used directly. It's always wrapped with ServerRequest. So the object coming from the parameter is a ServerRequest instance, always.
Second, the assert is not executed in production, and, for this reason it has zero cost.
As this is only there as a hint for static analysis, and we have certain that this is always true, it's better to use assert over the alternatives like creating a new instance or using Webmozart\Assert, that will have a cost only to satisfy SA tools.

williamdes (approved):

PR #18602, opened 2023-08-05
Refactor the DbTableExists class

MauricioFauth:
- Changes static methods to instance methods
- Extract response handling to the controllers
- Add unit tests for the DbTableExists class
- Remove the AbstractController::hasDatabase() method

### DIFF
--- a/libraries/classes/DbTableExists.php
+++ b/libraries/classes/DbTableExists.php
@@ -4,112 +4,88 @@ declare(strict_types=1);
 
 namespace PhpMyAdmin;
 
-use PhpMyAdmin\Controllers\Database\SqlController;
+use PhpMyAdmin\Identifiers\DatabaseName;
+use PhpMyAdmin\Identifiers\TableName;
 
-use function __;
-use function defined;
+use function in_array;
+use function sprintf;
+use function str_starts_with;
 
 final class DbTableExists
 {
+    /** @psalm-var list<non-empty-string> */
+    private array $databases = [];
+
+    /** @psalm-var list<non-empty-string> */
+    private array $tables = [];
+
+    public function __construct(private readonly DatabaseInterface $dbi)
+    {
+    }
+
+    public function hasDatabase(DatabaseName $databaseName): bool
+    {
+        if (in_array($databaseName->getName(), $this->databases, true)) {
+            return true;
+        }
+
+        if ($this->dbi->selectDb($databaseName) || $this->hasCommandsOutOfSyncError()) {
+            $this->databases[] = $databaseName->getName();
+
+            return true;
+        }
+
+        return false;
+    }
+
+    public function hasTable(DatabaseName $database, TableName $table): bool
+    {
+        if (in_array($database->getName() . '.' . $table->getName(), $this->tables, true)) {
+            return true;
+        }
+
+        if (
+            $this->hasCachedTableContent($database, $table)
+            || $this->isPermanentTable($table)
+            || $this->isTemporaryTable($table)
+        ) {
+            $this->tables[] = $database->getName() . '.' . $table->getName();
+
+            return true;
+        }
+
+        return false;
+    }
+
     /**
-     * Ensure the database and the table exist (else move to the "parent" script) and display headers.
+     * This "Commands out of sync" 2014 error may happen, for example after calling a MySQL procedure;
+     * at this point we can't select the db, but it's not necessarily wrong.
+     *
+     * @see https://dev.mysql.com/doc/mysql-errors/8.0/en/client-error-reference.html#error_cr_commands_out_of_sync
      */
-    public static function check(string $db, string $table, bool $isTransformationWrapper = false): void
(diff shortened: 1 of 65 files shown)


libraries/classes/DbTableExists.php
```diff
@@ -4,112 +4,90 @@
...
-            }
-
-            $result = $GLOBALS['dbi']->tryQuery('SHOW TABLES LIKE \'' . $GLOBALS['dbi']->escapeString($table) . '\';');
-            $GLOBALS['is_table'] = $result && $result->numRows();
-        }
-
-        if ($GLOBALS['is_table']) {
-            return;
-        }
+        return str_starts_with($this->dbi->getError(), '#2014 ');
```
kamil-tekiela, line 68: I am not a fan of this. This method belongs in Dbal if it belongs anywhere at all. We should not be lenient with errors, especially ones where we perform operations out of order.

libraries/classes/DbTableExists.php
```diff
@@ -4,112 +4,90 @@
...
-        /** @var SqlController $controller */
-        $controller = Core::getContainerBuilder()->get(SqlController::class);
-        $controller(Application::getRequest());
+    /**
+     * SHOW TABLES doesn't show temporary tables, so try select
+     * as it can happen just in case temporary table. It should be fast.
+     */
+    private function isTemporaryTable(TableName $table): bool
+    {
+        $result = $this->dbi->tryQuery(sprintf('SELECT COUNT(*) FROM %s;', Util::backquote($table)));
```
kamil-tekiela, line 89:
```suggestion
        $result = $this->dbi->tryQuery(sprintf('SELECT 1 FROM %s LIMIT 1;', Util::backquote($table)));
```
I believe this should be faster.

libraries/classes/DbTableExists.php
```diff
@@ -4,112 +4,90 @@
...
 
-        ResponseRenderer::getInstance()->callExit();
+        return $result !== false && $result->numRows() > 0;
```
kamil-tekiela, line 91:
```suggestion
        return $result !== false;
```

libraries/classes/DbTableExists.php
```diff
@@ -4,112 +4,90 @@
...
         }
 
-        $response = ResponseRenderer::getInstance();
-        if ($response->isAjax()) {
-            $response->setRequestStatus(false);
-            $response->addJSON(
-                'message',
-                Message::error(__('No databases selected.')),
-            );
+        if ($this->dbi->selectDb($databaseName) || $this->hasCommandsOutOfSyncError()) {
```
kamil-tekiela, line 32: This is an unexpected side effect from a method that looks like it just checks something and returns a result. But I don't know yet how to improve it, so let's leave it as it is and maybe something will come to me later.

kamil-tekiela (changes requested):
I like it! I wonder if we could have some helper method for:
```
$this->response->setRequestStatus(false);
$this->response->addJSON('message', Message::error(__('No databases selected.')));
```

libraries/classes/DbTableExists.php
```diff
@@ -4,112 +4,90 @@
...
-        Core::sendHeaderLocation('./index.php?route=/' . Url::getCommonRaw($urlParams, '&'));
-
-        $response->callExit();
+        return false;
     }
 
-    private static function checkTable(string $db, string $table, bool $isTransformationWrapper): void
+    /**
+     * This "Commands out of sync" 2014 error may happen, for example after calling a MySQL procedure;
+     * at this point we can't select the db, but it's not necessarily wrong.
```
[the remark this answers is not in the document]
MauricioFauth: This was introduced by b55ba34d43138b11b71d8eeae6b09ea160ded213. I don't want to remove this in this PR, but it's worth verifying if this could be removed later on.

libraries/classes/DbTableExists.php
MauricioFauth: If this is faster, then it can also improve the MySQL `sys.table_exists` procedure.

https://github.com/mysql/mysql-sys/blob/a1374ea7cf341b71fd1060136533ce01fe0bd57b/procedures/table_exists.sql#L141

kamil-tekiela: No, it won't. They don't execute it. They only prepare.

libraries/classes/DbTableExists.php
```diff
@@ -4,112 +4,90 @@
...
 
-        $urlParams = ['reload' => 1];
+        return false;
+    }
 
-        if (isset($GLOBALS['message'])) {
-            $urlParams['message'] = $GLOBALS['message'];
+    public function hasTable(DatabaseName $database, TableName $table): bool
+    {
+        if (in_array($table->getName(), $this->tables, true)) {
```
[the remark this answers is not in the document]
MauricioFauth: > I wonder what exactly is the need for memorization here?

This is similar as the global variable that was used before. There are some places that it can be checked more than once. This can be removed after fixed those cases.

PR #18603, opened 2023-08-06
Move Core::sendHeaderLocation() to ResponseRenderer::redirect()

MauricioFauth:
Removes the IIS specific logic.


libraries/classes/ResponseRenderer.php
```diff
@@ -506,4 +509,22 @@ public function callExit(string $message = ''): never
 
         exit;
     }
+
+    /**
+     * @psalm-param non-empty-string $url
+     * @psalm-param StatusCodeInterface::STATUS_*|null $statusCode
+     */
+    public function redirect(string $url, int|null $statusCode = null): void
```
kamil-tekiela, line 517: Instead of `$statusCode ?? StatusCodeInterface::STATUS_FOUND` 
```suggestion
    public function redirect(string $url, int|null $statusCode = StatusCodeInterface::STATUS_FOUND): void
```

kamil-tekiela (approved):

PR #18606, opened 2023-08-06
Fix: Cannot execute stored procedure with SELECT using the SQL tab

### LINKED ISSUE #18605: Cannot execute stored procedure with SELECT using the SQL tab
### Describe the bug

A stored procedure is a nasty thing in MySQL. Executing a stored procedure will always produce a result (although empty, it has metadata). If the stored procedure contains any result-producing queries, they will also produce results. The outcome is that calling a stored procedure can create one or more results. They need to be fetched whether needed or not, otherwise the connection is blocked. 

### To Reproduce

1. Go to any database and open SQL tab
2. Execute query `SELECT 'foo'; SELECT 'bar';` and see that it correctly produces two results.
3. Create a stored procedure with the same two queries.
4. Instead of executing the queries, execute the newly created SP in the SQL tab. 

### Expected behavior

Same output in both situations.

### Screenshots

image

> SET FOREIGN_KEY_CHECKS = ON;
> 
> MySQL said: Documentation
> 
> #2014 - Commands out of sync; you can't run this command now

### Server configuration

- phpMyAdmin version: 5.2

kamil-tekiela:
Fixes #18605 and #14614 but it's not an ideal fix. This is a workaround. 

What the logic here did was it checked if the executed query was a SELECT statement. If it was, it added additional stuff on UI, such as "create view" and bookmarks. My workaround enables the same for any query that produces a result set. It could be `SHOW TABLES`, `CALL` or `SELECT`. This obviously is wrong.

### DIFF
--- a/libraries/classes/Display/Results.php
+++ b/libraries/classes/Display/Results.php
@@ -4191,7 +4191,7 @@ class Results
         // display the Export link).
         if (
             ($analyzedSqlResults['querytype'] === self::QUERY_TYPE_SELECT)
-            && empty($analyzedSqlResults['procedure'])
+            && empty($analyzedSqlResults['is_procedure'])
         ) {
             if (count($analyzedSqlResults['select_tables']) === 1) {
                 $urlParams['single_table'] = 'true';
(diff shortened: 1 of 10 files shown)


williamdes (approved):
Wow, can we adjust something for phpstan to notice that we where using a key that did not exist ?

kamil-tekiela:
> Wow, can we adjust something for phpstan to notice that we where using a key that did not exist ?

We can forbid the use of `empty` which we already did.

kamil-tekiela:
I think this workaround is good for 5.2, but for master I will try to come up with something better. 
I don't know why Selenium tests are failing, so I will have to look into this.

MauricioFauth (approved):

PR #18629, opened 2023-08-15
Strict boolean conditions

kamil-tekiela:
Risky! This is done using Rector's set `STRICT_BOOLEANS`. I picked out the safer changes only, and even then this is risky because we relied on loose type checking before. 

This change treats all uses of strings in boolean conditions as comparing only empty string. Thus every loose comparison that intentionally compared for 0-string is considered a bug. 

Hopefully this fixes more bugs than it creates.

### DIFF
--- a/libraries/classes/Controllers/Table/ZoomSearchController.php
+++ b/libraries/classes/Controllers/Table/ZoomSearchController.php
@@ -373,7 +373,7 @@ class ZoomSearchController extends AbstractController
                 'where_clause' => $uniqueCondition,
                 'where_clause_sign' => Core::signSqlQuery($uniqueCondition),
             ];
-            $tmpData[$dataLabel] = $dataLabel ? $row[$dataLabel] : '';
+            $tmpData[$dataLabel] = $dataLabel !== '' ? $row[$dataLabel] : '';
             $data[] = $tmpData;
         }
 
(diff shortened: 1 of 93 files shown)


libraries/classes/Controllers/Table/ZoomSearchController.php
```diff
@@ -392,8 +392,8 @@ public function zoomSubmitAction(string $dataLabel, string $goto): void
                 $columnName,
             );
             if (
-                ! $this->foreigners
-                || ! $searchColumnInForeigners[$columnIndex]
+                $this->foreigners === []
+                || ($searchColumnInForeigners[$columnIndex] === [] || $searchColumnInForeigners[$columnIndex] === false)
```
MauricioFauth, line 396: These parenthesis are useless.

libraries/classes/DatabaseInterface.php
```diff
@@ -505,7 +505,7 @@ static function (array $a, array $b): int {
                     $needAnd = true;
                 }
 
-                if ($tableType) {
+                if ($tableType !== null && $tableType !== '') {
```
MauricioFauth, line 508: There is a similar condition a few lines earlier.

libraries/classes/Table.php
```diff
@@ -532,7 +532,7 @@ public static function generateFieldSpec(
             $query .= ' AS (' . $expression . ') ' . $virtuality;
         }
 
-        if (! $virtuality || $isVirtualColMysql) {
+        if ($virtuality === '' || $isVirtualColMysql) {
```
MauricioFauth, line 535: There is a similar condition a few lines earlier.

MauricioFauth (changes requested):

MauricioFauth (approved):

PR #18633, opened 2023-08-17
Introduce alias type mapping

### LINKED ISSUE #18613: Editing procedure with BOOL param changes type to INT
### Describe the bug

BOOL type is converted to INT when editing a procedure.
It works with BOOLEAN type.

### To Reproduce

1. Create procedure 
```sql
CREATE PROCEDURE `p` (IN b BOOL) BEGIN END;
``` 
2. Edit procedure
3. See type of `b` changed to INT

### Expected behavior

The BOOL data type should be preserved

### Screenshots

image

### Server configuration

phpMyAdmin version: 5.2 & 6.0

kamil-tekiela:
Fixes #18613

This will be backported to QA, but first we need to fix SQL parser.

### DIFF
--- a/src/Types.php
+++ b/src/Types.php
@@ -729,7 +729,7 @@ class Types
 
     public function mapAliasToMysqlType(string $alias): string
     {
-        return [
+        return match ($alias) {
             'BOOL' => 'TINYINT',
             'BOOLEAN' => 'TINYINT',
             'CHARACTER VARYING' => 'VARCHAR',
@@ -746,7 +746,8 @@ class Types
             'LONG' => 'MEDIUMTEXT',
             'MIDDLEINT' => 'MEDIUMINT',
             'NUMERIC' => 'DECIMAL',
-        ][$alias] ?? $alias;
+            default => $alias,
+        };
     }
 
     /**


M393:
Thanks for fixing this, @kamil-tekiela

I would prefer if the BOOLEAN type were kept, as it reveals something about the expected use case.
The other aliases make sense. :+1:

kamil-tekiela:
> Thanks for fixing this, @kamil-tekiela
> 
> I would prefer if the BOOLEAN type were kept, as it reveals something about the expected use case. The other aliases make sense. 👍

Right, but it's only visible at the creation time. As soon as you hit save in table creation, it will be converted to TINYINT. So it won't keep this information any longer than it is in your head.

M393:
Sure, but it helps knowing the alias when executing or editing a procedure.

williamdes:
Can this one be rebased ?
Does it need a sql-parser release ?

kamil-tekiela:
This needs to be fixed in SQL parser first. Currently using aliases crashes PMA.

williamdes:
> This needs to be fixed in SQL parser first. Currently using aliases crashes PMA.

Okay, can you manage this on the parser codebase ?
If needed you can backport fixes to QA sql-parser

kamil-tekiela:
Rebased. The `/database/routines` page no longer crashes and the routine using aliases is now editable. I don't know how to write a test for this though. The method `getParameterRow` isn't covered by unit tests as of now.

kamil-tekiela:
Also, I cannot test it on 5.2 yet because the page is still crashing due to non-updated SqlParser.

williamdes:
on 6.0 you mean?

kamil-tekiela:
It works fine on both now. Thanks.

MauricioFauth:
Could you rebase this PR so we can include the `match` in `Types::mapAliasToMysqlType()`.

MauricioFauth (approved):

PR #18638, opened 2023-08-20
Replace `$GLOBALS['cfg']['Server']` with `Config::$selectedServer`

MauricioFauth:
The `$GLOBALS['cfg']['Server']` array is not really a part of `$GLOBALS['cfg']` array, so it makes sense to be a Config's property.

`$GLOBALS['cfg']['Server']` represents the current selected server configuration. It's not the same as `$GLOBALS['cfg']['Servers'][$GLOBALS['server']]` because it includes changes from the authentication plugin. For example, for the 'cookie' auth type, the user and password from the login form are stored in `$GLOBALS['cfg']['Server']` and are not present in `$GLOBALS['cfg']['Servers'][$GLOBALS['server']]`.

Next step is to replace `Config::$selectedServer` array with an `Config\Settings\Server` object.


kamil-tekiela (approved):

williamdes (approved):
🎉 👏🏻

PR #18650, opened 2023-08-24
Fix result cell title attribute escape

MoonE:
An escaped string was displayed when browsing a table with foreign key relation where the displayed title for a foreign key link contained special characters.

### DIFF
--- a/libraries/classes/Display/Results.php
+++ b/libraries/classes/Display/Results.php
@@ -4484,9 +4484,9 @@ class Results
                 if ($relationalDisplay === self::RELATIONAL_KEY) {
                     // user chose "relational key" in the display options, so
                     // the title contains the display field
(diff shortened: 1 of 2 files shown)


kamil-tekiela (approved):

kamil-tekiela:
However, the single quote was not encoded into HTML entity until PHP 8.1. So either adjust the test or make the `htmlspecialchars` flags explicit in Generator.

MoonE:
> However, the single quote was not encoded into HTML entity until PHP 8.1. So either adjust the test or make the `htmlspecialchars` flags explicit in Generator.

Thanks, that helped me, @kamil-tekiela. I changed it to a double quote character. Any character that needs to be escaped will do here.

MauricioFauth (approved):

Issue #18512, opened 2023-06-24
New directory structure for phpMyAdmin 6.0


MauricioFauth:
As phpMyAdmin 6.0.0 has a directory structure change already with the introduction of the `public` directory, what about rethinking the entire structure for 6.0?

For example, when PHP classes were moved to a dedicated directory, instead of moving them to a root-level `src` directory, as it's common for PHP applications, it was chosen the keep them inside the libraries directory, as it was an already known directory to keep protected from direct access.

With the introduction of the `public` directory, `libraries` should not be a public directory anymore. So we can move the `libraries/classes` directory to a root-level `src` directory.

I found this standard ([php-pds/skeleton](https://github.com/php-pds/skeleton) and [php-pds/skeleton_research](https://github.com/php-pds/skeleton_research)) that we can use as a starting point:

> | If a package has a root-level directory for ... | ... then it MUST be named: |
> | ----------------------------------------------- | -------------------------- |
> | command-line executables                        | `bin/`                     |
> | configuration files                             | `config/`                  |
> | documentation files                             | `docs/`                    |
> | web server files                                | `public/`                  |
> | other resource files                            | `resources/`               |
> | PHP source code                                 | `src/`                     |
> | test code                                       | `tests/`                   |

So, based on this standard, the new directory structure for phpMyAdmin could be:

| Old path | New path |
| -------- | -------- |
| `build/` | `build/` |
| `doc/` | `docs/` |
| `examples/` | `examples/` |
| `js/` | `resources/js/` |
| `libraries/cache/` | `cache/` |
| `libraries/classes/` | `src/` |
| `locale/` | `resources/locale/` |
| `po/` | `resources/po/` |
| `public/` | `public/` |
| `scripts/` | `bin/` |
| `sql/` | `resources/sql/` |
| `templates/` | `resources/templates/` |
| `test/` | `tests/` |

What about the `libraries/*.php` files? Should we add a `config/` directory?

What do you think?

williamdes:
>What about the libraries/*.php files?

6.0 is probably the only right time to do such renames. 
So let's do it? 

I agree with the list of renames your proposed. 

cc @ibennetch 

> Should we add a config/ directory?

No, that's too much for users. Let's keep config.inc.php

MauricioFauth:
@williamdes should we rename the `libraries` directory to what name? `app`, `bootstrap`, `config` or another name?

My preference is `config` > `app` > `bootstrap`.

[...]

PR #18667, opened 2023-08-29
Rename the directory `libraries/classes/` to `src/`

MauricioFauth:
- Related to https://github.com/phpmyadmin/phpmyadmin/issues/18512


scripts/check-release-excludes.sh
```diff
@@ -110,6 +110,11 @@ validateExtension() {
             fi
         ;;
         libraries/*)
```
williamdes, line 112: it should be dropped as this script version is only for master

williamdes (approved):

scripts/check-release-excludes.sh
MauricioFauth: I left it because there are still some files in it.

williamdes: Oh okay

kamil-tekiela:
Is there any easy way to rebase my work-in-progress branches to avoid many conflicts?

MauricioFauth:
> Is there any easy way to rebase my work-in-progress branches to avoid many conflicts?

@kamil-tekiela What I usually do is `git checkout my-branch` and `git rebase master`. For me Git has been smart enough to resolve the rename, the only exceptions are the Psalm and PHPStan baselines, that I think it's easier to just regenerate them.
What issues do you have?

kamil-tekiela:
Ok, that worked, but I used cherry-pick instead. Rebasing was too messy for some reason. I picked the commits into a new branch and git handled the dir rename.

MauricioFauth:
Yes. Cherry-picking works great when rebasing is too messy. You want to keep the same branch, you can also create a new branch, go back to the branch, hard reset it, and then cherry-pick the commits.

PR #18668, opened 2023-08-29
Add number formatting to Total on Browse

### LINKED ISSUE #18666: add `number_format` function to rows count
please apply `number_format` function to count of query result for more readability:

kamil-tekiela:
Closes #18666

Should it be done in QA too?

### DIFF
--- a/libraries/classes/Display/Results.php
+++ b/libraries/classes/Display/Results.php
@@ -3709,13 +3709,13 @@ class Results
         if ($messageViewWarning === false) {
             if ($unlimNumRows != $total) {
                 $messageTotal = Message::notice(
-                    $preCount . __('%1$d total, %2$d in query'),
+                    $preCount . __('%1$s total, %2$s in query'),
                 );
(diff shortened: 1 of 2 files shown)


williamdes (approved):
Thanks!
No need for QA as it's a good change but clearly will maybe confuse users as they do not expect such changes

MauricioFauth (approved):

MauricioFauth:
> Should it be done in QA too?

I think this could count as an accessibility issue, instead of an enhancement, so it can go to QA as well.

MauricioFauth:
Let's keep it for 6.0.

liviuconcioiu:
Dumb question: shouldn't the formatting be added to between rows too?

Untitled

kamil-tekiela:
It should and I thought I did.

williamdes:
> Dumb question: shouldn't the formatting be added to in between rows too?
> 
> Untitled

I would not format this one, as other values can be there, it could also be strings or hex

PR #18670, opened 2023-08-29
Fix simulate query #2

MoonE:
Allows showing the updated values by clicking on the affected row number https://github.com/phpmyadmin/phpmyadmin/pull/18583#discussion_r1306655772.

- Did not work when setting value to other column value, e.g. `SET a=a+1`
- Show updated data when clicking on the affected rows number
  Now shows all table columns plus the new value as a column
- The button now also appears when using an update statement that uses schema.table format,
and delete statements like `DELETE a FROM a ...`

### DIFF
--- a/libraries/classes/Import/SimulateDml.php
+++ b/libraries/classes/Import/SimulateDml.php
@@ -8,7 +8,6 @@ use PhpMyAdmin\Core;
 use PhpMyAdmin\DatabaseInterface;
 use PhpMyAdmin\Html;
 use PhpMyAdmin\SqlParser\Parser;
-use PhpMyAdmin\SqlParser\Statement;
 use PhpMyAdmin\SqlParser\Statements\DeleteStatement;
 use PhpMyAdmin\SqlParser\Statements\UpdateStatement;
 use PhpMyAdmin\SqlParser\Utils\Query;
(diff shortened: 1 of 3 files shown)


libraries/classes/Import/SimulateDml.php
```diff
@@ -118,21 +120,39 @@ private function getSimulatedUpdateQuery(Parser $parser, UpdateStatement $statem
         Assert::count($tableReferences, 1, 'No joins allowed in simulation query');
         Assert::isNonEmptyList($statement->set, 'SET statements missing');
         Assert::notNull($parser->list, 'Parser list not set');
-        $newValues = [];
-        $oldValues = [];
+
+        preg_match('/^(?:`([^`]+)`\.)?`([^`]+)`$/i', $tableReferences[0], $matches);
```
kamil-tekiela, line 124: It would be nice to have a comment for this regex.

libraries/classes/Import/SimulateDml.php
```diff
@@ -102,9 +106,7 @@ private function getSimulatedDeleteQuery(Parser $parser, DeleteStatement $statem
             : ' ORDER BY ' . Query::getClause($statement, $parser->list, 'ORDER BY');
         $limit = $statement->limit === null ? '' : ' LIMIT ' . Query::getClause($statement, $parser->list, 'LIMIT');
 
-        return 'SELECT COUNT(*) FROM (' .
-            'SELECT 1 FROM ' . $tableReferences[0] . $where . $order . $limit .
-            ') AS `pma_tmp`';
+        return 'SELECT * FROM ' . $tableReferences[0] . $where . $order . $limit;
```
williamdes, line 109: I think this is too much as a behaviour change for QA
is there no other way?

MoonE: This reverts the change from #18583. It' the 5.2.1 behavior for `DELETE` statements.

libraries/classes/Import/SimulateDml.php
```diff
@@ -118,21 +117,18 @@ private function getSimulatedUpdateQuery(Parser $parser, UpdateStatement $statem
...
+            if (array_key_exists($column, $values)) {
                 continue;
             }
 
-            $handledColumns[$set->column] = true;
-            $oldValues[] = $set->column . ' AS ' . ($oldColumns[] = Util::backquote('o' . $i));
-            $newValues[] = $set->value . ' AS ' . ($newColumns[] = Util::backquote('n' . $i));
-            ++$i;
+            $oldColumns[] = Util::backquote($column);
+            $values[$column] = $set->value . ' AS ' . ($newColumns[] = Util::backquote($column . ' `new`'));
```
MoonE, line 131: This is less complex compared to [that](https://github.com/phpmyadmin/phpmyadmin/compare/c972deb0da5b69df0680a24b4131d0933465bee5..5d9f5c7a10aceb25864b84eb0c414adb50493993#diff-226f8cc74a3e66440e9d8251ebfd4cf15958c469010811d43feda5a09d4a06c5R131), with the caveat that the old  and new value columns are not next to each other and *if* someone has a table with column pairs named `[x]` and `` [x] `new` `` it breaks.

MauricioFauth (approved):

PR #18678, opened 2023-08-31
Optimize autocomplete

kamil-tekiela:
Technically a microoptimization, but it might help on large databases with lots of columns. 

On my test database, the payload size dropped from 25.7kB to 18.1kB. I know it's not a very scientific measurement, but in theory, it should improve performance on slow connections.

### DIFF
--- a/src/Controllers/Database/SqlAutoCompleteController.php
+++ b/src/Controllers/Database/SqlAutoCompleteController.php
@@ -11,6 +11,9 @@ use PhpMyAdmin\Http\ServerRequest;
 use PhpMyAdmin\ResponseRenderer;
 use PhpMyAdmin\Template;
 
+use function array_map;
+use function array_values;
+
 /**
  * Table/Column autocomplete in SQL editors.
  */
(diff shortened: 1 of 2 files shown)


MauricioFauth (approved):

src/Controllers/Database/SqlAutoCompleteController.php
```diff
@@ -29,7 +32,17 @@ public function __invoke(ServerRequest $request): void
             if ($db) {
                 $tableNames = $this->dbi->getTables($db);
                 foreach ($tableNames as $tableName) {
-                    $sqlAutocomplete[$tableName] = $this->dbi->getColumns($db, $tableName);
+                    $sqlAutocomplete[$tableName] = array_map(
+                        static fn (array $field): array => [
+                            'field' => $field['Field'],
+                            'columnHint' => $field['Type'] . match ($field['Key']) {
+                                'PRI' => ' | Primary',
```
williamdes, line 39: Should this be available to be translated?
If no maybe they should be full uppercase

PR #18683, opened 2023-09-02
Refactor Dbi::getColumn and getColumns

kamil-tekiela:
This is a follow up to #18445 and #18434.

This took a lot of work, but it's still not in the perfect state I wish it to be. More work will be required in future PRs. 
This PR adds a new DTO `Column`, which matches `ColumnFull`. These DTOs are now prepared by Dbi and used to transfer the data to the models. This data is readonly. This helps with type inference and makes the code easier to understand.


williamdes (approved):

MauricioFauth (approved):

MoonE:
@kamil-tekiela I'm getting an error when browsing to `Structure` > `Relation view` with these changes

> Internal error in ./templates/table/relation/common_form.twig#88
Twig\Error\RuntimeError: Impossible to access a key "Field" on an object of class "PhpMyAdmin\Column" that does not implement ArrayAccess interface.

MariaDB 10.6.14
PHP 8.1.20

PR #18693, opened 2023-09-05
Fix enum/set value escaping

kamil-tekiela:
This fixes the escaping bug related to enum/set values containing quotes and backslashes. It also improves performance of this function through a completely new implementation


kamil-tekiela:
This function mentioned JS equivalent, but I cannot find it. I wonder why we have a JS version of it. I've put the comment back, but I don't know which JS function I need to adjust.

MauricioFauth (approved):

PR #18696, opened 2023-09-06
Fix max execution time exceeded error in tests

MauricioFauth:
Runs these tests in isolation because they are causing a time out error when backing up the global state.


williamdes (approved):

kamil-tekiela:
All tests pass! 🚀

PR #18701, opened 2023-09-09
Fix user accout navigation by initials

MoonE:
Fixes two issues with the user account initials links and some refactoring

- Empty user names were not displayed properly in the list of initials and did not filter correctly when clicked
- User names starting with %, _, or \ were not properly filtered when clicking the initial.

Before / After
image image

To test, just create a bunch of  users (at least 20 for the initials to be displayed)
```sql
CREATE USER '\\';
CREATE USER '';
CREATE USER '_';
CREATE USER '%';
```


MauricioFauth:
@MoonE Could you rebase it, please?

MauricioFauth (approved):

PR #18722, opened 2023-09-20
Improve git hash info error handling in home.js

MauricioFauth:
Instead of showing a generic error message alert when an error occurs, it displays an error message at the same place where the success output goes.

Screenshot showing the error message


williamdes (approved):
Good idea

PR #18726, opened 2023-09-20
Remove obsolete code for stand-alone gis editor

MoonE:
This puts the coordinate form creation entirely to JS.
As I understand it there used  to be a non-js version of the gis editor,  which opended in a new window.


templates/gis_data_editor_form.twig
```diff
@@ -108,7 +109,7 @@
                             {% trans 'Y' %}
                                 <input type="text" name="gis_data[{{ a }}][{{ type }}][{{ i }}][y]" value="{{ gis_data[a][type][i]['y'] ?? '' }}"></label>
                             {% endfor %}
-                            <input type="submit" name="gis_data[{{ a }}][{{ type }}][add_point]" class="btn btn-secondary add addPoint" value="{% trans 'Add a point' %}">
+                            <a href="#" name="gis_data[{{ a }}][{{ type }}][add_point]" class="btn btn-secondary addJs addPoint">{% trans 'Add a point' %}</a>
```
MauricioFauth, line 112: You should use a `<button>` tag instead of `<a href="#">`.

templates/gis_data_editor_form.twig
```diff
@@ -192,34 +194,34 @@
...
-        <input class="btn btn-primary" type="submit" name="gis_data[save]" value="{% trans 'Go' %}">
 
         <div id="gis_data_output">
             <h3>{% trans 'Output' %}</h3>
-            <p>
-                {% trans 'Choose "GeomFromText" from the "Function" column and paste the string below into the "Value" field.' %}
-            </p>
             <textarea id="gis_data_textarea" rows="5">{{ result }}</textarea>
+            <br>
+            <br>
```
MauricioFauth, line 222: The `<br>` tag should only be used for line breaks in texts. Here, you should use a `<div>` tag.

MauricioFauth (changes requested):

templates/gis_data_editor_form.twig
MoonE: Yeah the current html structure here isn't great. What I did is simply apply the changes done in JS to the twig template. So the structure is exactly the same as before.

MauricioFauth (approved):

PR #18731, opened 2023-09-26
Improve support for collations in MariaDB 10.10+

Simbiat:
### Description

Some more collations were added and structural changes were done in MariaDB 10.10 as per https://jira.mariadb.org/browse/MDEV-27009 which requires changes in how we get the *full* list of collations if we are using this engine.
Since I would not consider this a *bug* submitting to `master` branch.
Since the change  only affects the kind of query which is run, and not the structure of the collation list, no extra tests are added and not existing tests are modified.

- [+] Make sure you have read our [CONTRIBUTING.md](https://github.com/phpmyadmin/phpmyadmin/blob/master/CONTRIBUTING.md) document.
- [+] Make sure you are making a pull request against the correct branch. For example, for bug fixes in a released version use the corresponding QA branch and for new features use the _master_ branch. If you have a doubt, you can ask as a comment in the bug report or on the mailing list.
- [+] Every commit has proper `Signed-off-by` line as described in our [DCO](https://github.com/phpmyadmin/phpmyadmin/blob/master/DCO). This ensures that the work you're submitting is your own creation.
- [+] Every commit has a descriptive commit message.
- [+] Every commit is needed on its own, if you have just minor fixes to previous commits, you can squash them.
- [+] Any new functionality is covered by tests.

Signed-off-by: Dmitrii Kustov <simbiat@outlook.com>

### DIFF
--- a/src/Charsets.php
+++ b/src/Charsets.php
@@ -77,6 +77,38 @@ class Charsets
             return;
         }
 
+        $sql = 'SELECT `CHARACTER_SET_NAME` AS `Charset`,'
+            . ' `DEFAULT_COLLATE_NAME` AS `Default collation`,'
+            . ' `DESCRIPTION` AS `Description`,'
+            . ' `MAXLEN` AS `Maxlen`'
+            . ' FROM `information_schema`.`CHARACTER_SETS`';
+
+        if ($disableIs) {
+            $sql = 'SHOW CHARACTER SET';
+        }
+
+        $res = $dbi->query($sql);
+
+        self::$charsets = [];
+        foreach ($res as $row) {
+            self::$charsets[$row['Charset']] = Charset::fromServer($row);
+        }
+
+        ksort(self::$charsets, SORT_STRING);
+    }
+
+    /**
+     * Loads collation data from the server
+     *
+     * @param bool $disableIs Disable use of INFORMATION_SCHEMA
+     */
+    private static function loadCollations(DatabaseInterface $dbi, bool $disableIs): void
+    {
+        /* Data already loaded */
+        if (self::$collations !== []) {
+            return;
+        }
+
         /* Check if we have `FULL_COLLATION_NAME`, which mean that we have MariaDB 10.10 or newer.
         Refer https://jira.mariadb.org/browse/MDEV-27009 */
         $sql = 'SELECT `COLUMN_NAME` FROM `information_schema`.`COLUMNS`'
@@ -90,7 +122,7 @@ class Charsets
         
         $check = $dbi->query($sql);
         
-        if (!empty($check)) {
+        if ($check !== []) {
             /* Use query to accomodate new structure of MariaDB collations
             Note, that SHOW COLLATION command is not applicable at the time of writing*/
             $sql = 'SELECT `collapp`.`FULL_COLLATION_NAME` AS `Collation`,'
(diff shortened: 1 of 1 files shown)


Simbiat:
Hm... I do have sign-off in commit message (I used GitHub to commit). Did I place it incorrectly somehow?

kamil-tekiela:
You have Dmitry in one place and Dmitrii in another.

src/Charsets.php
```diff
@@ -77,16 +77,45 @@ private static function loadCharsets(DatabaseInterface $dbi, bool $disableIs): v
...
+        /* Check if we have `FULL_COLLATION_NAME`, which mean that we have MariaDB 10.10 or newer.
+        Refer https://jira.mariadb.org/browse/MDEV-27009 */
+        $sql = 'SELECT `COLUMN_NAME` FROM `information_schema`.`COLUMNS`'
+            . ' WHERE `TABLE_SCHEMA`=\'information_schema\' AND'
+            . ' `TABLE_NAME`=\'COLLATION_CHARACTER_SET_APPLICABILITY\' AND'
+            . ' `COLUMN_NAME`=\'FULL_COLLATION_NAME\'';
+        
         if ($disableIs) {
-            $sql = 'SHOW CHARACTER SET';
+            $sql = 'SHOW COLUMNS FROM `information_schema`.`COLLATION_CHARACTER_SET_APPLICABILITY` LIKE \'FULL_COLLATION_NAME\'';
```
kamil-tekiela, line 88: This is inside `disableIs` so it should not use `information_schema`. Is there no way to do it without it?

src/Charsets.php
```diff
@@ -77,16 +77,45 @@ private static function loadCharsets(DatabaseInterface $dbi, bool $disableIs): v
...
         }
-
+        
+        $check = $dbi->query($sql);
+        
+        if (!empty($check)) {
```
kamil-tekiela, line 93: Please do not use `empty`.

Simbiat:
> You have Dmitry in one place and Dmitrii in another.

Hm, I did update my profile name to `Dmitrii` after the commit. Should I re-commit?

src/Charsets.php
Simbiat: `count($check) > 0`?

kamil-tekiela: No. It's also an antipattern. What is this condition supposed to check? Please make it explicit. The expected type should be evident by looking at the condition. If you expect to receive a non-empty array, you can just check `!== []`

src/Charsets.php
kamil-tekiela: Then do we need this `if` statement at all?

Simbiat: My understanding was that `$disableIs` means that we are not doing direct `SELECT * from information_schema.some_table` queries. In `SHOW COLUMNS` `FROM` is basically a condition, not an indication of a table from which we are selecting. But if  `$disableIs` means we are not referencing `information_schema` under any circumstances, then I can remove this `if`

kamil-tekiela: See https://docs.phpmyadmin.net/en/latest/config.html#cfg_Servers_DisableIS

Simbiat: I read it. And checked multiple places in code, where this condition is used, and based on what I see the goal of this setting is not to use `SELECT * from information_schema.some_table`, which we are not doing here, since we use `SHOW COLUMNS`, which technically complies with the setting's description, specifically with "use `SHOW` commands instead". But if you think that this `if` should still be remove - I can do that

Simbiat:
> The signoff message has to match your name and email configured in git. And an empty line may be required to separeate it form the commit message

I was using GitHub, but my name was mismatching, and I did not have a new line before sing-off. Will see how it works when I submit corrections in next commit.

Simbiat:
Closing due to mess I made with DCO message. Will recreate another PR

PR #18740, opened 2023-09-30
Move ./libraries/cache directory to ./cache

MauricioFauth:
- Related to #18512


williamdes (approved):
Cool, it will be easier to administer
What folder should be used for pre warmed caches?

MauricioFauth:
> Cool, it will be easier to administer
> What folder should be used for pre warmed caches? 

This is the directory for that (routes, Twig, etc). For other temp cache, I've been using the build directory.

PR #18741, opened 2023-10-01
Refactor Message class

kamil-tekiela:


src/Message.php
```diff
@@ -588,11 +503,7 @@ public static function decodeBB(string $message): string
     public function getHash(): string
     {
         if ($this->hash === null) {
-            $this->hash = md5(
-                $this->getNumber() .
-                $this->string .
-                $this->message,
-            );
+            $this->hash = md5($this->type . $this->string . $this->message);
```
williamdes, line 506:
```suggestion
            $this->hash = hash('crc32b', $this->type . $this->string . $this->message);
```

williamdes (approved):

src/Message.php
kamil-tekiela: What does this do? This will change the output, correct?

MauricioFauth (approved):

src/Message.php
williamdes: This uses another hash that is better that md5 for such non important things
IIRC this is only a unique message key

PR #18743, opened 2023-10-02
Replaced superglobals with ServerRequest in ChangeRowsController and FindReplaceController

dardvas:
### Description

Replaced superglobals with ServerRequest in ChangeRowsController

### DIFF
--- a/src/Controllers/Table/ChangeRowsController.php
+++ b/src/Controllers/Table/ChangeRowsController.php
@@ -31,8 +31,10 @@ final class ChangeRowsController extends AbstractController
 
         $rowsToDelete = $request->getParsedBodyParam('rows_to_delete');
 
-        if ((! $request->hasBodyParam('rows_to_delete') || ! is_array($rowsToDelete))
-            && $request->hasBodyParam('goto')) {
+        if (
+            (! $request->hasBodyParam('rows_to_delete') || ! is_array($rowsToDelete))
+            && $request->hasBodyParam('goto')
+        ) {
             $this->response->setRequestStatus(false);
             $this->response->addJSON('message', __('No row selected.'));
 
--- a/src/Controllers/Table/ChangeRowsController.php
+++ b/src/Controllers/Table/ChangeRowsController.php
@@ -29,7 +29,10 @@ final class ChangeRowsController extends AbstractController
         $GLOBALS['active_page'] ??= null;
         $GLOBALS['where_clause'] ??= null;
 
-        if (isset($_POST['goto']) && (! isset($_POST['rows_to_delete']) || ! is_array($_POST['rows_to_delete']))) {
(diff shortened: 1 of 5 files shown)


src/Controllers/Table/ChangeRowsController.php
```diff
@@ -29,7 +29,10 @@ public function __invoke(ServerRequest $request): void
         $GLOBALS['active_page'] ??= null;
         $GLOBALS['where_clause'] ??= null;
 
-        if (isset($_POST['goto']) && (! isset($_POST['rows_to_delete']) || ! is_array($_POST['rows_to_delete']))) {
+        $rowsToDelete = $request->getParsedBodyParam('rows_to_delete');
+        $rowsToDeleteIsValid = ! isset($rowsToDelete) || ! is_array($rowsToDelete);
```
MauricioFauth, line 33: In this case, `isset` is redundant, because it's only checking if `$rowsToDelete` is not `null`, which `is_array` also checks, so the `isset` can be removed.

MauricioFauth (changes requested):

src/Controllers/Table/FindReplaceController.php
```diff
@@ -99,16 +99,30 @@ public function __invoke(ServerRequest $request): void
...
+            $this->response->addJSON('preview', $preview);
 
             return;
         }
 
         $this->addScriptFiles(['table/find_replace.js']);
 
-        if (isset($_POST['replace'])) {
-            $this->replaceAction();
+        if (!empty($request->getParsedBodyParam('replace'))) {
```
MauricioFauth, line 116: Please don't use empty. We prefer stricter comparisons. In this case you can use `ServerRequest::hasBodyParam()`.

```suggestion
        if ($request->hasBodyParam('replace')) {
```

MauricioFauth (changes requested):

src/Controllers/Table/FindReplaceController.php
dardvas: Sure, makes sense, thanks!

MauricioFauth (changes requested):
You should fix the tests.
- https://github.com/phpmyadmin/phpmyadmin/actions/runs/6410026164/job/17535125700?pr=18743#step:8:94
- https://github.com/phpmyadmin/phpmyadmin/actions/runs/6410026151/job/17535124606?pr=18743#step:8:33

You can run the tests with:
```
composer test
```
For PHPStan and Psalm, you can run:
```
composer update:baselines
```

MauricioFauth (approved):

MauricioFauth:
Merged. Thanks!

dardvas:
Thanks @MauricioFauth ! Sorry, got a bit busy with things :sweat_smile:

PR #18744, opened 2023-10-03
Remove Jenkins build.xml

MauricioFauth:
Jenkins builds are not working for a long time.

https://ci.phpmyadmin.net/


williamdes (approved):

MauricioFauth:
After thinking about it for a while, I think we could bring Jenkins back to live and used it more, specially for things that we are not so sure about in using GitHub Actions, like doing releases.
What do you think, @williamdes and @ibennetch?
Anyway, I lose my access to it after a upgrade.

williamdes:
> After thinking about it for a while, I think we could bring Jenkins back to live and used it more, specially for things that we are not so sure about in using GitHub Actions, like doing releases. What do you think, @williamdes and @ibennetch? Anyway, I lose my access to it after a upgrade.

I am against bringing back this old stuff. There is very decent alternatives today, one of them I heard about is https://www.drone.io/

And we could run a gitea server and use their CI system that is quite near to the GitHub actions syntax: https://docs.gitea.com/usage/actions/comparison

I think that Jenkins is part of the old world, uses java and some ressources.
Go/Rust written programs are much better in terms of energy consumption.
I host gitea services, and the power needed is quite low. A RaspberryPi can run it easy ^^

PR #18747, opened 2023-10-04
Add composer.lock to version control

MauricioFauth:
This will allow more control over the dependencies and will be easier to investigate issues from past commits as the dependencies will be locked.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -152,6 +152,9 @@
             "dealerdirect/phpcodesniffer-composer-installer": true,
             "phpstan/extension-installer": true,
             "composer/package-versions-deprecated": true
+        },
+        "platform": {
+            "php": "8.1.99"
         }
     }
 }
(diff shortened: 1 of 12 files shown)


composer.json
```diff
@@ -152,6 +152,9 @@
             "dealerdirect/phpcodesniffer-composer-installer": true,
             "phpstan/extension-installer": true,
             "composer/package-versions-deprecated": true
+        },
+        "platform": {
+            "php": "8.1.99"
```
williamdes, line 157: I am not sure, maybe this will have to be removed for the composer released version

williamdes (approved):

composer.json
MauricioFauth: This is ignored when installing using Composer.

williamdes: 🤔 I have to check once more, in my previous tests for Doctum this had some importance on what version of libraries that declared they php version compatibility would get installed

PR #18752, opened 2023-10-09
Upgrade web-auth/webauthn-lib to version 4

MauricioFauth:
Next step will be to upgrade it to version 5. After that, a config directive to change the rpId will be added.

### DIFF
--- a/bin/internal/create-release.sh
+++ b/bin/internal/create-release.sh
@@ -295,7 +295,7 @@ security_checkup() {
         echo 'TCPDF should be installed, detection failed !'
         exit 1;
     fi
-    if [ ! -f vendor/web-auth/webauthn-lib/src/Server.php ]; then
+    if [ ! -f vendor/web-auth/webauthn-lib/src/PublicKeyCredential.php ]; then
         echo 'Webauthn-lib should be installed, detection failed !'
         exit 1;
     fi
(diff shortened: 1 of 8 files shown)


williamdes (approved):

MauricioFauth:
@williamdes, What do you think about requiring this lib for 6.0 and then dropping the custom server that I wrote? We already ship this lib for the website package.

williamdes:
> @williamdes, What do you think about requiring this lib for 6.0 and then dropping the custom server that I wrote? We already ship this lib for the website package. I can also work on using this lib to replace the code-lts/u2f-php-server for FIDO U2F.

I am not sure what you mean about website
But it looks like a good plan, let's make 6.0 clean we only have one shot

MauricioFauth:
> I am not sure what you mean about website

The release package distributed on the website

williamdes:
We can change this like we want for 6.0 :)

joostdebruijn:
Hi @MauricioFauth. I was thinking about proposing a change related to phpMyAdmin's WebAuthn-implementation. At this moment the rpId is 'hardcoded' the hostname of the current request. That is a sensible default in most situations, however in some situations it would be nice if you can define the rpId yourself via the config (e.g. when you want to share a WebAuthn-credential across multiple hosts on the same domain). I can propose a separate PR for this, but looking at the changes in this PR it's maybe better to incorporate it on this branch as well?

PR #18760, opened 2023-10-13
Improve collations support for MariaDB 10.10+

Simbiat:
### Description

Some more collations were added and structural changes were done in MariaDB 10.10 as per https://jira.mariadb.org/browse/MDEV-27009 which requires changes in how we get the *full* list of collations if we are using this engine.
Since I would not consider this a *bug* submitting to `master` branch.
Since the change  only affects the kind of query which is run, and not the structure of the collation list, no extra tests are added and not existing tests are modified.

- [+] Make sure you have read our [CONTRIBUTING.md](https://github.com/phpmyadmin/phpmyadmin/blob/master/CONTRIBUTING.md) document.
- [+] Make sure you are making a pull request against the correct branch. For example, for bug fixes in a released version use the corresponding QA branch and for new features use the _master_ branch. If you have a doubt, you can ask as a comment in the bug report or on the mailing list.
- [+] Every commit has proper `Signed-off-by` line as described in our [DCO](https://github.com/phpmyadmin/phpmyadmin/blob/master/DCO). This ensures that the work you're submitting is your own creation.
- [+] Every commit has a descriptive commit message.
- [+] Every commit is needed on its own, if you have just minor fixes to previous commits, you can squash them.
- [+] Any new functionality is covered by tests.

Signed-off-by: Dmitrii Kustov <simbiat@outlook.com>

### DIFF
--- a/src/Charsets.php
+++ b/src/Charsets.php
@@ -12,6 +12,7 @@ use PhpMyAdmin\Charsets\Collation;
 
 use function __;
 use function array_keys;
+use function count;
 use function explode;
 use function is_string;
 use function ksort;
@@ -28,7 +29,7 @@ class Charsets
      *
      * @var array<string, string>
      */
-    public static array $mysqlCharsetMap = [
+    public static $mysqlCharsetMap = [
         'big5' => 'big5',
         'cp-866' => 'cp866',
         'euc-jp' => 'ujis',
@@ -56,24 +57,27 @@ class Charsets
 
     /**
      * The charset for the server
+     *
+     * @var Charset|null
      */
-    private static Charset|null $serverCharset = null;
+    private static $serverCharset = null;
 
     /** @var array<string, Charset> */
-    private static array $charsets = [];
+    private static $charsets = [];
 
     /** @var array<string, array<string, Collation>> */
-    private static array $collations = [];
+    private static $collations = [];
 
     /**
      * Loads charset data from the server
      *
-     * @param bool $disableIs Disable use of INFORMATION_SCHEMA
+     * @param DatabaseInterface $dbi       DatabaseInterface instance
+     * @param bool              $disableIs Disable use of INFORMATION_SCHEMA
      */
     private static function loadCharsets(DatabaseInterface $dbi, bool $disableIs): void
     {
         /* Data already loaded */
-        if (self::$charsets !== []) {
+        if (count(self::$charsets) > 0) {
             return;
         }
 
@@ -100,27 +104,21 @@ class Charsets
     /**
      * Loads collation data from the server
      *
-     * @param bool $disableIs Disable use of INFORMATION_SCHEMA
+     * @param DatabaseInterface $dbi       DatabaseInterface instance
+     * @param bool              $disableIs Disable use of INFORMATION_SCHEMA
      */
     private static function loadCollations(DatabaseInterface $dbi, bool $disableIs): void
     {
         /* Data already loaded */
-        if (self::$collations !== []) {
+        if (count(self::$collations) > 0) {
             return;
         }
 
-        /* Check if we have `FULL_COLLATION_NAME`, which mean that we have MariaDB 10.10 or newer.
-        Refer https://jira.mariadb.org/browse/MDEV-27009 */
-        $sql = 'SELECT `COLUMN_NAME` FROM `information_schema`.`COLUMNS`'
-            . ' WHERE `TABLE_SCHEMA`=\'information_schema\' AND'
-            . ' `TABLE_NAME`=\'COLLATION_CHARACTER_SET_APPLICABILITY\' AND'
-            . ' `COLUMN_NAME`=\'FULL_COLLATION_NAME\'';
-
-        $check = $dbi->query($sql);
-
-        if ($check !== []) {
-            /* Use query to accomodate new structure of MariaDB collations
-            Note, that SHOW COLLATION command is not applicable at the time of writing*/
+        if ($dbi->isMariaDB() && version_compare($dbi->getVersionString(), '10.10', 'ge')) {
+            /* Use query to accommodate new structure of MariaDB collations.
+            Note, that SHOW COLLATION command is not applicable at the time of writing.
+            Refer https://jira.mariadb.org/browse/MDEV-27009
+            getVersionString() is used instead of getVersion() because of possibly variable integer length.*/
             $sql = 'SELECT `collapp`.`FULL_COLLATION_NAME` AS `Collation`,'
                 . ' `collapp`.`CHARACTER_SET_NAME` AS `Charset`,'
                 . ' `collapp`.`ID` AS `Id`,'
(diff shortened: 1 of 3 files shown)


Simbiat:
Hm... So I see several issues from tests:
1. Selenium one does not seem to be relevant, and is rather issue with the test
2. [analyse-php](https://github.com/phpmyadmin/phpmyadmin/actions/runs/6504493199/job/17668023491#logs) also complains about `$check !== []` but this was agreed in my initial attempt, so not clear what else can be used here
3. I am entirely unsure about Codecov report: does it want some extra test for the `if` branching or something?

src/Charsets.php
```diff
@@ -109,16 +109,39 @@ private static function loadCollations(DatabaseInterface $dbi, bool $disableIs):
...
-            . ' `CHARACTER_SET_NAME` AS `Charset`,'
-            . ' `ID` AS `Id`,'
-            . ' `IS_DEFAULT` AS `Default`,'
-            . ' `IS_COMPILED` AS `Compiled`,'
-            . ' `SORTLEN` AS `Sortlen`'
-            . ' FROM `information_schema`.`COLLATIONS`';
-
-        if ($disableIs) {
-            $sql = 'SHOW COLLATION';
+        /* Check if we have `FULL_COLLATION_NAME`, which mean that we have MariaDB 10.10 or newer.
```
MauricioFauth, line 112: I think is better to directly check if it's a MariaDB 10.10+ server using `$dbi->isMariaDB()` and `$dbi->getVersion()`.

MauricioFauth (changes requested):

Simbiat:
what the... I only change 1 small thing, but somehow there are lots of changes now and lots of complaints, too O_o

MauricioFauth:
> what the... I only change 1 small thing, but somehow there are lots of changes now and lots of complaints, too O_o
> 
> Ah, figured out, what was wrong, corrected and covered one complaint that was actually relevant

To help avoid these type of issues, is recommended to create a new branch before opening a pull request, this way will be easier to sync your `master` branch with the upstream remote.

Simbiat:
Yay, everything passed besides `codecov/patch`. I do not understand how to pass that one. I understand that it wants a test, but existing tests call `getCollations` function, and not the changed `loadCollations`, so to test this it would bee required to somehow force the MariaDB check to succeed, but where should I even start adding a test like that?

src/Charsets.php
```diff
@@ -109,16 +110,32 @@ private static function loadCollations(DatabaseInterface $dbi, bool $disableIs):
...
-            . ' `SORTLEN` AS `Sortlen`'
-            . ' FROM `information_schema`.`COLLATIONS`';
-
-        if ($disableIs) {
-            $sql = 'SHOW COLLATION';
+        if ($dbi->isMariaDB() && version_compare($dbi->getVersionString(), '10.10', 'ge')) {
+            /* Use query to accommodate new structure of MariaDB collations.
+            Note, that SHOW COLLATION command is not applicable at the time of writing.
+            Refer https://jira.mariadb.org/browse/MDEV-27009
+            getVersionString() is used instead of getVersion() because of possibly variable integer length.*/
```
MauricioFauth, line 117: What do you mean by variable integer length?

Simbiat: Like 10.9.10 should become 10910, which will be greater than 1010, and same if you have MariaDB 10.9.8, it's integer should be 1098, which will be greater than 1010.

MauricioFauth: The integer version is constructed in a way that newer versions are always bigger.
For example, 10.9.10 will become 100910. The 10.10 version, which is actually 10.10.0, will become 101000. The version 5.7.2 is represented as 50702. This way 101000 > 100910 > 100908 > 50702.

Simbiat: Ok, got it. Switched to `getVersion()`

williamdes:
> Yay, everything passed besides `codecov/patch`. I do not understand how to pass that one. I understand that it wants a test, but existing tests call `getCollations` function, and not the changed `loadCollations`, so to test this it would bee required to somehow force the MariaDB check to succeed, but where should I even start adding a test like that?

Most probably in: `test/classes/CharsetsTest.php`

And probably using a bit of
https://github.com/phpmyadmin/phpmyadmin/blob/83b1d6be45f104c50f21c43cd08d8254c83c9014/test/classes/Controllers/Server/Privileges/AccountUnlockControllerTest.php#L41-L42
https://github.com/phpmyadmin/phpmyadmin/blob/83b1d6be45f104c50f21c43cd08d8254c83c9014/test/classes/Controllers/Server/Privileges/AccountUnlockControllerTest.php#L60

Simbiat:
Added a test, hope it's enough for `codecov`

Simbiat:
And it failed. Probably because I need to feed something else into the stub, but I am unsure what (the whole query, I'm testing? just the expected results?)

test/classes/CharsetsTest.php
```diff
@@ -100,6 +101,20 @@ public function testGetCollationsWithIS(): void
...
     }
 
+    public function testGetCollationsMariaDB(): void
+    {
+        DatabaseInterface::$instance = $this->createDatabaseInterface();
+        $dbi = $this->createStub(DatabaseInterface::class);
+        $dbi->method('isMariaDB')->willReturn(true);
+        $dbi->method('getVersion')->willReturn(101000);
+        $collations = Charsets::getCollations($dbi, false);
+        $this->assertCount(4, $collations);
```
williamdes, line 111: 1) PhpMyAdmin\Tests\CharsetsTest::testGetCollationsMariaDB
Failed asserting that actual size 0 matches expected size 4.

Can you run the test on your workstation ?

`./vendor/bin/phpunit --no-coverage --filter=testGetCollationsMariaDB`

williamdes:
> And it failed. Probably because I need to feed something else into the stub, but I am unsure what (the whole query, I'm testing? just the expected results?)

🤔 Maybe there is something wrong, I need to take time to check this
I suspect that the mock DBI is not as great as the normal DbiDummy class that uses the query you added earlier

Simbiat:
I have not used PHPUnit before, so will need to set it up first, if I am to try running it, but I think the test fails because we fake MariaDB, while the environment still has MySQL, so the query fails, due to unknown column(s).  I see `testGetCollationsWithoutIS` is adding the results before running the test, perhaps the same trick should be done for the new test, too?

test/classes/CharsetsTest.php
```diff
@@ -100,6 +101,20 @@ public function testGetCollationsWithIS(): void
         }
     }
 
+    public function testGetCollationsMariaDB(): void
+    {
+        DatabaseInterface::$instance = $this->createDatabaseInterface();
+        $dbi = $this->createStub(DatabaseInterface::class);
+        $dbi->method('isMariaDB')->willReturn(true);
+        $dbi->method('getVersion')->willReturn(101000);
```
MauricioFauth, line 109:
```suggestion
        $dbi = $this->createDatabaseInterface();
        $dbi->setVersion(['@@version' => '10.10.0-MariaDB']);
```

MauricioFauth (approved):

MauricioFauth:

gabrielrbarbosa:
@MauricioFauth

PR #18776, opened 2023-10-21
Fix truncate operation for views

### LINKED ISSUE #18762: Unable to truncate all tables when a VIEW is included.
### Describe the bug

When using the "Check all" method in phpMyAdmin, options can be selected that will result into an error. For example, when selecting all tables, views will be included/selected as well. When choosing the option to empty/truncate them, an error appears.

### To Reproduce

1. Go to phpMyAdmin, create a new table, for example, something like this:

```CREATE TABLE `test` (
  `id` INT NOT NULL AUTO_INCREMENT , PRIMARY KEY (`id`)
  `firstname` varchar(50) COLLATE utf8mb4_unicode_ci NOT NULL,
  `lastname` varchar(50) COLLATE utf8mb4_unicode_ci NOT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;```

2. Also, create a VIEW, doesn't matter what, just create some, for example:

```CREATE VIEW `test2` AS SELECT * FROM `test`;```

3. Navigate to the database, choose "Check all" als select the option "Empty" in dropdown "With selected":

image

4. Click [ Yes ] to confirm the action:

image

5. See error

image

### Expected behavior

A view can't be truncated, so the statement ```TRUNCATE `test2`;``` can't be executed. However, I'm able to select it. I thins views shouldn't be included in the "Check all" option OR when views are checked, the option "Empty" should be disabled OR when views are included in the list with tables to empty/truncate, just skip them to prevent errors.

### PR DESCRIPTION
**Pull request by u7723606**
### Description

This pull request was for the issue #18762 
Currently, attempting to truncate a view results in an error because views cannot be truncated. I think it would be better to just skip over the views when they are included in the list of tables to truncate to prevent errors.

- Fixes #18762 

I added a check to truncate function so that if the target is a view, the TRUNCATE operation is not executed.

### DIFF
--- a/libraries/classes/Controllers/Database/Structure/EmptyTableController.php
+++ b/libraries/classes/Controllers/Database/Structure/EmptyTableController.php
@@ -14,6 +14,7 @@ use PhpMyAdmin\Message;
 use PhpMyAdmin\Operations;
 use PhpMyAdmin\ResponseRenderer;
 use PhpMyAdmin\Sql;
+use PhpMyAdmin\Table;
 use PhpMyAdmin\Template;
 use PhpMyAdmin\Transformations;
 use PhpMyAdmin\Util;
@@ -21,6 +22,7 @@ use PhpMyAdmin\Utils\ForeignKey;
 
 use function __;
 use function count;
+use function is_string;
 
 final class EmptyTableController extends AbstractController
 {
@@ -82,6 +84,10 @@ final class EmptyTableController extends AbstractController
         $selectedCount = count($selected);
 
         for ($i = 0; $i < $selectedCount; $i++) {
+            if (! is_string($selected[$i]) || Table::get($selected[$i], $GLOBALS['db'], $this->dbi)->isView()) {
+                continue;
+            }
+
             $aQuery = 'TRUNCATE ';
             $aQuery .= Util::backquote($selected[$i]);
 
--- a/libraries/classes/Partitioning/Maintenance.php
+++ b/libraries/classes/Partitioning/Maintenance.php
@@ -7,8 +7,10 @@ namespace PhpMyAdmin\Partitioning;
 use PhpMyAdmin\DatabaseInterface;
 use PhpMyAdmin\Dbal\DatabaseName;
 use PhpMyAdmin\Dbal\TableName;
+use PhpMyAdmin\Table;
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/Partitioning/Maintenance.php [hunk 1]
```diff
@@ -134,17 +134,34 @@ public function repair(DatabaseName $db, TableName $table, string $partition): a
...
+            $this->dbi->$table->getName()
         );
 
-        $this->dbi->selectDb($db);
-        $result = $this->dbi->tryQuery($query);
-
-        return [(bool) $result, $query];
+        $result = $this->dbi->fetchResult($query);
+        return !empty($result);
+    }
```
williamdes, line 147: You do not need this function, the Table class has isView()

src/Partitioning/Maintenance.php
```diff
@@ -134,17 +134,34 @@ public function repair(DatabaseName $db, TableName $table, string $partition): a
...
+    
+    public function truncate(DatabaseName $db, TableName $table, string $partition): array
+    {
+        if ($this->isView($db, $table)) {
```
williamdes, line 151: @MauricioFauth what is the best way to build a Table object from a TableName ?

MauricioFauth:
```suggestion
        if (\PhpMyAdmin\Table::get($table->getName(), $db->getName(), $this->dbi)->isView()) {
```

u7723606: Thanks! I gonna try this function and improve the truncate()

src/Partitioning/Maintenance.php
```diff
@@ -136,6 +136,10 @@ public function repair(DatabaseName $db, TableName $table, string $partition): a
      */
     public function truncate(DatabaseName $db, TableName $table, string $partition): array
     {
+        if (\PhpMyAdmin\Table::get($table->getName(), $db->getName(), $this->dbi)->isView()) {
+            return [false, 'Table is a view, cannot truncate.'];
```
williamdes, line 140:
```suggestion
            return [false, _('This table is a view, it can not be truncated.')];
```
Adds translation

MauricioFauth (approved):

MauricioFauth:

PR #18777, opened 2023-10-23
Fix search using enums

CommanderRoot:
### Description

#18693 broke the selection of enums on searches as it couldn't deal with the space after a comma anymore. This PR removes the added spaces which restores the ability to use the select inputs on searches again

Before:
grafik

After:
grafik

### DIFF
--- a/src/Controllers/Table/SearchController.php
+++ b/src/Controllers/Table/SearchController.php
@@ -125,9 +125,7 @@ class SearchController extends AbstractController
             }
 
             // reformat mysql query output
-            if (strncasecmp($type, 'set', 3) == 0 || strncasecmp($type, 'enum', 4) == 0) {
-                $type = str_replace(',', ', ', $type);
-            } else {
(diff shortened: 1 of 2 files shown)


kamil-tekiela (approved):
It would be good to have a test case for this too.

williamdes (approved):
Indeed a test would be great

CommanderRoot:
I just noticed that this "fix" doesn't fix some other issues I found. For example there are issues when the `enum` / `set` value contains a `,` as that will just get treated as option delimitor and not part of the value.

kamil-tekiela:
If you can provide a table schema as an example, I can take a look and fix them. Maybe the custom implementation had more going on than I thought.

CommanderRoot:
```
CREATE TABLE `test` (
  `test` enum('1','2,','3''','''4') NOT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;
```
Results in this on the search:
grafik

kamil-tekiela:
Thanks. I will take a look at it once I have some time. We might have to revert my PR.

PR #18778, opened 2023-10-24
Rename ./libraries directory to ./app

MauricioFauth:
Also moves the `./cache` directory to `./app/cache`.

- Related to https://github.com/phpmyadmin/phpmyadmin/issues/18512
- Related to https://github.com/phpmyadmin/phpmyadmin/pull/18740


test/doctum-config.php
```diff
@@ -13,7 +13,7 @@
 $iterator = Finder::create()
     ->files()
     ->name('*.php')
-    ->in($root . 'libraries');
+    ->in($root . 'src');
```
MauricioFauth, line 16: @williamdes Is this right?

williamdes (approved):
Nice!
The topic about permissions on the cache folder will come back to us if write permission is needed
if it is read only that's okay and we will have no problems
That's not related to your PR but to the folder existing in the extracted zip on the user's workstation
Something like that

test/doctum-config.php
williamdes: Yes, it must search in the folder where classes are

You can run it locally with -vvv I think
or double dash debug

PR #18788, opened 2023-11-03
Remove containerBuilder global variable

MauricioFauth:
Removes ContainerBuilder::setParameter() calls. These parameters are only written, but never used.


williamdes (approved):
🎉
Easier to replace the DI lib if needed

MauricioFauth:
> 🎉 Easier to replace the DI lib if needed

That's the plan!

PR #18789, opened 2023-11-07
Change icons to use SVG sprite instead of individual images

MauricioFauth:
- [x] Change icons to use SVG sprite instead of individual images
- [x] Inline SVG icons into the sprite instead of using `<image>` (This allows the SVG icons to use `currentColor` as fill)
- [x] Automate the sprite generation


liviuconcioiu:
@MauricioFauth can you check why there are a bunch of 404 errors? It happens on demo server and yesterday's snapshot.

MauricioFauth:
Thank you, @liviuconcioiu.
I [reverted the commit](https://github.com/phpmyadmin/phpmyadmin/commit/f9ab9d15aca6f06604243ad29df9a67cc45bc577) until I find a solution for this.
Looks like Chromium and Firefox have different opinions on how to resolve the relative URL of the images (Chromium being the wrong one in my opinion).

PR #18798, opened 2023-11-22
Rename scripts directory to bin

MauricioFauth:
- Related to #18512


williamdes (approved):
Can we have bin/internal/ for scripts that are only internal and not needed in the final vendor bundles ?

williamdes (approved):
Can we have bin/internal/ for scripts that are only internal and not needed in the final vendor bundles ?

MauricioFauth:
> Can we have bin/internal/ for scripts that are only internal and not needed in the final vendor bundles ?

Aren't all scripts internal? We remove this directory when creating a release.

williamdes:
> > Can we have bin/internal/ for scripts that are only internal and not needed in the final vendor bundles ?
> 
> Aren't all scripts internal? We remove this directory when creating a release.

No they are not
And we should not remove the directory on some type of bundles, the source one I think
That said such bundles do not need internal scripts like check release excludes or create release

bin/lang-cleanup.sh
williamdes, line 1: Internal

bin/locales-contributors
williamdes, line 1: internal

bin/set-version.sh
williamdes, line 1: internal

bin/upload-release
williamdes, line 1: very internal

bin/transformations_generator_plugin.sh
williamdes, line 1: internal

bin/check-release-excludes.sh
williamdes, line 1: very internal

MauricioFauth:
> No they are not
> And we should not remove the directory on some type of bundles, the source one I think
> That said such bundles do not need internal scripts like check release excludes or create release

Besides the transformations scripts (which should be removed in my opinion, and replaced by proper interfaces), all scripts are internal. That's why I don't see the point of having the internal directory.

I changed the create release script to not remove the bin directory for the source bundle. Should I also do that for 5.2?

williamdes:
> > No they are not
> > And we should not remove the directory on some type of bundles, the source one I think
> > That said such bundles do not need internal scripts like check release excludes or create release
> 
> Besides the transformations scripts (which should be removed in my opinion, and replaced by proper interfaces), all scripts are internal. That's why I don't see the point of having the internal directory.

Okay. For example in https://salsa.debian.org/phpmyadmin-team/phpmyadmin if I had the generate mo script it could re build the mo files from source too

> I changed the create release script to not remove the bin directory for the source bundle. Should I also do that for 5.2?

no, that's okay

PR #18804, opened 2023-11-27
Update Psalm baseline

kamil-tekiela:


composer.lock
```diff
@@ -5700,23 +5701,79 @@
             ],
             "time": "2023-11-08T05:53:05+00:00"
         },
+        {
+            "name": "rector/rector",
```
MauricioFauth, line 5705: This is a mistake.

kamil-tekiela: ohh good catch, I forgot to remove it

kamil-tekiela: Why is composer.lock committed anyway? This file should not be in the git tree.

MauricioFauth (approved):

composer.lock
MauricioFauth: One of the reasons is that is easier to investigate issues in old commits, as the dependencies for a specific commit will always be the same.

PR #18805, opened 2023-11-27
Remove globals in operations

kamil-tekiela:


composer.lock
```diff
@@ -5700,23 +5701,79 @@
             ],
             "time": "2023-11-08T05:53:05+00:00"
         },
+        {
+            "name": "rector/rector",
```
MauricioFauth, line 5705: Same mistake here.

MauricioFauth (approved):

Issue #18782, opened 2023-10-26
Impossible to manage users on AWS RDS Aurora Mysql


kamyweb:
### Describe the bug

On AWS RDS Aurora Mysql server, master user can't create users on phpmyadmin. 
User is displayed with only "USAGE" grant, but if I run a "SHOW GRANTS on 'admin'@'%'" is there another grant that refer to another user called "rds_superuser_role" that has all grant options.

### To Reproduce

1. create a new AWS RDS Aurora Mysql cluster
2. try connecting with latest phpmyadmin
3. go to "User accounts"
4. you can see "You do not have the privileges to administrate the users!" and you can't edit or create users
5. 'admin' user is show with only "USAGE" grant
6. try to run the query "SHOW GRANTS for 'admin'@'%';" and check the additional grant that doesn't shown

### Expected behavior

If user has and additional grants, phpmyadmin must add GRANTS of that to the user's one.

### Screenshots

users list without any edit/create
![image](https://github.com/phpmyadmin/phpmyadmin/assets/8101573/ec6955e9-c7c3-42c5-9ed7-c3dad1899909)
show grants output
![image](https://github.com/phpmyadmin/phpmyadmin/assets/8101573/b42c4c34-5ff5-4d32-b6fc-1f94dc9847be)

### Server configuration

- Operating system: official phpmyadmin Docker that is based on Debian 
- Web server: Apache/2.4.57
- Database version: RDS Aurora Mysql - 8.0.23
- PHP version: 8.2.11
- phpMyAdmin version: 5.2.1

williamdes:
Thank you, I will have a look

kamyweb:
@williamdes I made some other checkes and I think is a role-based auth used by RDS on new mysql-8 specifications. 
The "rds_superuser_role" is a role that is automatic created and is used to assign all grants to users.
![image](https://github.com/phpmyadmin/phpmyadmin/assets/8101573/1baedf3a-370c-41bf-baad-f9f1df9977dd)

I start a new mysql8.2 docker on my machine with latest phpmyadmin-6.0.0-dev and I run this queries:
```
CREATE ROLE `rds_superuser_role`;
GRANT ALL PRIVILEGES ON *.* TO 'rds_superuser_role'@'%' WITH GRANT OPTION;
CREATE USER 'testbug'@'%' IDENTIFIED BY 'testbug';
GRANT USAGE ON *.* TO 'testbug'@'%';
[...]

PR #18814, opened 2023-11-28
Add role based auth for MySQL 8.x and compatibles

kamyweb:
### Description

On MySQL 8.x engine implement a role based granting system. That system is used as default grant on AWS RDS Aurora MySQL where I hit this problem that I report on this issue https://github.com/phpmyadmin/phpmyadmin/issues/18782

As describe on MySQL 8.x documentation the roles maybe multiple and can be switched on runtime with `SET ROLE` command, that's why I implement it on that way and without any cache.

This PR is the first step to implement the new role based system and start to be compatible with this kind of grant system.

This changes may implemented also on phpMyAdmin 5.2 version.

### Testing

create a role and add that role to a test user without any grant; that user must create databases and add/change/revoke grants
```sql
CREATE ROLE `rds_superuser_role`;
GRANT ALL PRIVILEGES ON *.* TO 'rds_superuser_role'@'%' WITH GRANT OPTION;
CREATE USER 'testbug'@'%' IDENTIFIED BY 'testbug';
GRANT USAGE ON *.* TO 'testbug'@'%';
GRANT `rds_superuser_role`@`%` TO 'testbug'@'%';
SET DEFAULT ROLE 'rds_superuser_role' TO 'testbug'@'%';
```

refs:
general reference of roles starting from MySQL 8.x https://dev.mysql.com/doc/refman/8.0/en/roles.html
multiple roles allowed at the same time https://dev.mysql.com/doc/refman/8.0/en/information-functions.html#function_current-role
`SET ROLE` command https://dev.mysql.com/doc/refman/8.0/en/set-role.html

### DIFF
--- a/libraries/classes/DatabaseInterface.php
+++ b/libraries/classes/DatabaseInterface.php
@@ -50,6 +50,7 @@ use function openlog;
 use function reset;
 use function sprintf;
 use function str_contains;
+use function str_replace;
 use function str_starts_with;
 use function stripos;
 use function strlen;
@@ -120,6 +121,9 @@ class DatabaseInterface implements DbalInterface
 
     /** @var array Current user and host cache */
     private $currentUser;
+
+    /** @var array Current role and host cache */
+    private $currentRoleAndHost;
 
     /** @var string|null lower_case_table_names value cache */
     private $lowerCaseTableNames = null;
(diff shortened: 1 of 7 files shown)


kamyweb:
Marked as draft awaiting PR approval on main branch; after which I also rework this PR with the same suggestions/reworks that I made on the main.

williamdes (approved):
This diff looks nice, could you also port the test please ?

williamdes:
> Marked as draft awaiting PR approval on main branch; after which I also rework this PR with the same suggestions/reworks that I made on the main.

In fact when this PR is ready it will get merged into QA_5_2 and QA_5_2 into master
So the final diff of the master PR will be only master branch specific adjustments

williamdes (approved):
Thank you  !

MauricioFauth (approved):

MauricioFauth:

PR #18827, opened 2023-12-02
Fix params in ExportPlugin

kamil-tekiela:


src/Plugins/ExportPlugin.php
```diff
@@ -193,13 +190,12 @@ public function getTableDefStandIn(string $db, string $view, array $aliases = []
...
      * Outputs triggers
      *
-     * @param string    $db       database name
      * @param string    $table    table name
      * @param Trigger[] $triggers
      *
      * @return string Formatted triggers list
      */
-    protected function getTriggers(string $db, string $table, array $triggers): string
+    protected function getTriggers(string $table, array $triggers): string
```
williamdes, line 198: can we keep the database argument in case a new format needs it or some other export format provided by a 3rd party

kamil-tekiela: To be honest, I don't think we even need to keep this polymorphism at all. I've removed it completely now.

MauricioFauth (approved):

MauricioFauth (approved):

PR #18845, opened 2023-12-13
Refactor Favorites + DbTableExists

kamil-tekiela:


kamil-tekiela:
The tests are still failing and I don't know why.

MauricioFauth:
> The tests are still failing and I don't know why.

I added two commits that fixed the failing tests. These types of issues are usually related to global/static variables in conjunction with test doubles.

MauricioFauth (approved):

src/Controllers/Database/Structure/FavoriteTableController.php
```diff
@@ -170,7 +174,10 @@ private function synchronizeFavoriteTables(
             }
         }
 
-        $favoriteTables[$user] = $favoriteInstance->getTables();
+        $favoriteTables[$user] = array_map(
+            static fn (RecentFavoriteTable $table) => $table->toArray(),
+            $favoriteInstance->getTables(),
+        );
```
MauricioFauth, line 180: This was needed because DatabaseName and TableName doesn't implement JsonSerializable.

kamil-tekiela: Of course, and they don't use public properties either. Makes sense and I missed it. 

Thank you for fixing the tests too.

kamil-tekiela: I think this was not the only place where `json_encode` was used. Let me implement `JsonSerializable` then.

PR #18847, opened 2023-12-15
Refactor CentralColumnsController to use DatabaseName 

kamil-tekiela:
If I understand correctly, `$GLOBALS['db']` can be safely replaced with value from `$request` in controllers.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1506,38 +1506,8 @@ parameters:
 			path: src/Controllers/Database/CentralColumns/PopulateColumnsController.php
 
 		-
-			message: "#^Construct empty\\(\\) is not allowed\\. Use more strict comparison\\.$#"
-			count: 1
-			path: src/Controllers/Database/CentralColumnsController.php
-
-		-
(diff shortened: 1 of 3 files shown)


MauricioFauth (approved):

MauricioFauth:
Related to #17727 and #17769.

MauricioFauth:
> If I understand correctly, `$GLOBALS['db']` can be safely replaced with value from `$request` in controllers.

Yes, they are defined in DatabaseAndTableSetting middleware and getting db and table directly from request is preferable.

https://github.com/phpmyadmin/phpmyadmin/blob/5abc1a58702062335ef551926ef51f5187f414fb/src/Http/Middleware/DatabaseAndTableSetting.php#L32-L36

kamil-tekiela:
> > If I understand correctly, `$GLOBALS['db']` can be safely replaced with value from `$request` in controllers.
> 
> Yes, they are defined in DatabaseAndTableSetting middleware and getting db and table directly from request is preferable.
> 
> https://github.com/phpmyadmin/phpmyadmin/blob/5abc1a58702062335ef551926ef51f5187f414fb/src/Http/Middleware/DatabaseAndTableSetting.php#L32-L36

Would it make any sense to have methods for this in `ServerRequest`?

PR #18848, opened 2023-12-15
Remove dead code and globals

kamil-tekiela:
This PR removes a large chunk of code which in my opinion is not in use. I tried to verify if and how it's used, but I was not able to identify any code path that would lead to its execution. If it's actually used then it should be implemented in a better way. I am 90% sure this is dead code and neither of these globals should have been global.


src/Controllers/Sql/SqlController.php
```diff
@@ -172,13 +166,6 @@ public function __invoke(ServerRequest $request): void
             );
         }
 
-        /**
-         * Need to find the real end of rows?
-         */
-        if (isset($GLOBALS['find_real_end']) && $GLOBALS['find_real_end']) {
-            $GLOBALS['unlim_num_rows'] = $this->sql->findRealEndOfRows($GLOBALS['db'], $GLOBALS['table']);
-        }
```
MauricioFauth, line 180: This was introduced by 4766676ee2e95c1ddf3bd708acc0f3fba5beccb4. I think the only way it have worked was by enabling PHP's `register_globals` directive.

kamil-tekiela: Makes sense. Then I suppose it hasn't worked for a number of years.

resources/templates/display/results/table.twig
```diff
@@ -68,9 +68,6 @@
               <input type="hidden" name="pos" value="{{ navigation.pos_last }}">
               <input type="hidden" name="is_browse_distinct" value="{{ is_browse_distinct }}">
               <input type="hidden" name="goto" value="{{ goto }}">
-              {% if navigation.has_real_end_input %}
-                <input id="real_end_input" type="hidden" name="find_real_end" value="1">
```
MauricioFauth, line 72: There's more code related to this that should be removed as well.

https://github.com/phpmyadmin/phpmyadmin/blob/4a67d88e1dc73f250b39434154afa633a8661bb1/resources/js/src/database/structure.ts#L34

https://github.com/phpmyadmin/phpmyadmin/blob/4a67d88e1dc73f250b39434154afa633a8661bb1/resources/js/src/database/structure.ts#L422-L440

https://github.com/phpmyadmin/phpmyadmin/blob/4a67d88e1dc73f250b39434154afa633a8661bb1/src/Controllers/JavaScriptMessagesController.php#L68

MauricioFauth (changes requested):

MauricioFauth (approved):

PR #18849, opened 2023-12-16
Replace db and table globals with static class properties

MauricioFauth:
### Description

This improves type inference of these variables.


kamil-tekiela (approved):
I love it!

williamdes (approved):

PR #18856, opened 2023-12-20
Refactoring of getTablesFull and Query/Cache

kamil-tekiela:
I've decided to narrow down the type hints because we know that results from DB are either `string` or `null`. The mroonga calculations add `int` as another possible type. The cache may also have `ExactRows` which is an int too. Since we know the exact structure of the array, and by extension the structure of the cache, we can specify narrow type hints. This doesn't help with fetching the values from cache as it can be any part of the cache or the default value, so this remains `mixed`.

All changes in this PR attempt to simplify and clean up usage of `getTablesFull()` results, but there are still many places where the results are used in a complicated and confusing way. In future PRs I will try to untangle it as much as possible. I still have many open questions about how this works. For example, `getStatusInfo()` has param `$disableError`. I don't understand why this function triggers an error and why it can be silenced. It doesn't make any sense to me.

### DIFF
--- a/app/services_controllers.php
+++ b/app/services_controllers.php
@@ -1576,6 +1576,7 @@ return [
                 '$transformations' => '@transformations',
                 '$relation' => '@relation',
                 '$dbi' => '@dbi',
+                '$dbTableExists' => '@' . DbTableExists::class,
             ],
         ],
         UserPasswordController::class => [
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -6986,16 +6986,6 @@ parameters:
 			path: src/Database/Designer/Common.php
 
 		-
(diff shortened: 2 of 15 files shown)


MauricioFauth (approved):

MauricioFauth:
> For example, getStatusInfo() has param $disableError. I don't understand why this function triggers an error and why it can be silenced. It doesn't make any sense to me.

@kamil-tekiela The `trigger_error` was added by d650f5b74fc91ea60c4ed86213964f5450fa5aaf and `$disableError` was added by 7c818d046a4c9f01a8e94b55a0726a1cc3d6a60d. I don't really know why both were added. I think that `trigger_error` can be removed.

PR #18857, opened 2023-12-20
New implementation for SqlAutoCompleteController

kamil-tekiela:
This is an optimized implementation of SQL Autocomplete. It produces identical output but avoids N+1 problem. Testing with 10k tables, the time went down from 96s to 2s.


MauricioFauth (approved):

williamdes (commented):
@MauricioFauth do you agree we should fix this on QA ?

PR #18858, opened 2023-12-20
Optimize MultiTableQuery

kamil-tekiela:
I used a custom SQL query here to avoid N+1 problem. On 10k tables the difference is between 94s and 13s. Maybe there's a better way of doing this, but this seems simple enough.

### DIFF
--- a/src/Database/MultiTableQuery.php
+++ b/src/Database/MultiTableQuery.php
@@ -19,26 +19,19 @@ use PhpMyAdmin\Transformations;
 use PhpMyAdmin\Url;
 
 use function md5;
+use function sprintf;
 
 /**
  * Class to handle database Multi-table querying
  */
 class MultiTableQuery
 {
-    /**
-     * Table names
-     *
-     * @var array<int, string>
-     */
-    private array $tables;
-
     public function __construct(
         private DatabaseInterface $dbi,
         public Template $template,
         private string $db,
         private int $defaultNoOfColumns = 3,
     ) {
-        $this->tables = $this->dbi->getTables($this->db);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Database/MultiTableQuery.php
```diff
@@ -48,10 +41,17 @@ public function __construct(
...
+            'SELECT TABLE_NAME, COLUMN_NAME FROM information_schema.columns WHERE table_schema = %s',
+            $this->dbi->quoteString($this->db),
+        ));
+
         $tables = [];
-        foreach ($this->tables as $table) {
+        /** @var array{TABLE_NAME:string, COLUMN_NAME:string} $column */
+        foreach ($columnsInTables as $column) {
+            $table = $column['TABLE_NAME'];
             $tables[$table]['hash'] = md5($table);
```
MauricioFauth, line 53: This only needs to be done the first time.

MauricioFauth (approved):

src/Database/MultiTableQuery.php
kamil-tekiela: Is this better? I am not sure how to do it better.

MauricioFauth: This is a good way of doing it. Other way is to use null coalescing attribution, as this value is always string or undefined. I'm happy with both ways. I think both ways will be faster than calculate the same hash for every column.

kamil-tekiela: I like your suggestion. I haven't checked the performance, but it should not be a problem either way. `md5` is a fast hash.

kamil-tekiela: Fun fact: the null-coalesce operator is actually more opcodes in this particular case than array_key_exists, and on a nanoscale would save a few CPU cycles. Of course, it doesn't matter much and aesthetics play more important role here. If the left side operand of null-coalesce assignment was a more expensive operation, it would matter then.

MauricioFauth (approved):

PR #18870, opened 2023-12-27
Fix #17538 Index Rename Button generated incorrect sql

### LINKED ISSUE #17538: Index Rename Button generated incorrect sql
### Describe the bug
Index Rename Button generated incorrect sql
this index choice is 'INDEX' instead of 'UNIQUE' and it loses the Index Field

### Expected behavior

The expected generated command should looks like this:  

> ALTER TABLE db.table DROP INDEX idx_x, ADD INDEX idx_x (x) ;

I saw this [Issues14953](https://github.com/phpmyadmin/phpmyadmin/issues/14953), so I edited this  function `getSqlQueryForIndexCreateOrEdit()` ,replace `$index` with `$this->getIndex($oldIndex)`  &  $sql_query .=`ADD new_name` which in file `libraries/classes/Table.php` .
it seems to work
### Screenshots

Bug pic

### Server configuration

- Web server:nginx/1.20.2
- Database version: MariaDB 10.3.22
- PHP version: 7.3.4
- phpMyAdmin version: 5.1.2

### Client configuration

- Browser: Chrome
- Operating system: MacOS

kamil-tekiela:
Fix #17538

### DIFF
--- a/libraries/classes/Controllers/Table/IndexRenameController.php
+++ b/libraries/classes/Controllers/Table/IndexRenameController.php
@@ -53,7 +53,9 @@ final class IndexRenameController extends AbstractController
         if (isset($_POST['index'])) {
             if (is_array($_POST['index'])) {
                 // coming already from form
-                $index = new Index($_POST['index']);
+                $oldIndex = is_array($_POST['old_index']) ? $_POST['old_index']['Key_name'] : $_POST['old_index'];
+                $index = clone $this->dbi->getTable($this->db, $this->table)->getIndex($oldIndex);
+                $index->setName($_POST['index']['Key_name']);
             } else {
                 $index = $this->dbi->getTable($this->db, $this->table)->getIndex($_POST['index']);
             }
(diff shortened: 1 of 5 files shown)


williamdes (approved):
Thank you !
Can you add an unit test so we can ensure the query is clean now ?

kamil-tekiela:
> Thank you ! Can you add an unit test so we can ensure the query is clean now ?

I thought about it, but it seems to be quite a challenge. I'd rather submit the test on master branch where it would be easier to do it. But I also think that this should be redesigned on master. So I don't even know which behaviour should be tested. I did find tests that should cover this feature, but the expectations are wrong. 

Would you be able to help me with an easy unit test for 5.2?

MauricioFauth (approved):

MauricioFauth:
> Would you be able to help me with an easy unit test for 5.2?

I added an unit test that covers the path for the preview SQL action.

Issue #17878, opened 2022-11-07
A PRIMARY key should not be possible to rename


kamil-tekiela:
### Describe the bug

An index can be renamed. PhpMyAdmin has a dedicated button for this action on table/structure/ page. However, trying to rename a PRIMARY key yields a strange error. This is because the code doesn't expect this option to be available for PRIMARY key. 

### To Reproduce

1. Go to /table/structure
2. Click on 'rename' on a PRIMARY key
3. Give some name

### Expected behavior

The button should be disabled or invisible. 

- phpMyAdmin version: master

kamil-tekiela:
Now that I think about it... if you are using a version that doesn't have `RENAME` support, this feature is completely broken. Probably the button should only be available on versions that support this feature in the first place.

williamdes:
Should we disable the button or hide it @MauricioFauth?

PR #18872, opened 2023-12-28
New implementation for Indexes

kamil-tekiela:
This PR finally finishes the feature of renaming indexes. Code copied unnecessarily from the other controller is removed. Shared code is decoupled and cleaned up. Bugs are fixed and tests are adjusted. 

This PR fixes bugs:
- https://github.com/phpmyadmin/phpmyadmin/issues/17538
- #17878

While the bug fixes should go to 5.2, this PR's changes are too extensive. The first one already has a patch and the second one can receive a small patch too (see last commit). 

`IndexesController` still needs clean up and I think there's an issue with JS which triggers multiple AJAX requests, but I wasn't able to narrow it down.


kamil-tekiela:
I'm not sure why this one test failed.

MauricioFauth:
> I'm not sure why this one test failed.

Probably related to `PhpMyAdmin\Index::$registry`. Look at the test that I added to the other PR.

kamil-tekiela:
@MauricioFauth  Thanks. I think I got it now. But the tests got cancelled.

MauricioFauth (approved):

PR #18885, opened 2024-01-07
Refactor ListDatabase class

MauricioFauth:


src/ListDatabase.php
```diff
@@ -17,38 +19,34 @@
...
+    public function __construct(
+        private readonly DatabaseInterface $dbi,
+        private readonly Config $config,
+        private readonly CheckUserPrivileges $checkUserPrivileges,
+    ) {
         parent::__construct();
 
-        $checkUserPrivileges = new CheckUserPrivileges(DatabaseInterface::getInstance());
-        $checkUserPrivileges->getPrivileges();
+        $this->checkUserPrivileges->getPrivileges();
```
kamil-tekiela, line 35: Do you know what this does? Should this class be a property?

MauricioFauth: Good question. I'll investigate it.

PR #18891, opened 2024-01-09
Fix initialization of Relation and Tracker

kamil-tekiela:
This PR attempts to fix a circular dependency between Relation and Tracking. The logic is tidied up, and the correct sequence is applied. 

Tracking is dependent on Relation being initialized. Relation can only be initialized once the DB connection is established. Since we want to cover all SQL queries in Tracking, it should be enabled as soon as possible. I could not find a better place to enable it than in the Middleware `Authentication`. It's the only place that made sense. However, it should probably be split into dedicated middleware. 

I removed all other places where the Tracking or Relation was enabled. It should now be guaranteed that everything is properly initialized in middleware. 

Once the correct sequence was implemented, I was able to remove unnecessary code. `ZeroConfPostConnection` seems to have lost all purpose. Same with `postConnectControl()`. I am not sure what the purpose of `Cache` class is, so I changed it to a plain static property. 

Related to #18710. This should improve performance as it removes unnecessary setup of Relation. Instead of doing it 3 times, we only do it once now. 

@MauricioFauth Can you please review it and make sure I haven't messed anything up?


MauricioFauth (approved):
These looks like good changes for me. I didn't find something that could break.

PR #18893, opened 2024-01-09
Don't optimize count for UNION queries

### LINKED ISSUE #18890: Query using UNION show errors on phpmyadmin query count
### Describe the bug

Performing a query such as:
```sql
select * from t1 union  select * from t2;
```

where tables have at least 2 columns and numbers of result set rows is >= 25 ( or 'Numbers of rows' selected on 'Select' menu),
we get this error:
Error: #1222 The used SELECT statements have a different number of columns

because PHPMyadmin implicit perform this query to get the total rows from above query:

```sql
SELECT COUNT(*) FROM (SELECT 1 FROM t1 UNION SELECT * FROM t2 ) as cnt
```

is replacing "SELECT *" with "SELECT 1" only on first query, which obviously returns that error. 

### Expected behavior

Don't see that error at the bottom.

### Server configuration

- Operating system: Linux
- Web server:
- Database version: MariaDB 10.6.11
- PHP version: 8.2.9
- phpMyAdmin version: 5.2.1

kamil-tekiela:
Fixes #18890

### DIFF
--- a/libraries/classes/Sql.php
+++ b/libraries/classes/Sql.php
@@ -752,6 +752,7 @@ class Sql
                 if (
                     $analyzedSqlResults['is_group'] === false
                     && $analyzedSqlResults['distinct'] === false
+                    && $analyzedSqlResults['union'] === false
                     && count($statement->expr) === 1
                 ) {
                     $statement->expr[0] = new Expression();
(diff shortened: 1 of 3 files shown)


williamdes (approved):
Could you add a test case if this is possible in an easy way ?

MauricioFauth (approved):

PR #15565, opened 2019-11-11
#13912 and #15123 - check phpMyAdmin tables are accessible

williamdes:
### Description

The tables can be listed but not accessible.
Corrupt data (#13912) or permission issue (#15123)
Fixes #13912 and fixes #15123

PR #18896, opened 2024-01-09
Refactor relation

kamil-tekiela:
Related to #18710

This removes checks for access to PMA tables. I really don't think we need to do this. If the control user doesn't have the privilege to access them, then `SHOW TABLES` will not list them. I doubt they would be inaccessible for any other reason. This should improve performance on slow systems.

### DIFF
--- a/src/ConfigStorage/Relation.php
+++ b/src/ConfigStorage/Relation.php
@@ -247,7 +247,6 @@ class Relation
             || $config->selectedServer['pmadb'] === ''
             || ! $this->dbi->selectDb($config->selectedServer['pmadb'], Connection::TYPE_CONTROL)
         ) {
-            // No server selected -> no bookmark table
             $config->selectedServer['pmadb'] = '';
 
             return $relationParams;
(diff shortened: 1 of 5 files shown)


src/ConfigStorage/Relation.php
```diff
@@ -307,18 +279,6 @@ private function checkRelationsParam(): array
...
     }
 
-    /**
-     * Check if the table is accessible
-     *
-     * @param string $tableDbName The table or table.db
-     */
-    public function canAccessStorageTable(string $tableDbName): bool
-    {
-        $result = $this->dbi->tryQueryAsControlUser('SELECT NULL FROM ' . Util::backquote($tableDbName) . ' LIMIT 0');
```
williamdes, line 317: this is required because some weird setups do have the table listed but no select grant onto it

kamil-tekiela: I'd be interested in learning more about this. For me, this doesn't make sense. The MySQL specification says

> If you have no privileges for a base table or view, it does not show up in the output from [SHOW TABLES](https://dev.mysql.com/doc/refman/8.0/en/show-tables.html) or [mysqlshow db_name](https://dev.mysql.com/doc/refman/8.0/en/mysqlshow.html).
https://dev.mysql.com/doc/refman/8.0/en/show-tables.html

So if they don't have the SELECT grant on it, then it cannot show in the list. But if there's some weird setup that does this then I want to reproduce it and maybe find a better way.

MauricioFauth: Related to #15565.

kamil-tekiela: Thanks, @williamdes  and @MauricioFauth, for enlightening me. Now, I understand what circumstances can lead to this being a problem. To summarize this for posterity:
1. Create a user with only `GRANT` as global permission. 
2. Use that new user as a control user in config

What happens is that MySQL, for some strange reasons, considers ANY global permission as a permission to see the full list of databases and tables. Even though the user doesn't really have access to them, MySQL thinks they do. 

This, I believe, is a fault of MySQL documentation for ambiguous wording. I doubt it's a MySQL bug, because I think it makes some sense. A user that has global `GRANT` privilege technically has access to any table so that they can grant privileges on it. 

What we need is a way to identify if the user has **the necessary** permission for PMA tables. It's not just `SELECT`, but also for `INSERT`, `UPDATE`, `DELETE`, `CREATE` and probably more. While PMA seems to crash only without `SELECT`, it doesn't work properly without the remaining permissions. It spews out errors about missing privileges. 

Obviously, we should not merge my PR as it is.

PR #18910, opened 2024-01-12
Fix creation of no-password user

kamil-tekiela:
Here's another issue I've caused that I have just found. Previously if "no password" was selected then the SQL would set the password to an empty string. While refactoring, I missed the quotes and broke the resulting SQL.

### DIFF
--- a/src/Server/Privileges.php
+++ b/src/Server/Privileges.php
@@ -3103,7 +3103,7 @@ class Privileges
             if ($_POST['pred_password'] === 'keep') {
                 $createUserReal = sprintf($createUserStmt, $slashedPassword);
             } elseif ($_POST['pred_password'] === 'none') {
-                $createUserReal = sprintf($createUserStmt, '');
+                $createUserReal = sprintf($createUserStmt, "''");
             } else {
                 if (
                     ! ((Compatibility::isMariaDb() && $isMariaDBPwdPluginActive)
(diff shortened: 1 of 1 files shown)


MauricioFauth:
> Here's another issue I've caused that I have just found. Previously if "no password" was selected then the SQL would set the password to an empty string. While refactoring, I missed the quotes and broke the resulting SQL.

Doesn't this affects this line as well?
https://github.com/phpmyadmin/phpmyadmin/blob/ed8b58838245ba47f5f37c66d4974ef59f1b99f1/src/Server/Privileges.php#L3126-L3127

kamil-tekiela:
I don't have such an old MySQL server available to test, but presumably, yes. I have also changed this line now.

williamdes:
> I don't have such an old MySQL server available to test, but presumably, yes. I have also changed this line now.

Do you use Docker ?
There is containers for quite every version you could think of
What version do you need ?

kamil-tekiela:
> > I don't have such an old MySQL server available to test, but presumably, yes. I have also changed this line now.
> 
> Do you use Docker ? There is containers for quite every version you could think of What version do you need ?

I do not. :) I know, I know... but I like my setup and docker is annoying me. 

I think this requires MySQL 5.6 or older to test.

williamdes:
Since you are doing important work on phpMyAdmin, I can grant you a VM on my servers in an Europe datacenter.
@MauricioFauth you too

Email me if you want a VM for testing PHP/DB stuff.

MauricioFauth:
> Since you are doing important work on phpMyAdmin, I can grant you a VM on my servers in an Europe datacenter.
> @MauricioFauth you too
> 
> Email me if you want a VM for testing PHP/DB stuff.

Thank you for this. I'll send you a message.

MauricioFauth (approved):

PR #18913, opened 2024-01-13
Fix bug with numeric table names

kamil-tekiela:
The bug happens in 5.2 also but it only became apparent after my recent refactoring. 

To trigger this bug:
1. DisableIS = true
2. Create a DB with two tables: `0`, `42`. 
3. Open `/database/structure` and see PMA crash. 

The bug happens because of a little-known quirk of `array_multisort` that renames the numerical keys.

### DIFF
--- a/libraries/classes/DatabaseInterface.php
+++ b/libraries/classes/DatabaseInterface.php
@@ -25,6 +25,7 @@ use PhpMyAdmin\Utils\SessionCache;
 
 use function __;
 use function array_column;
+use function array_combine;
 use function array_diff;
 use function array_keys;
 use function array_map;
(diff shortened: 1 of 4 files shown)


kamil-tekiela:
Why is CI not starting? Did I bork my PR?

williamdes:
> Why is CI not starting? Did I bork my PR?

Weird target changing stuff. Sometime re-open helps

williamdes (approved):
Can we have a test for this special hack ?

kamil-tekiela:
Maybe. I need to think about how to add a test.

MauricioFauth (approved):

PR #18929, opened 2024-01-23
Fix navigation page after table edit

### LINKED ISSUE #18926: When a query is edited, it redirects to SQL tab (MariaDB)
### Describe the bug

When a query is edited, it redirects to SQL tab instead of table. This happens only for MariaDB. On 6.0.0-dev is even worse, won't show the query used on edit.

### To Reproduce

1. Go to https://demo.phpmyadmin.net/QA_5_2/index.php?route=/sql&db=sakila&table=actor&pos=0&server=1
2. Click on `Edit`
3. Clikc on `Go`
4. See redirecting to SQL tab.

### Expected behavior

It should have the same behaviour like MySQL on 5.2.2-dev and 6.0.0-dev.

MySQL (working):

https://github.com/phpmyadmin/phpmyadmin/assets/25424343/ed502224-360b-4d07-b9b3-997f4b8e4cb8

### Screenshots

MariaDB (not working):

https://github.com/phpmyadmin/phpmyadmin/assets/25424343/1c036f19-ff6b-4967-b608-0158e17fe1db

### Server configuration

- Database version: MariaDB

kamil-tekiela:
Fixes #18926

### DIFF
--- a/libraries/classes/InsertEdit.php
+++ b/libraries/classes/InsertEdit.php
@@ -1245,7 +1245,7 @@ class InsertEdit
             if (! preg_match('@^[a-z_]+\.php$@', $GLOBALS['goto'])) {
                 // this should NOT happen
                 //$GLOBALS['goto'] = false;
-                if ($GLOBALS['goto'] === 'index.php?route=/sql') {
+                if (str_contains($GLOBALS['goto'], 'index.php?route=/sql')) {
                     $gotoInclude = '/sql';
                 } else {
                     $gotoInclude = false;
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -3056,14 +3056,9 @@
     </TypeDoesNotContainType>
   </file>
   <file src="libraries/classes/Controllers/Table/ChartController.php">
-    <InvalidArgument occurrences="2">
-      <code>$rows</code>
+    <InvalidArgument occurrences="1">
       <code>$start</code>
     </InvalidArgument>
-    <InvalidScalarArgument occurrences="2">
-      <code>$_REQUEST['pos']</code>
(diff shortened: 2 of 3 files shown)


libraries/classes/InsertEdit.php
```diff
@@ -1245,7 +1245,7 @@ public function getGotoInclude($gotoInclude): string
             if (! preg_match('@^[a-z_]+\.php$@', $GLOBALS['goto'])) {
                 // this should NOT happen
                 //$GLOBALS['goto'] = false;
-                if ($GLOBALS['goto'] === 'index.php?route=/sql') {
+                if (str_contains($GLOBALS['goto'], 'index.php?route=/sql')) {
```
liviuconcioiu, line 1248: This could be an issue for 5.2.2-dev. Since `str_contains` is not available on PHP 7.2.

kamil-tekiela: We use the symfony/polyfill so that's not a problem. Besides, this function is already in use in multiple places in this file.

kamil-tekiela: @MauricioFauth  Now that I look at composer.json, I don't see `symfony/polyfill-php80` required by PMA. It's implicitly required by one of the dependencies, but not by the project. Should this be rectified?

MauricioFauth (approved):

libraries/classes/InsertEdit.php
williamdes, line 1248: it is more str_starts_with?

kamil-tekiela: I'm not sure. I don't know if it matters. I suppose you could change this as the start of the string should be the same every time, but why risk it.

PR #18935, opened 2024-01-24
Type hint and clean up $selected in controllers

kamil-tekiela:


MauricioFauth (approved):

src/Controllers/Database/Structure/AddPrefixTableController.php
```diff
@@ -29,15 +27,14 @@ public function __construct(
 
     public function __invoke(ServerRequest $request): void
     {
+        /** @var string[] $selected */
         $selected = $request->getParsedBodyParam('selected', []);
```
MauricioFauth, line 31: In future, it will be better to actually assert that when the value is from the request.

PR #18952, opened 2024-01-30
Implement natural sort for items in the navigation section

### LINKED ISSUE #18949: `$cfg['NaturalOrder']` is ignored on left sidebar
### Describe the bug

`$cfg['NaturalOrder']` is ignored on left sidebar.

### To Reproduce

1. Set `$cfg['NaturalOrder']` to `true`.
2. Navigate to a database with multiple tables with numbers.
3. See the difference on the sidebar and on database.

### Expected behavior

Left sidebar should respect the `$cfg['NaturalOrder']` value.

### Screenshots

order

### Server configuration

kamil-tekiela:
Fixes #18949

This might impact the performance.

### DIFF
--- a/libraries/classes/Navigation/Nodes/NodeDatabase.php
+++ b/libraries/classes/Navigation/Nodes/NodeDatabase.php
@@ -12,9 +12,10 @@ use PhpMyAdmin\Url;
 use PhpMyAdmin\Util;
 
 use function __;
+use function array_slice;
 use function in_array;
-use function intval;
 use function substr;
+use function usort;
 
 /**
  * Represents a database node in the navigation tree
@@ -439,7 +440,6 @@ class NodeDatabase extends Node
             }
 
             $query .= 'ORDER BY `TABLE_NAME` ASC ';
-            $query .= 'LIMIT ' . $pos . ', ' . $maxItems;
             $retval = $dbi->fetchResult($query);
         } else {
             $query = ' SHOW FULL TABLES FROM ';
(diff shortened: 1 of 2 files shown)


MauricioFauth:
MariaDB has the [NATURAL_SORT_KEY](https://mariadb.com/kb/en/natural_sort_key/) function.

libraries/classes/Navigation/Nodes/NodeDatabase.php
```diff
@@ -439,7 +440,6 @@ private function getTablesOrViews($which, int $pos, $searchClause)
             }
 
             $query .= 'ORDER BY `TABLE_NAME` ASC ';
```
MauricioFauth, line 442: Maybe using something like this for MariaDB 10.7+ will help with the performance.
```suggestion
            $query .= 'ORDER BY NATURAL_SORT_KEY(`TABLE_NAME`) ASC ';
```

kamil-tekiela: This is an interesting idea, but I think it should be part of a separate PR. It might even only be suitable for master branch. It requires more investigation and more code.

I did a quick test and I only observed 1-2s difference with 12k of tables. That's not terrible, but also not great. 

One more idea I had is to make the optimization conditional, so that people who do not use natural sort do not suffer the performance drawback. I will try to come up with some solution to this tomorrow.

kamil-tekiela: Ok, so my conclusion is that we don't do anything here. Let's not add any optimizations here. The difference is negligible. The query itself takes 1.5s-1.7s on my machine with 12k tables. It doesn't matter how many rows it returns. Technically on a slow connection this could be slightly worse, but it's too rare circumstance to worry prematurely about it. 

My findings so far are that the `usort` adds a significant overhead of about 1s, but given that the alternative is limited to MariaDB 10.7, we wouldn't be able to completely get rid of `usort`. Also, I use xDebug so it probably impacts the measurement too. 

The whole thing is slow because of redundant `count`. I think I cut that down in half on master if I am not mistaken, but this could probably be reduced to zero. That's something that can and should be refactored. 

I think this PR can be merged as is. Yes, it slightly worsens the performance, but not by that much.

MauricioFauth (approved):

PR #18956, opened 2024-01-30
Fix creation of mocks in unit tests

kamil-tekiela:
They are technically identical. The only difference is that `createMock` registers the created mock, but I don't know what exactly this means. This change helps with code navigation in IDEs. 
s/createStub/createMock/

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -16406,21 +16406,6 @@ parameters:
 			path: tests/classes/Controllers/Server/VariablesControllerTest.php
 
 		-
-			message: "#^Call to an undefined method PhpMyAdmin\\\\DatabaseInterface\\:\\:expects\\(\\)\\.$#"
-			count: 4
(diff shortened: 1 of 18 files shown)


MauricioFauth:
They maybe technically similar internally, but they are different semantically. Stubs are tests doubles that doesn't have any expectations, they are only used as a replacement. Mocks are like stubs, but expectations are defined before the code is executed.

kamil-tekiela:
You are right. Let's keep this semantic distinction. And if, in the future, PHPUnit changes the behaviour, we should be ready. 

However, I would like to repurpose this PR to fix the creation of mocks.

MauricioFauth:
> And if, in the future, PHPUnit changes the behaviour, we should be ready.

As this is not a PHPUnit specific terminology, I doubt that this will change any time soon.

MauricioFauth (approved):

PR #18963, opened 2024-02-01
Optimize navigation

kamil-tekiela:
This PR combines fetching count and records into a single operation and performs a single read from DB for tables and views. 

On my machine with 12k tables, the response time dropped from 5.3s to 2.7s, as measured by xDebug. 

Without xDebug, times measured in the browser:

| DisableIS |  before | after |
| ------------- |:-------------:| -----:|
| true    | 4.9s  | 1.8s |
| false | 4.9s | 1.7s |

Times include sorting with natSort, which adds about 150ms. With xDebug enabled, sorting adds up to 1s.

---

Reading the table `INFORMATION_SCHEMA.TABLES` takes about 1.5s. `SHOW FULL TABLES` is just a simple wrapper for `SELECT` from IS. This table has 2 columns that we want, the first one is the table name and the second is the type. We need to know the type to distinguish between view and normal table. MySQL will take the same amount of time to read from the table whether we want views, tables, or both. It will also take 1.5s to get the count of matching rows. 

The optimization here avoid doing the same thing 3 or 4 times*. Instead of reading the count, views, and tables separately, it makes a single call to DB to read everything into memory. We can filter the results in PHP and then use `count()` to count the records. It feels like this should be worse for performance, but it isn't. Memory is cheap and filling memory is quick. Since the fix I introduced on 5.2, we already read everything into memory anyway. So it was just a matter of reorganizing the code. This means that instead of spending ~4.5s (or 6s) reading from DB, the code spends only ~1.5s. 

\* I tested with a DB with ~12k table but no views. The code read 3 times from DB, but if I had even just a single view it would read 4 times. 

I still need to add unit tests for the new class.

### DIFF
--- a/src/Dbal/MysqliResult.php
+++ b/src/Dbal/MysqliResult.php
@@ -19,6 +19,7 @@ use function is_bool;
 use function is_string;
 
 use const MYSQLI_ASSOC;
+use const MYSQLI_NUM;
 
 /**
  * Extension independent database result
@@ -121,12 +122,12 @@ final class MysqliResult implements ResultInterface
     }
 
     /**
-     * Returns values from the first column of each row
+     * Returns values from the selected column of each row
      *
      * @return array<int, string|null>
      * @psalm-return list<string|null>
      */
-    public function fetchAllColumn(): array
+    public function fetchAllColumn(int|string $column = 0): array
     {
         if ($this->result === null) {
             return [];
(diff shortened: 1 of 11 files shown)


src/Dbal/MysqliResult.php
```diff
@@ -135,7 +136,13 @@ public function fetchAllColumn(): array
         // This function should return all rows, not only the remaining rows
         $this->result->data_seek(0);
 
-        return array_column($this->result->fetch_all(), 0);
+        if (is_string($column)) {
```
williamdes, line 139: you could one line this with a ternary on the flag

kamil-tekiela: I can't put the ternary on the flag because that would confuse Psalm. But I did combine this into a single line.

liviuconcioiu:
Awesome work! On my side, from 1.4s to 0.7s.

MauricioFauth (approved):

PR #18985, opened 2024-02-10
Refactor import

kamil-tekiela:
This is probably just part 1. This is such a big mess. The unit tests are passing locally, but I have not confirmed manually that any of these imports still work. If anyone wants to check them once, I would be grateful.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -11696,11 +11696,6 @@ parameters:
 			path: src/Plugins/Import/ImportOds.php
 
 		-
-			message: "#^Cannot access property \\$spreadsheet on mixed\\.$#"
-			count: 1
-			path: src/Plugins/Import/ImportOds.php
-
-		-
 			message: "#^Cannot call method __toString\\(\\) on mixed\\.$#"
 			count: 1
 			path: src/Plugins/Import/ImportOds.php
@@ -11713,11 +11708,6 @@ parameters:
 		-
 			message: "#^Cannot cast mixed to string\\.$#"
 			count: 3
-			path: src/Plugins/Import/ImportOds.php
-
-		-
-			message: "#^Construct empty\\(\\) is not allowed\\. Use more strict comparison\\.$#"
-			count: 1
 			path: src/Plugins/Import/ImportOds.php
 
 		-
@@ -11766,18 +11756,8 @@ parameters:
 			path: src/Plugins/Import/ImportOds.php
 
 		-
-			message: "#^Variable property access on \\(SimpleXMLElement\\|null\\)\\.$#"
-			count: 1
-			path: src/Plugins/Import/ImportOds.php
-
-		-
 			message: "#^Variable property access on SimpleXMLElement\\.$#"
 			count: 2
-			path: src/Plugins/Import/ImportOds.php
-
-		-
-			message: "#^Variable property access on mixed\\.$#"
-			count: 1
 			path: src/Plugins/Import/ImportOds.php
 
 		-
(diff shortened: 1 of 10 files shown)


liviuconcioiu:
Export/Import table with one column being a `binary(16)` - there's an issue with exporting binary that stores a hash for example (in .sql works):

Test table:
[actor.zip](https://github.com/phpmyadmin/phpmyadmin/files/14228660/actor.zip)

SQL:
`0xf080c845af056c10901a199450965e83`

Other formats:
`ð€ÈE¯l”P–^ƒ`

CSV:
1. Instead of `actor` table name, `actor_1` is created. I had `actor1`, `actor2` tables before importing. Doesn't happen with ODS/XML.
2. I get on another CSV file with `binary(16)` column - `Warning: #1366 Incorrect string value: \xF3\xA8\x96\xAE`
3. Even if `NULL` is specified, is ignored - empty values are created instead of `NULL` - check `first_name` column.

ODS:
1. `The XML file specified was either malformed or incomplete. Please correct the issue and try again.`

XML:
1. `The XML file specified was either malformed or incomplete. Please correct the issue and try again.`

MediaWiki:
1. Even if `NULL` is specified, is ignored - empty values are created instead of `NULL`. - check `first_name` column.

ShapeFile:
1. `table x` - x being the number of last table is created instead of the filename being imported? If `test.shp` is imported, shouldn't it be `test` table?
Test table:
[cs34_d00.zip](https://github.com/phpmyadmin/phpmyadmin/files/14228663/cs34_d00.zip)

Ideally, fixes should also be added to 5.2.2-dev.

Not related to this PR that I wanted to report a while ago:
Test table 1:
[vehicles.zip](https://github.com/phpmyadmin/phpmyadmin/files/14228665/vehicles.zip)

1. Partial import of CSV doesn't work (will log out - 30 seconds script execution).
3. Importing gives some warnings (`Warning: #1264 Out of range value for column`).
4. I see only `int` and `varchar` columns being created. At least where the value is float it could be a `float` with high precision?
5. Columns are using `DEFAULT NULL`, but when there's no data an empty value is added instead of `NULL` (NULL isn't specified in the file).

Test table 2:
[tires.zip](https://github.com/phpmyadmin/phpmyadmin/files/14228734/tires.zip)

1. When columns count from header doesn't match with count from rows, it gives `Warning: #1406 Data too long for column`.
2. Strangely some rows have `NULL` values.

I've noticed that when a .sql.zip is imported it takes 3 times more than .sql. This happens because of `mb_strcut`.

https://github.com/phpmyadmin/phpmyadmin/blob/be3c06c7dc8fb0de43dd45345d6bd1fdfd3b5ac0/src/File.php#L677

Is there a way to avoid it and speed up the import of zip? Only in PHP 8.4 or PHP 8.5 will be possible to read zip from memory (there's an open issue about this).

kamil-tekiela:
@liviuconcioiu  Thank you. I will look through all of them carefully. Are all of them because of my PR? I looked at the name generation and I think it was like this for some time already. I fixed and refactor it anyway, but I don't know if I should add it to this PR.

liviuconcioiu:
> @liviuconcioiu Thank you. I will look through all of them carefully. Are all of them because of my PR? I looked at the name generation and I think it was like this for some time already. I fixed and refactor it anyway, but I don't know if I should add it to this PR.

I haven't tested with your PR. just with latest 6.0.0-dev, and I think the same issues are in 5.2.2-dev.

kamil-tekiela:
Ahh ok. Then I will address them in next PRs. This one I feel is already big enough.

MauricioFauth:
@liviuconcioiu what PHP version you are using?

liviuconcioiu:
> @liviuconcioiu what PHP version you are using?

8.3.2

kamil-tekiela:
What I have observed is that XML import is completely broken on 6.0.0. I have tested by exporting a table and then importing it without changing any other settings.

MauricioFauth (approved):
:100:

PR #18987, opened 2024-02-12
Inject Config instead of using Config::getInstance()

MauricioFauth:


src/Header.php
```diff
@@ -79,18 +79,19 @@ class Header
 
     private bool $isTransformationWrapper = false;
 
-    /**
-     * Creates a new class instance
```
kamil-tekiela, line 83: That was such an awesome comment. :D

kamil-tekiela (approved):

src/Header.php
MauricioFauth: A long time ago, doc blocks were required for all functions (mainly for type hints), that's why we still have such comments here and there. haha

PR #18990, opened 2024-02-14
Remove dbi object from table/relation/common_form.twig

MauricioFauth:


src/Controllers/Table/RelationController.php
```diff
@@ -241,6 +238,54 @@ public function __invoke(ServerRequest $request): void
...
+            // Database dropdown
+            if (isset($relations[$column->field])) {
+                $foreignDb = $relations[$column->field]['foreign_db'];
+            } else {
+                $foreignDb = Current::$database;
+            }
+
+            // Table dropdown
+            $foreignTables = [];
+            if ($foreignDb) {
```
kamil-tekiela, line 259: Shouldn't this be `!== ''` to avoid a bug with a DB called `0`?

src/Controllers/Table/RelationController.php
```diff
@@ -241,6 +238,54 @@ public function __invoke(ServerRequest $request): void
...
+                if (isset($relations[$column->field])) {
+                    $foreignTable = $relations[$column->field]['foreign_table'];
+                }
+
+                $foreignTables = $this->dbi->getTables($foreignDb);
+            }
+
+            // Column dropdown
+            $uniqueColumns = [];
+            if ($foreignDb && $foreignTable) {
```
kamil-tekiela, line 269: Same here, no? I guess it would only require changing `$foreignTable = ''` or can the value from the array be also something other than a string?

PR #18993, opened 2024-02-16
Make PHPUnit's assertions stricter

MauricioFauth:
Only replaces assertEquals with assertSame where it doesn't causes the test to fail.


williamdes (approved):
Very good idea

PR #19038, opened 2024-03-08
Add tests for app/services* files

MauricioFauth:
Tests each service if it's possible to create it.

- Removes some initialization methods from constructor
- Fixes error introduced by https://github.com/phpmyadmin/phpmyadmin/pull/18625

### DIFF
--- a/app/services_controllers.php
+++ b/app/services_controllers.php
@@ -1252,7 +1252,7 @@ return [
             'arguments' => [
                 '$response' => '@response',
                 '$template' => '@template',
-                '$maintenance' => '@partitioning_maintenance',
+                '$model' => '@partitioning_maintenance',
             ],
         ],
         Table\Partition\CheckController::class => [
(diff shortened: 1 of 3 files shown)


src/Controllers/Table/FindReplaceController.php
```diff
@@ -132,7 +130,7 @@ public function __invoke(ServerRequest $request): void
     /**
      * Gets all the columns of a table along with their types.
      */
-    private function loadTableInfo(): void
+    public function loadTableInfo(): void
```
kamil-tekiela, line 133: Why does this need to be `public`? Is it used in tests?

MauricioFauth: Yes, there is a test for another method that also should be public, so I changed it to public for simplicity of this PR.

kamil-tekiela: Forgive me for asking, but I don't think I understand how this change is related to this PR. At the moment the method is private in all 3 controllers that call it from the constructor. 
image showing all 3 controllers

I understand if you want to move the initialization out of the constructor; it makes total sense and I am on board with it. But I would like to understand why this change is needed for this PR and why only one of these methods needed to be made public.

MauricioFauth: This is needed because, otherwise, the test will need a test setup just for these controllers. But, yeah. You're right. It's better to properly do that in a different PR.

kamil-tekiela (approved):

PR #19050, opened 2024-03-13
Refactor errors

kamil-tekiela:
This PR does the following things:
- replaces `trigger_error` that has `E_USER_ERROR` as a second argument with an exception
- removed unreachable statements
- fixes a potential crash in error handler
- refactors theme loading to avoid 2 additional warnings on top of the missing theme exception. If the path cannot be opened PHP will throw one of standard exceptions instead.


MauricioFauth (approved):
:100:

PR #19053, opened 2024-03-14
Refactor instantiation of Import plugin classes

kamil-tekiela:
I think the idea was to load the plugins based on the files in the directory or something like that. I am not sure why the code was so dynamic. But since the values were hardcoded anyway, it makes sense to just list all values in a single enum.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -2746,6 +2746,11 @@ parameters:
 			path: src/Controllers/Import/ImportController.php
 
 		-
+			message: "#^Static property PhpMyAdmin\\\\Import\\\\ImportSettings\\:\\:\\$importType \\(''\\|'database'\\|'query'\\|'server'\\|'table'\\) does not accept string\\.$#"
+			count: 1
+			path: src/Controllers/Import/ImportController.php
+
+		-
 			message: "#^Static property PhpMyAdmin\\\\Import\\\\ImportSettings\\:\\:\\$readLimit \\(int\\) does not accept float\\|int\\<min, \\-1\\>\\|int\\<1, max\\>\\.$#"
 			count: 1
 			path: src/Controllers/Import/ImportController.php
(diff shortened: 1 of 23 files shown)


MauricioFauth:
It's dynamic because the idea is to allow custom plugins to be installed. However I don't know why the format are hard coded. Seems like a bug to me. So an enumeration is probably not a good idea in this case.

kamil-tekiela:
How are the plugins installed? I cannot find any guide or documentation for this.

MauricioFauth:
> How are the plugins installed? I cannot find any guide or documentation for this.

The current way is not very good. You have to create a new class at the same directory of the phpMyAdmin plugins and use the PhpMyAdmin namespace. That is why the plugin system has some includes here and there, so a custom plugin could be loaded. This system was created before Composer was introduced.
A new plugin should definitely use Composer to load the plugins and the directory and namespace shouldn't matter. 

I thought about introducing a config directive that contains a allow list of plugins.

kamil-tekiela:
The current system is definitely not good, especially since it relies on a lot of globals. 

The 7 import plugins were hardcoded since forever. I assumed that someone could modify PMA source code to add new plugins, but they would need to add their new plugin to the hardcoded list anyway. This is why I decided upon an enum with a list of class names. My approach makes it slightly easier to add a new import plugin. You can put the plugin in any namespace you want and all you have to do is modify that single enum file. 

But with all the changes I already made and am planning to make in future, keeping the plugins up to date is already a problem for whoever is doing it. Not to mention that during an upgrade they would need to redo the modifications to the source code and adapt to the new directory structure. Modifying PMA source code to add extra functionality should not be encouraged. Loading plugins via composer would be the best idea but I don't know how to do that.

MauricioFauth (approved):

MauricioFauth:
I'm gonna think about how to properly do that. Anyway, this PR is better than the current code.

PR #19054, opened 2024-03-15
Remove register globals from import and privileges

kamil-tekiela:
I can't think of any better way to do it right now.


MauricioFauth (approved):
That's great!

MauricioFauth (approved):

PR #19068, opened 2024-03-20
Refactor Console settings handling

MauricioFauth:


src/Controllers/Console/UpdateConfigController.php
```diff
@@ -0,0 +1,56 @@
...
+        $key = $request->getParsedBodyParam('key');
+        $value = $request->getParsedBodyParam('value');
+        if (! in_array($key, $validKeys, true) || ! is_string($value)) {
+            $this->response->setRequestStatus(false);
+            $this->response->addJSON(['message' => Message::error()]);
+
+            return;
+        }
+
+        $result = $this->config->setUserValue(null, 'Console/' . $key, json_decode($value));
```
williamdes, line 48: should we use the new json validate?
or try catch decode before trying to save a broken value?

MauricioFauth: That's a good idea.

PR #19072, opened 2024-03-21
Remove $GLOBALS['allowDeny_forbidden']

kamil-tekiela:
We never set this global to any truthy value. If it was meant to serve as some sort of hook then it should be implemented properly not as a global.

### DIFF
--- a/src/Plugins/Auth/AuthenticationConfig.php
+++ b/src/Plugins/Auth/AuthenticationConfig.php
@@ -21,7 +21,6 @@ use function count;
 use function sprintf;
 use function trigger_error;
 
-use const E_USER_NOTICE;
 use const E_USER_WARNING;
 
 /**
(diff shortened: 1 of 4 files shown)


MauricioFauth:
Looks like a bug introduced by 2327dbf3c3c109b907ea478be098c77e80e1bf0b.
After a quick look at the code, it seems like it should be true only when the parameter `$failure` is `'allow-denied'`.

src/Plugins/Auth/AuthenticationConfig.php
```diff
@@ -95,47 +94,44 @@ public function showFailure(string $failure): never
         <tr>
             <td>';
         $config = Config::getInstance();
-        if (isset($GLOBALS['allowDeny_forbidden']) && $GLOBALS['allowDeny_forbidden']) {
```
MauricioFauth, line 98:
```diff
- if (isset($GLOBALS['allowDeny_forbidden']) && $GLOBALS['allowDeny_forbidden']) {
+ if ($failure === 'allow-denied') {
```

MauricioFauth: I think this fix should target 5.2.

kamil-tekiela: Would you like to do it? You can close my PR.

PR #19074, opened 2024-03-21
Fixing PHPUnit performance

kamil-tekiela:
I am trying to investigate why PHPUnit tests are so slow. It seems like the biggest cause is the separation of certain tests due to singletons/globals. I am opening this PR to test if we can remove some of the separation.


kamil-tekiela:
@MauricioFauth What do you think? Do we still need this test separation?

MauricioFauth:
> @MauricioFauth What do you think? Do we still need this test separation?

As the tests are ran in random order, sometimes they fail. Most of these were added as a quick fix. I agree to remove most of them. It's better to refactor the test or the tested code to avoid using the test isolation.

MauricioFauth (approved):

kamil-tekiela:
I have rerun the tests locally multiple times and I did not get any failures.

PR #19076, opened 2024-03-22
Remove $GLOBALS['no_constraints_comments']

kamil-tekiela:
There doesn't seem to be any way to enable it.


kamil-tekiela:
I just figured out what this was supposed to be.

Issue #19075, opened 2024-03-21
Language combobox is hard to read


M393:
### Is your feature request related to a problem? Please describe.

The languages in the combobox are sorted by the English language name, but this name is not horizontally aligned with the other languages. This makes it difficult to find a language.

### Describe the solution you'd like

Sort the combobox by the native language name. This name s what I'm looking for, the name of my language in my language.

![image](https://github.com/phpmyadmin/phpmyadmin/assets/56256405/a53cd167-e2be-4909-aa95-06bf3defde86)

### Describe alternatives you've considered

Put the English language name first.
![image](https://github.com/phpmyadmin/phpmyadmin/assets/56256405/6cc1cef6-2957-4fa3-91e1-82ca2938c92c)

### Additional context

- phpMyAdmin version: 5.2.1 / master
Current state:
![image](https://github.com/phpmyadmin/phpmyadmin/assets/56256405/bb33c107-a167-4123-b4ca-bd62845286a1)

anujeshify:
I wish to work on this issue.

M393:
The necessary change is quite small. And I already implemented it locally. (That's how I made the screenshots.)
Just waiting for someone to say they welcome this change.

anujeshify:
> The necessary change is quite small. And I already implemented it locally. (That's how I made the screenshots.) Just waiting for someone to say they welcome this change.

Thank you for implementing the change locally and for providing screenshots. I'm glad to hear that the necessary modification is relatively small. I've already submitted a pull request addressing the issue based on our discussion. You can find it [here.](https://github.com/phpmyadmin/phpmyadmin/pull/19078)

PR #19078, opened 2024-03-22
#19075 - Sort the combobox by native language name

anujeshify:
### Description

This pull request addresses issue #19075.
The language combobox was challenging to read due to the ordering of languages. To enhance usability, I have restructured the order of languages to align with their English names. This change ensures that languages are sorted more intuitively, making them easier to locate and facilitating a coherent viewing experience.

Changes Made
In src/Language.php, the following modifications were made:
```
public function getName(): string {
    if ($this->native !== '') {
        return $this->name . ' - ' . $this->native;
    }

    return $this->name;
}
```

Fixes
This pull request fixes the readability issue in the language combobox.

Signed-off-by: Anujesh Bansal <anujeshbansaltoo@gmail.com>

### DIFF
--- a/src/Language.php
+++ b/src/Language.php
@@ -69,7 +69,7 @@ class Language
     public function getName(): string
     {
         if ($this->native !== '') {
-            return $this->native . ' - ' . $this->name;
+            return $this->name . ' - ' . $this->native;
         }
 
         return $this->name;


M393:
The two suggested alternatives from #19075 were
1. Sort by the native language (my preferation)
2. Put the English language name first

The title of this pull request is from alternative 1, but it actually implements alternative 2.

Looks like you selected the wrong branch as the merge target, should be `master` instead of `QA_5_2`

PR #19094, opened 2024-04-03
Display correct effective size

### LINKED ISSUE #18771: Space usage information seems wrong
### Describe the bug

The effective byte value of a table is negative after deleting a lot of rows. The table originally had ~10M rows, after deleting ~7.5M rows the `Effective` value is negative. 

The .ibd file for this table is 10.764.288 KB

### Expected behavior

Not sure what this value is supposed to be. Also the formatting is in bytes instead of GB, likely because the value is negative.

### Screenshots
phpmyadmin:
image
mysql workbench:
image

### Server configuration

- Operating system: Windows
- Database version: MySQL 5.7.42
- PHP version: 8.1
- phpMyAdmin version: 5.2

### Client configuration

- Operating system: Windows

M393:
Fixes #18771

`Total` is the space needed for index and data, `Effective` is the file size on disk, which includes `'Data_free'`, so it should be added instead of subtracted.

### DIFF
--- a/libraries/classes/Controllers/Table/StructureController.php
+++ b/libraries/classes/Controllers/Table/StructureController.php
@@ -321,17 +321,17 @@ class StructureController extends AbstractController
 
         if (isset($showtable['Data_free'])) {
             [$free_size, $free_unit] = Util::formatByteDown($showtable['Data_free'], $max_digits, $decimals);
-            [$effect_size, $effect_unit] = Util::formatByteDown(
-                $showtable['Data_length']
-                + $showtable['Index_length']
-                - $showtable['Data_free'],
-                $max_digits,
-                $decimals
-            );
+            $effective = $showtable['Data_length'] + $showtable['Index_length'];
+            if ($tbl_storage_engine === 'INNODB') {
+                $effective += $showtable['Data_free'];
+            } else {
+                $effective -= $showtable['Data_free'];
+            }
+
+            [$effect_size, $effect_unit] = Util::formatByteDown($effective, $max_digits, $decimals);
         } else {
             [$effect_size, $effect_unit] = Util::formatByteDown(
-                $showtable['Data_length']
-                + $showtable['Index_length'],
+                $showtable['Data_length'] + $showtable['Index_length'],
                 $max_digits,
                 $decimals
             );
@@ -347,9 +347,7 @@ class StructureController extends AbstractController
         $avg_unit = '';
         if ($table_info_num_rows > 0) {
             [$avg_size, $avg_unit] = Util::formatByteDown(
-                ($showtable['Data_length']
-                + $showtable['Index_length'])
-                / $showtable['Rows'],
+                ($showtable['Data_length'] + $showtable['Index_length']) / $showtable['Rows'],
                 6,
                 1
             );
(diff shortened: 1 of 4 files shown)


libraries/classes/Controllers/Table/StructureController.php
```diff
@@ -322,16 +322,13 @@ protected function getTableStats(bool $isSystemSchema)
         if (isset($showtable['Data_free'])) {
             [$free_size, $free_unit] = Util::formatByteDown($showtable['Data_free'], $max_digits, $decimals);
             [$effect_size, $effect_unit] = Util::formatByteDown(
-                $showtable['Data_length']
-                + $showtable['Index_length']
-                - $showtable['Data_free'],
+                $showtable['Data_length'] + $showtable['Index_length'] + $showtable['Data_free'],
```
williamdes, line 325: a 20+ years bug ?
Seems to have been added in https://github.com/phpmyadmin/phpmyadmin/commit/84485775aa37612ba2bff73e5490648fa8501af1#diff-83e3a89d98949a8b6e583e77d428adf8b1f0d8b54abba4afaa97728b5f7b8edbR275

williamdes: I still can not find a conclusion using external sources.
See: https://tableplus.com/blog/2018/04/mysql-get-size-of-tables.html or https://www.a2hosting.com/kb/developer-corner/mysql/determining-the-size-of-mysql-databases-and-tables/

Do you have an idea on how to verify the new total is the right one ?

liviuconcioiu: @williamdes not sure if is related or not, but I've encountered many times a strange issue like, the total size of a db or table had a different size after duplicating it from phpMyAdmin.

MoonE: The effective size should match the .ibd file size.
> The .ibd file for this table is 10.764.288 KB

From https://github.com/phpmyadmin/phpmyadmin/issues/18771#issue-1954361499 roughly matches the sum of  Overhead and Total `(3.8 + 6.6) * 1024 * 1024` = 10,905,190.4 KB

M393: @williamdes You could create a table and insert a lot of data. Repeatedly check that the effective size matches the file size on disk after deleting some of the data.

MauricioFauth: All right! I was able to reproduce this and I understand now why this was implemented this way.
When this was implemented, Innodb wasn't really a thing on MySQL yet, so this logic is for MyISAM engine and it works fine with it. So this change breaks for MyISAM engine.
However, for Innodb, I don't know what to do. I think this "effective size" doesn't makes much sense. Because it's only a rough estimate.

MauricioFauth (changes requested):

M393:
> this logic is for MyISAM engine and it works fine with it.

So we need to test all different storage engines, to find out how they behave?

>I think this "effective size" doesn't makes much sense. Because it's only a rough estimate.

Even as an estimate it can be useful, so a table can be optimized when there is a lot of overhead.

PR #19104, opened 2024-04-08
Remove the Controllers\AbstractController abstract class

MauricioFauth:
This removes the hard dependency on `ResponseRenderer` and `Template` from the controllers.

The properties of the `Controllers\AbstractController` are not required by all controllers, and its methods can be moved to the `ResponseRenderer` class. These methods should be refactored to avoid using the `ResponseRenderer` class.


kamil-tekiela (approved):
Awesome

PR #19111, opened 2024-04-14
Extract recent table handling from Header into a middleware

MauricioFauth:
This removes the recent visited table handling from the `Header` class and moves it to a new middleware.

This also removes a duplicated update via front-end.


kamil-tekiela (approved):
I love it.

Issue #19112, opened 2024-04-16
Twig 3.9.0 does not work correctly with QA_5_2


MoonE:
### Describe the bug

The recently released Twig version 3.9.0 has some issues with the phpmyadmin translate tags. It doesn't strip the `notes` from rendering.
https://github.com/phpmyadmin/phpmyadmin/blob/ec78dfd088b13a8ab84829f2bfc4115f69338305/templates/javascript/variables.twig#L15
### To Reproduce

1. Run composer install on the QA_5_2 branch
2. Go the login page
3. The form is missing, only the logo and heading visible

### Expected behavior

The page should allow to login.

### Screenshots
![image](https://github.com/phpmyadmin/phpmyadmin/assets/4426597/4167a089-e93c-4f63-9b35-73522ab6fea8)

### Server configuration

- Operating system: Linux 
- Web server: apache 
- PHP version: 8.2.10-2ubuntu1
- phpMyAdmin version: QA_5_2

### Client configuration

- Browser: firefox
- Operating system: Linux

### Additional context

Twig version 3.8.0 works as expected.

Please consider adding composer.lock file also in the QA_5_2 branch to prevent random breakage like this.

williamdes:
@MauricioFauth okay for the lock file, could you check for a fix please?

AdamReece-WebBox:
This happens with QA_5_1 too. I've not tried QA_5_0 as this requires also going way back to PHP 7.1.

williamdes:
I pushed add463dfe0852e400dfddc5c82b7f64ef961bc30 to avoid the upgrade until some fix is done

MauricioFauth:
> I pushed [add463d](https://github.com/phpmyadmin/phpmyadmin/commit/add463dfe0852e400dfddc5c82b7f64ef961bc30) to avoid the upgrade until some fix is done

Thanks! I was doing that right now as well. :smile:

williamdes:
Added also in Debian: https://salsa.debian.org/phpmyadmin-team/phpmyadmin/-/commit/9427e6cd70e2cd29d5b37042856f576604dcb9ff

AdamReece-WebBox:
> I pushed [add463d](https://github.com/phpmyadmin/phpmyadmin/commit/add463dfe0852e400dfddc5c82b7f64ef961bc30) to avoid the upgrade until some fix is done

Oh cool I forgot about a ranged rule. I was just changing it to be clamped at "3.8.0" our end. :)

[...]

PR #19116, opened 2024-04-17
Disable Twig's text nodes optimization

MauricioFauth:
The text nodes optimization added in Twig 3.9.0 breaks the i18n extension. So let's disable it until a fix for the extension itself is created.

- Fixes https://github.com/phpmyadmin/phpmyadmin/issues/19112

### DIFF
--- a/.github/workflows/lint-and-analyse-php.yml
+++ b/.github/workflows/lint-and-analyse-php.yml
@@ -91,7 +91,7 @@ jobs:
       - name: Check Twig templates
         # Allow the previous check to fail but not abort
         if: always()
-        run: composer run twig-lint
+        run: php scripts/console lint:twig --ansi
 
   analyse-php:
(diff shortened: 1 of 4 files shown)


MauricioFauth:
This only partially fixes the issue, as I think there's a bug in Twig 3.9 where it can't set the a translation into a variable.

```twig
A {% set variable %}{% trans 'value' %}{% endset %} B {{ variable }} C
```
That will print:
```
A value B  C
```
Instead of the expected:
```
A  B value C
```

kamil-tekiela:
Why not keep 5_2 locked to <3.9?

williamdes (approved):
Changes look okay to me

MauricioFauth:
Closing this in favor of https://github.com/phpmyadmin/twig-i18n-extension/pull/12.

PR #19121, opened 2024-04-21
Fix linter for sql query box with multiple statements

### LINKED ISSUE #18389: Analyzed statement's error not shown immediately
### Describe the bug

Analyzed statement's error is not shown immediately. It only shows up when there is an error two consecutive times.

### To Reproduce

1. Go to  SQL
2. Type single letter or paste an invalid statement
3. Observe no error being shown
4. Type one more letter
5. Now an error is displayed

### Expected behavior

It should display error as soon as the statement is not valid

### Screenshots

image
image

### Server configuration

- phpMyAdmin version: 6  514d536 

### Client configuration

- Browser: Firefox 112.0.2
- Operating system: Linux

### LINKED ISSUE #18487: Unexpected token. (near ;)
### Describe the bug

Something has changed in the past weeks that is causing a `Unexpected token. (near ;)` error on phpMyAdmin 5.2.2-dev and 6.0.0-dev for every query that is edited and has `;` at the end, no matter what you type.

Strange, it doesn't happen on the demo server.

### To Reproduce

1. Go to https://www.phpmyadmin.net/downloads/
2. Download phpMyAdmin 5.2+snapshot
3. Edit a query with `;` at the end

### Expected behavior

No `Unexpected token. (near ;)` error should be given.

### Screenshots

https://github.com/phpmyadmin/sql-parser/assets/25424343/bcd882ba-226b-44b3-ab5e-fbf855bef697

### Server configuration

### LINKED ISSUE #18511: Parser error shown with multiple statements in query box
### Describe the bug

An error is shown when multiple SQL statements are entered in the query box.

### To Reproduce

1. Go to the query box 
2. Input 
```sql
SELECT 1; SELECT 1; 
```
3. See error

### Expected behavior

No error shown.

### Screenshots
image

### Server configuration

- phpMyAdmin version: 5.2 and 6.0

MoonE:
### Description

Only change the delimiter for routine/event/trigger editor. The query editor should use the default delimiter.

Fixes #18389, fixes  #18487, fixes #18511,

### DIFF
--- a/libraries/classes/Controllers/LintController.php
+++ b/libraries/classes/Controllers/LintController.php
@@ -48,14 +48,14 @@ class LintController extends AbstractController
             $options = $params['options'];
 
             if (! empty($options['routineEditor'])) {
-                $sqlQuery = 'CREATE PROCEDURE `a`() ' . $sqlQuery;
+                $sqlQuery = "DELIMITER $$\nCREATE PROCEDURE `a`() " . $sqlQuery;
             } elseif (! empty($options['triggerEditor'])) {
-                $sqlQuery = 'CREATE TRIGGER `a` AFTER INSERT ON `b` FOR EACH ROW ' . $sqlQuery;
+                $sqlQuery = "DELIMITER $$\nCREATE TRIGGER `a` AFTER INSERT ON `b` FOR EACH ROW " . $sqlQuery;
             } elseif (! empty($options['eventEditor'])) {
-                $sqlQuery = 'CREATE EVENT `a` ON SCHEDULE EVERY MINUTE DO ' . $sqlQuery;
+                $sqlQuery = "DELIMITER $$\nCREATE EVENT `a` ON SCHEDULE EVERY MINUTE DO " . $sqlQuery;
             }
         }
 
-        echo json_encode(Linter::lint("DELIMITER $$\n" . $sqlQuery . "$$\nDELIMITER ;\n"));
+        echo json_encode(Linter::lint($sqlQuery));
     }
 }
--- a/libraries/classes/Controllers/LintController.php
+++ b/libraries/classes/Controllers/LintController.php
@@ -10,6 +10,8 @@ namespace PhpMyAdmin\Controllers;
 use PhpMyAdmin\Core;
 use PhpMyAdmin\Linter;
 
+use function is_array;
+use function is_string;
 use function json_encode;
 
 /**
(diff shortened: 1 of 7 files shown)


libraries/classes/Controllers/LintController.php
```diff
@@ -48,14 +48,14 @@ public function __invoke(): void
...
 
             if (! empty($options['routineEditor'])) {
-                $sqlQuery = 'CREATE PROCEDURE `a`() ' . $sqlQuery;
+                $sqlQuery = "DELIMITER $$\nCREATE PROCEDURE `a`() " . $sqlQuery . '$$';
             } elseif (! empty($options['triggerEditor'])) {
-                $sqlQuery = 'CREATE TRIGGER `a` AFTER INSERT ON `b` FOR EACH ROW ' . $sqlQuery;
+                $sqlQuery = "DELIMITER $$\nCREATE TRIGGER `a` AFTER INSERT ON `b` FOR EACH ROW " . $sqlQuery . '$$';
             } elseif (! empty($options['eventEditor'])) {
-                $sqlQuery = 'CREATE EVENT `a` ON SCHEDULE EVERY MINUTE DO ' . $sqlQuery;
+                $sqlQuery = "DELIMITER $$\nCREATE EVENT `a` ON SCHEDULE EVERY MINUTE DO " . $sqlQuery . '$$';
```
kamil-tekiela, line 55: Do we need the `$$` at the end?

williamdes: Yes, multi line content would break otherwise

williamdes (approved):
Awesome, thank you

libraries/classes/Controllers/LintController.php
MoonE: I don't think it is necessary, In these cases only one statement will be in the editor and a delimiter at the very end is not required.

kamil-tekiela: I'm not sure how multiline content would break. I am testing it right now. A delimiter is supposed to delimit two queries apart. If there is no other query appended to it then the delimiter should is redundant.

kamil-tekiela: I can't get it to break and it seems to work fine without it. The reported issue is also gone. I think we should merge it without the delimiter at the end. It was needed before because the delimiter was reset back to `;` but this is also unnecessary.

MoonE:
The second commit fixes #18389, the problem was that the annotations were generated for non-existing lines as the linted sql was not the same as the sql in the editor.

With this fixed, linting routines/events/triggers **never** (except when the sql length exceeds 10000 characters) produces lint results as the body of these statements is not analyzed. It is simply skipped:

Routine: https://github.com/phpmyadmin/sql-parser/blob/011fa18a4e55591fac6545a821921dd1d61c6984/src/Statements/CreateStatement.php#L719-L726
Trigger: https://github.com/phpmyadmin/sql-parser/blob/011fa18a4e55591fac6545a821921dd1d61c6984/src/Statements/CreateStatement.php#L786-L793
Event: https://github.com/phpmyadmin/sql-parser/blob/011fa18a4e55591fac6545a821921dd1d61c6984/src/Statements/CreateStatement.php#L795-L802

So I'm not sure what the reason was for adding the sql prefix in the first place. Wouldn't it work better without it? Parsing the body of these statements as normal sql.

libraries/classes/Controllers/LintController.php
```diff
@@ -10,52 +10,59 @@
...
+        $prefix = is_string($editorType) ? self::EDITOR_SQL_PREFIX[$editorType] ?? '' : '';
 
-        // Disabling standard response.
-        $this->response->disable();
-
-        Core::headerJSON();
+        $lints = Linter::lint($prefix . $sqlQuery);
+        if ($prefix !== '') {
+            // Adjust positions to account for prefix
+            foreach ($lints as &$lint) {
```
kamil-tekiela, line 50: Stylistic nit, but I would prefer `foreach ($lints as $key => $lint) {` and then `$lints[$key] = `

kamil-tekiela:
> So I'm not sure what the reason was for adding the sql prefix in the first place. Wouldn't it work better without it? Parsing the body of these statements as normal sql.

I have asked myself this question this morning too. I suppose there was a scenario which made this fix necessary. It seems to be related to this fix https://github.com/phpmyadmin/sql-parser/pull/438

libraries/classes/Controllers/LintController.php
williamdes: Can you state versions for the record? "needed before"

williamdes:
> > So I'm not sure what the reason was for adding the sql prefix in the first place. Wouldn't it work better without it? Parsing the body of these statements as normal sql.
> 
> I have asked myself this question this morning too. I suppose there was a scenario which made this fix necessary. It seems to be related to this fix [phpmyadmin/sql-parser#438](https://github.com/phpmyadmin/sql-parser/pull/438)

That's correct
The parser made changes required in this repo

williamdes:
> So I'm not sure what the reason was for adding the sql prefix in the first place. Wouldn't it work better without it? Parsing the body of these statements as normal sql.

@Tithugues

williamdes (approved):
This looks good

kamil-tekiela (approved):

williamdes (approved):

PR #19126, opened 2024-04-23
Add `t` (trans) function to Twig\I18nExtension

MauricioFauth:
For translations, a Twig function is more versatile than a Twig tag and has the benefit of auto escaping.

As the `phpmyadmin/twig-i18n-extension` package is a fork of the Twig's `I18nExtension`, the custom function was added to the phpMyAdmin main repository, this way the cases not used by phpMyAdmin don't need to be implemented. For example, only the default domain is used and MoTranslator is always available.

The idea is to replace all usage of `trans` tags and filter with this new `t` function.

```diff
- {% trans 'Message' %}
+ {{ t('Message') }}
```

```diff
- {% trans %}Message{% endtrans %}
+ {{ t('Message') }}
```

```diff
- {% trans %}Message{% context %}Context{% endtrans %}
+ {{ t('Message', context = 'Context') }}
```

```diff
- {% trans %}Message{% notes %}Notes{% endtrans %}
+ {{ t('Message', notes = 'Notes') }}
```

```diff
- {% apply format(number_of_tables) %}
-     {% trans %}%d table{% plural number_of_tables %}%s tables{% endtrans %}
- {% endapply %}
+ {{ t('%d table', '%s tables', number_of_tables)|format(number_of_tables) }}
```

### DIFF
--- a/tests/unit/Twig/I18nExtensionTest.php
+++ b/tests/unit/Twig/I18nExtensionTest.php
@@ -1,0 +1,99 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpMyAdmin\Tests\Twig;
+
+use PhpMyAdmin\Config;
+use PhpMyAdmin\Template;
+use PhpMyAdmin\Tests\AbstractTestCase;
+use PhpMyAdmin\Twig\I18nExtension;
+use PhpMyAdmin\Twig\Node\Expression\TransExpression;
+use PHPUnit\Framework\Attributes\CoversClass;
+use ReflectionProperty;
+use Twig\Loader\FilesystemLoader;
+
+use const TEST_PATH;
+
+#[CoversClass(I18nExtension::class)]
+#[CoversClass(TransExpression::class)]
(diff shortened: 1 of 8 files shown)


williamdes (approved):
Let's have our own functions that extend the library, good PR

tests/unit/Twig/I18nExtensionTest.php
```diff
@@ -0,0 +1,99 @@
...
+    {
+        parent::tearDown();
+
+        (new ReflectionProperty(Template::class, 'twig'))->setValue(null, null);
+    }
+
+    public function testMessage(): void
+    {
+        $expected = <<<'HTML'
+Message
```
kamil-tekiela, line 41: Any reason for this strange indentation? Why not indent these lines properly?

MauricioFauth: You're right. I always forgot that nowdocs can be indented. And since PHP 7.3! Wow!

PR #19128, opened 2024-04-25
Replace the `trans` tag and filter with the `t` function in templates

MauricioFauth:
Removes the `phpmyadmin/twig-i18n-extension` package as well.

- Follow up of https://github.com/phpmyadmin/phpmyadmin/pull/19126


williamdes (commented):
💯

PR #19131, opened 2024-04-26
Fix #18240 - inserting value with UNIX_TIMESTAMP() without param

### LINKED ISSUE #18240: UNIX_TIMESTAMP -> Incorrect datetime value: ''
5.2.1 on Docker

Trying to insert a row in a table:

image

I select "UNIX_TIMESTAMP" in the dropdown for this INT(10) field.

I get this error:

> #1292 - Incorrect datetime value: ''

Because the SQL was generated like this:

`VALUES (NULL, NULL, UNIX_TIMESTAMP(''), NULL, NULL, NULL, NULL,`

Should be just:

`VALUES (NULL, NULL, UNIX_TIMESTAMP(), NULL, NULL, NULL, NULL,`

without the quotes inside `()`

MoonE:
### Description

At this point the value has not been escaped so to check whether the textfield was empty compare to empty `''` string instead of a quoted empty string `"''"`.

Was introduced here: [`6d0a30a`](https://github.com/phpmyadmin/phpmyadmin/commit/6d0a30a391cfb78cb51a476b4f925a625070c4b6#diff-97b13ff1169ee3b4fa17e6067276e7950186fa99e1056282568fc82d63350fb3L1613-L1629)

Fixes #18240

### DIFF
--- a/libraries/classes/InsertEdit.php
+++ b/libraries/classes/InsertEdit.php
@@ -1614,7 +1614,7 @@ class InsertEdit
 
         if (
             ! in_array($multiEditFuncs[$key], $funcNoParam)
-            || ($currentValue != "''"
+            || ($currentValue !== ''
                 && in_array($multiEditFuncs[$key], $funcOptionalParam))
         ) {
             if (


kamil-tekiela (approved):
It makes sense to me.

williamdes (approved):

PR #19135, opened 2024-04-27
Export views after indexes

### LINKED ISSUE #18566: Error when importing exported view with USE INDEX hint
### Describe the bug

CREATE VIEW statements are written to the exported sql before indexes, when the view uses `USE INDEX (...)` import fails.

### To Reproduce

```sql
CREATE SCHEMA s;
USE `s`;

CREATE TABLE `t` (
  `name` varchar(45) DEFAULT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `t_name` (`name`)
);
CREATE VIEW `v`  AS SELECT `t`.`id` AS `id`, `t`.`name` AS `name` FROM `t` USE INDEX (`t_name`) ;
```

1. Execute the sql above
2. Select the schema
3. Click on Export tab
4. Click on Export button
5. Re-create schema: `DROP SCHEMA s; CREATE SCHEMA s; USE s;`
6. Execute the exported script

### Expected behavior

Should be able to execute exported sql without error.

### Server configuration

- Database version: MySQL 5.7.38
- phpMyAdmin version: 5.2.1

MoonE:
### Description

When a view uses USE INDEX syntax import failed because of the missing index.
Store the view structure sql in a global variable and put them after the indexes. This is handled similarly for the indexes.

Fixes #18566

### DIFF
--- a/libraries/classes/Plugins/Export/ExportSql.php
+++ b/libraries/classes/Plugins/Export/ExportSql.php
@@ -67,6 +67,9 @@ class ExportSql extends ExportPlugin
      */
     private $sentCharset = false;
 
+    /** @var string */
+    private $sqlViews = '';
+
     protected function init(): void
     {
         // Avoids undefined variables, use NULL so isset() returns false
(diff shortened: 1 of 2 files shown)


kamil-tekiela:
Could you please use a property instead of a global? Even a public property would be better. See https://github.com/phpmyadmin/phpmyadmin/pull/19124/commits/34caf80af1757741f7cf2a0a0f39e25b086773f7

MoonE:
Makes a lot more sense with a property.

williamdes (approved):

PR #19136, opened 2024-04-27
Fix #18674 - formatted sql in grid should be inline text

### LINKED ISSUE #18674: Formatted sql in browse table result has linebreak after each token
### Describe the bug

Formatted sql when browsing pma__history has `display: block` applied to each token.

### To Reproduce

1. Browse pma__history table
2. See values in column sqlquery

### Expected behavior

There shouldn't be a new line for each token

### Screenshots

image

### Server configuration

- Operating system: Linux
- PHP version: 8.1
- phpMyAdmin version:  5.2.1

### Client configuration

- Browser: Firefox / Chrome
- Operating system: Windows

MoonE:
### Description

The css expects all table cells to have a wrapping span, but the wrapping was only done when no span existed in the cell. If the formatting is done before the wrapping there already exist nested spans and no top-level span is created. In this case the style that makes the sql displayed inline did not apply.

Fixes #18674

### DIFF
--- a/js/src/makegrid.js
+++ b/js/src/makegrid.js
@@ -2188,7 +2188,7 @@ var makeGrid = function (t, enableResize, enableReorder, enableVisib, enableGrid
 
     // wrap all truncated data cells with span indicating the original length
     // todo update the original length after a grid edit
-    $(t).find('td.data.truncated:not(:has(span))')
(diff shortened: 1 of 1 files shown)


williamdes (approved):
Thank you

PR #19149, opened 2024-05-06
Remove ResponseRenderer::disable() from controllers

MauricioFauth:
Creates a new Response object instead of using the response object from ResponseRenderer class.

This is an attempt to simplify the ResponseRenderer class by removing the ResponseRenderer::disable() method.


williamdes (approved):

MauricioFauth:
The most simple cases were removed. The more complex ones will be removed in other pull requests.

Issue #16197, opened 2020-06-15
Replacing the master/slave terminology


mauriciofauth:
We should replace the _master_ and _slave_ terminology when possible. We can't completely remove this terms because it's a MySQL/MariaDB terminology.

MariaDB began replacing the terms with the _primary_ and _replica_ pair and already [added `replica` as an alias for `slave`](https://github.com/MariaDB/server/commit/23686b2bafe2f3f9614d68923dfd4aea7641dbcc).

MySQL related issue: https://bugs.mysql.com/bug.php?id=99806
MariaDB related issue: https://jira.mariadb.org/browse/MDEV-18777

Update: MariaDB is replacing the terms with _primary_ and _replica_ and MySQL is replacing the terms with _source_ and _replica_.

rajat315315:
So, is it as simple as replacing all ocurrences of `master` in UI to `primary` and `slave` to `replica`?
Can I do it?

williamdes:
Sure @rajat315315 you can do it but be cautious about some cases where the words can not be replaced :)

mauriciofauth:
MariaDB is replacing the terms with _primary_ and _replica_ and MySQL is replacing the terms with _source_ and _replica_.

- https://mariadb.com/kb/en/replication-overview/
- https://dev.mysql.com/doc/refman/8.0/en/replication.html

PR #19153, opened 2024-05-09
Update to fix replication that broke with MySQL 8.4.0 removal of master and slave commands

dlmorri75:
### Description

Fixes and issue with replication that broke with MySQL 8.4.0 removal of depreciated master and slave replication commands.

- #19281

### DIFF
--- a/libraries/classes/Replication.php
+++ b/libraries/classes/Replication.php
@@ -62,7 +62,11 @@ class Replication
             return -1;
         }
 
-        return $dbi->tryQuery($action . ' REPLICA ' . $control . ';', $link);
+        if ($dbi->isMySql() && $dbi->getVersion() >= 80400) {
+            return $dbi->tryQuery($action . ' REPLICA ' . $control . ';', $link);
+        }
+
+        return $dbi->tryQuery($action . ' SLAVE ' . $control . ';', $link);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


MauricioFauth:
- Related to #16197

MauricioFauth (changes requested):

These changes will break for older versions of MySQL, as phpMyAdmin supports MySQL/MariaDB >= 5.5.

So, it should check if the server is MySQL >= 8.4 to use the new statements, otherwise use the old ones.

libraries/classes/Replication.php
```diff
@@ -95,16 +99,29 @@ public function replicaChangePrimary(
...
-            'MASTER_USER=\'' . $user . '\',' .
-            'MASTER_PASSWORD=\'' . $password . '\',' .
-            'MASTER_LOG_FILE=\'' . $pos['File'] . '\',' .
-            'MASTER_LOG_POS=' . $pos['Position'] . ';',
-            $link
-        );
+        if ($dbi->isMySql() && $dbi->getVersion() >= 80400) {
+            $out = $dbi->tryQuery(
+                'CHANGE REPLICATION SOURCE TO ' .
+                'SOURCE_HOST=\'' . $host . '\',' .
```
kamil-tekiela, line 105: Shouldn't this use `quoteString()`?

MauricioFauth: I thought so, but 5.2 doesn't have `quoteString` and those variables were escaped earlier using `escapeString`.

MauricioFauth:
@dlmorri75 Merged! Thanks!

PR #19154, opened 2024-05-09
Remove ResponseRenderer::callExit() calls from auth plugins

MauricioFauth:
Refactors authentication plugins to return a response object instead of throwing `ExitException`.

### DIFF
--- a/src/Exceptions/AuthenticationFailure.php
+++ b/src/Exceptions/AuthenticationFailure.php
@@ -1,0 +1,78 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpMyAdmin\Exceptions;
+
+use RuntimeException;
+use Throwable;
+
+use function __;
+
+final class AuthenticationFailure extends RuntimeException
+{
+    public const SERVER_DENIED = 'server-denied';
+    public const ALLOW_DENIED = 'allow-denied';
+    public const ROOT_DENIED = 'root-denied';
+    public const EMPTY_DENIED = 'empty-denied';
+    public const NO_ACTIVITY = 'no-activity';
+
+    /** @psalm-param self::* $failureType */
+    public function __construct(
+        public readonly string $failureType,
+        string $message = '',
+        int $code = 0,
+        Throwable|null $previous = null,
+    ) {
+        parent::__construct($message, $code, $previous);
+    }
+
+    /**
+     * Database server denied user login
+     */
+    public static function serverDenied(): self
+    {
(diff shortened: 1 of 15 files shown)


src/Exceptions/AuthenticationFailure.php
```diff
@@ -0,0 +1,78 @@
...
+namespace PhpMyAdmin\Exceptions;
+
+use RuntimeException;
+use Throwable;
+
+use function __;
+
+final class AuthenticationFailure extends RuntimeException
+{
+    public const SERVER_DENIED = 'server-denied';
```
kamil-tekiela, line 14: Is this a good candidate for an enum?

src/Exceptions/AuthenticationFailure.php
```diff
@@ -0,0 +1,78 @@
...
+     */
+    public static function rootDenied(): self
+    {
+        return new self(self::ROOT_DENIED, __('Access denied!'));
+    }
+
+    /**
+     * Empty password is denied
+     */
+    public static function emptyDenied(): self
```
kamil-tekiela, line 57:
```suggestion
    public static function emptyPasswordDenied(): self
```

src/Exceptions/AuthenticationFailure.php
```diff
@@ -0,0 +1,78 @@
...
+     */
+    public static function allowDenied(): self
+    {
+        return new self(self::ALLOW_DENIED, __('Access denied!'));
+    }
+
+    /**
+     * User 'root' is denied in configuration
+     */
+    public static function rootDenied(): self
```
kamil-tekiela, line 49:
```suggestion
    public static function rootDeniedByConfiguration(): self
```
Can the comment be merged into the method name? Does it have to be this name?

src/Exceptions/AuthenticationFailure.php
```diff
@@ -0,0 +1,78 @@
...
+        return new self(
+            self::EMPTY_DENIED,
+            __('Login without a password is forbidden by configuration (see AllowNoPassword).'),
+        );
+    }
+
+    /**
+     * Automatically logged out due to inactivity
+     */
+    public static function noActivity(): self
```
kamil-tekiela, line 68:
```suggestion
    public static function loggedOutDueToInactivity(): self
```

src/Exceptions/AuthenticationFailure.php
MauricioFauth: I thought about it, but a custom auth plugin may want to add its own failure type.

PR #19169, opened 2024-05-16
Add support for statement messages in FlashMessenger class

MauricioFauth:
Besides standard feedback messages, is also common to have a feedback message with the executed statement after an action is executed. This allows to use HTTP redirection in those cases.

### DIFF
--- a/resources/templates/export.twig
+++ b/resources/templates/export.twig
@@ -9,9 +9,13 @@
 
   {% for flash_key, flash_messages in flash() %}
     {% for flash_message in flash_messages %}
-      <div class="alert alert-{{ flash_key }}" role="alert">
-        {{ flash_message }}
-      </div>
+      {% if flash_message.statement is empty %}
+        <div class="alert alert-{{ flash_key }}" role="alert">
+          {{ flash_message.message }}
+        </div>
+      {% else %}
+        {{ statement_message(flash_message.message, flash_message.statement, flash_key) }}
+      {% endif %}
     {% endfor %}
   {% endfor %}
 
--- a/resources/templates/export.twig
+++ b/resources/templates/export.twig
@@ -7,16 +7,14 @@
   {{ page_settings_error_html|raw }}
   {{ page_settings_html|raw }}
 
-  {% for flash_key, flash_messages in flash() %}
-    {% for flash_message in flash_messages %}
-      {% if flash_message.statement is empty %}
-        <div class="alert alert-{{ flash_key }}" role="alert">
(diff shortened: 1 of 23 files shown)


src/FlashMessenger.php
```diff
@@ -8,14 +8,14 @@
...
+final class FlashMessenger
 {
-    private const STORAGE_KEY = 'flashMessages';
+    private const STORAGE_KEY = 'FlashMessenger';
 
     /** @var mixed[] */
     private array $storage;
 
-    /** @var array<string, string[]> */
+    /** @psalm-var list<array{context: string, message: string, statement: string}> */
```
kamil-tekiela, line 18: I don't think there's anything Psalm-specific here anymore. Both list and array shape are universally supported.

resources/templates/export.twig
```diff
@@ -9,9 +9,13 @@
...
     {% for flash_message in flash_messages %}
-      <div class="alert alert-{{ flash_key }}" role="alert">
-        {{ flash_message }}
-      </div>
+      {% if flash_message.statement is empty %}
+        <div class="alert alert-{{ flash_key }}" role="alert">
+          {{ flash_message.message }}
+        </div>
+      {% else %}
+        {{ statement_message(flash_message.message, flash_message.statement, flash_key) }}
```
kamil-tekiela, line 17: Where is `flash_key` coming from? Shouldn't it be something like `flash_message.context`?

src/Twig/MessageExtension.php
```diff
@@ -35,4 +37,10 @@ public function getFilters(): array
             ),
         ];
     }
+
+    /** @inheritDoc */
+    public function getFunctions(): array
+    {
+        return [new TwigFunction('statement_message', Generator::getMessage(...), ['is_safe' => ['html']])];
```
kamil-tekiela, line 44: That `match` statement would fit here more than inside the function. You can do an anonymous function and as a default throw an exception so that we avoid typos.

src/FlashMessenger.php
MauricioFauth: I think sudo-bot/action-doctum does not support this format yet. Am I right, @williamdes?

src/Twig/MessageExtension.php
MauricioFauth: I will left `MessageType::Notice` as default for now, as I think `Generator::getMessage()` should be refactored to extract all the HTML from it, and maybe a view helper for `MessageType` to match with the Bootstrap classes.

PR #19182, opened 2024-05-28
Refactor GIS editor modal

MauricioFauth:
Extracts OpenLayers JS from the GIS PHP classes.


MoonE:
Thanks you for the updated layout of the gis editor. It looks a lot better.

I found some bugs resulting from these changes:
- labels are missing in the OpenLayers visualization
- the Start row and number of rows inputs no longer work in the OpenLayers visualization 
- in the editor the textarea should not be resizable horizontally
- in the editor when selecting gemetrycollection as type the visualization still shows the last geometry, but should be empty, as the default is an empty geometry collection

The editor modal should not be closable by clicking outside of it. All the form input data will be lost in this case. It's just too easy to accidentally close it.

PR #19211, opened 2024-06-25
Replace jqPlot with Chart.js in table zoom search page

MauricioFauth:
Adds dependency on chartjs-plugin-zoom to support the zoom feature.


williamdes (approved):
🎉

PR #19230, opened 2024-07-14
Clean up of the public/index.php and app/constants.php files

MauricioFauth:
- [x] Remove the `PHPMYADMIN` constant
- [x] ~Remove runtime check of the PHP version~ Extract runtime check of the PHP version from `public/index.php`
- [x] Remove duplicated code from `public/setup/index.php`
- [x] Remove constants defined in `app/constants.php`


public/index.php
```diff
@@ -9,13 +9,8 @@
     define('ROOT_PATH', dirname(__DIR__) . DIRECTORY_SEPARATOR);
 }
 
-define('PHPMYADMIN', true);
 // phpcs:enable
 
-if (PHP_VERSION_ID < 80102) {
```
williamdes, line 15: I am not sure why you removed it
I will have to add it back on Debian

MauricioFauth: Sorry, I added the why in the commit message, but not in the PR description.
This check is part of the [Composer autoloader](https://getcomposer.org/doc/07-runtime.md#platform-check). As the Debian package uses a custom autoloader, I think this check can be added to it, instead of adding it back to `public/index.php`.

MauricioFauth: Even for 5.2, instead of patching index.php to increase the required version, this check could go into the custom autoload file.

MauricioFauth:
@williamdes I also though about removing the ROOT_PATH constant and maybe TEST_PATH as well, because constants are really hard to deal with when testing. Well, at least the `if (! defined('ROOT_PATH'))` part, I don't really know why I added it.
But I'm not going to do this in this PR, and if I would do that, I'll make sure it works for the Debian scenario to not need a patch like the TEST_PATH one.
What do you think?

williamdes:
> @williamdes I also though about removing the ROOT_PATH constant and maybe TEST_PATH as well, because constants are really hard to deal with when testing. Well, at least the `if (! defined('ROOT_PATH'))` part, I don't really know why I added it. But I'm not going to do this in this PR, and if I would do that, I'll make sure it works for the Debian scenario to not need a patch like the TEST_PATH one. What do you think?

Well, you can remove them but tests need to survive having the tests folder being moved into a random place and still testing the actual phpmyadmin via composer autoload magic PSR-4

public/index.php
williamdes: I really think this check written in pure php is the best thing ever. Relying on some more complex code may break in some ways that the user will not have the information
I would like to keep this check

MauricioFauth: It's not complex. Composer uses the required PHP version defined in composer.json to generate basically the same code with a similar message and it does this check before any autoload code. But this is only for Composer 2, that's why I added the composer API requirement. That means it will not be possible to use Composer 1 to install the dependencies.

williamdes: Okay, can you post the first lines of the generated autoload ?

MauricioFauth: It's basically this:

`vendor/autoload.php`:
```php
require_once __DIR__ . '/composer/autoload_real.php';
return ComposerAutoloaderInitSomeHash::getLoader();
```
`vendor/composer/autoload_real.php`:
```php
class ComposerAutoloaderInitSomeHash
{
    private static $loader;
    public static function getLoader()
    {
        if (null !== self::$loader) {
            return self::$loader;
        }
        require __DIR__ . '/platform_check.php';
        // autoload stuff
        return self::$loader;
    }
}
```
`vendor/composer/platform_check.php`:
```php
<?php
$issues = array();

if (!(PHP_VERSION_ID >= 80102)) {
    $issues[] = 'Your Composer dependencies require a PHP version ">= 8.1.2". You are running ' . PHP_VERSION . '.';
}

if ($issues) {
    if (!headers_sent()) {
        header('HTTP/1.1 500 Internal Server Error');
    }
    if (!ini_get('display_errors')) {
        if (PHP_SAPI === 'cli' || PHP_SAPI === 'phpdbg') {
            fwrite(STDERR, 'Composer detected issues in your platform:' . PHP_EOL.PHP_EOL . implode(PHP_EOL, $issues) . PHP_EOL.PHP_EOL);
        } elseif (!headers_sent()) {
            echo 'Composer detected issues in your platform:' . PHP_EOL.PHP_EOL . str_replace('You are running '.PHP_VERSION.'.', '', implode(PHP_EOL, $issues)) . PHP_EOL.PHP_EOL;
        }
    }
    trigger_error(
        'Composer detected issues in your platform: ' . implode(' ', $issues),
        E_USER_ERROR
    );
}
```

MauricioFauth: I noticed that is not common to include PHP version checks in the autoload files in Debian. I can revert that commit if you think that it's worth to keep that check in index.php. It's possible to completely disable the Composer's platform check by adding `"platform-check": false` to composer's config.

MauricioFauth:
> > @williamdes I also though about removing the ROOT_PATH constant and maybe TEST_PATH as well, because constants are really hard to deal with when testing. Well, at least the `if (! defined('ROOT_PATH'))` part, I don't really know why I added it. But I'm not going to do this in this PR, and if I would do that, I'll make sure it works for the Debian scenario to not need a patch like the TEST_PATH one. What do you think?
> 
> Well, you can remove them but tests need to survive having the tests folder being moved into a random place and still testing the actual phpmyadmin via composer autoload magic PSR-4

Removing TEST_PATH was easier than I thought. On the other hand, ROOT_PATH is more difficult, but I have an idea that might work. I will try it later. 

- #19231

public/index.php
MauricioFauth: I reverted the commit. See if you like the new solution.

MauricioFauth:
Besides the removal of the PHPMYADMIN constant, I'm not happy the other changes. I'm closing this.

public/index.php
```diff
@@ -9,24 +9,17 @@
...
-    die('<p>PHP 8.1.2+ is required.</p><p>Currently installed version is: ' . PHP_VERSION . '</p>');
-}
-
-require_once ROOT_PATH . 'app/constants.php';
+require ROOT_PATH . 'app/platform_check.php';
```
williamdes, line 14: That was not too bad

PR #19239, opened 2024-07-25
Remove the PHPMYADMIN constant

MauricioFauth:
This constant was used to avoid non-entry-points scripts being called directly. As phpMyAdmin has a public directory now, directly calling scripts outside public is not possible anymore.

Introduced by afbb2a9dc2ff6f612e01f86e85788610e19a0338.

### DIFF
--- a/bin/console
+++ b/bin/console
@@ -16,7 +16,6 @@ if (! defined('ROOT_PATH')) {
     define('ROOT_PATH', dirname(__DIR__) . DIRECTORY_SEPARATOR);
 }
 
-define('PHPMYADMIN', true);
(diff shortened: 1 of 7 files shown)


kamil-tekiela (approved):

bin/console
```diff
@@ -16,7 +16,6 @@ if (! defined('ROOT_PATH')) {
     define('ROOT_PATH', dirname(__DIR__) . DIRECTORY_SEPARATOR);
 }
 
-define('PHPMYADMIN', true);
```
williamdes, line 19: we should add a security check that sapi is CLI

PR #19241, opened 2024-07-27
Remove the Functions JS global object

MauricioFauth:
Exports each function individually and removes the `Functions` global object.

### DIFF
--- a/resources/js/src/import.ts
+++ b/resources/js/src/import.ts
@@ -1,6 +1,6 @@
 import $ from 'jquery';
 import { AJAX } from './modules/ajax.ts';
-import { Functions } from './modules/functions.ts';
+import { formatBytes } from './modules/functions.ts';
 import { Navigation } from './modules/navigation.ts';
 import { CommonParams } from './modules/common.ts';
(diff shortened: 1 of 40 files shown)


resources/js/src/import.ts
```diff
@@ -1,6 +1,6 @@
 import $ from 'jquery';
 import { AJAX } from './modules/ajax.ts';
-import { Functions } from './modules/functions.ts';
+import { formatBytes } from './modules/functions.ts';
```
williamdes, line 3:
```suggestion
import * as Functions from './modules/functions.ts';
```
That would also help stack traces be readable?
For the Functions is undefined bug

MauricioFauth: > That would also help stack traces be readable?

That will make the import.ts import all exported stuff from functions.ts instead of only the formatBytes function. I think the stack trace issue has more to do with the source maps not being used by the report tool.

> For the Functions is undefined bug

Are you talking about a specific bug? `Functions` doesn't exist anymore with this PR.

williamdes: > I think the stack trace issue has more to do with the source maps not being used by the report tool.

Definitly

PR #19243, opened 2024-07-31
Bump minimum PHP version to 8.2.0

MauricioFauth:
- https://github.com/phpmyadmin/phpmyadmin/pull/18078


williamdes (approved):
For the record, I agree to this bump because thee support of previous versions ends up very quickly from this date. One year is not much.

PR #19255, opened 2024-08-13
Backport some fixes for PHPUnit 10 compatibility

MauricioFauth:
- Fixes https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1070623

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -106,7 +106,7 @@
         "phpstan/phpstan": "^1.4.8",
         "phpstan/phpstan-phpunit": "^1.0",
         "phpstan/phpstan-webmozart-assert": "^1.0",
-        "phpunit/phpunit": "^8.5.16 || ^9.5",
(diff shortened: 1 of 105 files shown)


williamdes (approved):
Thanks
Is it possible to have the phpunit 11 compat for annotations ?

MauricioFauth:
> Is it possible to have the phpunit 11 compat for annotations ?

Is possible, yes. This PR fixes all errors for PHPUnit 11. Does you think is worth to fix the deprecations as well?

williamdes:
> > Is it possible to have the phpunit 11 compat for annotations ?
> 
> Is possible, yes. This PR fixes all errors for PHPUnit 11. Do you think is worth to fix the deprecations as well?

Yes definitely, adding the second syntax will make us ready for newer phpunit versions

Issue #19177, opened 2024-05-23
Edit privileges doesn't work


liviuconcioiu:
### Describe the bug

Edit privileges doesn't work, neither when a new user is created. It won't save the changes. It resets them to `USAGE` only.

### To Reproduce

1. Go to https://demo.phpmyadmin.net/master-config/public/index.php?route=/server/privileges
2. Edit privileges of a user
3. Privileges aren't updated.

### Expected behavior

Edit privileges should work as expected.

### Screenshots

https://github.com/phpmyadmin/phpmyadmin/assets/25424343/1d63a47a-597e-4149-9d2b-a41fdf72599e

### Server configuration

- phpMyAdmin version: 6.0.0-dev

PR #19262, opened 2024-08-15
Fix privileges

kamil-tekiela:
While using PMA today I noticed I broke the privilege assignment page in https://github.com/phpmyadmin/phpmyadmin/pull/19054

This fixes the issue by replacing the `$GLOBALS` with `$_POST`. I still don't know why we used GLOBALS before so there could still be something I missed. But it seems to fix the issue I introduced earlier.


liviuconcioiu:
This will fix #19177.

MauricioFauth (approved):

PR #19263, opened 2024-08-16
Fix #17834 - Add a button for wrapping queries into a transaction

igronus:
### Description

 Execute multiple queries from the query textbox inside one transaction. 

Fixes #17834

## Fixes

Added a checkbox and implemented logic for wrapping SQL queries into a transaction.

### DIFF
--- a/resources/templates/sql/query.twig
+++ b/resources/templates/sql/query.twig
@@ -132,6 +132,13 @@
 
           <div class="col-12">
             <div class="form-check form-check-inline">
+              <input class="form-check-input" type="checkbox" name="transaction_query" value="1" id="transaction_query" tabindex="135">
+              <label class="form-check-label" for="transaction_query">{{ t('Wrap in one transaction') }}</label>
+            </div>
+          </div>
+
+          <div class="col-12">
+            <div class="form-check form-check-inline">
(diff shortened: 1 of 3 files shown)


MauricioFauth:
@igronus Thank you for your contribution.

The Signed-off-by line is missing from your commit message. You can add it to your commits or add o comment here in this pull request with a line similar to `Signed-off-by: Author Name <authoremail@example.com>`.

https://github.com/phpmyadmin/phpmyadmin/pull/19263/checks?check_run_id=28935514516

resources/templates/sql/query.twig
```diff
@@ -130,6 +130,13 @@
...
+          <div class="col-12">
+            <div class="form-check form-check-inline">
+              <input class="form-check-input" type="checkbox" name="transaction_query" value="1" id="transaction_query" tabindex="135">
+              <label class="form-check-label" for="transaction_query">{{ t('Wrap in one transaction') }}</label>
+            </div>
+          </div>
+
           <div class="col-12">
             <div class="form-check form-check-inline">
               <input class="form-check-input" type="checkbox" name="rollback_query" value="1" id="rollback_query" tabindex="134">
```
williamdes, line 142: it should be tabindex 134 and the one after 135

PR #19287, opened 2024-09-16
add optional support for binary data in database search

GregorLohaus:
Signed-off-by: Gregor Lohaus <lohausgregor@gmail.com>

### Description

Recently I had to search for some binary data in a database, globally. 
Unfortunately neither LIKE nor REGEXP work with hex literals, so i added an option to include matches on hexed column data in the database search module. 
I think some people could benefit from this.

### DIFF
--- a/src/Database/Search.php
+++ b/src/Database/Search.php
@@ -45,6 +45,13 @@ class Search
     private int $criteriaSearchType;
 
     /**
+     * Options to apply to search
+     *
+     * @var bool criteriaSearchOptionsIncludeHex
+     */
+    private bool $criteriaSearchOptionsIncludeHex;
+
+    /**
      * Already set search type's description
      */
     private string $searchTypeDescription = '';
@@ -75,6 +82,7 @@ class Search
             4 => __('the exact phrase as whole field'),
             5 => __('as regular expression'),
         ];
+        $this->criteriaSearchOptionsIncludeHex = false;
         // Sets criteria parameters
         $this->setSearchParams();
     }
@@ -115,6 +123,11 @@ class Search
             $this->criteriaColumnName = '';
         } else {
             $this->criteriaColumnName = $_POST['criteriaColumnName'];
+        }
+
+        //phpcs:ignore SlevomatCodingStandard.ControlStructures.EarlyExit.EarlyExitNotUsed
+        if (isset($_POST['criteriaSearchOptionIncludeHex']) && $_POST['criteriaSearchOptionIncludeHex'] === '1') {
+            $this->criteriaSearchOptionsIncludeHex = true;
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/Database/Search.php
```diff
@@ -44,6 +44,13 @@ class Search
      */
     private int $criteriaSearchType;
 
+    /**
+     * Options to apply to search
+     *
+     * @var mixed[]
+     */
+    private array $criteriaSearchOptions;
```
kamil-tekiela, line 52: Why is it an array of mixed values if you have only a single boolean?

src/Database/Search.php
```diff
@@ -116,6 +124,12 @@ private function setSearchParams(): void
         } else {
             $this->criteriaColumnName = $_POST['criteriaColumnName'];
         }
+
+        if (isset($_POST['criteriaSearchOptionIncludeHex']) === false) {
+            $this->criteriaSearchOptions['include_hex_matches'] = false;
+        } else {
+            $this->criteriaSearchOptions['include_hex_matches'] = true;
+        }
```
kamil-tekiela, line 132:
```suggestion
        $this->criteriaSearchOptions['include_hex_matches'] = isset($_POST['criteriaSearchOptionIncludeHex']);
```

src/Database/Search.php
```diff
@@ -189,6 +203,14 @@ private function getWhereClause(string $table): string
                 $likeClausesPerColumn[] = 'CONVERT(' . Util::backquote($column->field) . ' USING utf8)'
                     . ' ' . $likeOrRegex . ' '
                     . $this->dbi->quoteString($automaticWildcard . $searchWord . $automaticWildcard);
+
+                if (! ((bool) $this->criteriaSearchOptions['include_hex_matches'])) {
```
kamil-tekiela, line 207: The cast is unnecessary since the only possible value is a boolean.

tests/unit/Database/SearchTest.php
```diff
@@ -88,40 +89,93 @@ public function testGetWhereClause(string $type, string $expected): void
...
             ],
             [
                 '4',
+                ['include_hex_matches' => null],
                 " WHERE (CONVERT(`column1` USING utf8) LIKE 'search string'"
                 . " OR CONVERT(`column2` USING utf8) LIKE 'search string')",
             ],
             [
                 '5',
+                ['include_hex_matches' => null],
```
kamil-tekiela, line 127: I see no reason to test `null`. Your condition checks if the value is set, so `['include_hex_matches' => null],` is unnecessary.

kamil-tekiela (changes requested):

GregorLohaus:
@kamil-tekiela i have refactored this a bit adding a hidden input so the arraykey will always be set in the post request, i thought using an array might be a good idea if someone wanted to add more options in the future but i have changed this to use a boolean property for now

src/Database/Search.php
GregorLohaus: i think i did this to satisfy phpstan when the value could be unset but i removed the cast now

src/Database/Search.php
GregorLohaus: the value will now always be set to either '1' or '0' due to the added hidden input, i still have to check for isset because on the first page render the value is not set the and the empty check used for searchType is disallowed by phpstan config

tests/unit/Database/SearchTest.php
GregorLohaus: this was also done to satisfy phpstan

MauricioFauth (approved):

MauricioFauth:
Merged. Thank you!

PR #19294, opened 2024-09-22
[Query Generator] simplify IN, NOT IN criteria text

faissaloux:
Hellooo 👋🏻 

### Current
We need to pass criteria text with quotes for each value to generate a valid query.

image

### After

image

### DIFF
--- a/resources/js/src/database/multi_table_query.ts
+++ b/resources/js/src/database/multi_table_query.ts
@@ -35,6 +35,8 @@ AJAX.registerTeardown('database/multi_table_query.js', function () {
 
     $('#update_query_button').off('click');
     $('#add_column_button').off('click');
+    $('body').off('click', 'input.add-option');
+    $('body').off('click', 'input.remove-option');
 });
 
 AJAX.registerOnload('database/multi_table_query.js', function () {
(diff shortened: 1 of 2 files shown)


js/src/functions.js
```diff
@@ -441,6 +441,15 @@ Functions.escapeSingleQuote = function (s) {
     return s.replaceAll('\\', '\\\\').replaceAll('\'', '\\\'');
 };
 
+/**
+ * @param {string} s
+ * @param {string} char
+ * @return {string}
+ */
+Functions.wrap = function (s, char) {
```
MauricioFauth, line 449: If this function is only used in a single place, it doesn't need to be in functions.js.

MauricioFauth (changes requested):
Does it still accepts using single quotes? What about a single value that has a comma?
I'm not sure if this should go to QA_5_2.

MauricioFauth:
> No it does not, if using single quotes it escapes them like if they belong to the value like any other character.

> This won't work because it counts as two separate values. Do you have any idea to solve this?

These are breaking changes.

The only way to accept not needing a quoted value is to also accept quoted values, as a quoted value fixes the comma issue. But accepting both ways could be very confusing.

I think a good solution is to only accept quoted values and add a button to edit the list, like the enum/set editor that exists when creating/editing a table.

faissaloux:
> I think a good solution is to only accept quoted values and add a button to edit the list, like the enum/set editor that exists when creating/editing a table.

Yeah this is the kind of the solution I thought about, but for addition too.

We need to be able to add values one by one (without quotes, quotes can be part of the value) and then can be edited or removed there on a list of values.

faissaloux:
@MauricioFauth

https://github.com/user-attachments/assets/166fd993-f5f0-417b-9476-1cd8b5707c84

MauricioFauth:
That's nice!

MauricioFauth (approved):
Great work! Thank you!

Could you please change the target branch to `master` instead of `QA_5_2` and squash your commits?

faissaloux:
@MauricioFauth

Retargeted to `master` and commits squashed!

PR #19305, opened 2024-09-28
Foreigners refactoring

kamil-tekiela:
The goal here was to split the `getForeigners()` method and introduce more type safety in the process. It has only been a partial success as `'both'` is still allowed. It's difficult to untangle the whole thing.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -8,16 +8,16 @@
     "packages": [
         {
             "name": "composer/ca-bundle",
-            "version": "1.5.1",
+            "version": "1.5.2",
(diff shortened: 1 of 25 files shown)


MauricioFauth (approved):

MauricioFauth:
Great work!

> It has only been a partial success as 'both' is still allowed.

What about extracting a new method like getForeignersInternal() and getForeigners() is always both by merging getForeignersInternal() and getForeignKeysData()? At least the $source param will be removed.

PR #19315, opened 2024-10-10
Fix #17262 - Add warning when setting a password to a blank value if AllowNoPassword is false

### LINKED ISSUE #17262: Changing password should respect AllowNoPassword

jacekbarecki:
### Description

Fixes #17262 : Changing password should respect `AllowNoPassword`.

If `AllowNoPassword` is `false` then a warning modal will be displayed, both under the "Change password" link on the main page and in the change password section on the "User accounts" tab.

r1
r2

### DIFF
--- a/src/DatabaseInterface.php
+++ b/src/DatabaseInterface.php
@@ -2044,11 +2044,6 @@ class DatabaseInterface implements DbalInterface
         return $this->databaseList;
     }
 
-    public function getConfig(): Config
-    {
-        return $this->config;
-    }
-
     /**
      * Returns the number of warnings from the last query.
      */
--- a/src/DatabaseInterface.php
+++ b/src/DatabaseInterface.php
@@ -2044,6 +2044,11 @@ class DatabaseInterface implements DbalInterface
         return $this->databaseList;
     }
 
+    public function getConfig(): Config
+    {
+        return $this->config;
+    }
+
     /**
      * Returns the number of warnings from the last query.
      */
--- a/app/services.php
+++ b/app/services.php
@@ -158,7 +158,7 @@ return [
         'server_plugins' => ['class' => Plugins::class, 'arguments' => ['@dbi']],
         'server_privileges' => [
             'class' => Privileges::class,
-            'arguments' => ['@template', '@dbi', '@relation', '@relation_cleanup', '@server_plugins'],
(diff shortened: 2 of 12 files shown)


jacekbarecki:
Hi, I'm not sure what should be the target branch, if it needs to be changed then please let me know.
Also, if a mention in the changelog should be added, I would be thankful for guidance in which part of the file to add it.
Thanks!

src/DatabaseInterface.php
```diff
@@ -2044,6 +2044,11 @@ public function getDatabaseList(): ListDatabase
         return $this->databaseList;
     }
 
+    public function getConfig(): Config
+    {
+        return $this->config;
+    }
```
kamil-tekiela, line 2050: Isn't there a better way of doing this?

jacekbarecki: Hi Kamil, do you have any suggestion here?
 
In the `Privileges` class, `DatabaseInterface` is already injected as a dependency and by adding this method, it can expose config related to currently selected server which is required in this case.

Initially I was thinking of injecting `Config` as another dependency in the `Privileges` class but this seemed to be redundant to me if  `DatabaseInterface` with loaded config is already injected. 

I could also implement a very specific method to fetch only this value, and access it like `$this->dbi->isAllowNoPasswordEnabled()` but given that there's plenty of config values that may be required in different places in the code, creating a separate method for fetching each of them doesn't seem too elegant for me.

kamil-tekiela: Yeah, I think adding Config as a dependency is better.

jacekbarecki: Sure! I pushed the changes.

MauricioFauth (approved):
That's great! Thank you!

PR #19321, opened 2024-10-15
Suppress deprecation message of E_STRICT constant

MauricioFauth:
The `E_STRICT` constant was deprecated in PHP 8.4.

It's still a possible error level when using PHP 7.2 and PHP 7.3.

See: https://wiki.php.net/rfc/deprecations_php_8_4#remove_e_strict_error_level_and_deprecate_e_strict_constant

### DIFF
--- a/libraries/classes/Error.php
+++ b/libraries/classes/Error.php
@@ -38,7 +38,6 @@ use const E_ERROR;
 use const E_NOTICE;
 use const E_PARSE;
 use const E_RECOVERABLE_ERROR;
-use const E_STRICT;
 use const E_USER_DEPRECATED;
 use const E_USER_ERROR;
 use const E_USER_NOTICE;
@@ -54,7 +53,7 @@ class Error extends Message
     /**
      * Error types
      *
-     * @var array
+     * @var array<int, string>
      */
     public static $errortype = [
         0 => 'Internal error',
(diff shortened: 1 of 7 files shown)


libraries/classes/Error.php
```diff
@@ -316,7 +312,11 @@ public function getLine(): int
      */
     public function getType(): string
     {
-        return self::$errortype[$this->getNumber()];
+        if ($this->getNumber() === @E_STRICT) {
```
kamil-tekiela, line 315: Why not keep it in the array and just use a number instead with a comment beside what the number represents?

libraries/classes/ErrorHandler.php
```diff
@@ -281,7 +281,7 @@ public function addError(
         }
 
         switch ($error->getNumber()) {
-            case E_STRICT:
+            case @E_STRICT:
```
kamil-tekiela, line 284:
```suggestion
            case 2048; // E_STRICT
```

libraries/classes/Error.php
MauricioFauth: I don't like to use the numbers directly, but I think it's better in this case, as `E_STRICT` should be removed in `master` branch.

kamil-tekiela (approved):

kamil-tekiela:
Out of curiosity, if it's only a deprecation why does it go to 5.2 branch? The code still works, right?

MauricioFauth:
> Out of curiosity, if it's only a deprecation why does it go to 5.2 branch? The code still works, right?

Yes, it still works.

But in this case, when in development environment, as the deprecation is triggered in the ErrorHandler when handling an exception, the deprecation is converted to an ErrorException and it won't be caught.

Also, besides deprecation messages not being shown to users, I think it's worth "fixing" it if it's possible.

PR #19326, opened 2024-10-18
Add getParsedBodyParamAsString

kamil-tekiela:
I propose we add two methods `getParsedBodyParamAsString` and `getParsedBodyParamAsStringOrNull` (maybe 2 more for GET if needed). I would like to hear what is your opinion. 

The type can be `string|null|array` for `getParsedBodyParam`, but I am proposing that we add only two variants. Arrays can be nested and mixed, so they require more specific validation. However, getting a string from Request is a very common pattern. We could reduce a lot of Psalm and PHPStan issues as well as a lot of type hints. Stronger type validation is rarely a bad thing. 

If we proceed with this, I can refactor the rest.


MauricioFauth (approved):
That's great! Thanks!

PR #19352, opened 2024-11-02
[Query Generator] support for `BETWEEN` and `NOT BETWEEN`

faissaloux:
Hello 👋🏻 

I have added support for `BETWEEN` and `NOT BETWEEN` for the Query Generator.

### Current
Screenshot 2024-11-02 183108

### After
Screenshot 2024-11-02 183233

It generates the needed inputs depending on the choosen op:
- op that accepts multiple arguments (eg: IN / NOT IN): It allows to add multiple inputs.
- op that accepts two arguments (eg: BETWEEN / NOT BETWEEN): It generates only the needed two inputs.
- op that accepts one argument (eg: LIKE / NOT LIKE): It generates only the needed one input.
- op that accepts no argument (eg: IS NULL / IS NOT NULL): it removes all inputs.

https://github.com/user-attachments/assets/d5bb1327-5c4d-4e49-872e-5e6de2d9e990

### DIFF
--- a/resources/js/src/database/multi_table_query.ts
+++ b/resources/js/src/database/multi_table_query.ts
@@ -51,6 +51,10 @@ AJAX.registerOnload('database/multi_table_query.js', function () {
         return ['IN (...)', 'NOT IN (...)'];
     }
 
+    function opsWithTwoArgs (): string[] {
+        return ['BETWEEN', 'NOT BETWEEN'];
+    }
+
     $('#update_query_button').on('click', function () {
         var columns = [];
         var tableAliases = {};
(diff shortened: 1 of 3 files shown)


resources/js/src/database/multi_table_query.ts
```diff
@@ -212,23 +219,40 @@ AJAX.registerOnload('database/multi_table_query.js', function () {
...
+        const criteriaInput: JQuery<HTMLElement> = critetiaCol.find('input').first();
+
+        if (critetiaCol.find('.binary').length == 0) {
+            critetiaCol.empty();
+
+            critetiaCol.append(`
+                <div class="options binary">
+                    <input type="text" class="val" placeholder="Enter first value" value="${criteriaInput.val()}" />
+                    <input type="text" class="val" placeholder="Enter second value" />
+                </div>
```
williamdes, line 234: the placeholders are missing translations

resources/js/src/database/multi_table_query.ts
```diff
@@ -212,23 +219,40 @@ AJAX.registerOnload('database/multi_table_query.js', function () {
         });
     });
 
+    function showTwoFields (opSelect: JQuery<HTMLElement>) {
+        const critetiaRow: JQuery<HTMLElement> = opSelect.closest('table').find('.rhs_text');
+        const critetiaCol: JQuery<HTMLElement> = critetiaRow.find('td').last();
+        const criteriaInput: JQuery<HTMLElement> = critetiaCol.find('input').first();
+
+        if (critetiaCol.find('.binary').length == 0) {
```
williamdes, line 227:
```suggestion
        if (critetiaCol.find('.binary').length === 0) {
```

resources/js/src/database/multi_table_query.ts
williamdes: Hi !

> and the translation will be generated by the translationbot right?

Weblate will pull any new commits, merge it into the system and propose the translation to translators.
The translator changes are pushed into the repo branches as changes to .po files.

> How can I generate the lang files locally?

Sure, run: `./bin/generate-mo`

> And how can I use langs locally?

After you run `./bin/generate-mo` the .po files get transformed into .mo files.
And the interface should show you the languages in a list.

williamdes (approved):
Thank you for this contribution !

PR #19356, opened 2024-11-03
Refactor LanguageManager class

MauricioFauth:
Moves the `Language::active()` method to the `LanguageManager` class, as this method does not belong to the `Language` value object.
Some issues reported by static analysis were fixed in the `LanguageManager` class.

### DIFF
--- a/tests/unit/I18n/LanguageManagerTest.php
+++ b/tests/unit/I18n/LanguageManagerTest.php
@@ -13,6 +13,7 @@ use PHPUnit\Framework\Attributes\DataProvider;
 use PHPUnit\Framework\Attributes\Large;
 
 use function _ngettext;
+use function array_map;
 use function count;
 use function file_exists;
 use function is_readable;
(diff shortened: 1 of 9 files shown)


tests/unit/I18n/LanguageManagerTest.php
```diff
@@ -21,47 +22,47 @@
...
         parent::tearDown();
 
         // Ensure we have English locale after tests
-        $lang = $this->manager->getLanguage('en');
-        if ($lang === false) {
-            return;
-        }
+        $languageManager = new LanguageManager(new Config());
+        $lang = $languageManager->getLanguage('en');
+        self::assertNotFalse($lang);
```
kamil-tekiela, line 45: Should we perform assertions in `tearDown`?

MauricioFauth: It's okay to perform assertions in `setUp` and `tearDown`, as they are part of the test case. Of course, it's advisable not to overdo it. In this case, the return being `false` is an issue that should cause the test to fail.

PR #19357, opened 2024-11-03
Don't abuse name attribute of exported gis svg elements

MoonE:
The [name attribute](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/name) should not be used for custom data.


MauricioFauth (approved):

MauricioFauth:
Thank you!

PR #19361, opened 2024-11-05
Fix #19155 - Column sorting does not work when names contain dots

### LINKED ISSUE #19155: Column sorting does not work when names contain dots
### To Reproduce

1. Create a table like this:
```sql
CREATE TABLE `as.df` (
  `i.d` int(11) NOT NULL AUTO_INCREMENT,
  `a.b.c` int(11) NOT NULL,
  PRIMARY KEY (`i.d`)
);

INSERT INTO `as.df` (`i.d`, `a.b.c`) VALUES
(1, 1),
(2, 2);
```
2. Browse table
3. Click column header to change sorting
4. Nothing happens

### Expected behavior

The column should be sorted.

### Server configuration

- Operating system: Linux
- Web server: Apache 2.4
- Database version: MariaDB 10.11.6
- PHP version: 8.2
- phpMyAdmin version: 5.2.2 da10e0b19b7dba9b9f1cb5af841eef9b9b6777da

### Client configuration

- Browser: Firefox 125.0.3
- Operating system: Linux

### Additional context

The code uses `str_contains($clause, '.')` to check whether the name has table and column parts and `explode('.', $clause)` to split  table from column.
https://github.com/phpmyadmin/phpmyadmin/blob/cfb05d22b9512abd4b3c5075bc85c2641f375616/libraries/classes/Display/Results.php#L1758-L1760
The name should have never been concatenated in the first place, better use an array.

This problem affects a few other places where `explode` is used, too.

kamil-tekiela:
Fixes #19155

@MauricioFauth Do you know if there is a better way to write the `parseStringIntoTableAndColumn` function?

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -6729,7 +6729,7 @@ parameters:
 		-
 			message: '#^Cannot access property \$type on PhpMyAdmin\\SqlParser\\Token\|null\.$#'
 			identifier: property.nonObject
-			count: 5
+			count: 10
 			path: src/Display/Results.php
 
 		-
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -6739,6 +6739,18 @@ parameters:
 			path: src/Display/Results.php
 
 		-
+			message: '#^Cannot access property \$type on PhpMyAdmin\\SqlParser\\Token\|null\.$#'
+			identifier: property.nonObject
+			count: 3
+			path: src/Display/Results.php
+
+		-
+			message: '#^Cannot access property \$value on PhpMyAdmin\\SqlParser\\Token\|null\.$#'
+			identifier: property.nonObject
+			count: 6
+			path: src/Display/Results.php
+
+		-
 			message: '#^Cannot cast mixed to int\.$#'
 			identifier: cast.int
 			count: 6
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -6985,12 +6985,6 @@ parameters:
 			path: src/Display/Results.php
 
 		-
-			message: '#^Parameter \#4 \$sortDirection of method PhpMyAdmin\\Display\\Results\:\:getSingleAndMultiSortUrls\(\) expects array\<string\>, array\<mixed\> given\.$#'
-			identifier: argument.type
-			count: 1
-			path: src/Display/Results.php
-
-		-
 			message: '#^Parameter \#6 \$colVisib of method PhpMyAdmin\\Display\\Results\:\:getRowValues\(\) expects array\<mixed\>\|bool\|string, mixed given\.$#'
 			identifier: argument.type
 			count: 1
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -6753,7 +6753,7 @@ parameters:
 		-
 			message: '#^Construct empty\(\) is not allowed\. Use more strict comparison\.$#'
 			identifier: empty.notAllowed
-			count: 24
+			count: 23
 			path: src/Display/Results.php
 
 		-
@@ -6765,7 +6765,7 @@ parameters:
 		-
 			message: '#^Loose comparison via "\=\=" is not allowed\.$#'
 			identifier: equal.notAllowed
-			count: 8
+			count: 6
 			path: src/Display/Results.php
 
 		-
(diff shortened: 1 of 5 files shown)


kamil-tekiela:
I think I fixed all issues. Examples of some issues I fixed:
- columns/tables containing a period in their name
- names containing brackets `(` or `)` in their name
- names containing backticks in their name
- identical names from joined columns
- and more

I still haven't fixed an issue when removing an identical name from a selection. I don't know which code is responsible for that. 

It would be nice to backport this bug fix to 5.2 but it's going to be very annoying.

kamil-tekiela:
@liviuconcioiu If you could also test that this fixes your issue I would appreciate.

liviuconcioiu:
Awesome work!

I've found a small bug, which is also present in `QA_5_2`.

1. Create a table named `
2. Go to SQL - query is OK - `SELECT * FROM ```` WHERE 1`
3. Click on SELECT * - query is NOT OK (it misses a `) - SELECT * FROM ``` WHERE 1

same for all other buttons `SELECT, INSERT, UPDATE, DELETE`.

M393:
There's still an issue when using a column name including schema and table `s`.`t`.`c`.

Thanks for fixing this!

M393:
I'd suggest parsing the column names here into a single array `list<array{order:'ASC'|'DESC', columnParts: list<string>}>`.
This might simplify the method signatures and avoid lexing the same name multiple times.
https://github.com/phpmyadmin/phpmyadmin/blob/e328ec450b2a13e7ce2535ff86d7fa7f9d02e699/src/Display/Results.php#L3097-L3114

kamil-tekiela:
@liviuconcioiu Good find. It looks unrelated to sorting so I will handle that in a separate PR.

kamil-tekiela:
@M393 Thanks for the suggestion. I refactored it all and fixed more bugs.

kamil-tekiela:
@liviuconcioiu I have analysed the issues you found. Could you please create a GH issue for that issue with the buttons? I don't know how to fix it right now. The other issue you linked to is related to Code Mirror.

liviuconcioiu:
> Could you please create a GH issue for that issue with the buttons?

Done.

MoonE:
Just one small problem when manually adding the order by clause without  quotes, the column is not beeing detected as sorted.
This is because in this case the token type is None instead of Symbol.

```sql
SELECT Country FROM opengis.`uk` ORDER BY Country ASC;
```

MoonE:
I wonder if it is really necessary to always prefix the column in the order by clause by the table name.

I kind of miss the new line before the order by clause, I always make new lines in my queries for readability and now it is just appended to the  last line.
Maybe add a new line but only if it is already a multi-line query?

Anyways, thanks for fixing this. :+1:

kamil-tekiela:
>I kind of miss the new line before the order by clause, I always make new lines in my queries for readability and now it is just appended to the last line.
Maybe add a new line but only if it is already a multi-line query?

I have removed the newline because it felt odd that everything else was in one line and the ORDER BY was on its own line. I suppose if there already are newlines in the query then it makes sense to add a newline.

kamil-tekiela:
> Just one small problem when manually adding the order by clause without quotes, the column is not beeing detected as sorted. This is because in this case the token type is None instead of Symbol.
> 
> ```sql
> SELECT Country FROM opengis.`uk` ORDER BY Country ASC;
> ```

Right, good catch! This is not as easy to fix as it seems. I can cover the most common scenarios but covering the edge cases becomes unwieldy. Consider a query such as this:

```
SELECT from_uNixtime(p.temps) FROM `mytable` AS `p` ORDER BY from_unixtime(p.temps) DESC;
```

The result is sorted by its column. I can cover the scenario when the case is identical in both places, but when the case is mismatched it becomes difficult. Table aliases are only case-insensitive on Windows. If the code ignores the case, not only it would  be more logic to be added into the code, but it could also be wrong sometimes. 

Another edge case is using some ridiculous column names:
```
SELECT temps as `from_unixtime(p.temps)` FROM `mytable` AS `p` ORDER BY from_unixtime(p.temps) DESC;
```
The sort condition does not exist in the results table. But how do I figure this out in the code?

I don't know what's a sensible compromise. I propose that I only accept the Token of type None and do not concern myself with the more exotic scenarios.

kamil-tekiela:
> I wonder if it is really necessary to always prefix the column in the order by clause by the table name.

Maybe it's not, but it's easier if it is. I wonder if it's even 100% correct. It is working for the most part. If I try to figure out now when to add the table and when not, I feel like it would complicate the code unnecessarily and potentially break more things.

liviuconcioiu:
@kamil-tekiela can you double check the issue? I think it stopped working after latest PR merges. Now it gives `Error: Object of class PhpMyAdmin\SqlParser\Components\OrderSortKeyword could not be converted to string` when entering a table.

https://github.com/user-attachments/assets/ef72b873-dd89-4ba8-9525-a81c61ca0d48

```SQL
CREATE TABLE ```` (
  `id` int(10) UNSIGNED NOT NULL,
  `{}` varchar(255) DEFAULT NULL,
  `()` varchar(255) DEFAULT NULL,
  ```col``` varchar(255) DEFAULT NULL,
  `.` timestamp NOT NULL DEFAULT current_timestamp(),
  `..` varchar(255) DEFAULT NULL,
  `````` varchar(255) DEFAULT NULL,
  ` ``` int(10) UNSIGNED NOT NULL,
  ` .` int(11) NOT NULL
) ENGINE=MyISAM DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci ROW_FORMAT=DYNAMIC;

INSERT INTO ```` (`id`, `{}`, `()`, ```col```, `.`, `..`, ``````, ` ```, ` .`) VALUES
(1, '2', '4', '7', '2024-11-06 07:29:01', NULL, NULL, 0, 0),
(2, '1', '5', '6', '2024-11-06 07:29:02', NULL, NULL, 0, 0),
(3, 'a', 'b', 'c', '2024-11-06 07:31:08', NULL, NULL, 0, 0);

ALTER TABLE ````
  ADD PRIMARY KEY (`id`);

ALTER TABLE ````
  MODIFY `id` int(10) UNSIGNED NOT NULL AUTO_INCREMENT, AUTO_INCREMENT=4;
COMMIT;
```

kamil-tekiela:
@liviuconcioiu  Are you sure you are using the current PR state? I cannot reproduce this error. I can imagine this happening if you didn't re-pull the PR correctly.

liviuconcioiu:
> @liviuconcioiu Are you sure you are using the current PR state? I cannot reproduce this error. I can imagine this happening if you didn't re-pull the PR correctly.

You were right. I had an older commit.

williamdes (approved):

williamdes:
should this be rebased ?

MauricioFauth (approved):
Thank you!

PR #19362, opened 2024-11-05
Fix issue when updating privileges for various databases

MauricioFauth:
The sql_query global variable should be a string, and it can be a string in this case, which will cause an error as the string will not be converted to array.

### DIFF
--- a/libraries/classes/Server/Privileges.php
+++ b/libraries/classes/Server/Privileges.php
@@ -2539,7 +2539,7 @@ class Privileges
         $message = null;
         $queries = null;
         $queriesForDisplay = null;
-        $sqlQuery = null;
+        $sqlQuery = '';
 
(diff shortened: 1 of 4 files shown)


libraries/classes/Server/Privileges.php
```diff
@@ -2539,7 +2539,7 @@ public function addUser(
         $message = null;
         $queries = null;
         $queriesForDisplay = null;
-        $sqlQuery = null;
+        $sqlQuery = '';
```
kamil-tekiela, line 2542: What was this line for anyway?

MauricioFauth: It's serves no purpose besides setting a default value. This change is a backport of 256a847af73a8eee3c00be4bf1122cca48e98a95. It was easier to trigger the issue with this change.

MauricioFauth: Yes! And the other ones as well.

PR #19378, opened 2024-11-11
Drop $GLOBALS['error_message']

kamil-tekiela:
I couldn't find any assignments.


MauricioFauth (approved):
It indeed looks like dead code. Thank you!

PR #19379, opened 2024-11-11
Don't use exceptions for flow control

kamil-tekiela:
I wonder how much sense it makes to still check for an old exploit from PHP 4 era. Overwriting GLOBALS should no longer be possible without register_globals.


MauricioFauth (approved):
Thank you!

PR #19382, opened 2024-11-11
Refactor TokenRequestParamChecking

kamil-tekiela:
It was a bit tricky to follow the logic here, but I tried to keep the old logic as much as possible. I have to admit I don't fully understand what's going on there.


src/Http/Middleware/TokenRequestParamChecking.php
```diff
@@ -28,43 +30,51 @@ final class TokenRequestParamChecking implements MiddlewareInterface
...
-        if (isset($_POST['token']) && is_scalar($_POST['token']) && (string) $_POST['token'] !== '') {
-            $GLOBALS['token_provided'] = true;
-            $GLOBALS['token_mismatch'] = ! @hash_equals($_SESSION[' PMA_token '], (string) $_POST['token']);
-        }
-
-        if (! $GLOBALS['token_mismatch']) {
-            return;
+        $token = $request->getParsedBodyParamAsString('token', '');
+        if ($token !== '') {
+            if (@hash_equals($_SESSION[' PMA_token '], $token)) {
```
MauricioFauth, line 49: Does it makes sense to have an error suppression here?

src/Http/Middleware/TokenRequestParamChecking.php
```diff
@@ -28,43 +30,51 @@ final class TokenRequestParamChecking implements MiddlewareInterface
...
+            return $responseRenderer->response();
+        }
+
         /**
          * We don't allow any POST operation parameters if the token is mismatched
          * or is not provided.
          */
-        $allowList = ['ajax_request'];
-        Sanitize::removeRequestVars($allowList);
+        $_REQUEST = $_POST = $_GET = $_COOKIE = [];
```
MauricioFauth, line 76: I don't understand this part. Is it even possible to reach this point? If it's possible, does it makes sense to clear the global variables and not request object?

MauricioFauth (commented):
Makes sense to merge both middleware. Thank you!

It's indeed confusing. I'll do more tests before merging this PR.

src/Http/Middleware/TokenRequestParamChecking.php
kamil-tekiela: Yes, it is possible to reach this point, e.g. when the request is not AJAX. 

I don't know if it makes sense not to clear the request object. I don't know if it even makes sense to do this at all. From what I saw the purpose was to empty the variables so that the authorization doesn't succeed. But if the user is already authenticated then I don't know what happens.

src/Http/Middleware/TokenRequestParamChecking.php
kamil-tekiela: No, this function will never produce a supressable warning. I'll remove it.

williamdes (approved):
Looks good to me too

MauricioFauth (approved):

PR #19392, opened 2024-11-16
[PHP 8.4] `str_getcsv`: Fix `$escape` parameter deprecation

Ayesh:
In PHP 8.4, not passing the `$escape` parameter to the `str_getcsv` function emits a deprecation error. This fixes the notice by passing an empty string which disables the buggy escaping mechanism.

 - [Initiating RFC](https://wiki.php.net/rfc/kill-csv-escaping)
 - [Deprecation RFC](https://wiki.php.net/rfc/deprecations_php_8_4#deprecate_proprietary_csv_escaping_mechanism)
 - [PHP 8.4: CSV: The `$escape` parameter must be provided](https://php.watch/versions/8.4/csv-functions-escape-parameter)

### DIFF
--- a/libraries/classes/Util.php
+++ b/libraries/classes/Util.php
@@ -1913,7 +1913,7 @@ class Util
         // It future replace str_getcsv with $dbi->fetchSingleRow('SELECT '.$expressionInBrackets[1]);
 
         preg_match('/\((.*)\)/', $definition, $expressionInBrackets);
-        $matches = str_getcsv($expressionInBrackets[1], ',', "'");
+        $matches = str_getcsv($expressionInBrackets[1], ',', "'", '\\');
 
         $values = [];
         foreach ($matches as $value) {


williamdes (approved):
Thank you!
we should send this into QA_5_2

Ayesh:
Thank you @williamdes, I will check the CI failure and rebase to QA 5.2 branch.

MauricioFauth (approved):

Ayesh:
Thank you for reviewing this @MauricioFauth and @williamdes . 
I rebased these changes against `QA_5_2` branch and force-pushed. We do indeed have tests for the workarounds we have in place for the `str_getcsv` buggy behavior tests passed locally. Thank you.

kamil-tekiela:
I don't know about this change. I wrote this code before I knew that this escaping mechanism is buggy. In fairness, it's all MySQL's fault. What you are proposing is just a temporary fix. If PHP goes ahead with the removal of this feature in PHP 9 then we will be left with no solution for this. We'd probably need a custom parser for this like we had before but less broken. Or we can refactor all of this code and heed my advise from the code comments:

```
        // This should really be delegated to MySQL but since we also want to HTML encode it,
        // it is easier this way.
        // It future replace str_getcsv with $dbi->fetchSingleRow('SELECT '.$expressionInBrackets[1]);
```

Also, I'm not too fond of the fact that this code is duplicated in JS.

williamdes (approved):

Ayesh:
@kamil-tekiela yeah you are right this is just silencing the symptom. In fact, my first attempt at "fixing" this was to pass `escape: ''` to disable the escaping mechanism altogether, but it turned out that phpmyadmin is aware and accounting for this. Thank you.

PR #19406, opened 2024-11-23
PHPUnit fixes

kamil-tekiela:

PR #19407, opened 2024-11-24
Fix ListDatabaseTest

MoonE:
As @kamil-tekiela discovered in https://github.com/phpmyadmin/phpmyadmin/pull/19406/commits/f0a8b5ce776c5a26790e75748ce7ac713fb78ed1, testCheckHideDatabase did not check anything as checkHideDatabase's return type is void.
Though instead of removing the test, it should be fixed.

checkHideDatabase is called in the constructor of ListDatabase, so all that's left  to do is asserting the list is empty.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -22221,15 +22221,6 @@ parameters:
 				Use dependency injection instead\.$#
 			'''
 			identifier: staticMethod.deprecated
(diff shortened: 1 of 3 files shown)


kamil-tekiela (approved):
Good idea! Thanks.

MauricioFauth (approved):

PR #19409, opened 2024-11-25
Drop Protocol version

kamil-tekiela:
This information is no longer relevant. It might have been relevant 20 years ago when protocol 9 was still used, but nowadays I doubt PHP can even support protocol 9. I am not aware of any further iterations of this protocol. This information can be easily confused with the one line above it.


MauricioFauth (approved):

MauricioFauth:
Thank you!

PR #19416, opened 2024-12-02
Add PHPStan stricter rules

MauricioFauth:


MauricioFauth:
> Shouldn't `config.sample.inc.php` and ` examples/` be excluded?

Makes sense to also ignore the sample config, but I think examples should be analyzed.

PR #19422, opened 2024-12-04
Add check fox Sodium extension

kamil-tekiela:
The link doesn't work. I tried to solve it, but to be honest, I do not understand why it's not a direct link to php.net. It sends the URL to the server only for it to generate a JavaScript to redirect to php.net.


liviuconcioiu:
> I do not understand why it's not a direct link to php.net. It sends the URL to the server only for it to generate a JavaScript to redirect to php.net.

Direct links aren't allowed, to prevent phishing and also cross-domain referer leakage.

MauricioFauth (approved):
Thank you!

williamdes (commented):
Thank you !

Fixes https://github.com/phpmyadmin/phpmyadmin/issues/19404#issuecomment-2509742736

PR #19438, opened 2024-12-09
Replace sql_query global var with Current::$sqlQuery

MauricioFauth:


kamil-tekiela (approved):
The changes look fine, but the tests are failing.

src/Controllers/Sql/SqlController.php
```diff
@@ -119,20 +123,20 @@ public function __invoke(ServerRequest $request): Response
...
         // (needed for browsing from DefaultTabTable)
-        if (empty($GLOBALS['sql_query']) && Current::$table !== '' && Current::$database !== '') {
-            $GLOBALS['sql_query'] = $this->sql->getDefaultSqlQueryForBrowse(Current::$database, Current::$table);
+        if (Current::$sqlQuery === '' && Current::$table !== '' && Current::$database !== '') {
+            Current::$sqlQuery = $this->sql->getDefaultSqlQueryForBrowse(Current::$database, Current::$table);
 
             // set $goto to what will be displayed if query returns 0 rows
             UrlParams::$goto = '';
-        } elseif (! $this->response->checkParameters(['sql_query'])) {
+        } elseif (! $this->response->checkParameters(['sql_query'], true)) {
```
MauricioFauth, line 131: @kamil-tekiela Tests failed because of this method call. Without `true`, it was checking using the $GLOBALS instead of $_REQUEST. I think this method should be replaced altogether.

PR #19446, opened 2024-12-13
Removal of $GLOBALS['errorUrl']

kamil-tekiela:
The only place where it was used was in `ImportCsv`. Everywhere else it was either a local use only or completely unused. This feels so wrong and I have a feeling like I missed something obvious, but I really couldn't find any use for this apart from what I mentioned. Is it possible that so much code was dead code?


MauricioFauth (approved):

src/Controllers/Export/ExportController.php
```diff
@@ -170,20 +167,13 @@ public function __invoke(ServerRequest $request): Response
...
             Assert::isArray($tableNames);
             Assert::allString($tableNames);
         } elseif ($GLOBALS['export_type'] === 'table' && Current::$database !== '' && Current::$table !== '') {
-            $GLOBALS['errorUrl'] = Url::getFromRoute('/table/export', [
-                'db' => Current::$database,
-                'table' => Current::$table,
-            ]);
         } elseif ($GLOBALS['export_type'] === 'raw') {
-            $GLOBALS['errorUrl'] = Url::getFromRoute('/server/export', ['sql_query' => Current::$sqlQuery]);
         } else {
```
MauricioFauth, line 177: This created some empty if statements.

MauricioFauth (changes requested):

src/Controllers/Export/ExportController.php
kamil-tekiela: I've changed these checks to match the style of what you have done recently.

MauricioFauth (approved):

PR #19451, opened 2024-12-17
Remove some export plugins global variables

MauricioFauth:


williamdes (approved):

williamdes:
Awesome work !

Do you have a list of all our globals and files with globals ?

MauricioFauth:
> Do you have a list of all our globals and files with globals ?

All globals are using the $GLOBALS superglobal for access. I'm trying to remove the non obvious cases first, like `$GLOBALS[$GLOBALS['what'] . '_null']`.

PR #19454, opened 2024-12-17
Upgrade PHPStan and others

kamil-tekiela:


williamdes (approved):
This is misleading as it also bumps TCPDF

MauricioFauth (approved):

PR #19461, opened 2024-12-19
Creation of fetchResultSimple and fetchSingleColumn

kamil-tekiela:
The method `fetchResult` was a complex method that could very easily be misused. Two new methods were created to more precisely express the intent. A number of related changes were made possible and even necessary after these changes. I believe it also fixes a bug in normalization which was introduced with native parameter types.


MauricioFauth (approved):
That's awesome! Thank you!

PR #19470, opened 2024-12-23
Remove DatabaseInterface::getColumnsFull()

kamil-tekiela:
I've decided to merge the two methods together. I see no evidence here that the `information_schema` SQL would be better than the `SHOW COLUMNS`. There is some extra overhead of getting the indexes but I think it's acceptable.


williamdes (approved):

MauricioFauth (approved):
That's much better! Thank you!

PR #19495, opened 2025-01-03
Config refactoring

kamil-tekiela:
There were two goals of this PR:
- Refactor loading of the config file
- Removal of the Config::get() method

The Config class is very confusing. It stores the same information redundantly many times. It also has a couple of ways of accessing the same information and they are not always in sync with each other. The `get()` method was the least useful of all as it returns a value of type `mixed` which might not even exist in Config, and it can't access deep entries. The config storage was also used for non-config information. I got rid of them, turning them into either new classes, methods or public properties. To remove the `get()` method completely, I had to ensure that the entries that used it were set using the `set()` method which was the best way to keep them in sync despite the significant performance drawback.


williamdes (approved):
it looks great !
not very sure why we parse user agents

kamil-tekiela:
Neither do I. I think the user agent string is sent in the error report anyway and the parsing looks very inaccurate. 

I will rebase it tomorrow.

src/Config/Settings.php
```diff
@@ -504,19 +503,6 @@ final class Settings
...
-     *
-     * ```php
-     * $cfg['OBGzip'] = 'auto';
-     * ```
-     *
-     * @link https://docs.phpmyadmin.net/en/latest/config.html#cfg_OBGzip
-     *
-     * @psalm-var 'auto'|bool
-     */
-    public string|bool $OBGzip;
```
MauricioFauth, line 518: Could you also update the docs? For example:

https://github.com/phpmyadmin/phpmyadmin/blob/4858e7912d0017bda6ddf2501d783630fc8ffe51/docs/config.rst?plain=1#L3686-L3696

MauricioFauth (approved):

src/Config/Settings.php
kamil-tekiela: I have added a new commit. Is this what you meant? I am not very familiar with the syntax of this file. There are also mentions of this in faq.

PR #19503, opened 2025-01-08
Remove global variables

MauricioFauth:
Removes almost all global variables and replaces them with static properties.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -3505,45 +3505,9 @@ parameters:
 			path: src/Controllers/Server/PrivilegesController.php
 
 		-
-			message: '#^Loose comparison via "\=\=" is not allowed\.$#'
-			identifier: equal.notAllowed
-			count: 2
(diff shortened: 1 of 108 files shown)


kamil-tekiela (approved):
I agree with the core concept, but I am not sure if I want to review it in detail. My web browser is lagging. The test is failing though.

williamdes (approved):
Do we have no more globals after this PR ?

MauricioFauth:
> I agree with the core concept, but I am not sure if I want to review it in detail. My web browser is lagging. The test is failing though.

Sorry about this huge pull request. I hyper-focused on this and it end up bigger than I expected.
I basically changed as little as possible to not change behavior. So there's some static properties that don't make much sense and need a refactor. But it's easier to catch these cases now.

> Do we have no more globals after this PR ?

There are only a few globals now. That should be removed in another PR.

williamdes:
PS: I have a work in progress to fix the bot, will try to finish it. So it stops commenting again and again

Issue #19014, opened 2024-02-27
Clearing up the 2FA confusion


codemasher:
Hey there! I just wanted to suggest some changes to the 2FA subsystem.

One of the reasons is that the dependency [`pragmarx/google2fa-qrcode`](https://github.com/antonioribeiro/google2fa-qrcode) is outdated and even seems to be abandoned (a [PR of mine](https://github.com/antonioribeiro/google2fa-qrcode/pull/12), among others, hasn't received a response in over 3 years).

Another reason is to split the 2FA and QR Code libraries in order to permanently add a 2FA dependency, so that baseline security is guaranteed - no matter how PMA was installed (i.e. #16779).

The QR Code library becomes an optional dependency as it is not vital to the functionality of the 2FA backend - it is merely a helper to ease installation on mobile devices - the secret string however will always be displayed during installation, so that it can be copy/pasted or entered manually into an authenticator app.

Further, I have implemented a backup-code functionality (WIP), where a one-time backup code (HOTP) is displayed during installation, so that a user who has lost their authenticator can use this code to gain access.
This flow is supposed to redirect the user to a page where they get a new backup code for the next time they might lose access (not implemented yet).

Now that the QR Code library is entirely decoupled from vital functionality and has only very little use in the backend, it could be moved to the front-end javascript, where it can be used for other things, such as displaying queries or field values etc.

[I made a proof-of-concept](https://github.com/codemasher/phpmyadmin/commit/6f6bf11b88540b3340ceaf04b5ef54cc0fa1594b) intended for v6.x, that with slight changes also works well with PMA v5.2.x.
For this POC i used my own libraries, but any library that satisfies the needs can be used.

- [`chillerlan/php-authenticator`](https://github.com/chillerlan/php-authenticator)
- [`chillerlan/php-qrcode`](https://github.com/chillerlan/php-qrcode)
- [`chillerlan/js-qrcode`](https://github.com/chillerlan/js-qrcode) (as front-end QR Code suggestion)

There are still things to clean up, such as:

- restructure/clean-up the 2FA installation UI
- maybe add 2FA library settings, e.g. hash algo, secret length, ...
- move the "missing library" warning from the 2-factor main page to 2FA installation
- add a redirect/page to obtain a new backup code

Unfortunately the DCO stops me from submitting this as a PR, however, you have my full permission to use any of the suggested code.

Below a screenshot from my local v5.2.1 installation:

[...]

Issue #17567, opened 2022-05-30
No QR code appearing in settings


netfiesta:
I've upgraded from phpMyAdmin 5.0.2 to 5.2.0. Everything seems fine however after the upgrade the QR code to enable 2FA does not show up anymore. It shows a complete white block.
I did a inspect on the generated HTML code and notices this issue:

```
<svg xmlns="http://www.w3.org/2000/svg" version="1.1" width="200" height="200" viewBox="0 0 200 200">
<rect x="0" y="0" width="200" height="200" fill="#fefefe"></rect>
<g transform="scale(3,509)"><g transform="translate(4,4)"><path fill-rule="evenodd" d="[LONG QR CODE]" fill="#000000"></path></g></g></svg>
```

This <g transform="scale(3,509)"> seems to be incorrect. If I change the comma here to a dot, so 3.509 the QR shows fine.
I now made a workaround to fix this but of course it's a workaround...

I've added the following CSS to \templates\preferences\two_factor\configure.twig:

```
	<style>
	/* Added because QR would not appear because of scale(3,509) value instead of 3.509 (dot instead of comma). */
	svg > g {scale: 3.5;}
	</style>
```

williamdes:
Hi @netfiesta 
Could you share the server specific informations ?
What processor architecture do you have ?

Here is the source: https://github.com/Bacon/BaconQrCode/blob/d70c840f68657ce49094b8d91f9ee0cc07fbf66c/src/Renderer/Image/SvgImageBackEnd.php#L100

Could you debug what the input values are ?

netfiesta:
> Hi @netfiesta Could you share the server specific informations ? What processor architecture do you have ?
> 
> Here is the source: https://github.com/Bacon/BaconQrCode/blob/d70c840f68657ce49094b8d91f9ee0cc07fbf66c/src/Renderer/Image/SvgImageBackEnd.php#L100
> 
> Could you debug what the input values are ?

Hello @williamdes,

Thansk for your quick reply. Sure, my environment is a Windows Server 2019 with PHP 7.4.29 x64 NTS and MySQL 8.0.2.
Hmmm, i'm not really a PHP guru, however I have skills in HTML, CSS, JS and C# (.NET) so no clue how I can debug these input values, but maybe you can help me a little with that? ;-)

williamdes:
> > Hi @netfiesta Could you share the server specific informations ? What processor architecture do you have ?
[...]

Issue #16779, opened 2021-03-30
How to Active Two-Factor Auth on phpmyadmin


mrmmg:
Hi
I installed PHPMyAdmin with the terminal on my ubuntu 20 server. when I want to active 2-factor auth this errors appear:

![image](https://user-images.githubusercontent.com/30490118/113034884-d986fb00-91a7-11eb-9bfd-26b7b3235a49.png)

my question is that where I must install these 3 packages with composer? which directory in my server?

Thanks

ibennetch:
Could you please tell us how you have installed phpMyAdmin? Did you
download directly from phpmyadmin.net, use the package manager, or are you
using the Fit repository directly?

mrmmg:
@ibennetch I guess I was used ubuntu package manager (apt)

williamdes:
> my question is that where I must install these 3 packages with composer? which directory in my server?
Thanks

Hi @mrmmg 

There is no way to use 2FA/U2F with the Ubuntu/Debian version.
I am sorry for that, It requires much more work on Debian to enable this.

Ref: https://salsa.debian.org/phpmyadmin-team/phpmyadmin/-/issues/3

As an alternative you can install phpMyAdmin using the files from our [website](https://www.phpmyadmin.net/downloads/).

stratege1401:
> > my question is that where I must install these 3 packages with composer? which directory in my server?
> > Thanks
> 
> Hi @mrmmg
> 
> There is no way to use 2FA/U2F with the Ubuntu/Debian version.
> I am sorry for that, It requires much more work on Debian to enable this.


NOP.  Manual install of phpmyadmin release 5.1.0 over Debian Buster/Bullseye even /sid works for 2FA
Just don't install via apt-get but use phpMyAdmin-latest-all-languages.tar.gz.

williamdes:
> Is there still no way to do this with the version installed through apt?

No, until now I did not manage to create the new packages :/

williamdes:
Hi Debian/Ubuntu users,

I spent some time packaging repositories recently. The next Debian version will have support for Fido U2F: https://tracker.debian.org/news/1399554/accepted-phpmyadmin-4514dfsg1-2-source-into-unstable/

It might also end up on our Ubuntu PPA in some time soon: [Wiki for install methods on Ubuntu/Debian](https://github.com/phpmyadmin/phpmyadmin/wiki/DebianUbuntu#related-issues-and-comments-on-install-methods)

[...]

PR #19506, opened 2025-01-08
2FA: Decouple authenticator and qrcode, permanently add an authenticator library (#19014)

codemasher:
## Description

**Implements #19014**

- Decouples the 2FA and QR Code libraries in order to permanently add a 2FA ("Google Authenticator") dependency, so that baseline security is guaranteed - no matter how PMA was installed  (i.e. #16779, #17567, among others).
- Introduces a backup-code functionality, so that users who have lost their authenticator can regain access. (WIP)
- The QR Code dependency becomes optional and can even be moved to the frontend Javascript and be used for other purposes, as it has only very little use in the backend (no requirements for image generating PHP extensions anymore).
- Removes an outdated, possibly abandoned dependency in the process.

### Authenticator

- Extract into an interface to ease switching 3rd party libraries? (With that said, I'd extract all 3rd party functionality in the subclasses of `PhpMyAdmin\Plugins\TwoFactorPlugin` into interfaces, so that they remain untouched in case a library changes)
- Offer global options for secret length and number of allowed adjacent codes (grace period), and maybe time offset?
- Backup code functionality (currently WIP without function aside of being displayed in the setup screen):
  - I'm not exactly sure what whould be the best course of action in case a backup code was used. Just redirect to the setup screen, so that the user can scan/enter the secret once again? Provide a new secret? Unknown third thing??
  - I'm unsure how to faciliate a redirect in that case, and I might a little help/pointing into the right direction here.
  - Does PMA have functionality to rate limit certain endpoints (to avoid the 2FA endpoint being brute forced), or do we just use a good ol' internal counter (via settings)?

Suggested library: [chillerlan/php-authenticator](https://github.com/chillerlan/php-authenticator)

### QR Code

- Leave in backend?
  - if kept:
    - Extract into an interface to ease switching 3rd party libraries and allow enhanced configuration and/or user defined classes?
      - Add an option for a user-defined class, provide one by default, with an optional composer dependency.
    - Maybe move the "missing qrcode library" message to the setup page, so that a user knows why no QR Code is shown.
  - if not:
    - Figure out possible issues frontend QR Codes could cause (e.g. mobile compatibility: SVG, HTML Canvas, performance, package size).
    - Explore other possible uses, e.g. display of queries and field values (up to ~3kb).

Suggestes libraries: [chillerlan/php-qrcode](https://github.com/chillerlan/php-qrcode), [chillerlan/js-qrcode](https://github.com/chillerlan/js-qrcode)

*I'm going to submit this as a draft for now as there are several open questions that need to be discussed.*

Before submitting the pull request, please review the following checklist:

- [ ] ~~Every commit has proper `Signed-off-by` line as described in our [DCO](https://github.com/phpmyadmin/phpmyadmin/blob/master/DCO). This ensures that the work you're submitting is your own creation.~~ (see https://github.com/phpmyadmin/phpmyadmin/issues/19014#issuecomment-2573927213)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -56,6 +56,7 @@
         "ext-pcre": "*",
         "ext-sodium": "*",
         "ext-xml": "*",
+        "chillerlan/php-authenticator": "^5.2.1",
         "composer/ca-bundle": "^1.2",
         "fig/http-message-util": "^1.1",
         "google/recaptcha": "^1.3",
@@ -80,8 +81,6 @@
         "williamdes/mariadb-mysql-kbs": "^1.2"
     },
     "conflict": {
-        "bacon/bacon-qr-code": "<2.0",
-        "pragmarx/google2fa-qrcode": "<2.1",
         "tecnickcom/tcpdf": "<6.4.4"
     },
     "suggest": {
@@ -93,13 +92,12 @@
         "ext-gd2": "For image transformations",
         "ext-mbstring": "For best performance",
         "tecnickcom/tcpdf": "For PDF support",
-        "pragmarx/google2fa-qrcode": "^2.1 or ^3.0 - For 2FA authentication",
-        "bacon/bacon-qr-code": "^2.0 - For 2FA authentication",
+        "chillerlan/php-qrcode": "^5.0 or higher - For 2FA authentication setup via QR Code",
         "code-lts/u2f-php-server": "For FIDO U2F authentication",
         "web-auth/webauthn-lib": "For better WebAuthn/FIDO2 authentication support"
     },
     "require-dev": {
-        "bacon/bacon-qr-code": "^2.0",
+        "chillerlan/php-qrcode": "^5.0",
         "code-lts/u2f-php-server": "^1.2",
         "guzzlehttp/psr7": "^2.5",
         "httpsoft/http-message": "^1.1",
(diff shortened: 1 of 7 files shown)


composer.json
```diff
@@ -93,13 +92,12 @@
...
         "tecnickcom/tcpdf": "For PDF support",
-        "pragmarx/google2fa-qrcode": "^2.1 or ^3.0 - For 2FA authentication",
-        "bacon/bacon-qr-code": "^2.0 - For 2FA authentication",
+        "chillerlan/php-qrcode": "^5.0 or higher - For 2FA authentication setup via QR Code",
         "code-lts/u2f-php-server": "For FIDO U2F authentication",
         "web-auth/webauthn-lib": "For better WebAuthn/FIDO2 authentication support"
     },
     "require-dev": {
-        "bacon/bacon-qr-code": "^2.0",
+        "chillerlan/php-qrcode": "^5.0",
```
williamdes, line 100: I would perfer that we keep the existing libraries, changing it creates more packaging work for Debian and some other distributions.

codemasher: You can just keep the other library, but then again `chillerlan/php-qrcode` is quite popular and chances are that it already made it into the linux distros - and if not, I'd gladly help.

williamdes: Well, as a Debian maintainer one thing I am sure of is that [baconqrcode](https://tracker.debian.org/pkg/baconqrcode) is already packaged (I maintain it since years)
See: https://repology.org/project/baconqrcode/versions
I searched for your project but did not find it on repology, that might indicate that it is not in any know distributions ?

I respect your work, that is not what I am saying 😃

PR #19511, opened 2025-01-12
Fix Undefined global variable $col_priv

liviuconcioiu:
### Description

Fix `Undefined global variable $col_priv`.

Before:

https://github.com/user-attachments/assets/1d89f0f5-cf66-4ab0-9042-92da207ba7a1

After:

https://github.com/user-attachments/assets/c5ee4158-551e-439c-b482-313a76875b41

Fixes #

### DIFF
--- a/libraries/classes/Normalization.php
+++ b/libraries/classes/Normalization.php
@@ -201,7 +201,7 @@ class Normalization
             'max_rows' => intval($GLOBALS['cfg']['MaxRows']),
             'char_editing' => $GLOBALS['cfg']['CharEditing'],
             'attribute_types' => $this->dbi->types->getAttributes(),
-            'privs_available' => $GLOBALS['col_priv'] && $GLOBALS['is_reload_priv'],
+            'privs_available' => Util::getValueByKey($GLOBALS, 'col_priv', false) && Util::getValueByKey($GLOBALS, 'is_reload_priv', false),
             'max_length' => $this->dbi->getVersion() >= 50503 ? 1024 : 255,
             'charsets' => $charsetsList,
         ]);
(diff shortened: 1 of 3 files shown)


libraries/classes/Normalization.php
```diff
@@ -201,7 +201,7 @@ public function getHtmlForCreateNewColumn(
             'max_rows' => intval($GLOBALS['cfg']['MaxRows']),
             'char_editing' => $GLOBALS['cfg']['CharEditing'],
             'attribute_types' => $this->dbi->types->getAttributes(),
-            'privs_available' => $GLOBALS['col_priv'] && $GLOBALS['is_reload_priv'],
+            'privs_available' => Util::getValueByKey($GLOBALS, 'col_priv', false) && Util::getValueByKey($GLOBALS, 'is_reload_priv', false),
```
kamil-tekiela, line 204: I'd rather prefer that we use `??` here. Or even better figure out why it's undefined in the first place.

williamdes: Agreed, let us know

MauricioFauth, line 204: That's the same as:

```suggestion
            'privs_available' => ($GLOBALS['col_priv'] ?? false) && ($GLOBALS['is_reload_priv'] ?? false),
```

MauricioFauth (changes requested):

libraries/classes/Normalization.php
MauricioFauth: These globals being missing means that `CheckUserPrivileges::getPrivileges()` was not called.

MauricioFauth (approved):

libraries/classes/Normalization.php
kamil-tekiela: Should it be called? Why is the code checking for these variables if the function isn't called?

PR #19515, opened 2025-01-13
Remove the fields_meta global variable

MauricioFauth:
This feature is broken since phpMyAdmin 4.1.0, so it's better to remove that option for now.


williamdes:
> This feature is broken since phpMyAdmin 4.1.0, so it's better to remove that option for now.

Can you edit the config.rst ?

MauricioFauth:
> Can you edit the config.rst ?

There no config directive for this option.

williamdes (approved):

kamil-tekiela (approved):

PR #19517, opened 2025-01-14
Fix #18530 - Fix error when there are no favorite tables

### LINKED ISSUE #18530: TypeError: PhpMyAdmin\Controllers\Database\Structure\FavoriteTableController::synchronizeFavoriteTables(): Argument #3 ($favoriteTables) must be of type array, null given
### Describe the bug

Internal error in .\libraries\classes\Controllers\Database\Structure\FavoriteTableController.php#146
TypeError: PhpMyAdmin\Controllers\Database\Structure\FavoriteTableController::synchronizeFavoriteTables(): Argument #3 ($favoriteTables) must be of type array, null given, called in .\libraries\classes\Controllers\Database\Structure\FavoriteTableController.php on line 66

Introduced by the yesterday commits.

### To Reproduce

1. Log in to phpMyAdmin

### Expected behavior

No error should happen.

### Server configuration

- phpMyAdmin version: 5.2.2-dev+20230714.a0d30e3de1

liviuconcioiu:
### Description

@williamdes please double check this to make sure is okay.

This check is the same as the one from here https://github.com/phpmyadmin/phpmyadmin/blob/fe25c103d7cf78895fdb98cdfa9f404ca5e9c4ee/js/src/navigation.js#L533-L540.

The value should be `null`, because this is what we check in PHP:

https://github.com/phpmyadmin/phpmyadmin/blob/fe25c103d7cf78895fdb98cdfa9f404ca5e9c4ee/libraries/classes/Controllers/Database/Structure/FavoriteTableController.php#L39
https://github.com/phpmyadmin/phpmyadmin/blob/fe25c103d7cf78895fdb98cdfa9f404ca5e9c4ee/libraries/classes/Controllers/Database/Structure/FavoriteTableController.php#L53

I've tested and it works as expected. The error no longer shows, and if there are no favorite tables, there's no request made.

Tested it with favorite tables too, and I can see them, both in the top-left corner and also in table list, and also can be removed.

Fixes #18530

### DIFF
--- a/js/src/functions.js
+++ b/js/src/functions.js
@@ -3820,15 +3820,14 @@ AJAX.registerOnload('functions.js', function () {
 
     // Sync favorite tables from localStorage to pmadb.
     if ($('#sync_favorite_tables').length) {
-        var favoriteTables = '';
-        if (isStorageSupported('localStorage')
-            && typeof window.localStorage.favoriteTables !== 'undefined'
-            && window.localStorage.favoriteTables !== 'undefined') {
-            favoriteTables = window.localStorage.favoriteTables;
-            if (favoriteTables === 'undefined') {
-                // Do not send an invalid value
-                return;
-            }
+        var hasLocalStorage = isStorageSupported('localStorage') &&
+            typeof window.localStorage.favoriteTables !== 'undefined' &&
+            window.localStorage.favoriteTables !== 'undefined';
+        var favoriteTables = hasLocalStorage ? window.localStorage.favoriteTables : null;
+
+        if (favoriteTables === null) {
+            // Do not send an invalid value
+            return;
         }
         $.ajax({
             url: $('#sync_favorite_tables').attr('href'),
--- a/js/src/navigation.js
+++ b/js/src/navigation.js
@@ -531,13 +531,14 @@ $(function () {
         }
 
         var hasLocalStorage = isStorageSupported('localStorage') &&
-            typeof window.localStorage.favoriteTables !== 'undefined';
+            typeof window.localStorage.favoriteTables !== 'undefined' &&
+            window.localStorage.favoriteTables !== 'undefined';
         $.ajax({
             url: $self.attr('href'),
             cache: false,
             type: 'POST',
             data: {
-                'favoriteTables': hasLocalStorage ? window.localStorage.favoriteTables : '',
+                'favoriteTables': hasLocalStorage ? window.localStorage.favoriteTables : null,
                 'server': CommonParams.get('server'),
             },
             success: function (data) {


js/src/navigation.js
```diff
@@ -531,13 +531,14 @@ $(function () {
...
-            typeof window.localStorage.favoriteTables !== 'undefined';
+            typeof window.localStorage.favoriteTables !== 'undefined' &&
+            window.localStorage.favoriteTables !== 'undefined';
         $.ajax({
             url: $self.attr('href'),
             cache: false,
             type: 'POST',
             data: {
-                'favoriteTables': hasLocalStorage ? window.localStorage.favoriteTables : '',
+                'favoriteTables': hasLocalStorage ? window.localStorage.favoriteTables : null,
```
kamil-tekiela, line 541: Does it actually make any difference? I am not that familiar with the JS.

liviuconcioiu: In PHP it checks for null, and here is string.

williamdes (approved):
This looks good !
A nice fix

williamdes:
` if (! empty($parameters['favoriteTables'])) {` would have done the trick, but it is not clean

js/src/navigation.js
kamil-tekiela: Yes, but how does this value get to PHP? Where is the logic that converts the JS null to PHP null?

liviuconcioiu: I could not find it neither the logic, just changed it to be similar with the one I fixed, in case the code will be changed in the future. Easier to find the same piece of code.

I also think there's some rendundant code, because I removed this code, and it fixes the issue and also favorites work, evern after clearing browser cache, restart it, private window, etc. Did not dig any further.

https://github.com/phpmyadmin/phpmyadmin/pull/19518

kamil-tekiela: Ok, so I just tested this myself and there is no mechanism to transfer `null` to PHP. So this should be an empty string as it was.

js/src/functions.js
```diff
@@ -3820,15 +3820,14 @@ AJAX.registerOnload('functions.js', function () {
...
-            if (favoriteTables === 'undefined') {
-                // Do not send an invalid value
-                return;
-            }
+        var hasLocalStorage = isStorageSupported('localStorage') &&
+            typeof window.localStorage.favoriteTables !== 'undefined' &&
+            window.localStorage.favoriteTables !== 'undefined';
+        var favoriteTables = hasLocalStorage ? window.localStorage.favoriteTables : null;
+
+        if (favoriteTables === null) {
```
kamil-tekiela, line 3828: This looks like it could fix the issue (I don't know JS well enough and I don't know how to test it) but the PHP code needs to be fixed too. This action should call its own controller and not reuse the database structure one. Let me see if I can tidy this up a little.

liviuconcioiu:
I think this should be closed, in favor of https://github.com/phpmyadmin/phpmyadmin/pull/19520

PR #19519, opened 2025-01-14
Remove plugin global variables

MauricioFauth:
Those are the last global variables in the code base. So after #19515, #19516 and this PR, all global variables (`$GLOBALS`) will be removed.

- #19515
- #19516

### DIFF
--- a/src/Plugins.php
+++ b/src/Plugins.php
@@ -32,15 +32,14 @@ use function class_exists;
 use function count;
 use function htmlspecialchars;
 use function is_array;
+use function is_string;
 use function is_subclass_of;
 use function mb_strtolower;
 use function mb_strtoupper;
 use function mb_substr;
 use function method_exists;
-use function preg_match_all;
 use function sprintf;
 use function str_contains;
-use function str_replace;
 use function str_starts_with;
 use function strcasecmp;
 use function usort;
@@ -175,18 +174,6 @@ class Plugins
     }
 
     /**
-     * Returns locale string for $name or $name if no locale is found
-     *
-     * @param string|null $name for local string
-     *
-     * @return string  locale string for $name
-     */
-    public static function getString(string|null $name): string
-    {
-        return $GLOBALS[$name] ?? $name ?? '';
-    }
-
-    /**
      * Returns html input tag option 'checked' if plugin $opt
      * should be set by config or request
      *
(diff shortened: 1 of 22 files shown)


src/Plugins/ExportPlugin.php
```diff
@@ -342,4 +342,9 @@ protected function setStructureOrData(
             ?? StructureOrData::tryFrom(is_string($valueFromConfig) ? $valueFromConfig : '')
             ?? $defaultValue;
     }
+
+    public function getTranslatedText(string $text): string
+    {
```
williamdes, line 347: Can you refresh the readme.md of plugins & wiki page?

williamdes (approved):

src/Plugins.php
```diff
@@ -576,7 +553,7 @@ public static function getOptions(string $section, array $list): string
 
             $ret .= '<div id="' . $pluginName
                 . '_options" class="format_specific_options">';
-            $ret .= '<h3>' . self::getString($text) . '</h3>';
+            $ret .= '<h3>' . $plugin->getTranslatedText($text) . '</h3>';
```
kamil-tekiela, line 556: Why some of them pass through `htmlspecialchars` and some don't?

src/Plugins.php
```diff
@@ -488,22 +463,22 @@ public static function getHtmlForProperty(
                 $ret .= '<li class="list-group-item">' . "\n";
                 $ret .= '<label for="select_' . $pluginName . '_'
                     . $pitem->getName() . '" class="form-label">'
-                    . self::getString($pitem->getText()) . '</label>';
+                    . $plugin->getTranslatedText($pitem->getText() ?? '') . '</label>';
```
kamil-tekiela, line 466: This is nullable for no reason it seems. The only null check is in `getOneOption` and I doubt it was intentional.

src/Plugins.php
MauricioFauth: htmlspecialchars was extracted from Plugins::getDefault(). There wasn't htmlspecialchars for Plugins::getString() and I didn't tested if it's necessary.

src/Plugins/ExportPlugin.php
MauricioFauth: There are some other changes I want to do first.

src/Plugins.php
kamil-tekiela: Would it cause any harm to have it there anyway? It looks very inconsistent now.

PR #19520, opened 2025-01-14
Avoid passing null to synchronizeFavoriteTables

kamil-tekiela:
Fixes #18530 This PR DOES NOT invalidate the other one.

PR #19521, opened 2025-01-14
Create a new controller for synchronizing favourites

kamil-tekiela:
Related to #19520

@liviuconcioiu Could you please test it?


kamil-tekiela:
I don't know why Psalm is acting up.

liviuconcioiu:
> Could you please test it?

Can't find any issues. Favorites work.

src/Controllers/SyncFavoriteTablesController.php
```diff
@@ -0,0 +1,76 @@
...
+
+        $favoriteInstance = RecentFavoriteTables::getInstance(TableType::Favorite);
+
+        $favoriteTables = json_decode($request->getParsedBodyParamAsString('favoriteTables'), true);
+        if (! is_array($favoriteTables)) {
+            $favoriteTables = [];
+        }
+
+        // Required to keep each user's preferences separate.
+        $user = sha1(Config::getInstance()->selectedServer['user']);
```
williamdes, line 46:
```suggestion
        $user = hash('sha1', Config::getInstance()->selectedServer['user']);
```

kamil-tekiela: I can change it but why? What difference does it make?

williamdes: not sure, I would say we should only use hash to hash everything
removing the need for a specific function

MauricioFauth (approved):

PR #19523, opened 2025-01-15
Fix #19384 - varbinary equal search

### LINKED ISSUE #19384: Searching on a VARBINARY field
### Describe the bug

VARBINARY are by default displayed as hex, but copying that hex and searching on it doesn't work. Search adds quotes around the hex search string, which searches for the string of the hex instead of the hex.

```SELECT * FROM `log` WHERE `ip_address` = '0x7f000001'; ```

### Server configuration

- Operating system: Linux 5.15.0-119-generic Ubuntu 22.04
- Web server: Apache/2.4.52
- Database version:  libmysql - mysqlnd 8.1.29 -  10.6.18-MariaDB-0ubuntu0.22.04.1
- PHP version:  8.1.29 
- phpMyAdmin version:  5.1.1deb5ubuntu1

### Additional context

related: #8046, #4374

### PR DESCRIPTION
**Pull request by faissaloux**
Fixes #19384

While I was investigating this issue, I have found that all ops does not work with `varbinary` as it should.

In this PR I have fixed the equal op with `varbinary` field as it's the one reported now, other ops should be fixed on other PRs.

#### In case the value starts with `0x`
image

#### In case the value is a simple string
image

### CODE REVIEW

File: libraries/classes/Table/Search.php [hunk 1]
```diff
@@ -197,6 +197,8 @@ private function getWhereClause(
                 && $func_type !== 'BETWEEN'
                 && $func_type !== 'NOT BETWEEN'
             ) {
+                $quot = $names === 'varbinary' ? '' : $quot;
```
williamdes, line 200: should it not be in the line above ` preg_match('@char|binary|blob|text|set|date|time|year|uuid@i', $types)` ?

faissaloux: Yeah you are right! I didn't see that line, I'll check it tonight.

kamil-tekiela:
But that breaks the current behaviour. What if I really do want to search for a string like I can do with all other types?

faissaloux:
> But that breaks the current behaviour. What if I really do want to search for a string like I can do with all other types?

Do you mean you need to search for a sting on a `varbinary` field?

kamil-tekiela:
> > But that breaks the current behaviour. What if I really do want to search for a string like I can do with all other types?
> 
> Do you mean you need to search for a sting on a `varbinary` field?

Yes, that should be possible by default.

faissaloux:
> > > But that breaks the current behaviour. What if I really do want to search for a string like I can do with all other types?
> > 
> > 
> > Do you mean you need to search for a sting on a `varbinary` field?
> 
> Yes, that should be possible by default.

Isn't this what `= "` operator is for?
- `=` shouldn't add any quotes.
- `="` should add quotes.

Can you confirm this.

image

kamil-tekiela:
> Isn't this what = " operator is for?

No, that's not what this is for. This option allows you to search for an empty string. Try it, you can't enter the value because the condition will always be an empty string.

faissaloux:
> > Isn't this what = " operator is for?
> 
> No, that's not what this is for. This option allows you to search for an empty string. Try it, you can't enter the value because the condition will always be an empty string.

Oh it's confusing, so this is another issue that I should fix later on. Noted!

Can you give me an example where you gonna need quotes on a binary field please.

kamil-tekiela:
> Can you give me an example where you gonna need quotes on a binary field please.

I am not sure what you're asking but basically any text would need quotes. 

```
SELECT * FROM t WHERE vbin = 'foo';
```

faissaloux:
So I'll add the quotes only if the inserted value is a simple string like `foo`, if it's not (eg: `0xfoo`) quotes should not be used.

kamil-tekiela:
Yeah, that could work but how will you know when it's a special value?

faissaloux:
> Yeah, that could work but how will you know when it's a special value?

Yeah exactly that's what I'm thinking about, the only idea I have now is check if the string starts with `0x` we don't add the quotes.

libraries/classes/Table/Search.php
```diff
@@ -167,7 +167,10 @@ private function getWhereClause(
             // strings to numbers and numbers to strings as necessary
             // during the comparison
             if (
-                preg_match('@char|binary|blob|text|set|date|time|year|uuid@i', $types)
+                (
+                    preg_match('@char|binary|blob|text|set|date|time|year|uuid@i', $types)
```
williamdes, line 171: this new version of the check is confusing me

faissaloux: Can you elaborate.

williamdes: Yes, sorry
why was the varbinary type not added to the list ?
is there a case where varbinary is different from binary and not used in the 0x form ?

faissaloux: > why was the varbinary type not added to the list ?

I didn't have to add it since there is `binary` (substring) that detects it.

> is there a case where varbinary is different from binary and not used in the 0x form ?

As @kamil-tekiela said and I tested it yes, it can be stored as normal string.

williamdes (approved):
Okay, thank you both for the review and work !

libraries/classes/Table/Search.php
kamil-tekiela: I still don't know about this. What if I search for a string `0x`? Then the search query will be broken and now you have another more serious bug. I'd rather either not change anything or create a separate option allowing the user to provide a self-quoted string/numerical value.

kamil-tekiela:
In other words, with this change, you're going to break the current functionality. All the available options such as `LIKE`, `NOT LIKE`, `REGEXP` are string searches. If you want to introduce a numerical search then it should be a separate option.

PR #19524, opened 2025-01-15
Fix #19351 - paging with approximate number of rows

### LINKED ISSUE #19351: [BUG] PMA doesn't show all/last rows of a "larger" table > 125k rows
### Describe the bug

PMA doesn't show all and/or last rows of a table when showing the table and using "go to last page".

When examining the structure of the database PMA tells me that for this table there is ~144.879 rows, clicking on this triggers the real count 146.020 like described in the PMA FAQ point 3.11. The problem when viewing this table and using last page, PMA still refers to this approximately number resulting in "missing rows".

Now to the funpart, changing the count of datasets from standard 25 to 250 on this last page resulting in the same info output

`Zeige Datensätze 144875 - 144878 (144879 insgesamt, Die Abfrage dauerte 0.1275 Sekunden.)` (sry PMA is set to german but i think you get it)

BUT i got 225 more rows shown, according to the last primary key id.

Im not into your sourcecode, the query shown is `SELECT * FROM table_name` but i think internally you will do something like that to query and show the data `SELECT * FROM table_name LIMIT $pagenumber * $count dataset, $count_dataset;`

In my case after switching to 250 datasets to show, last page is 580 which is the rounded output of (144.879/250), but as mentioned, this is not the real number. The table has a normal primary key with autoincrement and PMA also shows the correct next Autoindex when examining the structure of the table.

When manually running `SELECT * FROM table_name limit 146000, 250;` i get the last 20 rows `Zeige Datensätze 146000 - 146019 (20 insgesamt, Die Abfrage dauerte 0.1280 Sekunden.)`

### used software
- latest PMA dockerimage 5.2.1
- mariadb dockerimage
- customer database (inno db)

### Expected behavior

When the last page query result in a "full page result" like 25 of 25 datasets or 250 of 250, PMA should assume, that there is or could be still another page and should set the last page to last page + 1 as long there as there are rows returned.

i tested other databases and tables and the difference between the approximately number and the real number of rows was around 1000-2000 max. so when looping with the obove logic and 25 datasets this would result in 40-80 iterations of that loop or maybe using another approach internally like using a greater limit internally like 100 which would result in lesser iterations.

Or PMA should query a `select count(*) from table_name` to get the real number for calculating the lastpage. I found the Hint, that $cfg['MaxExactCount'] Changed in version 4.8.0: The default value was lowered to 50000 for performance reasons. So the count query isn't the way but maybe another logic?

### other Infos

Examining the database with other tools like Adminer or HeidiSQL is working "normal" and last page or last dataset is really the last row of the table, but customer is used to PMA.

kamil-tekiela:
Fixes #19351

My brain isn't working today, but I believe I managed to fix this bug. I'll try to come up with a cleaner version on master later.

### DIFF
--- a/libraries/classes/Display/Results.php
+++ b/libraries/classes/Display/Results.php
@@ -840,12 +840,16 @@ class Results
         // Move to the next page or to the last one
         $moveForwardButtons = '';
         if (
+            ! (
             // view with unknown number of rows
-            ($this->properties['unlim_num_rows'] === -1 || $this->properties['unlim_num_rows'] === false)
-            || (! $isShowingAll
-            && intval($_SESSION['tmpval']['pos']) + intval($_SESSION['tmpval']['max_rows'])
-                < $this->properties['unlim_num_rows']
-            && $this->properties['num_rows'] >= $_SESSION['tmpval']['max_rows'])
+            $this->properties['unlim_num_rows'] !== -1
+            && $this->properties['unlim_num_rows'] !== false
+            && ($isShowingAll
+                || ( $this->isExactCount()
+                    && (int) $_SESSION['tmpval']['pos'] + (int) $_SESSION['tmpval']['max_rows']
+                    >= $this->properties['unlim_num_rows'])
+                || $this->properties['num_rows'] < $_SESSION['tmpval']['max_rows'])
+            )
         ) {
             $moveForwardButtons = $this->getMoveForwardButtonsForTableNavigation(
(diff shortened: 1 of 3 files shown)


kamil-tekiela:
This is driving me insane, I am giving up.

williamdes:
> This is driving me insane, I am giving up.

The diff is a headache, will check once more soon ^^

MauricioFauth (approved):
That's confusing, but looks like the right fix.

PR #19532, opened 2025-01-17
Merge ColumnFull & Column

kamil-tekiela:
I did this 3 months ago but I forgot about it. I rebased it now.
I require it for further PRs.


williamdes (approved):
At some point DisableIS will have no more sense or implementations.
Right ?

kamil-tekiela:
> At some point DisableIS will have no more sense or implementations. Right ?

Yes, that's correct.

MauricioFauth (approved):

ibennetch:
> Do we know why this was important and implemented ?

DisableIS was originally introduced to work around some limitations in older MySQL versions. When `information_schema` was first introduced, it had some major performance problems, especially with large numbers of databases or tables (such as in a shared hosting situation).

Whether MySQL fixed that or if MariaDB was ever affected, I'm not sure, but it seems to be better now than it was at first.

Ref:
* https://bugs.mysql.com/bug.php?id=19588
* https://stackoverflow.com/questions/42465926/why-is-using-information-schema-too-much-slow
* https://www.percona.com/blog/solving-information_schema-slowness/
* https://dba.stackexchange.com/questions/334569/information-schema-tables-take-so-long-when-used-in-certain-ways

kamil-tekiela:
I am still not keen on outright getting rid of that setting, but more and more it makes less sense. While these two options are meant to be replacements for each other, they are not identical. The differences make it difficult to allow the use of both of them.

liviuconcioiu:
@kamil-tekiela can you double check this? I've narrowed down to this PR, some strange bug. It doesn't recognize `NULL` values:

null

liviuconcioiu:
When the column is edited, it shows NULL as default value
edit

liviuconcioiu:
```SQL
CREATE TABLE `tbl_test` (
  `col1` varchar(255) DEFAULT NULL,
  `col2` varchar(255) DEFAULT NULL,
  `col3` varchar(20) NOT NULL,
  `col4` text DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=MyISAM DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci
```

PR #19538, opened 2025-01-18
Fix #19537 - Make sure rowCount is always int

### LINKED ISSUE #19537: TypeError: PhpMyAdmin\Util::formatNumber(): Argument #1 ($value) must be of type string|int|float, null given
### Describe the bug

demo server:
```
TypeError: PhpMyAdmin\Util::formatNumber(): Argument #1 ($value) must be of type string|int|float, null given, called in /home/demo/.pma-demo-web/master-config/src/Controllers/Database/Structure/RealRowCountController.php on line 76
```

localhost:

```
Internal error in .\src\Util.php#365
TypeError: PhpMyAdmin\Util::formatNumber(): Argument #1 ($value) must be of type string|int|float, null given, called in E:\GitHub\phpmyadmin\src\Controllers\Database\Structure\RealRowCountController.php on line 76
```

### To Reproduce

1. Go to https://demo.phpmyadmin.net/master-config/public/index.php?route=/database/structure&db=information_schema&server=1
2. Click on row count at the bottom or `Rows` column
3. See error

### Expected behavior

An exact count of the rows should be given, without errors.

### Screenshots

https://github.com/user-attachments/assets/4290c803-30cf-4d3b-bcfe-167ae7a1347a

### Server configuration

liviuconcioiu:
### Description

Fixes #19537

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -3505,12 +3505,6 @@ parameters:
 			path: src/Controllers/Server/Status/StatusController.php
 
 		-
-			message: '#^Parameter \#1 \$value of static method PhpMyAdmin\\Util\:\:formatNumber\(\) expects float\|int\|string, mixed given\.$#'
-			identifier: argument.type
-			count: 5
-			path: src/Controllers/Server/Status/StatusController.php
-
-		-
 			message: '#^Binary operation "\*" between 0\.01 and float\|int\|string results in an error\.$#'
 			identifier: binaryOp.invalid
 			count: 1
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1234,9 +1234,6 @@
     <PossiblyInvalidCast>
       <code><![CDATA[$parameters['table']]]></code>
     </PossiblyInvalidCast>
-    <PossiblyNullArgument>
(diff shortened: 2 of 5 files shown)


faissaloux:
This method is used as `int` everywhere, so I think it's better to change it from here and make sure it always returns `int`.

https://github.com/phpmyadmin/phpmyadmin/blob/38bbaff2af3009de6d94d5fa19858f200d8b643a/src/Table/Table.php#L1745

kamil-tekiela:
You can even make this function simpler by using `fetchValue` and removing `Util::backquote('row_count')`.

kamil-tekiela (approved):
That's very good, but I am thinking now whether we should actually show some information to the user that it's impossible to fetch the row count. For example, instead of displaying 0 show a message such as N/A.

liviuconcioiu:
Wouldn't we run then in more potential bugs? There are other functions that have to be changed for this change.

Also, instead of N/A, probably `unknown` can be used, since it's also used on `Size` column.

https://github.com/user-attachments/assets/438009ea-028a-41b2-b2ea-439aea66a7fe

williamdes (approved):
the diff is okay

MauricioFauth (approved):

PR #19541, opened 2025-01-18
Optimization: Memoize SpecialSchemaLinks

kamil-tekiela:
Assuming that the Config doesn't change during the execution of `getRowValues()` the values should remain constant. There's no need to compile the list again and again for every row. This should improve the performance of the `route=/sql` by up to 10%.


MauricioFauth (approved):
Great! Thank you!

PR #19542, opened 2025-01-18
Made Search operators `= ''/ != ''` more obvious

faissaloux:
First time I have seen these two operators in the `Search` I was confused, I thought they add quotes to the input ( https://github.com/phpmyadmin/phpmyadmin/pull/19523#issuecomment-2593885922 ).

image

So in this PR I made them more clear by replacing them with `EMPTY` and `NOT EMPTY`.

image

#### Example
image

#### Generated query
image


js/src/table/select.js
```diff
@@ -92,8 +92,8 @@ AJAX.registerOnload('table/select.js', function () {
         var unaryFunctions = [
             'IS NULL',
             'IS NOT NULL',
-            '= \'\'',
-            '!= \'\''
+            'EMPTY',
+            'NOT EMPTY',
```
MauricioFauth, line 96: I don't like this. Looks like `EMPTY` and `NOT EMPTY` are SQL statements.

kamil-tekiela: I agree. The way it is now it fine

faissaloux: Emm yeah it makes sense, I'll add it this weekend!

faissaloux:
@MauricioFauth @kamil-tekiela What about this?

image

libraries/classes/Types.php
```diff
@@ -105,8 +105,8 @@ public function getTextOperators()
             'REGEXP',
             'REGEXP ^...$',
             'NOT REGEXP',
-            "= ''",
-            "!= ''",
+            '= \'\' (empty)',
+            '!= \'\' (not empty)',
```
MauricioFauth, line 109: I don't think **empty** and **not empty** should be included in the value.

I think this will be better if only the display value of the `select` show that.
This can be changed in the method `Types::getTypeOperatorsHtml()`. And **empty** and **not empty** should be translatable.

Also, as this pull request is an enhancement, it should target the `master` branch.

MauricioFauth (changes requested):

MauricioFauth (approved):
Thank you!

PR #19551, opened 2025-01-24
Small refactoring

kamil-tekiela:


src/Config/Settings.php
```diff
@@ -1832,18 +1832,18 @@ final class Settings
     /**
      * You can select here which functions will be used for character set conversion.
      * Possible values are:
-     *      auto   - automatically use available one (first is tested iconv, then mbstring)
-     *      iconv  - use iconv or libiconv functions
-     *      mb     - use mbstring extension
-     *      none   - disable encoding conversion
+     *      auto     - automatically use available one (first is tested iconv, then mbstring)
+     *      iconv    - use iconv or libiconv functions
+     *      mbstring - use mbstring extension
```
MauricioFauth, line 1837: You should also update the docs.

https://github.com/phpmyadmin/phpmyadmin/blob/b6a51a0b906d6d53e2b585acceefd273e2b59cb8/docs/config.rst?plain=1#L3003

src/Display/Results.php
```diff
@@ -112,9 +112,9 @@ class Results
     /**
      * the total number of rows returned by the SQL query without any appended "LIMIT" clause programmatically
      *
-     * @var int|numeric-string|false
+     * @var int|numeric-string
      */
-    private int|string|false $unlimNumRows = 0;
+    private int|string $unlimNumRows = 0;
```
MauricioFauth, line 117: This could cause some issues. See:
- #17059
- #17055
- #17356

MauricioFauth (changes requested):

src/Config/Settings.php
kamil-tekiela: I didn't change anything there. The value in `config.inc` is still `mb`.

src/Display/Results.php
kamil-tekiela: That is true. It's broken currently. I am working on a fix, but could you please review #19524 first? I have removed this commit from this PR for now. I will create a new one separately.

MauricioFauth (approved):

Issue #19337, opened 2024-10-27
Singe quotes around unix_timestamp(current_timestamp(6))


liviuconcioiu:
### Describe the bug

There are single quotes on Insert, when the default value is `unix_timestamp(current_timestamp(6))`.

### To Reproduce

1. Create table:
```SQL
CREATE TABLE `tbl_test` (
  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
  `created_at` double(16,6) DEFAULT unix_timestamp(current_timestamp(6)),
  PRIMARY KEY (`id`)
) ENGINE=MyISAM DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;
```
2. Insert values
3. See error

### Expected behavior

There should be no singe quotes around `unix_timestamp(current_timestamp(6))` when inserting.

```SQL
INSERT INTO `tbl_test` (`id`, `created_at`) VALUES (NULL, unix_timestamp(current_timestamp(6)));
```

### Screenshots

https://github.com/user-attachments/assets/2f8d9f3a-47e1-40ee-8aa0-91ec69ba105b

### Server configuration

- phpMyAdmin version: 5.2.2-dev, 6.0.0-dev

kamil-tekiela:
Why do you expect there to be no quotes? It makes perfect sense to me for it to have quotes when the type is double.

liviuconcioiu:
> Why do you expect there to be no quotes? It makes perfect sense to me for it to have quotes when the type is double.

I understand this and you are absolutely right, but since that table schema is fine, there's no way to insert data from phpMyAdmin. In cases like this for example, I think there should be no single quotes, otherwise UX is broken.

williamdes:
Can we say this is fixed or still a bug to fix ?

kamil-tekiela:
I'd say keep it open. I stopped working on it because I found a security issue and then I forgot about it. I'll come back to it.

PR #19552, opened 2025-01-24
Fix default values

kamil-tekiela:
This is not finished but it should fix the majority of the problem. @liviuconcioiu Could you please check if this fixes the problem?

Related to #19532


liviuconcioiu:
Yes! It works!

williamdes (approved):

kamil-tekiela:
This is the next small step towards implementing #19337

MauricioFauth (approved):

PR #19555, opened 2025-01-26
Add button for Copy Query

faissaloux:
https://github.com/user-attachments/assets/125ce4a8-0d57-483f-bab0-31b48a28ad3f

### DIFF
--- a/resources/js/src/database/multi_table_query.ts
+++ b/resources/js/src/database/multi_table_query.ts
@@ -194,8 +194,8 @@ AJAX.registerOnload('database/multi_table_query.js', function () {
     $('#copy_query').on('click', function (): void {
         const query: string = editor.getDoc().getValue();
 
-        copyToClipboard(query, '<textarea>');
-        displayCopyNotification();
+        const copyStatus = copyToClipboard(query, '<textarea>');
+        displayCopyNotification(copyStatus);
     });
 
     $('#add_column_button').on('click', function () {
--- a/resources/js/src/database/multi_table_query.ts
+++ b/resources/js/src/database/multi_table_query.ts
@@ -185,14 +185,14 @@ AJAX.registerOnload('database/multi_table_query.js', function () {
         });
     });
 
-    editor.getDoc().on('change', function () {
-        const query = editor.getDoc().getValue();
+    editor.getDoc().on('change', function (): void {
+        const query: string = editor.getDoc().getValue();
 
         $('#copy_query').prop('disabled', query === '');
     });
 
-    $('#copy_query').on('click', function () {
-        const query = editor.getDoc().getValue();
+    $('#copy_query').on('click', function (): void {
+        const query: string = editor.getDoc().getValue();
 
         copyToClipboard(query, '<textarea>');
         ajaxShowMessage(window.Messages.strCopyQueryButtonSuccess, 1000);
--- a/resources/js/src/database/multi_table_query.ts
+++ b/resources/js/src/database/multi_table_query.ts
@@ -195,6 +195,7 @@ AJAX.registerOnload('database/multi_table_query.js', function () {
         const query = editor.getDoc().getValue();
 
         navigator.clipboard.writeText(query);
+        ajaxShowMessage(window.Messages.strCopyQueryButtonSuccess, 1000);
     });
 
     $('#add_column_button').on('click', function () {
--- a/resources/js/src/database/multi_table_query.ts
+++ b/resources/js/src/database/multi_table_query.ts
@@ -185,15 +185,16 @@ AJAX.registerOnload('database/multi_table_query.js', function () {
         });
     });
 
+    editor.getDoc().on('change', function () {
+        const query = editor.getDoc().getValue();
+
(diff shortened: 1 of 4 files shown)


src/Controllers/JavaScriptMessagesController.php
```diff
@@ -336,6 +336,8 @@ private function getMessages(): array
             /* For query editor */
             'strFormatting' => __('Formatting SQL…'),
             'strNoParam' => __('No parameters found!'),
+            'strQueryCopiedSuccess' => __('Query copied successfully'),
```
williamdes, line 339: maybe we should re-use strCopyQueryButtonSuccess instead of having a new translation

src/Controllers/JavaScriptMessagesController.php
```diff
@@ -336,6 +336,8 @@ private function getMessages(): array
...
+            'strQueryCopyEmpty' => __('Query is empty!'),
```
williamdes, line 340:
```suggestion
            'strQueryCopyEmpty' => __('There is no query to copy.'),
```
Or you can avoid that and disable the button if there is no query to copy

resources/js/src/database/multi_table_query.ts
```diff
@@ -184,6 +185,17 @@ AJAX.registerOnload('database/multi_table_query.js', function () {
...
+    $('#copy_query').on('click', function () {
+        const query = editor.getDoc().getValue();
+
+        if (query !== '') {
+            navigator.clipboard.writeText(query);
+            ajaxShowMessage(window.Messages.strQueryCopiedSuccess, 1000);
+        } else {
+            ajaxShowMessage(window.Messages.strQueryCopyEmpty, 1000);
+        }
+    });
```
williamdes, line 197: can you try to re-use the code from https://github.com/phpmyadmin/phpmyadmin/blob/89d9817628e52bcdf5d51165329374da2f1d3a40/resources/js/src/export_output.ts#L20-L21

to avoid re-creating the copy logic

faissaloux:
Done, video updated!

resources/js/src/database/multi_table_query.ts
```diff
@@ -184,6 +185,19 @@ AJAX.registerOnload('database/multi_table_query.js', function () {
...
+        const query: string = editor.getDoc().getValue();
+
+        $('#copy_query').prop('disabled', query === '');
+    });
+
+    $('#copy_query').on('click', function (): void {
+        const query: string = editor.getDoc().getValue();
+
+        copyToClipboard(query, '<textarea>');
+        ajaxShowMessage(window.Messages.strCopyQueryButtonSuccess, 1000);
```
williamdes, line 198:
```suggestion
        var copyStatus = copyToClipboard(query, '<textarea>');
        displayCopyStatus(this, copyStatus); 
```

faissaloux: I have tried this but when the message shows up, it pushes the `Submit Query` button to the right which is not good. That's why I ended up doing it as a popup message. What do you think?

williamdes: Indeed, I did not read it correctly.
Can you clone the function displayCopyStatus to displayCopyNotification
And show the message like you did
That will be re-usable

faissaloux: Yeah sure 👌🏻

resources/js/src/modules/functions.ts
```diff
@@ -1482,6 +1482,13 @@ export function displayCopyStatus (copyButton: JQuery<HTMLInputElement>|HTMLElem
     }, 2000);
 }
 
+/**
+ * Displaying notification of copy to clipboard action.
+ */
+export function displayCopyNotification (): void {
+    ajaxShowMessage(window.Messages.strCopyQueryButtonSuccess, 1000);
+}
```
williamdes, line 1490: that's not a copy of the logic of the other function

resources/js/src/database/multi_table_query.ts
```diff
@@ -184,6 +185,19 @@ AJAX.registerOnload('database/multi_table_query.js', function () {
...
+        const query: string = editor.getDoc().getValue();
+
+        $('#copy_query').prop('disabled', query === '');
+    });
+
+    $('#copy_query').on('click', function (): void {
+        const query: string = editor.getDoc().getValue();
+
+        copyToClipboard(query, '<textarea>');
+        displayCopyNotification();
```
williamdes, line 198: here there was a clear variable to pass to the notification to handle the copy error case

resources/js/src/modules/functions.ts
faissaloux: Do you mean it should handle the error case?

williamdes: Yes, that's what is missing in your PR vs the existing functions that handle copying text

faissaloux: Done!

williamdes (approved):
Nice, this looks ready for future uses 🎉

MauricioFauth (approved):

PR #19559, opened 2025-01-27
Improve Plugins page UI

faissaloux:
### Before
image

### After
image


faissaloux:
The eyes don't move so much to find something, it's more comfortable.

ibennetch:
I don't personally care for this change.

williamdes:
I find this more messy to look at

MauricioFauth:
This page needs some care, but this is not a good improvement in my opinion, as it doesn't looks good on wide screens.

PR #19569, opened 2025-01-28
Add an error instead of triggering warning

kamil-tekiela:
When in development mode and you enter wrong credentials, the error is displayed on a blank page. This is quite annoying and inconsistent with production. But if we add the error directly then it behaves the same in both modes. 

The line and file name doesn't make any difference in this case. 

@MauricioFauth If this makes sense, I would like to refactor more to tidy this up.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -6982,12 +6982,6 @@ parameters:
 			path: src/Error/ErrorHandler.php
 
 		-
-			message: '#^Loose comparison via "\!\=" is not allowed\.$#'
-			identifier: notEqual.notAllowed
-			count: 1
-			path: src/Error/ErrorHandler.php
-
-		-
 			message: '#^Parameter \#1 \$value of function count expects array\|Countable, mixed given\.$#'
 			identifier: argument.type
 			count: 1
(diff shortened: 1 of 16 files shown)


MauricioFauth (approved):
That's great! Is it possible to replace all trigger_error with that?

kamil-tekiela:
> That's great! Is it possible to replace all trigger_error with that?

I'm going to check now.

MauricioFauth:
> I'm going to check now.

I can't test it right now, but if it's possible, that would be great.

kamil-tekiela:
I found two more. We go from this:
image
to this:
image
And this is how it looks like in development mode before this PR:
image

kamil-tekiela:
So I went ahead and removed all but one of `trigger_error`. 

- E_USER_WARNING is now converted to `addUserError`. It no longer calls the default PHP error handler.
- E_USER_NOTICE is now converted to `addNotice`. It no longer calls the default PHP error handler.
- E_USER_ERROR was replaced with exceptions. Since we do not handle errors in the error handler, this error would stop the script execution so it is kind of pointless to use a `trigger_error` for something that can be done more easily with an exception and achieve the exact same effect.
- E_USER_DEPRECATED remains because I don't know how to handle it. It's probably the least problematic of all. It's handled in a slightly strange way but it's used in only one place.

The next step is to improve the error handling code.

williamdes (approved):
Good idea !

MauricioFauth (approved):
Thank you!

PR #19583, opened 2025-01-30
Update column_row.twig

### LINKED ISSUE #19582: Uncaught SyntaxError: missing ) after argument list
### Describe the bug

`Uncaught SyntaxError: missing ) after argument list` when selecting enum values.

Introduced in https://github.com/phpmyadmin/phpmyadmin/commit/aaa3587332766764fb5d5cd015bfa7d13fe4cfca

### To Reproduce

1. Select a value from enum
2. See error

### Expected behavior

No error.

### Screenshots

https://github.com/user-attachments/assets/7c591f66-c78f-4d6e-9335-0b4876a990d2

### Server configuration

### Additional context

The problem is it should be only enum without values.

Before:

```HTML
<select name="fields[multi_edit][1][da45ec4be6574774008df9be683a4778]" class="textfield valid" tabindex="4" id="field_4_3" onchange="return verificationsAfterFieldChange('da45ec4be6574774008df9be683a4778', '1', 'enum')" aria-invalid="false">
```

After:

```HTML
<select name="fields[multi_edit][1][da45ec4be6574774008df9be683a4778]" class="textfield valid" tabindex="4" id="field_4_3" onchange="return verificationsAfterFieldChange('da45ec4be6574774008df9be683a4778', '1', 'enum('111','222','333')')" aria-invalid="false">
```

kamil-tekiela:
Fixes #19582

Enum and set are the only types that can have quotes in their definition. I didn't pay attention to what the previous code was doing but it had code that handled these two types specifically instead of using the true type in the JS call.

### DIFF
--- a/resources/templates/table/insert/column_row.twig
+++ b/resources/templates/table/insert/column_row.twig
@@ -17,7 +17,7 @@
       <td class="text-center">--</td>
     {% else %}
       <td>
-        <select name="funcs[multi_edit][{{ row_id }}][{{ column.md5 }}]" onchange="return verificationsAfterFieldChange('{{ column.md5|escape('js') }}', '{{ row_id|escape('js') }}', '{{ column.type }}')" id="field_{{ id_index }}_1">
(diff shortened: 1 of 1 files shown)


MauricioFauth (approved):
Thank you!

PR #19591, opened 2025-02-04
Removed some mixed type hints

kamil-tekiela:


src/Controllers/Database/DataDictionaryController.php
```diff
@@ -110,4 +110,23 @@ public function __invoke(ServerRequest $request): Response
 
         return $this->response->response();
     }
+
+    /**
+     * @param mixed[][] $indexes index data
+     *
+     * @return true[] processes index data
```
williamdes, line 117:
```suggestion
     * @return array<string,true> The list of primary keys
```

williamdes (approved):

src/Controllers/Database/DataDictionaryController.php
```diff
@@ -110,4 +110,23 @@ public function __invoke(ServerRequest $request): Response
 
         return $this->response->response();
     }
+
+    /**
+     * @param mixed[][] $indexes index data
+     *
+     * @return array<true> The list of primary key columns
```
williamdes, line 117:
```suggestion
     * @return array<string,true> The list of primary key columns
```
I think documenting the key is important

kamil-tekiela: Ok, but then it has to be `array<array-key, true>` or `array<string|int, true>`. Psalm doesn't allow me to make it as `string` only because the key can be either `int` or `string`.

MauricioFauth (approved):

PR #19499, opened 2025-01-04
Fix #17730 - Make LIKE %...% as default search operator

liviuconcioiu:
### Description

Make `LIKE %...%` as default search operator. As it can be seen in the issue, this can cause confusion.

Fixes #17730 

Before submitting pull request, please review the following checklist:

- [x] Make sure you have read our [CONTRIBUTING.md](https://github.com/phpmyadmin/phpmyadmin/blob/master/CONTRIBUTING.md) document.
- [x] Make sure you are making a pull request against the correct branch. For example, for bug fixes in a released version use the corresponding QA branch and for new features use the _master_ branch. If you have a doubt, you can ask as a comment in the bug report or on the mailing list.
- [x] Every commit has proper `Signed-off-by` line as described in our [DCO](https://github.com/phpmyadmin/phpmyadmin/blob/master/DCO). This ensures that the work you're submitting is your own creation.
- [x] Every commit has a descriptive commit message.
- [x] Every commit is needed on its own, if you have just minor fixes to previous commits, you can squash them.
- [ ] Any new functionality is covered by tests.

Issue #17730, opened 2022-09-19
search for html tags does not work


staabm:
### Describe the bug

some columns in our database contain html text, e.g. `<b>Hallo World</b>`.
when searching this column for `<b>` the phpmyadmin search does not yield any results.

![grafik](https://user-images.githubusercontent.com/120441/190994093-ef973775-7326-4a83-b0c9-5e869bd68ab3.png)

search with a plain sql string yields results though:
```
SELECT * FROM `table` WHERE columnA like '%<b>%'
``` 

I would guess there is some broken html special char conversion involved, which breaks the search pattern

### To Reproduce

1. open a db, and a table in the phpmyadmin web ui
2. Click on 'Search'
3. type the searchterm `<b>`
4. submit 

### Expected behavior

result should be found

### Screenshots

If applicable, add screenshots to help explain the bug.

### Server configuration

- Operating system:
- Web server: apache 2.4
- Database version:  5.7.39-0ubuntu0.18.04.2-log - (Ubuntu) 
- PHP version: 7.2.34-33+ubuntu18.04.1+deb.sury.org+1 
- phpMyAdmin version: 5.2.0

 libmysql - mysqlnd 5.0.12-dev - 20150407

### Client configuration

- Browser: firefox nightly 106 
- Operating system: windows 11

williamdes:
Thank you for reporting this issue !

liviuconcioiu:
@williamdes I think there's not bug here. @staabm when you search, `LIKE` will generate `LIKE '<b>'`. You have to select `LIKE %...%` in order to generate `LIKE '%<b>%'`.

@williamdes probably `LIKE %...%` should be used by default instead of `LIKE`, to avoid another step in selecting from the drop-down.

I have attached a video.

https://user-images.githubusercontent.com/25424343/191057059-42ab9407-c35a-4d3e-ad1a-fba662da084b.mp4

staabm:
good catch... thanks for investigating.

maybe it would even make sense to add a notice/warning when the user does not input one of the LIKE wildcard characters?


> probably `LIKE %...%` should be used by default instead of `LIKE`

sounds great to me...

liviuconcioiu:
> another idea

There's an issue open about this https://github.com/phpmyadmin/phpmyadmin/issues/14542.

Issue #19556, opened 2025-01-27
Move default search operator for varchar columns into configurable options


CrazyHackGUT:
### Is your feature request related to a problem? Please describe.
Two weeks ago, in phpMyAdmin arrived changes that change the default operator for searching by string columns in tables from `LIKE` to `LIKE "%...%"`. #17730 
Search with `LIKE "%...%"` can overload MySQL DB because search content in the middle of the line cannot be performed via indexes:
```
mysql> explain select count(*) from xf_user where username like 'Kruzya';
+----+-------------+---------+------------+-------+-------------------------+----------+---------+------+------+----------+--------------------------+
| id | select_type | table   | partitions | type  | possible_keys           | key      | key_len | ref  | rows | filtered | Extra                    |
+----+-------------+---------+------------+-------+-------------------------+----------+---------+------+------+----------+--------------------------+
|  1 | SIMPLE      | xf_user | NULL       | range | username,staff_username | username | 202     | NULL |    1 |   100.00 | Using where; Using index |
+----+-------------+---------+------------+-------+-------------------------+----------+---------+------+------+----------+--------------------------+
1 row in set, 1 warning (0.00 sec)

mysql> explain select count(*) from xf_user where username like '%Kruzya%';
+----+-------------+---------+------------+-------+---------------+----------+---------+------+-------+----------+--------------------------+
| id | select_type | table   | partitions | type  | possible_keys | key      | key_len | ref  | rows  | filtered | Extra                    |
+----+-------------+---------+------------+-------+---------------+----------+---------+------+-------+----------+--------------------------+
|  1 | SIMPLE      | xf_user | NULL       | index | NULL          | username | 202     | NULL | 21366 |    11.11 | Using where; Using index |
+----+-------------+---------+------------+-------+---------------+----------+---------+------+-------+----------+--------------------------+
1 row in set, 1 warning (0.00 sec)
```

### Describe the solution you'd like
Move default search operator to config. Anyone who needs to search in the middle of a line knows how to do it. And this change now forces make an additional 2 clicks to search, so as not to accidentally overload MySQL by searching a large table, even if there is an index.

### Describe alternatives you've considered
Rollback #17730.

### Additional context
- phpMyAdmin version: 5.2.2

williamdes:
I can understand the issue, @liviuconcioiu and @staabm what do you think

staabm:
Thanks for bringing the issue up. 

[...]

PR #19612, opened 2025-02-24
Fix #19611 - restore text operators order

### LINKED ISSUE #19611: Default TEXT operator should not be LIKE %...%
### Describe the bug

Since v5.2.2, operators LIKE and LIKE %...% for the TEXT fields have been switched, which makes LIKE %...% the default option in the "Find" form. This is a bad idea as there is a great chance that the resulting query will not use any index and it will lead to performance problems in large tables. Plus, it is not coherent with operators for NUMERIC or UUID fields where LIKE is the default before LIKE %...%.

### To Reproduce

1. Open any table with a text field
2. Open the 'Search' tab
3. See that the default operator for text fields is now LIKE %...%

### Expected behavior

The default operator should be LIKE to avoid performance issues

setaou:
### Description

Restore original order for text operators, making LIKE the default instead of LIKE %...%.

Fixes #19611
Reverts #17730 / #19499

### DIFF
--- a/libraries/classes/Types.php
+++ b/libraries/classes/Types.php
@@ -96,8 +96,8 @@ class Types
     public function getTextOperators()
     {
         return [
-            'LIKE %...%',
             'LIKE',
+            'LIKE %...%',
             'NOT LIKE',
             'NOT LIKE %...%',
             '=',
--- a/test/classes/TypesTest.php
+++ b/test/classes/TypesTest.php
@@ -74,8 +74,8 @@ class TypesTest extends AbstractTestCase
     public function testgetTextOperators(): void
     {
         self::assertSame([
-            'LIKE %...%',
             'LIKE',
+            'LIKE %...%',
             'NOT LIKE',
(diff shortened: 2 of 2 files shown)


kamil-tekiela:
See https://github.com/phpmyadmin/phpmyadmin/issues/19556#issuecomment-2620062424 which I think would be a better solution.

setaou:
That is a suggestion that is worth discussing, however my proposal was for now to restore the previous behavior which has been working for years for most of the users, which makes sense to me for a QA branch.

williamdes:
@liviuconcioiu are you okay for a revert ?
For me let's go and we can clean this up in a way that is not changing the default behavior

liviuconcioiu:
> @liviuconcioiu are you okay for a revert ? For me let's go and we can clean this up in a way that is not changing the default behavior

YES! PLEASE MERGE IT! I was about to ask for this to be merged for 5.2.3. In 6.0.0 we can do some enhancement configurable in config

PR #19624, opened 2025-03-08
Fix column move

kamil-tekiela:
This should fix at least 3 issues:
- When the UI goes out of sync with the table structure, e.g. you change the structure in another tab
- When your column names are numerical
- When your column names are quoted
It should probably be fixed in 5.2 but I don't feel like working on 5.2 anymore. If someone wants to fix it there too, feel free to do it.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -4225,12 +4225,6 @@ parameters:
 			path: src/Controllers/Table/Structure/ChangeController.php
 
 		-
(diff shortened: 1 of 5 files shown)


MauricioFauth (approved):
Thank you!

MauricioFauth:
> * When your column names are quoted
>       It should probably be fixed in 5.2 but I don't feel like working on 5.2 anymore. If someone wants to fix it there too, feel free to do it.

Could you please open an issue about it for 5.2? With a reference to this PR.
Thank you!

PR #19628, opened 2025-03-14
Fix mysql default value insert

### LINKED ISSUE #19623: insert default value generates an invalid query
### Describe the bug

on Insert, phpMyAdmin generates invalid query for enum fields - missing quotes around the value.

### To Reproduce

1. Go to `Insert`
2. Set `enum` fields.
4. Click `Go` or `Preview SQL`.

### Current behavior

```sql
INSERT INTO `reservations` (`id`, `space_type`, `status`) VALUES (NULL, room, waiting);
```

### Expected behavior

```sql
INSERT INTO `reservations` (`id`, `space_type`, `status`) VALUES (NULL, 'room', 'waiting');
```

### Screenshots

Image

Image

### Server configuration

kamil-tekiela:
This should fix #19623 without affecting MariaDB since it doesn't set the flag. It doesn't fix the issue with escaped quotes on MySQL. 

@faissaloux Can you please check if it fixes the issue for you?

### DIFF
--- a/src/InsertEdit.php
+++ b/src/InsertEdit.php
@@ -1865,7 +1865,11 @@ class InsertEdit
     /** @return array<string|null> */
     public function getColumnDefaultValues(string $database, string $table): array
     {
-        $sql = 'SELECT COLUMN_NAME, COLUMN_DEFAULT FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_NAME = '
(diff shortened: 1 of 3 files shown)


faissaloux (approved):

MauricioFauth:
Should this be back ported to QA as well?

kamil-tekiela:
> Should this be back ported to QA as well?

I doubt we have this issue in 5.2. Maybe there is a similar problem but I haven't checked.

faissaloux:
This issue happens only on `6.0.0-dev`.

MauricioFauth (approved):

PR #19629, opened 2025-03-16
This is a modified version for PHP 8.4.5

mohammadismaily:
Hello, I am Mohammad Esmaeili from Iran, and I fixed the compatibility issue of phpMyAdmin 5.2.2 version to work with PHP 8.4.5.
This version is reliable and tested and I am using this version myself
I would be happy to confirm.

Fixed an issue using PHP 8.4.5 as a web server for phpMyAdmin version 5.2.2.


williamdes:
Hi
Thank you for opening a pull request, I appreciate the effort
what are you trying to contribute?
the diff looks like you did pull a final version into git.
I changed the target branch to reduce the diff.
what files did you intend in changing?

mohammadismaily:
> Hi Thank you for opening a pull request, I appreciate the effort what are you trying to contribute? the diff looks like you did pull a final version into git. I changed the target branch to reduce the diff. what files did you intend in changing?

Thank you, there were some files that were causing problems when executing the 'null' variable, there were a lot of files and I modified them so that they would not cause problems when they were null.

MauricioFauth:
However, there's no actual change here. The PHP 8.4 deprecations are from vendor files, which aren't present in this repository. Git will ignore changes to files inside the `vendor` folder.
So we can't merge this pull request.

mohammadismaily:
It's not my problem, the vendor folder is one of the folders based on the phpMyAdmin files that we download from your website, so the problem is yours, goodbye.

williamdes:
> It's not my problem, the vendor folder is one of the folders based on the phpMyAdmin files that we download from your website, so the problem is yours, goodbye.

I get that you can be confused, but normally users should not change vendor folders.
See: https://github.com/phpmyadmin/phpmyadmin/issues/19404#issuecomment-2727478757
A solution will be found soon

Goodbye

mohammadismaily:
You say that the vendor folder should not be tampered with, but the main problem is with the vendor folder. Thanks anyway, I delete my branch, I was expecting a different answer. Bye

PR #19634, opened 2025-03-18
Fix enum editor with empty values

faissaloux:
When I submit enum values from the `ENUM/SET` editor with empty fields, it generates empty strings as enum values. I have fixed this on this PR.

image

### Current
image

### After
image


MauricioFauth:
Duplicate of #12318

MauricioFauth:
I know this is annoying, but it should be possible to add an empty string as a value.

faissaloux:
In this case we should remove duplicates.

PR #19635, opened 2025-03-19
remove duplicates from enum editor

faissaloux:
Following https://github.com/phpmyadmin/phpmyadmin/pull/19634, this will avoid generating duplicate values.

image

#### Current
image

#### After
image

### DIFF
--- a/js/src/functions.js
+++ b/js/src/functions.js
@@ -3142,10 +3142,14 @@ AJAX.registerOnload('functions.js', function () {
         $('#enumEditorGoButton').on('click', function () {
             // When the submit button is clicked,
             // put the data back into the original form
-            var valueArray = [];
+            let valueArray = [];
             $('#enumEditorModal').find('.values input').each(function (index, elm) {
-                var val = elm.value.replace(/\\/g, '\\\\').replace(/'/g, '\'\'');
-                valueArray.push('\'' + val + '\'');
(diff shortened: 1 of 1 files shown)


kamil-tekiela:
Why can't enum have duplicates? I believe only set cannot have duplicates.

faissaloux:
> Why can't enum have duplicates? I believe only set cannot have duplicates.

1- MySQL does not allow duplicates on ENUM.
2- What's the point from having duplicates in ENUM?

kamil-tekiela:
MySQL does allow duplicate enum values:
```
CREATE TABLE `duplenums` (
  `id` int NOT NULL AUTO_INCREMENT,
  `duplenums` enum('a','a','b','b') NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci
```

I don't know what the point is because I don't even know what the point of using enums is. But for someone who knows how to use it, they would probably appreciate that PMA doesn't limit them.

kamil-tekiela:
I'm using 8.0.39

faissaloux:
Are you disabling strict mode?

kamil-tekiela:
Yes, that was it. Still, if it's technically possible to do it then I don't think PMA should restrict it.

MauricioFauth:
Removing the duplicating values this way could be unexpected for some users.
What could be nice though is to add a visual warning about duplicate values.

PR #19638, opened 2025-03-21
Improve issue templates

faissaloux:
In this PR, I have tried to improve the issue templates, see [GitHub Documentation](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository).


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,100 @@
...
+      placeholder: Apache/2.4.63
+  - type: dropdown
+    id: database-server
+    attributes:
+      label: Database Server
+      description: In which database server this bug appears?
+      multiple: true
+      options:
+        - MySQL
+        - MariaDB
```
williamdes, line 65:
```suggestion
        - MariaDB
        - PerconaDB
        - AuroraDB (by AWS)
        - Other
```

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,100 @@
...
+    validations:
+      required: true
+  - type: input
+    id: database-version
+    attributes:
+      label: Database version
+      placeholder: 8.0.33
```
williamdes, line 72: I think we can hint a query for the user
`SELECT @@version;` if I am correct

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,100 @@
...
+      placeholder: Firefox v136.0.2
+    validations:
+      required: true
+  - type: textarea
+    id: additional-context
+    attributes:
+      label: Additional context
+      description: Add any other context about the bug here.
+    validations:
+      required: false
```
williamdes, line 100: Same, not required

williamdes (approved):
Very nice change, what does it look like in screenshots?
Also, let's think about how much effort filling bug reports is. And make only the minimum required required.

@liviuconcioiu for comments

.github/ISSUE_TEMPLATE/bug_report.yml
faissaloux: It is not required (`false`).

faissaloux: I have removed it since it's the default value.

ibennetch:
I really like this idea, this is a great feature and nice improvement.

I do agree that we should only require very minimal information.

I also think the form should be as short as possible; if users see a long list they are more likely to leave instead of filing a report. We do need to capture relevant information, but users often don't know or can't be bothered — and a vague report is _sometimes_ better than no report.

This is coming along nicely.

liviuconcioiu:
Looks nice!

faissaloux:
> Very nice change, what does it look like in screenshots?

image

> I also think the form should be as short as possible

I agree I find it huge and we def need to make it shorter, can anyone help me with the fields that you find not helpful so I can remove them.

faissaloux:
> I do agree that we should only require very minimal information.

I agree I find it huge and we def need to make it shorter, can anyone help me with the fields that you find not helpful so I can remove them.

faissaloux:
One more thing is the labels:

- Bug Report label is `bug`. (We have this)
- Feature Request label is `feature`. (We don't have this)
- Security Report label is `security.` (We don't have this)

@williamdes @MauricioFauth You have to add `feature` and `security` labels for this or let me know which labels would help you guys so I link them with the templates.

MauricioFauth:
> * Bug Report label is `bug`. (We have this)
> * Feature Request label is `feature`. (We don't have this)
> * Security Report label is `security.` (We don't have this)

For bug and security reports, the [Bug](https://github.com/phpmyadmin/phpmyadmin/labels/Bug) label can be used (having a security label doesn't make much sense for me, as we don't handle security issues here), and for feature request the label [Enhancement](https://github.com/phpmyadmin/phpmyadmin/labels/Enhancement) should be used.

MauricioFauth (approved):

PR #19642, opened 2025-03-26
Fix #19641 - Allow optional param for NOW

### LINKED ISSUE #19641: error on {INSERT} to fields {timestamp(6)}
### Describe the bug
using:
phpMyAdmin-5.2.1-all-languages
using also:
phpMyAdmin-5.2.2-all-languages

create any table with field `**type**  timestamp(6)`, `**default** None**`

when you try [INSERT] *using the form PMA* with value `*now(6)*` for the field `moment` PMA send the QUERY:

`INSERT INTO `table` (id`, `moment`) VALUES (NULL, NOW());`

( NOTE: we send using the GUI FORM PMA, not the SQL QUERY, not the CONSOLE )

correct QUERY:

`INSERT INTO `table` (id`, `moment`) VALUES (NULL, NOW(6));`

### Server configuration

- Operating system: w10
- Web server: Apache/2.4.43 (Win64) OpenSSL/1.1.1g
- Database version: 10.11.11-MariaDB
- PHP version: 8.1.14
- phpMyAdmin version: phpMyAdmin-5.2.2-all-languages

liviuconcioiu:
### Description

Allow optional param for NOW.

https://github.com/user-attachments/assets/ef81e6ca-655b-4967-a121-44f741aedd06

https://github.com/user-attachments/assets/ba1f065d-4b8f-4046-91be-7f3cbeb80ceb

Fixes #19641

### DIFF
--- a/libraries/classes/InsertEdit.php
+++ b/libraries/classes/InsertEdit.php
@@ -30,6 +30,7 @@ use function implode;
 use function in_array;
 use function is_array;
 use function is_file;
+use function is_numeric;
 use function is_string;
 use function max;
 use function mb_stripos;
@@ -1630,7 +1631,10 @@ class InsertEdit
                     . $this->dbi->escapeString($multiEditSalt[$key]) . "')";
             }
 
-            if ($multiEditFuncs[$key] === 'NOW') {
+            if (
+                $multiEditFuncs[$key] === 'NOW'
+                && (is_numeric($currentValue) && $currentValue >= 0 && $currentValue <= 6)
+            ) {
                 return $multiEditFuncs[$key] . '(' . $this->dbi->escapeString($currentValue) . ')';
             }
 
--- a/libraries/classes/InsertEdit.php
+++ b/libraries/classes/InsertEdit.php
@@ -1630,6 +1630,10 @@ class InsertEdit
                     . $this->dbi->escapeString($multiEditSalt[$key]) . "')";
             }
 
+            if ($multiEditFuncs[$key] === 'NOW') {
+                return $multiEditFuncs[$key] . "(" . $this->dbi->escapeString($currentValue) . ")";
+            }
+
             return $multiEditFuncs[$key] . "('" . $this->dbi->escapeString($currentValue) . "')";
         }
 
--- a/libraries/classes/InsertEdit.php
+++ b/libraries/classes/InsertEdit.php
@@ -1631,7 +1631,7 @@ class InsertEdit
             }
 
             if ($multiEditFuncs[$key] === 'NOW') {
-                return $multiEditFuncs[$key] . "(" . $this->dbi->escapeString($currentValue) . ")";
+                return $multiEditFuncs[$key] . '(' . $this->dbi->escapeString($currentValue) . ')';
             }
 
             return $multiEditFuncs[$key] . "('" . $this->dbi->escapeString($currentValue) . "')";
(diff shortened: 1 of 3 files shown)


faissaloux (commented):
`NOW()` should allow only from 0 to 6.

libraries/classes/InsertEdit.php
```diff
@@ -1630,6 +1631,13 @@ public function getCurrentValueAsAnArrayForMultipleEdit(
                     . $this->dbi->escapeString($multiEditSalt[$key]) . "')";
             }
 
+            if (
+                $multiEditFuncs[$key] === 'NOW'
+                && (is_numeric($currentValue) && $currentValue >= 0 && $currentValue <= 6)
+            ) {
+                return $multiEditFuncs[$key] . '(' . $this->dbi->escapeString($currentValue) . ')';
+            }
+
```
faissaloux, line 1640: This will not allow `NOW('7')` or any number not between 0 and 6.

```suggestion
            if ($multiEditFuncs[$key] === 'NOW') {
                if ((is_numeric($currentValue) && $currentValue >= 0 && $currentValue <= 6)) {
                    return $multiEditFuncs[$key] . '(' . $this->dbi->escapeString($currentValue) . ')';
                }

                return $multiEditFuncs[$key] . '()';
            }

```

faissaloux (changes requested):

libraries/classes/InsertEdit.php
liviuconcioiu: Yeah. but wouldn't it be best to let the user know that the value is wrong? I think we should not change the behaviour, and pass the value to the query.

`INSERT INTO `test` (`id`, `ts`) VALUES (NULL, NOW('7'))`

with your change we will have

`INSERT INTO `test2` (`id`, `ts`) VALUES (NULL, NOW())`

faissaloux: Maybe show a hint when choosing `NOW()` that the value should be between 0 and 6, and while writing the value don't allow to enter something other than numbers between 0 and 6.

liviuconcioiu: @faissaloux can you check it with last changes? I also added a second video to PR description.

MauricioFauth (approved):

williamdes (approved):

libraries/classes/InsertEdit.php
faissaloux: From the video it appears that it shows a wrong alert message `Please enter a value less than or equal to 0` when you entered 3.

And I think it's better to be `Please enter a value between 0 and 6` as from `less than or equal 6`   it appears that it allows negative numbers too.

faissaloux: @liviuconcioiu When I enter a valid value on a timestamp without length, it shows a false error message.
image

This is due to `data-decimals` attribute which value is 0 on timestamps without length.

liviuconcioiu: Not sure about this, because even if you put 5 there, it won't store the precision to the timestamp. So the message should be correct. The only accepted value is empty for `NOW()` or 0 for `NOW(0)`.

faissaloux: No, actually for `timestamp`, `NOW()` accepts any value between 0 and 6.

liviuconcioiu: Yeah, but isn't stored.

faissaloux: Yes true, I got you now. https://github.com/phpmyadmin/phpmyadmin/pull/19648#issuecomment-2759971521

PR #19648, opened 2025-03-27
Fix timestamp max

faissaloux:
This PR fixes this issue.

image

timestamp without length should support from 0 to 6, but it shows an error `Please enter a value less than or equal to 0` this is due to `data-decimals` attribute that equals 0 on timestamp without length columns.

Related to https://github.com/phpmyadmin/phpmyadmin/pull/19642
See https://github.com/phpmyadmin/phpmyadmin/pull/19642#discussion_r2017499761


liviuconcioiu:
Not sure about this, because even if you put 5 there, it won't store the precision to the timestamp. So the message should be correct. The only accepted value is empty for `NOW()` or 0 for `NOW(0)`.

faissaloux:
> Not sure about this, because even if you put 5 there, it won't store the precision to the timestamp. So the message should be correct. The only accepted value is empty for `NOW()` or 0 for `NOW(0)`.

Yeah I tested it, I got you now you are right!

Screenshot from 2025-03-28 01-53-51
Screenshot from 2025-03-28 01-54-17

For `timestamp` it's always stored as `NOW(0)` even though MySQL accepts other numbers between 0 and 6.

And now I am confused, should we stick with MySQL and not restrict users, or the other way for better user experience. Can someone answer me for this? @kamil-tekiela @williamdes

MauricioFauth:
From [MySQL documentation](https://dev.mysql.com/doc/refman/8.0/en/fractional-seconds.html):
> The fsp value, if given, must be in the range 0 to 6. A value of 0 signifies that there is no fractional part. If omitted, the default precision is 0. (This differs from the standard SQL default of 6, for compatibility with previous MySQL versions.)

So, if a restriction is applied when there is a precision part, I think, for consistency, a restriction should be applied as well if the precision is omitted.

PR #19660, opened 2025-04-03
Remove .noSelect() jQuery plugin

MauricioFauth:
It works as a polyfill for the `user-select` CSS property. Can be safely replaced by the `user-select-none` class from Bootstrap.


williamdes (approved):
yay !

PR #19675, opened 2025-04-17
Synchronize routines UX

kamil-tekiela:
- #16487 

@MauricioFauth This is what I was talking about, though I am not very familiar with the HTML structure.

### DIFF
--- a/resources/templates/database/routines/execute_form.twig
+++ b/resources/templates/database/routines/execute_form.twig
@@ -6,57 +6,62 @@
   <div class="card">
     <div class="card-header">{{ ajax ? t('Routine parameters') : routine['item_name'] }}</div>
 
-    <div class="card-body">
-      <table class="table w-auto rte_table{{ not ajax ? ' caption-top' }}">
-        {% if not ajax %}
-          <caption>{{ t('Routine parameters') }}</caption>
-        {% endif %}
-        <tr>
-          <th>{{ t('Name') }}</th>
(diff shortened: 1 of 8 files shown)


resources/templates/database/routines/execute_form.twig
```diff
@@ -3,60 +3,71 @@
...
+
+  <code>
+    <div class="card">
+      <div class="card-header">{{ ajax ? t('Routine code') : routine['item_name'] }}</div>
 
-    {% if not ajax %}
-      <div class="card-footer">
-        <input class="btn btn-primary" type="submit" name="execute_routine" value="{{ t('Go') }}">
+      <div class="card-body">
+        <pre>{{ routine['item_definition']|raw }}</pre>
```
MauricioFauth, line 60: Why displaying the definition here?

MauricioFauth (commented):
That's great!

resources/templates/database/routines/execute_form.twig
kamil-tekiela: Because otherwise it would be an empty modal. And the definition can be useful to see before execution.

MauricioFauth: That's redundant with the Export button. Maybe just a message that there's no parameters. At least until a better design for this whole page is found.

MauricioFauth (approved):

MauricioFauth (approved):

PR #19683, opened 2025-04-27
Access property directly

kamil-tekiela:
This is in anticipation of doctrine/coding-standard upgrade. See https://github.com/phpmyadmin/coding-standard/pull/10. I actually can't tell if Psalm is correct here or not.


williamdes (approved):
I use the other syntax at work, they are both the same
But the current one makes IDEs happy

MauricioFauth (approved):

PR #19686, opened 2025-04-30
Use native MySQL paging

kamil-tekiela:
https://github.com/phpmyadmin/phpmyadmin/issues/16487#issuecomment-2814048056

This PR changes how Routines are fetched from DB. The MySQL shorthand commands are dropped and all information is fetched from IS, which allows us to limit and sort the results. I was not able to observe any performance impact but the page opens pretty fast for me either way with 1000 routines. Maybe someone on a worse DB connection with more routines can test the performance. In theory the performance should remain more or less the same.


MauricioFauth (approved):
Thank you!

PR #19688, opened 2025-05-01
Fix bugs in relations tab

kamil-tekiela:
I don't know if they have been reported but they annoyed me. Fixed issues:
- Using a database/table/column name which is zero-string
- Replacing an existing relation with an invalid exception which causes a rollback to happen (passing null to string parameter error)

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -15226,6 +15226,12 @@ parameters:
 			path: src/Table/Table.php
 
 		-
+			message: '#^Parameter \#2 \$field of method PhpMyAdmin\\Table\\Table\:\:getSQLToCreateForeignKey\(\) expects array\<string\>, array\<int\|string\> given\.$#'
+			identifier: argument.type
+			count: 1
+			path: src/Table/Table.php
+
+		-
 			message: '#^Parameter \#2 \$string of function explode expects string, mixed given\.$#'
 			identifier: argument.type
 			count: 1
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -15132,7 +15132,7 @@ parameters:
 		-
 			message: '#^Construct empty\(\) is not allowed\. Use more strict comparison\.$#'
 			identifier: empty.notAllowed
-			count: 13
+			count: 2
 			path: src/Table/Table.php
 
 		-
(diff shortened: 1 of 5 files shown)


MauricioFauth:
Could you explain more on how to reproduce it? I want to backport it to 5.2 if it's affected.

kamil-tekiela:
### Using a database/table/column name which is zero-string

Create these two tables:
```
CREATE TABLE IF NOT EXISTS `0` (
  `id` int NOT NULL AUTO_INCREMENT,
  `0` int NOT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY (`0`)
) ENGINE=InnoDB AUTO_INCREMENT=5 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;

CREATE TABLE IF NOT EXISTS `foo` (
  `Id` int NOT NULL AUTO_INCREMENT,
  `0` int NOT NULL,
  `1` int NOT NULL,
  PRIMARY KEY (`Id`)
) ENGINE=InnoDB AUTO_INCREMENT=2 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;
```

Open the "structure" tab of the table `foo`. Go to "relation view". Create a new relation on column `0` pointing to the column `0` in table `0`. 

image

The new relation should be created and when the page is refreshed it should be shown. Without changing anything press "save". The newly created relation should not disappear. 

### Replacing an existing relation with an invalid exception which causes a rollback to happen (passing null to string parameter error)

Building on the same steps as above. Add a new row in the table `foo` with value `2` for the column `1`. Insert another row in the table `0` with a **different** value. In the "relation view" of the `foo` table change the existing relation from column `0` to column `1`. 
image

There should be an error about a mismatched value and no changes should be made to the relations. 

image


I realise now that I haven't fully fixed the last part as I was testing with a different error condition. I will check what else is missing there.

kamil-tekiela:
@MauricioFauth I think I fixed it, but it requires https://github.com/phpmyadmin/sql-parser/pull/624 to be merged. Could you please merge it first?

kamil-tekiela:
When changing the column of the relation `$masterField` will have a different value for the new key. When we roll back we need to use the old column which is in `$existrelForeign[$masterFieldMd5]->indexList`.

The baselines were updated so you can merge it first and then the PR from sql-parser.

MauricioFauth (approved):

PR #19700, opened 2025-05-12
Improve error template

kamil-tekiela:
The `<pre>` element is block display while `<code>` is an inline element. Block elements should not be within inline elements. Semantically, it also makes sense for `<code>` to be inside of `<pre>` and not the other way around. I believe that the CSS I removed was an attempt to fix a problem caused by this issue. 

The `<pre>` by default has a bottom margin which looks odd IMHO and existing PMA users are probably not used to it, so I have removed it. 

As a side effect, the text inside `<code>` is now more readable. 

The error message template has been updated to make it possible to use `<p>` for delimiting paragraphs instead of `<br>`.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -7692,7 +7692,7 @@ parameters:
 		-
 			message: '#^Construct empty\(\) is not allowed\. Use more strict comparison\.$#'
 			identifier: empty.notAllowed
-			count: 11
+			count: 7
 			path: src/Html/Generator.php
 
 		-
@@ -14070,7 +14070,7 @@ parameters:
 		-
 			message: '#^Construct empty\(\) is not allowed\. Use more strict comparison\.$#'
 			identifier: empty.notAllowed
-			count: 11
+			count: 10
 			path: src/Sql.php
 
 		-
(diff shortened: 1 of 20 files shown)


MauricioFauth (approved):

liviuconcioiu:
@kamil-tekiela can you double check those two issues?

1. After you click `Edit inline` buttons are messed up, and also if you click `Cancel`, the query doubles.
2. `Create PHP code` loses code highlight

Before:

https://github.com/user-attachments/assets/eb1451fb-dace-4344-b182-283b528e7842

After:

https://github.com/user-attachments/assets/3b4aeacc-5c3f-406e-8911-d3de8c663613

kamil-tekiela:
I will check the first issue. 

There was no code highlight on the PHP code. It was just a pink tint that was a result of mismatched CSS rules. It can be restored but I don't think it should.

liviuconcioiu:
@kamil-tekiela found another issue, when you drop an index, the css is set to `display: none` and the query is hidden.

https://github.com/user-attachments/assets/02c18bdb-78b2-4f4d-ab8f-d71a81890d43

PR #19708, opened 2025-05-19
Fix duplicate Set-Cookie

### LINKED ISSUE #19702: [Bug]: Duplicated Set-Cookie headers causing NGINX proxy to fail
### Describe the bug

It seems that after executing statements on `SQL` page phpmyadmin returns same duplicated `Set-Cookie` header line for each statement in submitted query.

This makes HTTP response header big if you have a lot of SQL statements.

In our case it started to cause errors `upstream sent too big header while reading response header from upstream` in our reverse proxy, which is Kubernetes Ingress NGINX Controller.

We use `phpmyadmin:5.2.1-apache` Docker image.

Workaround I used is to increase `proxy-buffer-size` in ingress config for phpmyadmin:

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: phpmyadmin
  annotations:
    # Workaround for "upstream sent too big header while reading response header from upstream"
    # https://www.getpagespeed.com/server-setup/nginx/tuning-proxy_buffer_size-in-nginx
    nginx.ingress.kubernetes.io/proxy-buffer-size: "64k" # Default: 4k
```

### How to Reproduce

1. Go to `/server/sql` page and submit multiple SQL statements
2. Check response of `/import` in browser network tab - it will have multiple `Set-Cookie` header lines with same values

### Expected behavior

I expected to not have duplicated `Set-Cookie` headers.

### Screenshots

Example: response headers of `/import`

Image

### Operating System

Linux

### Web Server

Apache/2.4.62 (Debian)

### Database Server

MySQL

### Database version

8.0.41

### PHP version

8.2.26

### phpMyAdmin version

5.2.1

### browser

Firefox/138.0

### Additional context

_No response_

kamil-tekiela:
Fixes #19702

This should fix the linked issue, but in my opinion, this is a PHP bug. PHP should not resend the same header without changes. I have spent the whole day working on this and I still could not understand why PHP sessions behave this way. I raised an issue https://github.com/php/php-src/issues/18601

However, I believe I have found a workaround for this. Most of the session logic is in Session.php. The session is closed and then reopened to avoid blocking the user from opening another tab when a long-running query is executed. The code doesn't seem to be doing any session-related stuff after that. If we reopen the session without cookie support, then it will not set the header. **We cannot do it in Session.php** because the code will call `session_regenerate_id()` later and that needs to send a new cookie. I have added a comment explaining that. 

The actual fix for the issue is only the change in Sql.php and it can be backported to QA5_2. The rest is just code quality refactoring and removing dead code.

### DIFF
--- a/src/Session.php
+++ b/src/Session.php
@@ -78,7 +78,7 @@ class Session
      *
      * @throws SessionHandlerException
      */
-    private static function sessionFailed(array $errors): void
+    private static function sessionFailed(array $errors): never
     {
         $messages = [];
         foreach ($errors as $error) {
@@ -115,7 +115,7 @@ class Session
     /** @throws SessionHandlerException */
     public static function setUp(Config $config, ErrorHandler $errorHandler): void
     {
-        if (! empty(ini_get('session.auto_start')) && session_name() !== 'phpMyAdmin' && ! empty(session_id())) {
+        if (ini_get('session.auto_start') && session_name() !== 'phpMyAdmin' && session_id()) {
             // Do not delete the existing non empty session, it might be used by
             // other applications; instead just close it.
             if ($_SESSION === []) {
@@ -173,7 +173,7 @@ class Session
         session_cache_limiter('private');
 
         $httpCookieName = $config->getCookieName('phpMyAdmin');
-        @session_name($httpCookieName);
+        session_name($httpCookieName);
 
         // Restore correct session ID (it might have been reset by auto started session
         if ($config->issetCookie('phpMyAdmin')) {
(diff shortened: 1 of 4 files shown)


src/Session.php
```diff
@@ -192,15 +192,6 @@ public static function setUp(Config $config, ErrorHandler $errorHandler): void
...
         }
 
-        unset($origErrorCount, $sessionResult);
-
-        /**
-         * Disable setting of session cookies for further session_start() calls.
-         */
-        if (session_status() !== PHP_SESSION_ACTIVE) {
-            ini_set('session.use_cookies', 'true');
-        }
```
kamil-tekiela, line 202: I dug into the history of this statement and I concluded that it was added with a typo (true should have been false) and then later someone added an if statement, effectively turning it into dead code. The code above asserts that this condition will always be false. Anyway, to make it work it should have been added after `session_write_close` or as an argument to `session_start` like I have done.

MauricioFauth (approved):

PR #19727, opened 2025-06-05
Replace jQuery UI's datetimepicker with tempus-dominus

MauricioFauth:


williamdes (approved):
Do we loose precision support?
can you post screenshots?

MauricioFauth:
> Do we loose precision support?

Most libraries doesn't support the fraction part. This library uses the Date JS class that only support three digits of precision, which means that it only supports milliseconds and not microseconds. Most datepickers are really awful. This is the least frustrating library that I've found. At least it has support for plugins that I'm probably gonna use it.

MauricioFauth:
And I don't want to replace the current date/time picker with a half-backed one just for the sake of using a "modern" library. If it doesn't work well, I'm going to close this PR.

PR #19730, opened 2025-06-08
Replace equal operator with identical operator in GIS classes

MauricioFauth:
Replaces equal operator (`==`) with identical operator (`===`) in GIS classes.


williamdes (commented):
Nice

Issue #19367, opened 2024-11-08
Renaming a timestamp(6) column doesn't work: it quotes the value


liviuconcioiu:
### Describe the bug

Renaming a `timestamp(6)` column with `DEFAULT current_timestamp(6)` doesn't work, because it adds single quotes around `current_timestamp(6)`.

### To Reproduce

1. Create table:

```SQL
CREATE TABLE `test` (
  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
  `created_at` timestamp(6) NOT NULL DEFAULT current_timestamp(6),
  PRIMARY KEY (`id`)
) ENGINE=MyISAM DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci ROW_FORMAT=DYNAMIC;
```
2. Rename `created_at`
3. See error

### Expected behavior

Renaming should work.

### Screenshots

https://github.com/user-attachments/assets/10093d81-d29e-47f6-9fb4-1158e449cde6

### Server configuration

- phpMyAdmin version: 5.2.2-dev, 6.0.0-dev

PR #19739, opened 2025-06-24
Fix to Renaming Timestamp issue #19367

yashpunwani:
### Description

This pull request will solves the bug that does not let's us rename the timestamp column with a current timestamp as it's default value as reported in Issue [#19367](https://github.com/phpmyadmin/phpmyadmin/issues/19367)

It was happening because the query builder was quoting the `current_timestamp()` function of SQL that was raising a Query Error of Invalid Default Value.

### DIFF
--- a/app/services.php
+++ b/app/services.php
@@ -219,7 +219,8 @@ return [
             'class' => TrackingChecker::class,
             'arguments' => ['$dbi' => '@dbi', '$relation' => '@relation'],
         ],
-        'transformations' => ['class' => Transformations::class],
+        'transformations' => ['class' => Transformations::class, 'arguments' => ['@dbi', '@relation']],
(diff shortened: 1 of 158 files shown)


kamil-tekiela:
Apart from the obvious issues with too many commits and non-adherence to the coding standard, I don't know if this PR fixes the issue. The first question to answer is why is the value quoted at all? Address the problem not the symptom.

yashpunwani:
Sorry for the mess I created. I found out that the query builder function that is used to generate the alter command was quoting the current timestamp function and I was fixing it in a wrong way.

yashpunwani:
@kamil-tekiela can you please take a look.

yashpunwani:
I'll create a new branch to fix it

PR #19753, opened 2025-07-06
Make some comparisons stricter

MauricioFauth:
Replaces equal operators (`==`) with identical operators (`===`).


kamil-tekiela:
Did you check all the types? As far as I remember all of the remaining ones were a bit trickier to figure out.

MauricioFauth:
> Did you check all the types? As far as I remember all of the remaining ones were a bit trickier to figure out.

Yes, these are the ones I was able to quickly check the types. The remaining ones are indeed trickier to figure out.

PR #19762, opened 2025-07-12
Collect routes from the Route attribute in controller classes

MauricioFauth:
Add the route information with the new `PhpMyAdmin\Routing\Route` attribute to each controller instead of declaring the routes in a separated file.

The difference in the routes cache file is just the order of the paths and is the same as seen in 1e334d53c62429c08ef5554e9793d9e112ecd95c.


williamdes (approved):
🎉

PR #19767, opened 2025-07-15
Add static analysis stub for PSR ContainerInterface

MauricioFauth:
Hints static analysis tools that when passing a `class-string` to `Psr\Container\ContainerInterface::get()`, it will return an object of that class.

- https://phpstan.org/user-guide/stub-files
- https://psalm.dev/docs/running_psalm/configuration/#stubs
- https://www.jetbrains.com/help/phpstorm/ide-advanced-metadata.html

### DIFF
--- a/.phpstorm.meta.php
+++ b/.phpstorm.meta.php
@@ -1,0 +1,5 @@
+<?php
+
+namespace PHPSTORM_META {
+    override(\Psr\Container\ContainerInterface::get(0), map(['@']));
+}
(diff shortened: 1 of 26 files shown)


.phpstorm.meta.php
```diff
@@ -0,0 +1,5 @@
+<?php
```
williamdes, line 1: This should be added to check release excludes and removed manually while building the release
right?

MauricioFauth: You're right! Thank you!

PR #19786, opened 2025-07-27
Refactor the Engines\Innodb::getPageBufferpool() method

MauricioFauth:
Use `Engines\Innodb\BufferPool` value object for `Innodb_buffer_pool_*` variables.

### DIFF
--- a/src/Engines/Innodb.php
+++ b/src/Engines/Innodb.php
@@ -7,11 +7,10 @@ namespace PhpMyAdmin\Engines;
 use PhpMyAdmin\Dbal\DatabaseInterface;
 use PhpMyAdmin\Engines\Innodb\BufferPool;
 use PhpMyAdmin\StorageEngine;
-use PhpMyAdmin\Util;
+use PhpMyAdmin\Template;
 
 use function __;
 use function htmlspecialchars;
-use function implode;
 
 /**
  * The InnoDB storage engine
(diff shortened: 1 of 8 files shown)


src/Engines/Innodb.php
```diff
@@ -190,55 +180,41 @@ public function getPageBufferpool(): string
...
                 ? '---'
-                : htmlspecialchars(
-                    Util::formatNumber(
-                        $status['Innodb_buffer_pool_reads'] * 100
-                        / $status['Innodb_buffer_pool_read_requests'],
-                        3,
-                        2,
-                    ),
-                ) . ' %') . "\n"
+                : htmlspecialchars(Util::formatNumber($status->reads * 100 / $status->readRequests, 3, 2)) . ' %')
```
kamil-tekiela: The usage of `htmlspecialchars` seems very inconsistent in this method.

src/Engines/Innodb/BufferPool.php
```diff
@@ -0,0 +1,69 @@
...
+            self::getNumeric($result['Innodb_buffer_pool_read_requests'] ?? null) ?? 0,
+            self::getNumeric($result['Innodb_buffer_pool_reads'] ?? null) ?? 0,
+            self::getNumeric($result['Innodb_buffer_pool_wait_free'] ?? null) ?? 0,
+            self::getNumeric($result['Innodb_buffer_pool_write_requests'] ?? null) ?? 0,
+            self::getNumeric($result['Innodb_buffer_pool_pages_latched'] ?? null),
+        );
+    }
+
+    /** @return float|int|numeric-string|null */
+    private static function getNumeric(mixed $value): float|int|string|null
```
kamil-tekiela: Shouldn't this function just take string|null and return a nuneric-string?

MauricioFauth: You're right. It's always string or null.

PR #19798, opened 2025-08-04
Fix #17947 - table prefix group link

### LINKED ISSUE #17947: Click on table prefix doesn't work anymore
### Describe the bug

Phpmyadmin groups tables with the same prefix in the left menu. Until recently I could click on the prefix to get a list of all tables with that prefix in the main window. That doesn't work anymore. If I click on the prefix nothing happens. The prefix isn't even a link anymore. Note that I'm talking about clicking the prefix itself, not the plus sign to the left of the prefix.

### To Reproduce

Create a few tables with the same prefix so that they becomes grouped in the left menu. Then click on the name of the prefix (not the plus sign to the left

### Expected behavior

You should get a list of all tables with the specific prefix in the main window

### Screenshots

If applicable, add screenshots to help explain the bug.
phpmyadmin_bug

### Server configuration

- Operating system: Centos 7
- Web server: Apache
- Database version: 10.10.2-MariaDB - MariaDB Server
- PHP version: 8.1.13
- phpMyAdmin version: 5.2.0

### Client configuration

- Browser: Chromium
- Operating system: Linux

### Additional context

deg026:
Fixes [#17947](https://github.com/phpmyadmin/phpmyadmin/issues/17947)
bug with table prefix group link disapear

### DIFF
--- a/libraries/classes/Navigation/NavigationTree.php
+++ b/libraries/classes/Navigation/NavigationTree.php
@@ -1100,7 +1100,7 @@ class NavigationTree
             $linkHasAjaxClass = $parent[0]->type == Node::CONTAINER
                 && (in_array($parent[0]->realName, $haveAjax) || $isNewView);
 
-            if (! $node->isGroup) {
+            if (! $node->isGroup || $node->links['text']['route']) {
                 $args = [];
                 $parents = $node->parents(true);
                 foreach ($parents as $parent) {
--- a/templates/navigation/tree/node.twig
+++ b/templates/navigation/tree/node.twig
@@ -19,11 +19,16 @@
       <div class="fst-italic">
     {% endif %}
 
-    {% if node.isGroup %}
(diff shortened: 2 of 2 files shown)


liviuconcioiu:
Hi @deg026! Can you double check if this works? I tried your PR but it doesn't seem to work. See:

https://github.com/user-attachments/assets/37aa6c6b-0f0a-4d98-827a-beb464370a53

deg026:
> Hi @deg026! Can you double check if this works? I tried your PR but it doesn't seem to work. See:

did you update NavigationTree.php file too?
there is the main fix...

liviuconcioiu:
> > Hi @deg026! Can you double check if this works? I tried your PR but it doesn't seem to work. See:
> 
> did you update NavigationTree.php file too? there is the main fix...

Yes.

williamdes:
I have put this back to draft, if testing shows that is does not work maybe there is more to be done

PR #19802, opened 2025-08-08
Fix #19381 - Implement csv empty to null

### LINKED ISSUE #19381: Make empty values NULL when importing a CSV file

There's no option to make empty values `NULL` when importing a CSV file.

Also, even if a value is NULL, is treated as empty value. Not sure if it's a bug or not, since exporting works with `NULL` values, I would assume, importing should work too? If so, it should also be fixed in 5.2.2-dev.

A checkbox after `The first line of the file contains the table column names (if this is unchecked, the first line will become part of the data)`.

N/A

### Additional context

DarshBhavnani:
This PR adds a feature to add a checkbox in import which allows to treat empty values as null
Fixes [#19381](https://github.com/phpmyadmin/phpmyadmin/issues/19381)

This pull request introduces a new optional feature for CSV data imports. A new checkbox, Import empty values as NULL, has been added to the import settings. When this option is selected, the system will now treat any empty field in the CSV file as a SQL NULL value during the import.

Previously, empty fields were imported as empty strings (''), which could cause issues with database constraints and data integrity for columns intended to be nullable. This enhancement provides users with greater control over their imported data, allowing for more precise data representation.

The changes ensure consistent behavior for both new table creation (the "analyze" mode) and importing into existing tables.

### DIFF
--- a/src/Plugins/Import/ImportCsv.php
+++ b/src/Plugins/Import/ImportCsv.php
@@ -394,11 +394,6 @@ class ImportCsv extends AbstractImportCsv
                         $i += $csvTerminatedLen - 1;
                     }
 
-                    // unquoted NULL string
-                    if ($needEnd === false && $value === 'NULL') {
-                        $value = null;
-                    }
-
                     if ($fail) {
                         $i = $fallbacki;
                         $ch = mb_substr($buffer, $i, 1);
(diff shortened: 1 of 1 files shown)


liviuconcioiu:
Thank you @DarshBhavnani! It works as intended.

**Before:**

before


**After:**

after1


**After (with `Import empty values as NULL` checked):**

after2

williamdes (approved):
We could have a test for this

williamdes:
Thank you for your work @DarshBhavnani 
Merged as 2c03905eeafdd096c85b89aad863aaca18637ff5 with some changes and new tests

PR #19809, opened 2025-08-14
Fix showing progress value in processes view

M393:
#19779 was not correctly merged into the master branch. The array keys are all lowercase here.


williamdes (approved):
Thank you, too bas phpstan did not catch this

MauricioFauth (approved):

PR #19681, opened 2025-04-24
Fix thecodingmachine/safe deprecations using a fork

MauricioFauth:
Forces Composer to use a fork of thecodingmachine/safe to fix its PHP 8.4 deprecations.

https://github.com/thecodingmachine/safe/compare/v1.3.3...1.x
https://github.com/MauricioFauth/thecodingmachine-safe/compare/v1.3.3...v1-php8.4

- Fixes https://github.com/phpmyadmin/phpmyadmin/issues/19404

PR #19853, opened 2025-09-10
Replace web-auth/webauthn-lib with web-auth/webauthn-framework

AhmedAlaa4611:
Description
---
The current phpMyAdmin dependency on `web-auth/webauthn-lib` indirectly requires `thecodingmachine/safe`, which produces a large number of deprecation notices in modern PHP versions.

According to the maintainers, `web-auth/webauthn-lib` is a read-only sub-repository of the Web Authentication Framework. The main repository, `web-auth/webauthn-framework`, does not depend on `thecodingmachine/safe`, making it the recommended package to use moving forward.

This PR updates the dependency from `web-auth/webauthn-lib` to `web-auth/webauthn-framework` to:

- Remove the indirect dependency on `thecodingmachine/safe`.
- Reduce deprecation notices on newer PHP versions.
- Align with the upstream maintainers' guidance that webauthn-lib should not be used directly.

This change should be backward-compatible since the framework is the canonical package for the lib.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -98,7 +98,7 @@
         "pragmarx/google2fa-qrcode": "^2.1 or ^3.0 - For 2FA authentication",
         "bacon/bacon-qr-code": "^2.0 - For 2FA authentication",
         "code-lts/u2f-php-server": "For FIDO U2F authentication",
-        "web-auth/webauthn-lib": "For better WebAuthn/FIDO2 authentication support"
+        "web-auth/webauthn-framework": "For better WebAuthn/FIDO2 authentication support"
     },
     "require-dev": {
         "bacon/bacon-qr-code": "^2.0 || ^3.0",
@@ -122,7 +122,7 @@
         "symfony/console": "^7.1",
         "tecnickcom/tcpdf": "^6.4.4",
         "vimeo/psalm": "^6.0",
-        "web-auth/webauthn-lib": "^4.9"
+        "web-auth/webauthn-framework": "^5.2.2"
     },
     "extra": {
         "branch-alias": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "fa3a0d01a4eb2d8c3590344eaa02ca12",
+    "content-hash": "9842d8f17cf580386d2b77bc61d8230f",
     "packages": [
         {
             "name": "composer/ca-bundle",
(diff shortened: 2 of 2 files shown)


AhmedAlaa4611:
I created this PR using the GitHub web interface, so I don't have a way to amend the commits to include the DCO sign-off. Could you please squash the commits and add the `Signed-off-by` line when merging? I'd really appreciate it.

williamdes:
> I created this PR using the GitHub web interface, so I don't have a way to amend the commits to include the DCO sign-off. Could you please squash the commits and add the `Signed-off-by` line when merging? I'd really appreciate it.

Can you comment it here please ?
I have put `Signed-off-by: Ahmed Alaa <92916738+AhmedAlaa4611@users.noreply.github.com>` for now

williamdes (changes requested):
Well, this does not work
See: https://github.com/phpmyadmin/phpmyadmin/actions/runs/17625015013/job/50079590724?pr=19853

AhmedAlaa4611:
Hello @williamdes 

I've been digging into why the tests fail, and I think we should tackle the issues one by one.

The first blocker is the `PublicKeyCredentialSourceRepository` interface:

In webauthn-lib, it lives under `Webauthn\PublicKeyCredentialSourceRepository` and looks like this:

https://github.com/web-auth/webauthn-lib/blob/1b02740ab8539f025419380c9e4c41b090c6cf47/src/PublicKeyCredentialSourceRepository.php

In webauthn-framework, the equivalent is `Webauthn\Bundle\Repository\PublicKeyCredentialSourceRepositoryInterface` and looks like this:

https://github.com/web-auth/webauthn-framework/blob/a218bc98b3014153ac56255274ad2704b33e94ba/src/symfony/src/Repository/PublicKeyCredentialSourceRepositoryInterface.php

So, there are two differences:

- The namespace and name have changed.
- The `saveCredentialSource()` method is no longer part of the interface.

In our codebase, I only see `saveCredentialSource()` being implemented once [in src/WebAuthn/WebauthnLibServer.php:363](https://github.com/phpmyadmin/phpmyadmin/blob/5c6bcaa22089dd6491a160564d6dc6314fb4f5ca/src/WebAuthn/WebauthnLibServer.php#L363). That means if we switch to the new interface, we can keep `saveCredentialSource()` as an internal helper, but it won't be enforced by the interface anymore.

Would you prefer to keep the method locally (even though it's not part of the interface anymore), or Drop it entirely if it's not actually being used elsewhere in phpMyAdmin?

MauricioFauth:
Hi @AhmedAlaa4611, thank you for your contribution.

Both `web-auth/webauthn-framework` v3.x and `web-auth/webauthn-lib` v3.x depends on `thecodingmachine/safe`. This dependency was removed from both in version 4.0+.

The webauthn-lib v3 is used in the branch `QA_5_2` (phpMyAdmin 5.2.x) and the version 4 is used in the branch `master` (phpMyAdmin 6.0.x). It's not possible to upgrade webauthn-lib in the branch `QA_5_2` because it'll be a breaking change. This issue was fixed by using a fork of `thecodingmachine/safe`. See https://github.com/phpmyadmin/phpmyadmin/pull/19681.

Also, the `web-auth/webauthn-framework` package has a lot of extra stuff that's not used by phpMyAdmin. Only `web-auth/webauthn-lib` is necessary.

> Align with the upstream maintainers' guidance that webauthn-lib should not be used directly.

That's not correct. The `web-auth/webauthn-lib` repository is read-only because it's only used by Composer to download the code. This strategy is also used by Symfony. For example, you can get the Config component by installing `symfony/config` or `symfony/symfony`, but with the later you'll get the full framework.

> I've been digging into why the tests fail, and I think we should tackle the issues one by one.

The tests are failing because you upgraded the webauthn package from ^4.9 to ^5.2. You can help migrating `web-auth/webauthn-lib` from version 4 to version 5 in another pull request if you want. That'll be great!

You can start with the migration guide: https://webauthn-doc.spomky-labs.com/v4.9/migration/from-v3.x-to-v4.0-1

> I created this PR using the GitHub web interface, so I don't have a way to amend the commits to include the DCO sign-off. 

You can manually include the Signed-off-by line as the last line of the commit message.

I'm closing this PR because the deprecation issue was fixed by https://github.com/phpmyadmin/phpmyadmin/pull/19681.

AhmedAlaa4611:
Ok, thank you for your time.

PR #19621, opened 2025-03-06
Fix #19109 - Do not enforce character encoding when checking roles

sunshinedaniel:
### Description

The functions `getInformationSchemaDataForCreateRequest()` and `getInformationSchemaDataForGranteeRequest()` generate a query that fails with the error `COLLATION 'utf8mb4_general_ci' is not valid for CHARACTER SET 'utf8mb3'` on some servers that got migrated from MySql 5.7. This lead to admin users not being able to create new users via PhpMyAdmin's UI. Instead, the error message "You do not have the privileges to administrate the users!" is being shown. When not forcing the encoding to `utf8mb4_general_ci` in the query, everything works fine.

Fixes #19109

PR #19864, opened 2025-09-17
Fix collation error when checking for grant permission

MauricioFauth:
- Related to #19621

### DIFF
--- a/libraries/classes/DatabaseInterface.php
+++ b/libraries/classes/DatabaseInterface.php
@@ -1794,8 +1794,10 @@ class DatabaseInterface implements DbalInterface
             return $hasGrantPrivilege;
         }
 
+        $collation = $this->getServerCollation();
+
         [$user, $host] = $this->getCurrentUserAndHost();
-        $query = QueryGenerator::getInformationSchemaDataForGranteeRequest($user, $host);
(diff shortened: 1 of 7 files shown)


libraries/classes/DatabaseInterface.php
```diff
@@ -1851,8 +1856,13 @@ public function isCreateUser(): bool
             return $hasCreatePrivilege;
         }
 
+        $collation = (string) $this->getVariable('collation_server');
```
kamil-tekiela: Why not use `getServerCollation`?

libraries/classes/DatabaseInterface.php
```diff
@@ -1794,8 +1794,13 @@ public function isGrantUser(): bool
             return $hasGrantPrivilege;
         }
 
+        $collation = (string) $this->getVariable('collation_server');
+        if ($collation === '') {
+            $collation = 'utf8mb4_general_ci';
```
kamil-tekiela: Maybe `getDefaultCollation`?

williamdes (approved):

kamyweb:
@MauricioFauth I think you need to improve test file test/classes/Stubs/DbiDummy.php where I hardcoded "utf8mb4_general_ci".

PR #19867, opened 2025-09-18
Remove background color from select on focus

### LINKED ISSUE #19817: Select theme arrow disappears when focused
### Describe the bug

When you click on a select in the default theme (pmahomme), the arrow disappears from there.

Image

I found out that this is because a background color is set for select when focused ([see here](https://github.com/phpmyadmin/phpmyadmin/blob/master/public/themes/original/scss/_common.scss#L1028)). When this is removed, the arrow no longer disappears.

Image

I would have done a PR, but maybe there is a reason why the background color is set?

### How to Reproduce

_No response_

### Expected behavior

_No response_

### Screenshots

_No response_

### Operating System

Windows

### Web Server

Nginx 1.29.1

### Database Server

MariaDB

### Database version

_No response_

### PHP version

8.3.24

### phpMyAdmin version

5.2.2

### browser

Google Chrome 139.0.7258.128

### Additional context

_No response_

robinmiau:
Fixes #19817 

*Theoretically, border could also be removed, as both is already set in Bootstrap.*

### DIFF
--- a/public/themes/pmahomme/scss/_common.scss
+++ b/public/themes/pmahomme/scss/_common.scss
@@ -169,7 +169,6 @@ input {
 
 select:focus {
   border: 1px solid #7c7c7c;
-  background: #fff;
 }
 
 input {


MauricioFauth (approved):

williamdes (approved):
To merge into 5.2, or not ?

PR #19868, opened 2025-09-19
Improve documentation about PHP extensions

### LINKED ISSUE #19594: Is the sodium extension required?
When upgrading my installation to 5.2.2, I diffed config.sample.inc.php from the old version and noticed that it now mentions FAQ 2.10 for generating `$cfg['blowfish_secret']` (git ref: 69bc866).  That FAQ entry refers to using `sodium_hex2bin()`, as does the [cookie authentication documentation](https://docs.phpmyadmin.net/en/release_5_2_2/config.html#cookie-authentication-options):

> Changed in version 5.2.0: Since version 5.2.0, phpMyAdmin uses the [sodium_crypto_secretbox](https://www.php.net/sodium_crypto_secretbox) and [sodium_crypto_secretbox_open](https://www.php.net/sodium_crypto_secretbox_open) PHP functions to encrypt and decrypt cookies, respectively.

However, the [PHP Extension requirements](https://docs.phpmyadmin.net/en/latest/require.html#php) do not mention sodium.  I don't currently have sodium installed on my system, so using `sodium_hex2bin()` in my config file would certainly fail. And yet, with blowfish_secret set to a 32-character string, logins are working fine, despite the lack of sodium_crypto_secretbox!

I'd be happy to `zypper install php8-sodium` if needed, but please document this. 😃  When search for existing issues/PRs, all I could find was #19422 adding a hard requirement for it in master.

williamdes:
### Description

Fixes #19594

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -883,13 +883,10 @@ Here are a few points to check:
 ---------------------------------
 
 To be able to see a progress bar during your uploads, your server must
-have the `uploadprogress <https://pecl.php.net/package/uploadprogress>`_ extension, and
-you must be running PHP 5.4.0 or higher. Moreover, the JSON extension
-has to be enabled in your PHP.
+have the `uploadprogress <https://pecl.php.net/package/uploadprogress>`_ extension enabled.
 
-If using PHP 5.4.0 or higher, you must set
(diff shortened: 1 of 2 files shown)


MauricioFauth (approved):

ibennetch:
This looks great to me, go ahead and merge.

PR #19869, opened 2025-09-20
sql.ts:checkSavedQuery() clean up logic

anzz1:
### Description

The logic introduced in aa73769 is flawed and unnecessarily checks the cookie value even if localstorage check already returns true, plus new unnecessary variables are introduced.

I didn't really understand the license part but I hereby relinquish all and every rights to this piece of code and it can be used as however anyone sees fit. I do not need attribution or anything like that if this contribution wasn't done properly just close the PR and copy-paste the code and commit it yourself.

### DIFF
--- a/resources/js/sql.ts
+++ b/resources/js/sql.ts
@@ -1301,17 +1301,14 @@ function checkSavedQuery () {
     let key = Sql.getAutoSavedKey();
     let buttonGetAutoSavedQuery = $('#saved');
 
-    let isAutoSavedInLocalStorage = isStorageSupported('localStorage') && (typeof window.localStorage.getItem(key) === 'string');
-    // @ts-ignore
-    let isAutoSavedInCookie = window.Cookies.get(key, { path: CommonParams.get('rootPath') });
-
-    if (isAutoSavedInLocalStorage || isAutoSavedInCookie) {
+    if (isStorageSupported('localStorage') && typeof window.localStorage.getItem(key) === 'string') {
         buttonGetAutoSavedQuery.prop('disabled', false);
-
-        return;
+    // @ts-ignore
+    } else if (window.Cookies.get(key, { path: CommonParams.get('rootPath') })) {
+        buttonGetAutoSavedQuery.prop('disabled', false);
+    } else {
+        buttonGetAutoSavedQuery.prop('disabled', true);
     }
-
-    buttonGetAutoSavedQuery.prop('disabled', true);
 }
 
 AJAX.registerOnload('sql.js', function () {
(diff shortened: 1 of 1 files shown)


resources/js/sql.ts
```diff
@@ -1301,17 +1301,14 @@ function checkSavedQuery () {
...
-    let isAutoSavedInCookie = window.Cookies.get(key, { path: CommonParams.get('rootPath') });
-
-    if (isAutoSavedInLocalStorage || isAutoSavedInCookie) {
+    } else if (window.Cookies.get(key, { path: CommonParams.get('rootPath') })) {
         buttonGetAutoSavedQuery.prop('disabled', false);
-
-        return;
+    } else {
+        buttonGetAutoSavedQuery.prop('disabled', true);
     }
```
MoonE, line 1308: It seems the `//@ts-ignore` is hiding a bug and the second argument to `Cookies.get()` is never used.
https://github.com/js-cookie/js-cookie/blob/ab3f67fc4fad88cdf07b258c08e4164e06bf7506/src/api.mjs#L50-L53
Should use `withAttributes` instead.
https://github.com/js-cookie/js-cookie/blob/ab3f67fc4fad88cdf07b258c08e4164e06bf7506/src/api.mjs#L89-L91

```suggestion
    const key = Sql.getAutoSavedKey();
    const haveAutoSavedQuery =
        (isStorageSupported('localStorage') && typeof window.localStorage.getItem(key) === 'string') ||
        window.Cookies.withAttributes({ path: CommonParams.get('rootPath') }).get(key);

    $('#saved').prop('disabled', !haveAutoSavedQuery);
```

anzz1: If that second parameter was never used nor needed, wouldn't the correct way be to just scrap the rootPath attribute altogether?

MoonE: The second parameter is never used with the current version of js-cookie. There probably was some breaking change at some point.
The rootPath is needed when you have more than one phpmyadmin installation on the same server or something else uses the same cookie names. It filters the cookies to only include its own.

anzz1: Alright, it seems that GitHub updated the review UI at some point, it used to be good before but now it's incredibly bugged and shows the diff of your suggestion in relation to the original code and not to my modifications, plus the format is now all weird. Interesting how often development goes that you get the lightning in the bottle and make a great thing, but for some reason you then break it with updates.

Anyway, sorry for the rant, I think I got what you meant with your suggestion extracted.

Now I get though why some projects still use mailing lists and .patch files.
There is no way an eager intern at some company can go and break your workflow, .patch files will always work and nobody can break it. Unless they literally break into your house and steal your diffing tools from your hard drive.

williamdes:
Hi @anzz1 
Do you have a patch file to send here ?
I can apply it.
Did you take the last suggestion in your changes ?

PR #19873, opened 2025-09-28
Add a changelog file using the Keep a Changelog format

MauricioFauth:
Only includes the changes for the phpMyAdmin 6.0 release series.

I added the release series version in the name of the file to keep the changelog file cleaner. This way the process of removing/archiving changes from old versions won't be necessary. For example, for phpMyAdmin 6.1, a file `CHANGELOG-6.1.md` will be added and the file `CHANGELOG-6.0.md` will be removed.

The changes are duplicated in the file `ChangeLog` for now, because this file is used by the `ChangeLogController` and in the file `app/vendor_config.php`. Both should be changed to this new changelog file.


williamdes (approved):
Okay, this looks good
could you adapt the wiki page on changelogs afterwards?

I will miss the old file, but sometimes stuff needs to change
that said, maybe remove the links in the markdown file
the php code can do them?
it will be complicated to do entries otherwise

This tool to do changelogs automatically is quite good https://github.com/orhun/git-cliff

ibennetch:
I like how this breaks apart changes from new features or things that have been removed. I'm not one to stand in the way of progress, but I will also miss the old format. I really like breaking each major release in to its own file, that's a great idea.

I think this is much better than the old system (which we've gotten away from for some time anyway) where we used the `-` or `+` to indicate bug fixes or new features.

Personally, I would not need the github links, but if you think that's better I won't block it (which I suppose means it wouldn't need to be Markdown). We preprocess the file in some places to add the links, but whether that's better than putting the links in Markdown is debatable and not something I mind either way.

We should adapt the tool that generates https://demo.phpmyadmin.net/master-config/public/index.php?route=/changelog to use the new format.

Thanks for coming up with this.

MauricioFauth:
> maybe remove the links in the markdown file

> Personally, I would not need the github links

The links are not part of the Keep a Changelog format, they are not mandatory. So if you want to only add a simple entry without a link or issue/PR number you can do it.

I put the links there for reference only, because the issue/PR number itself has very little meaning outside GitHub. Even GitHub itself doesn't automatically creates links for issues in all places. 

> This tool to do changelogs automatically is quite good https://github.com/orhun/git-cliff

I don't like the idea of automatically generating changelogs. A meaningful change could be consisted of a single commit, or a pull request, or even multiple pull requests, so it's hard to auto generate a changelog file that's worth reading.

bin/internal/check-release-excludes.sh
```diff
@@ -144,7 +144,7 @@ validateExtension() {
                 foundFileExt
             fi
         ;;
-        ChangeLog)
+        CHANGELOG-[5-9].[0-9].md)
         ;;
```
williamdes, line 148: You can removed this one now

MauricioFauth: Won't it accuse to have found the `CHANGELOG-6.0.md` file?

williamdes: Oh sorry it highlighted the wrong one. I did mean the old format

PR #19881, opened 2025-10-14
Documentation improvements - remove outdated information and improve some areas

ibennetch:
### Description

There are a lot of places where we reference when a feature was introduced, especially in the 2.x versions. I think it's unnecessary, adds extra length to the documentation, and sometimes makes us look outdated (because of references to something from 2010 as "new").

I'm working on removing those in preparation of the 6.0 release, but before I put a lot of work in to it wanted to show a sample of what's in progress to make sure everyone is in agreement about this change.

For example, [here](https://github.com/phpmyadmin/phpmyadmin/pull/19881/files#diff-f0a5822170a0cb40ccc8b537d36cac7e27a986d9911b3a08c03feab8f9a47a13L639) or [here](https://github.com/phpmyadmin/phpmyadmin/pull/19881/files#diff-f0a5822170a0cb40ccc8b537d36cac7e27a986d9911b3a08c03feab8f9a47a13L937).

This is just a draft right now.


docs/setup.rst
```diff
@@ -787,7 +787,7 @@ Setup script on openSUSE
 Some openSUSE releases do not include setup script in the package. In case you
 want to generate configuration on these you can either download original
 package from <https://www.phpmyadmin.net/> or use setup script on our demo
-server: <https://demo.phpmyadmin.net/master/setup/>.
+server: <https://demo.phpmyadmin.net/master/public/setup/>.
```
williamdes, line 790: This is on my long todo list, for 6.0 we should avise users to use public as the root.
As should the demo server


```suggestion
with the ``/setup`` suffix. The changes are not saved to the server, you need to
use the :guilabel:`Download` button to save them to your computer and then upload
to the server.

@@ -787,7 +787,7 @@ Setup script on openSUSE
Some openSUSE releases do not include setup script in the package. In case you
want to generate configuration on these you can either download original
package from <https://www.phpmyadmin.net/> or use setup script on our demo
server: <https://demo.phpmyadmin.net/master/setup/>.
```

MauricioFauth (approved):
:100:

docs/config.rst
```diff
@@ -296,7 +296,7 @@ Server connection settings
     :type: string
     :default: ``''``
 
-    The path to the socket to use. Leave blank for default. To determine
+    The path to the socket to use. Leave blank for the PHP default socket. To determine
```
williamdes: It's not the default PHP socket but maybe more the default socket file used by the PHP mysqli driver?

docs/config.rst
```diff
@@ -749,10 +735,12 @@ Server connection settings
     :type: string
     :default: ``''``
 
-    Only useful when using phpMyAdmin with multiple server entries. If
-    set, this string will be displayed instead of the hostname in the
-    pull-down menu on the main page. This can be useful if you want to
-    show only certain databases on your system, for example. For HTTP
+    Sets a custom, descriptive label for the server which will be shown
+    in multiple locations instead of the host name.
+    With auth_type cookie,
```
williamdes:
```suggestion
    With `auth_type` `cookie`,
```
To avoid translations
maybe use a full example?

Or use `config:option:: $cfg['Servers'][$i]... `

docs/config.rst
```diff
@@ -811,9 +797,7 @@ Server connection settings
     :type: string or false
     :default: ``''``
 
-    .. versionadded:: 2.2.4
-
-    Since release 2.2.4 you can describe, in a special 'relation' table,
+    You can describe, in a special 'relation' table,
```
williamdes:
```suggestion
    You can describe, in a special ``relation`` table,
```

Backtick it like the other one, avoid a translation

docs/config.rst
```diff
@@ -1498,6 +1472,10 @@ Server connection settings
     ``SHOW`` commands instead), because of speed issues when many
     databases are present.
 
+    The performance penalty for using ``INFORMATION_SCHEMA`` has largely been
+    resolved in most database servers, so for most users keeping this set to
+    false is recommended.
```
williamdes:
```suggestion
    ``false`` is recommended.
```

docs/config.rst
```diff
@@ -2786,6 +2747,24 @@ Export and import settings
 
     Default export format.
 
+    Value values are:
+
+    * ``'codegen'`` NHibernate code generation output
+    * ``'csv'`` :term:`Comma--separated text<CSV>`
+    * ``'excel'`` CSV file optimized for use with Microsoft Excel
+    * ``'htmlexcel'`` **DOES NOT SEEM TO WORK**
```
williamdes: Removed now

docs/config.rst
```diff
@@ -2786,6 +2747,24 @@ Export and import settings
...
+    * ``'latex'`` :term:`LaTeX<LATEX>` typesetting markup file
+    * ``'ods'`` OpenDocument format spreadsheet as used in LibreOffice
+    * ``'odt'`` OpenDocument format text document as used in LibreOffice
+    * ``'pdf'`` :term:`PDF`
+    * ``'sql'`` :term:`SQL` is the standard import and export format for database servers
+    * ``'texytext'`` Text format for the Texy! markup language
+    * ``'xml'`` :term:`eXtensible Markup Language<XML>` format for sharing documents and marking up text
+    * ``'yaml'`` Another variation of markup language used for data serialization, designed to be easier for humans to read
+
+    ** MISSING MediaWiki table and PHP array AS OPTIONS **
```
williamdes: Todo

docs/setup.rst
```diff
@@ -741,7 +741,7 @@ setup feature. The file can be generated using the setup and you can download it
 for upload to the server.
 
 Next, open your browser and visit the location where you installed phpMyAdmin,
-with the ``/setup`` suffix. The changes are not saved to the server, you need to
+with the ``/public/setup`` suffix. The changes are not saved to the server, you need to
```
williamdes, line 744: They should not have public in the path, else it means the public prefix root was not well deployed

docs/setup.rst
```diff
@@ -1208,9 +1208,7 @@ Config authentication mode
   loading the phpMyAdmin site. This is by design but could allow any
   user to access your installation. Use of some restriction method is
   suggested, perhaps a :term:`.htaccess` file with the HTTP-AUTH directive or disallowing
-  incoming HTTP requests at one’s router or firewall will suffice (both
-  of which are beyond the scope of this manual but easily searchable
-  with Google).
+  incoming HTTP requests at one’s router or firewall.
```
williamdes:
```suggestion
  incoming HTTP requests at one's router or firewall.
```

docs/setup.rst
```diff
@@ -1227,13 +1225,13 @@ are always ways to make your installation more secure:
...
-* Remove the ``tests`` directory from phpMyAdmin, unless you are developing and need a test suite.
-* Remove the ``setup`` directory from phpMyAdmin, you will probably not
+* Remove the :file:`tests/` directory from phpMyAdmin, unless you are developing and need a test suite.
+* Remove the :file:`public/setup/` directory from phpMyAdmin, you will probably not
   use it after the initial setup.
 * Properly choose an authentication method - :ref:`cookie`
   is probably the best choice for shared hosting.
-* Deny access to auxiliary files in :file:`./libraries/` or
-  :file:`./templates/` subfolders in your webserver configuration.
+* Deny access to auxiliary files in :file:`libraries/` or
```
williamdes, line 1233: They did not configure the server to have public as the root
red flag

docs/setup.rst
```diff
@@ -1248,6 +1246,8 @@ are always ways to make your installation more secure:
   phpMyAdmin, you can use :config:option:`$cfg['Servers'][$i]['AllowDeny']['rules']` to limit them
   or :config:option:`$cfg['Servers'][$i]['AllowRoot']` to deny root user access.
 * Enable :ref:`2fa` for your account.
+* Consider hiding connection errors from the login page by setting
+  :config:option:`$cfg['Servers'][$i]['hide_connection_errors']` to true.
```
williamdes:
```suggestion
  :config:option:`$cfg['Servers'][$i]['hide_connection_errors']` to ``true``.
```

docs/setup.rst
ibennetch: I think that this is the current behavior of both my system and the demo server, is that incorrect?

docs/setup.rst
ibennetch: Should we change this section, then?

ibennetch:
Thanks for the great feedback. I've incorporated most of your comments in https://github.com/phpmyadmin/phpmyadmin/pull/19881/commits/e8446253aa6d6a5903373df01838be82e2d1070c which I just pushed along with some other work that's in progress.

docs/setup.rst
ibennetch: That's a great catch, although I'm not sure it really matters in rst. I have fixed it anyway to be safe.

docs/setup.rst
williamdes: yes, but it is a webserver configuration issue
We should make sure our users understand the public folder should be the root

docs/setup.rst
williamdes: We can keep this section to list all folders to deny access for, if users are not able to some reason to set the root folder to public/

ibennetch:
I think this is ready to go except I think we should decide how to handle the phpmyadmin/setup vs phpmyadmin/public/setup section from https://github.com/phpmyadmin/phpmyadmin/pull/19881#pullrequestreview-3337446202.

I believe what I've written here matches the current behavior, so if that's true my suggestion is to improve/fix the suggestion to remove the /public/ part in another commit.

PR #19895, opened 2025-10-19
Fix #19888 - Update documentation of `NavigationTreeDefaultTabTable`

### LINKED ISSUE #19888: [Feat]: Configuration directives should allow English values as well as localized values
### phpMyAdmin version

6.0.0-dev probably, but affects all versions

### Describe the feature

According to the documentation, when phpMyAdmin is using a translation instead of the default English, `$cfg['NavigationTreeDefaultTabTable']` and `$cfg['NavigationTreeDefaultTabTable2']` require the values to be in the localized text (so for instance, in Czech the configuration value of "struktura" instead of "structure".

This makes it difficult to document and in the case of a partial translation or a translation that changes between versions, the users can be confused about the expected value.

There are probably other places this occurs, too.

### Describe the solution

I propose that we allow the user to use the English values regardless of their localized language, in addition to the localized translated word.

### Describe the alternatives

We could leave it how it is as not many users have complained about this.

### Additional context

_No response_

matt-h:
### Description

Update documentation of `NavigationTreeDefaultTabTable` to not specify local language.

The current code allows the English words only and not localized words. https://github.com/phpmyadmin/phpmyadmin/blob/9eb9df6846cc0805ebc62fd615784f11d4a29293/src/Config/Settings.php#L3916-L3950

From what I can tell even from back when the doc was updated to include this, the config options only supported English in the config file even though the display of them in the UI would show localized versions. https://github.com/phpmyadmin/phpmyadmin/commit/1cac9d349505930f39b537578cb97f375870d263

Fixes #19888

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -2347,8 +2347,7 @@ Navigation panel setup
     :default: ``'structure'``
 
     Defines the tab displayed by default when clicking the small icon next
-    to each table name in the navigation panel. The possible values are the
-    localized equivalent of:
+    to each table name in the navigation panel. The possible values are:
 
     * ``structure``
     * ``sql``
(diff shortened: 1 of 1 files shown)


williamdes (approved):

ibennetch:
Thanks!

PR #19910, opened 2025-10-29
Switch the theme used when building our HTML documents

### LINKED ISSUE #19909: [Feat]: Documentation - improve styling for generated HTML files
### phpMyAdmin version

5.2.3

### Describe the feature

This is observed and reported against PMA 5.2.3, but has been the case for some time.

When we build the HTML files for our documentation, the theme used is the sphinx "default" one, which is rather simplistic. In particular, the `:guilabel:` role is not styled any differently than the default text (the only style applied is to set the font-family to sans-serif, which is already the body style).

Further, our documentation doesn't match between what's generated internally and what's available online at readthedocs. The styles are very different.

### Describe the solution

I think the best solution is to update our generated documentation to use the readthedocs style. This is easily accomplished as the rtd theme is packaged for Debian and is also available for others via pip, it unifies our look across all the documentation, and it gives a very nice style to the `:guilabel:` items.

### Describe the alternatives

We could modify the existing theme or select a completely different theme (for both RTD and our own generated docs).

"sphinxdoc" in particular is not a bad alternate choice, but I still prefer the rtd theme over it.

### Additional context

_No response_

ibennetch:
Switch the theme used when building our HTML documents to the Read The Docs theme (rtd)

This is more consistent with the look of our online documentation.

### Description

This switches our generated HTML documentation to match the theme used by Read The Docs for our online documentation.

Fixes #19909

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -98,7 +98,7 @@ pygments_style = 'sphinx'
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
-html_theme = 'default'
+html_theme = 'sphinx_rtd_theme'
 
 # See: https://docs.readthedocs.io/en/stable/reference/environment-variables.html#envvar-READTHEDOCS
 on_rtd = os.environ.get('READTHEDOCS', None) == 'True'


ibennetch:
I've created this but we can first discuss it at #19909 before merging this.

MauricioFauth (approved):

docs/conf.py
```diff
@@ -98,7 +98,7 @@
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
-html_theme = 'default'
+html_theme = 'sphinx_rtd_theme'
```
williamdes, line 101: See my fix below for READTHEDOCS env

Maybe you can remove it
or make the try catch the default behaviour
some users may build docs themselves and not want or have the theme

williamdes:
I merged this as 72287e2ebfaa2bc318c8075765ecbbf3ebd3ea8d
With a hint for Debian based users to know what package to install to have the theme.

PR #19913, opened 2025-10-30
Set MYSQLI_OPT_LOCAL_INFILE before mysqli->real_connect.

MegaManSec:
This is required for this to actuall work.

This bug was found with ZeroPath.

### DIFF
--- a/src/Dbal/DbiMysqli.php
+++ b/src/Dbal/DbiMysqli.php
@@ -89,6 +89,8 @@ class DbiMysqli implements DbiExtension
             $host = $server->host;
         }
 
+        $mysqli->options(MYSQLI_OPT_LOCAL_INFILE, (int) defined('PMA_ENABLE_LDI'));
+
         try {
             $mysqli->real_connect(
                 $host,
@@ -126,8 +128,6 @@ class DbiMysqli implements DbiExtension
 
             throw new ConnectionException($errorNumber . ': ' . $errorMessage, $errorNumber, $exception);
         }
-
-        $mysqli->options(MYSQLI_OPT_LOCAL_INFILE, (int) defined('PMA_ENABLE_LDI'));
 
         mysqli_report(MYSQLI_REPORT_OFF);
 


MegaManSec:
Wasn't quite sure which branch to create this PR for; any pointers appreciated.

kamil-tekiela:
That is not true. I don't know where you got this information from but the source must be wrong about this. I have never heard of `ZeroPath` but maybe you can report it to them.

MegaManSec:
It's from here: https://www.php.net/manual/en/mysqli.options.php

>mysqli_options() should be called after [mysqli_init()](https://www.php.net/manual/en/mysqli.init.php) and before [mysqli_real_connect()](https://www.php.net/manual/en/mysqli.real-connect.php).

The sentiment of this is mirrored in, for example, https://www.php.net/manual/en/mysqli.construct.php:

>Libmysqlclient uses the default charset set in the my.cnf or by an explicit call to [mysqli_options()](https://www.php.net/manual/en/mysqli.options.php) prior to calling [mysqli_real_connect()](https://www.php.net/manual/en/mysqli.real-connect.php), but after [mysqli_init()](https://www.php.net/manual/en/mysqli.init.php).

Are the PHP docs wrong?

kamil-tekiela:
Not wrong, just slightly inaccurate. Not all options need to be set before connecting. Only the options that affect the connection need to be set before. This part is not mentioned anywhere in the PHP manual.

MegaManSec:
https://bugs.php.net/bug.php?id=77496

it's actually a bug in php, huh. this pr (my pr) would have undone the specific workaround for this bug (https://github.com/phpmyadmin/phpmyadmin/commit/c5e01f84ad48c5c626001cb92d7a95500920a900). would recommend documenting this

kamil-tekiela:
Bug or an undocumented feature. I wouldn't call it a workaround. It's in exactly the right spot where it should be. But you are right that PHP manual should be clearer about this. I don't feel like submitting a PR there now, but feel free to do so if you wish.

MegaManSec:
Its confirmed as a bug that was introduced by accident in a specific commit, phpmyadmin is specifically called out in that bug report

Issue #19877, opened 2025-10-08
[Bug]: Root user cannot manage user accounts (edit privileges or create new users) in phpMyAdmin 5.2.3


fabiotnt:
### Describe the bug

After updating to phpMyAdmin 5.2.3, I can no longer manage user accounts when logged in as root.
The User accounts page loads, but shows the message:

"You do not have the privileges to edit users."

As a result, I cannot:

Edit existing user privileges

Create new users

This behavior does not occur in version 5.2.2, using the same environment and configuration.

### How to Reproduce

Steps to reproduce

1. Log in to phpMyAdmin 5.2.3 as root

2. Open the User accounts tab
→ The interface shows a message stating you lack privileges

### Expected behavior

The root user should be able to view and manage user accounts — edit privileges and create new users — as in version 5.2.2.

### Operating System

Linux

### Web Server

nginx/1.28.0

### Database Server

MySQL

### Database version

5.7.44-google-log

### PHP version

8.3.26

### phpMyAdmin version

5.2.3

### browser

Chrome – Version 141.0.7390.55 (Official Build) (x86_64)

### Additional context

It's a serverless deploy in GCR using CloudSQL and Docker based in php:8.3-fpm-alpine

phpMyAdmin: 5.2.3
Server: (Localhost via UNIX socket)
Server type: MySQL
Server version: 5.7.44-google-log (Google)
Protocol version: 10
User: root
Server charset: UTF-8 Unicode (utf8)

Web server: nginx/1.28.0
PHP version: 8.3.26
Database client version: libmysql - mysqlnd 8.3.26
PHP extensions: mysqli, curl, mbstring, sodium
Deployment: Google Cloud Run

williamdes:
Related to https://github.com/phpmyadmin/phpmyadmin/pull/19864 and #19109 
What is your server collation?

shdon:
As a quick workaround until the fix is released, in `config.inc.php`, set:
```$cfg ['Servers'][$i]['DisableIS'] = true;```

rodehoed:
I'm seeing this also, my server collation is: utf8mb4_unicode_ci

<img width="2356" height="214" alt="Image" src="https://github.com/user-attachments/assets/86c1181f-a5f2-4df7-b690-2cbf2b899668" />

williamdes:
@MauricioFauth could you please check if there is a new way to fix this ?

matt-h:
[...]

PR #19923, opened 2025-11-07
Fix collation error when checking for grant permission

MauricioFauth:
- Fixes https://github.com/phpmyadmin/phpmyadmin/issues/19877


williamdes (approved):
Thank you!

Issue #14129, opened 2018-03-27
[4.8 RC1] Allow Multiple U2F Sticks.


My1:
### Steps to reproduce
1. Configure one U2F Stick
2. go back to 2FA Settings
3. try to configure another one

### Expected behaviour
Be able to add another few as backups, as basically anything that allows U2F can do, because you can lose U2F keys, they can break etc etc.

### Actual behaviour
It just tells the user 2FA is enabled with an option to disable.

### Server configuration
**Operating system**: Win 8.1 Pro

**Web server:** Caddy 0.10.11

**Database:** MySQL 5.6.21

**PHP version:** 7.2.0

**phpMyAdmin version:** 4.8 RC1

### Client configuration
**Browser:** Opera 52

**Operating system:** Win 8.1 Pro

williamdes:
This really makes sense, I added it to the 5.2 milestone so it gets some attention

williamdes:
Indeed !

That would a be a nice to have for 6.0

PR #19925, opened 2025-11-08
Fix #19900 - two factor auth disappears

### LINKED ISSUE #19900: [Bug]: Two-factor authentication disappears
### Describe the bug

When I change the 「Remember table's sorting」 setting, two-factor authentication (Hardware Security Key (WebAuthn/FIDO2)) disappears.

### How to Reproduce

1. go to $your_phpmyadmin_domain/index.php?route=/preferences/two-factor
2. Configure two-factor authentication choose Authentication Application (2FA) or Hardware Security Key (WebAuthn/FIDO2)
3. Select a table and click the settings icon in the upper right corner -> Page-related settings
4. Change Remember table's sorting -> Apply
5. Then Two-factor authentication disappears 🤣

### Expected behavior

Changing the 「Remember table's sorting」should not affect the 「two-factor authentication」 functionality

### Screenshots

Image

Image

Image

Image

### Operating System

Linux

### Web Server

_No response_

### Database Server

_No response_

### Database version

8.0.31

### PHP version

8.3.26

### phpMyAdmin version

5.2.3

### browser

_No response_

### Additional context

_No response_

efturtle:
Fixes issue where two-factor authentication settings were being lost when saving page-specific user preferences through the settings UI.

Fixes #19900 

### DIFF
--- a/libraries/classes/UserPreferences.php
+++ b/libraries/classes/UserPreferences.php
@@ -125,6 +125,16 @@ class UserPreferences
      */
     public function save(array $config_array)
     {
+        // Smart merge: If this looks like a partial save (missing 2fa but we have it stored),
+        // automatically merge to prevent accidental overwrites
+        if (!isset($config_array['2fa'])) {
+            $existingPrefs = $this->load();
+            if (isset($existingPrefs['config_data']['2fa'])) {
+                // This is likely a partial save from page settings - merge to preserve 2fa
+                $config_array = array_merge($existingPrefs['config_data'], $config_array);
+            }
+        }
+
         global $dbi;
 
         $relationParameters = $this->relation->getRelationParameters();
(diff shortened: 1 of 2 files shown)


efturtle:
will work on the tests

williamdes (approved):
This fix is a nice one, I am wondering if it should be more general or not

efturtle:
Yeah, for 2FA, this is good. Since both DB and session storage use the same array structure, it works consistently. Maybe separating 2FA config in the future could be worth considering?

williamdes:
> Yeah, for 2FA, this is good. Since both DB and session storage use the same array structure, it works consistently. Maybe separating 2FA config in the future could be worth considering?

Probably there is an idea to find, 2FA need to support multiple hardware tokens: https://github.com/phpmyadmin/phpmyadmin/issues/14129

kamil-tekiela:
The Selenium failure was legit. You might have fixed 2fa with this hack, but you bricked the rest of the settings. 

I don't want to investigate the proper fix for this now, but it seems to me like there is much more broken than this simple fix implies. I am a bit confused as to why 2fa is saved in config even though it doesn't seem to be a config option. Perhaps it should have its own table. The whole feature of UserPreferences is one of the most confusing features of PMA for me. 

@williamdes Do you think it would be a good idea to revert this PR for the time being?

williamdes:
> @williamdes Do you think it would be a good idea to revert this PR for the time being?

Well, if there is no other way.

> but you bricked the rest of the settings.

can you describe this a bit more ?

kamil-tekiela:
I didn't investigate it fully, but from what I understand, is that if a setting is cleared out (empty string) then it is removed from the UserPreferences. This will not happen with this fix because it reads the old values again and only replaces a setting if it has a value coming from the form. This means it's not possible to remove certain settings as in the test.

MauricioFauth:
> The Selenium failure was legit. You might have fixed 2fa with this hack, but you bricked the rest of the settings.

I opened a pull request with the fix. 

- https://github.com/phpmyadmin/phpmyadmin/pull/20047

> I am a bit confused as to why 2fa is saved in config even though it doesn't seem to be a config option. Perhaps it should have its own table. The whole feature of UserPreferences is one of the most confusing features of PMA for me.

And `2fa` is not saved using `UserPreferences`, so it does not follow the same format.

PR #19926, opened 2025-11-09
Fix #19855 - Allow non-SQL export plugins to export table structure

### LINKED ISSUE #19855: [6.0]: Exporting table structure doesn't work
### Describe the bug

Exporting table structure ~~as PDF~~ doesn't work in 6.0.0-dev. It seems all formats that allow exporting structure or data and structure, don't show the structure.

https://github.com/user-attachments/assets/414e60b3-fef4-4914-91fe-db27cb8079e2

### How to Reproduce

1. Go to `Export`
2. Select `Custom` export method
3. Select `PDF` format
4. Check `structure`
5. Click on `Export`
6. See missing table structure in the exported PDF

### Expected behavior

_No response_

### Screenshots

_No response_

### Operating System

_No response_

### Web Server

_No response_

### Database Server

_No response_

### Database version

_No response_

### PHP version

_No response_

### phpMyAdmin version

6.0.0-dev

### browser

_No response_

### Additional context

_No response_

efturtle:
- Modified conditions on lines 822 and 826 in exportTable()
- Added OR clause to allow non-SQL plugins (PDF, Excel, etc.)
- Restores structure export for plugins without hasCreateTable()/hasCreateView()

The conditional logic in exportTable() was specific to ExportSql,
preventing other export plugins (PDF, Excel, etc.) from calling
exportStructure(). Modified conditions on lines 822 and 826 to allow
non-SQL plugins to export structure without requiring SQL-specific
methods like hasCreateTable() and hasCreateView().

Fixes #19855

### DIFF
--- a/src/Export/Export.php
+++ b/src/Export/Export.php
@@ -627,12 +627,12 @@ class Export
                 if ($isView) {
                     if (
                         $separateFiles === ''
-                        && $exportPlugin instanceof ExportSql && $exportPlugin->hasCreateView()
+                        && $exportPlugin->hasCreateView()
                         && ! $exportPlugin->exportStructure($db->getName(), $table, 'stand_in', $aliases)
                     ) {
                         break;
                     }
-                } elseif ($exportPlugin instanceof ExportSql && $exportPlugin->hasCreateTable()) {
+                } elseif ($exportPlugin->hasCreateTable()) {
                     $tableSize = self::$maxSize;
                     // Checking if the maximum table size constrain has been set
                     // And if that constrain is a valid number or not
@@ -679,7 +679,7 @@ class Export
             // now export the triggers (needs to be done after the data because
             // triggers can modify already imported tables)
             if (
-                ! ($exportPlugin instanceof ExportSql && $exportPlugin->hasCreateTrigger())
+                ! $exportPlugin->hasCreateTrigger()
                 || $structureOrData === StructureOrData::Data
                 || ! in_array($table, $tableStructure, true)
             ) {
@@ -697,7 +697,7 @@ class Export
             $this->saveObjectInBuffer('table_' . $table, true);
         }
 
-        if ($exportPlugin instanceof ExportSql && $exportPlugin->hasCreateView()) {
+        if ($exportPlugin->hasCreateView()) {
             foreach ($views as $view) {
                 // no data export for a view
                 if ($structureOrData === StructureOrData::Data) {
@@ -818,12 +818,12 @@ class Export
         $isView = $tableObject->isView();
         if ($structureOrData !== StructureOrData::Data) {
             if ($isView) {
-                if (! $exportPlugin instanceof ExportSql || $exportPlugin->hasCreateView()) {
+                if ($exportPlugin->hasCreateView()) {
                     if (! $exportPlugin->exportStructure($db, $table, 'create_view', $aliases)) {
                         return;
                     }
                 }
-            } elseif (! $exportPlugin instanceof ExportSql || $exportPlugin->hasCreateTable()) {
+            } elseif ($exportPlugin->hasCreateTable()) {
                 if (! $exportPlugin->exportStructure($db, $table, 'create_table', $aliases)) {
                     return;
                 }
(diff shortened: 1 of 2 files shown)


liviuconcioiu:
Not sure about this change because

```PHP
if ($exportPlugin instanceof ExportSql && $exportPlugin->hasCreateView() || ! $exportPlugin instanceof ExportSql) {
```

is the same as

```PHP
if ($exportPlugin->hasCreateView()) {
```

efturtle:
Not necessarily.
The first part of the condition

```php
if ($exportPlugin instanceof ExportSql && $exportPlugin->hasCreateView())
``` 

checks whether the plugin is using ExportSql and has the “create view” flag enabled.

The new part added in this PR —
```php
|| ! $exportPlugin instanceof ExportSql
``` 
— ensures the condition also passes when the plugin is something other than ExportSql (like ExportPdf, ExportOdt, etc.).

Because of operator precedence in PHP, the `&&` part will be evaluated before the `||`, so it should behave correctly when SQL is selected in the dropdown for export.

src/Export/Export.php
```diff
@@ -818,12 +818,12 @@ public function exportTable(
         $isView = $tableObject->isView();
         if ($structureOrData !== StructureOrData::Data) {
             if ($isView) {
-                if ($exportPlugin instanceof ExportSql && $exportPlugin->hasCreateView()) {
+                if ($exportPlugin instanceof ExportSql && $exportPlugin->hasCreateView() || ! $exportPlugin instanceof ExportSql) {
```
kamil-tekiela:
```suggestion
                if (! $exportPlugin instanceof ExportSql || $exportPlugin->hasCreateView()) {
```

src/Export/Export.php
```diff
@@ -818,12 +818,12 @@ public function exportTable(
...
                     if (! $exportPlugin->exportStructure($db, $table, 'create_view', $aliases)) {
                         return;
                     }
                 }
-            } elseif ($exportPlugin instanceof ExportSql && $exportPlugin->hasCreateTable()) {
+            } elseif ($exportPlugin instanceof ExportSql && $exportPlugin->hasCreateTable() || ! $exportPlugin instanceof ExportSql) {
```
kamil-tekiela:
```suggestion
            } elseif (! $exportPlugin instanceof ExportSql || $exportPlugin->hasCreateTable()) {
```

kamil-tekiela (changes requested):

src/Export/Export.php
efturtle: Made new commit with simplified condition thank you @kamil-tekiela

src/Export/Export.php
efturtle: Made new commit with simplified condition thank you @kamil-tekiela

williamdes (approved):
@kamil-tekiela are you okay with the new version ?

Old one: https://github.com/phpmyadmin/phpmyadmin/commit/038c59b1bf7cddd3634a665c864de976ba1eecf9

kamil-tekiela:
I am very confused on this one. The change from the linked commit should have been fine. It makes sense that `sql_*` options belong to the ExportSql plugin. What doesn't make sense is everything else. Why does Export class check for that? Why are all the options send over when the user selects PDF for example? 
image

It's possible that this PR fixes the issue, but WHY? This needs more investigation. I'll try to do it later this week. Or maybe @MauricioFauth has an idea on how this works.

efturtle:
I've refactored the solution by adding default implementations of methods hasCreateTable(), hasCreateView(), .. methods to the ExportPlugin base class, instead of relying on insteanceof.
Here is a video evidence: https://drive.google.com/file/d/19_Pa04C61BAEjUlj5bmhNKWyO_zc0OPP/view?usp=sharing

williamdes (approved):
Thanks, this is much better to understand
Maybe the methods should be abstract

williamdes:
Can you fix the test or implementation for `ExportOdsTest::testExportData` ?

kamil-tekiela:
But what if the user selects "data" only?

kamil-tekiela:
Also, I am not getting these fields on the UI. 
image

efturtle:
i'll look at the test
@kamil-tekiela I was going to add the evidence screencast here, but video weighed over 10MB. I uploaded to drive, I show data export https://drive.google.com/file/d/19_Pa04C61BAEjUlj5bmhNKWyO_zc0OPP/view?usp=sharing

kamil-tekiela:
Thank you very much for the PR and additional details. I investigated this a bit more. Your fix looks correct, but I don't think we should do it this way. I will propose an alternate PR with more changes. If you'd like, you can propose a regression test for this.

efturtle:
it's not a problem, thank you. sounds good I'll into the tests

efturtle:
closing this, follow up on https://github.com/phpmyadmin/phpmyadmin/pull/19996

PR #19929, opened 2025-11-11
Fix some PHP 8.5 deprecations

MauricioFauth:


libraries/classes/Image/ImageWrapper.php
```diff
@@ -145,6 +151,7 @@ public function destroy(): bool
             return true;
         }
 
+        // phpcs:ignore Generic.PHP.DeprecatedFunctions.Deprecated
         return imagedestroy($this->image);
```
kamil-tekiela, line 155: Why here without the version guard?

williamdes (approved):
Thanks
good question on the PR review comments

Can we agree to use `<username>/<name>` branches?
that way I can mute the git commit bot that send emails for PR from this repo

libraries/classes/Image/ImageWrapper.php
MauricioFauth: A version guard already exists as an early return in this case.

MauricioFauth:
> Can we agree to use / branches? that way I can mute the git commit bot that send emails for PR from this repo

What do you mean by to use / branches?

libraries/classes/Image/ImageWrapper.php
```diff
@@ -66,13 +66,19 @@ public static function create(int $width, int $height, ?array $background = null
 
         $backgroundColor = imagecolorallocate($image, $background['red'], $background['green'], $background['blue']);
         if ($backgroundColor === false) {
-            imagedestroy($image);
+            if (PHP_VERSION_ID < 80000) {
+                // phpcs:ignore Generic.PHP.DeprecatedFunctions.Deprecated
+                imagedestroy($image);
```
kamil-tekiela, line 71: I am actually unsure why we are using this function in these two places. Why does the image need to be destroyed before returning from the function?

MauricioFauth: I'm not sure as well. But these has been removed in the `master` branch.

MauricioFauth:
> Can we agree to use / branches? that way I can mute the git commit bot that send emails for PR from this repo

Sorry! I just noticed that I pushed a new branch to the phpmyadmin repo. It was a mistake.

williamdes:
> > Can we agree to use / branches? that way I can mute the git commit bot that send emails for PR from this repo
> 
> What do you mean by to use / branches?

The tags got removed automatically, comment edited
does it sound better?

PR #19837, opened 2025-08-31
Add support for bacon-qr-code v3

liviuconcioiu:
### Description

Before:

https://github.com/user-attachments/assets/27e9f753-bcc3-46ff-bea2-c4c26de272d5

After:

https://github.com/user-attachments/assets/ecb12a94-534f-4b79-bf09-1e3c1e450349

PR #19932, opened 2025-11-11
Suppress PHP deprecations from third-party packages

MauricioFauth:


kamil-tekiela:
If it's just a deprecation why are you suppressing it on 5_2? It's not an error.

MauricioFauth:
> If it's just a deprecation why are you suppressing it on 5_2? It's not an error.

When in development mode, deprecations are converted to exceptions, causing some functionalities to not work correctly.

Example from https://github.com/phpmyadmin/phpmyadmin/pull/19837:

https://github.com/user-attachments/assets/27e9f753-bcc3-46ff-bea2-c4c26de272d5

MauricioFauth:
I'm doing this for these third-party packages only, because they released fixes for these deprecations in newer major versions.
For the `strftime` function I did it to just remove the `@` operator and to suppress only the deprecation.

kamil-tekiela:
What do you think about making the code not convert deprecations to exceptions? I think the concern is that the deprecation might get burried in the HTML response and go unnoticed, but I don't think that's a problem. On 6.0 branch we can still have them throw, but even then, if they are silenced they can still be forgotten. We could also make the deprecations go away in development altogether.

williamdes:
I like the idea, but would rather take it directly to the deprecation error render
and block the render
most users complain because they render
I am unsure if we should throw them in dev mode
but it will be such a pain that we will fix them more quickly

Issue #19904, opened 2025-10-22
phpMyAdmin 5.2.3 - Queries without LIMIT return no results with 'Failed to parse JSON' error on PHP 8.4.13


sebas-cemic:
phpMyAdmin: 5.2.3
PHP: 8.4.13
MariaDB: 12.0.2
Windows XAMPP
Console error
Is ok in phpMyAdmin 5.2.2

<img width="1152" height="667" alt="Image" src="https://github.com/user-attachments/assets/b202bd6e-46a6-4629-8ee4-9dfc64641260" />

kamil-tekiela:
What is being sent instead of JSON? That's the important part, though I suppose if you aren't in development mode you won't get a meaningful reply.

sebas-cemic:
Query: SELECT * FROM `resultados` WHERE `codigoLaboratorio`=684

data send

db=progba&table=resultados&token=4c7b572961352c6e532473426344345d&is_js_confirmed=0&pos=0&goto=index.php%3Froute%3D%2Ftable%2Fsql&message_to_show=Su+consulta+se+ejecut%C3%B3+con+%C3%A9xito.&prev_sql_query=SELECT+*+FROM+%60resultados%60+WHERE+1&sql_query=SELECT+*+FROM+%60resultados%60+WHERE+%60codigoLaboratorio%60%3D684&bkm_label=&sql_delimiter=%3B&retain_query_box=1&fk_checks=0&SQL=Continuar&ajax_request=true&ajax_page_request=true&_nocache=1761137970619842576&token=4c7b572961352c6e532473426344345d

[image: image.png]

kamil-tekiela:
Sorry, I mean what is sent in the response. You know, the response that could not be parsed as JSON. It's clearly HTML, probably an HTML error page.

sebas-cemic:
I can't identify if there is an error page, when replacing it with the
phpmyadmin 5.2.2 directory in Xampp it works without problems.

M393:
Please provide the response of the failing request.

In the Chrome developer tools go to Network, select the failing request (has probably red font color) and go to the Response tab:
<img width="690" height="279" alt="Image" src="https://github.com/user-attachments/assets/d3357a28-d0f7-411f-bc28-315120e4ce2d" />

sebas-cemic:
[image: image.png]

kamil-tekiela:
Could you please open the GitHub page and take a look at what you posted? I don't think it works when you reply via email.

[...]

PR #19934, opened 2025-11-11
Show full error message and hide backtrace

kamil-tekiela:
There are two problems with how phpMyAdmin reports errors at the moment in production:
- it shows backtrace which isn't that useful to a developer, and especially not useful to the user, but it could expose some sensitive information to a potential attacker
- it doesn't show where the error/exception happened. 

This PR should help avoid situations like this: #19904

When a warning happens it will show it like this:
image

and when an error happens, it will show an alternative error page:
image


williamdes:
I am not sure to understand the after, can you share a screenshot?

kamil-tekiela:
This is the before:

Warning:
image

Error:
image

src/Error/ErrorHandler.php
```diff
@@ -224,11 +224,7 @@ public function handleError(
      */
     public function handleException(Throwable $exception): void
     {
-        $this->hideLocation = Config::getInstance()->config->environment !== 'development';
-        $message = $exception::class;
-        if (! ($exception instanceof \Error) || ! $this->hideLocation) {
-            $message .= ': ' . $exception->getMessage();
-        }
+        $message = $exception::class . ': ' . $exception->getMessage();
```
MauricioFauth: There are some error messages that shows a full path. That's why the message is not included. If there's a way to sanitize the message to remove full paths, that'll be great.

MauricioFauth (changes requested):

src/Error/ErrorHandler.php
kamil-tekiela: I can sanitize the path, e.g. with `basename()` but I would need to know when that happens. I can't think of a situation when that would happen now.

MauricioFauth: A `TypeError` is an example.

Example from https://github.com/phpmyadmin/phpmyadmin/issues/19537:
```
TypeError: PhpMyAdmin\Util::formatNumber(): Argument #1 ($value) must be of type string|int|float, null given, called in /home/demo/.pma-demo-web/master-config/src/Controllers/Database/Structure/RealRowCountController.php on line 76
```

kamil-tekiela: I never noticed until now. You are right that these two messages contain the full path. https://github.com/php/php-src/blob/master/Zend/zend_exceptions.c#L740

Issue #19578, opened 2025-01-29
After creating an event/procedure shown query is missing DELIMITER $$ to be executable again


M393:
### Describe the bug

After creating an event/procedure, the executed query is shown. If you try to copy it or Edit inline and execute it again it will be unsuccessful because it is missing the `DELIMITER $$` part.

### To Reproduce

1. Use the form to create a new event which includes a semicolon in the body
2. See the message for successful execution
3. It will not execute like this on a different server

### Screenshots

![Image](https://github.com/user-attachments/assets/d0ca5e2e-7d11-4580-9c7b-416ce5fd0187)
Should include `DELIMITER $$`

### Server configuration

- Operating system: Ubuntu 22.04
- Web server: Apache 2.4
- Database version: MariaDB 11.4
- PHP version: 8.1
- phpMyAdmin version: 5.2.2

### Client configuration

- Browser: Chromium
- Operating system: Windows 10

williamdes:
Related to https://github.com/phpmyadmin/phpmyadmin/pull/19942

PR #19942, opened 2025-11-14
Fix #17285 - Add DELIMITER $$ when exporting triggers

### LINKED ISSUE #17285: Exported Triggers not usable for import again
### Describe the bug

When I try to export all my triggers for any database schema, no delimiter is added to the output. Because of this it's impossible to import those triggers again.

Why do I need this? Because I needed to change the creator (user) of that triggers. The initial one was removed and no updates have been possible on that database.

### To Reproduce

1. Go to a database with many triggers
2. Click on menu item 'Trigger'
3. Select all trigger on very end of the page
4. Klick export on the very end of page

### Expected behavior

A sql file is downloaded OR the modal which opens has delimiters in generated sql code.

### Screenshots

none.

### Server configuration

phpMyAdmin: Versionsinformationen: 5.1.1
Operating system: Linux Ubuntu
Server: Localhost via UNIX socket
Server-Typ: MariaDB
Server-Verbindung: SSL wird nicht verwendet Dokumentation
Server-Version: 10.2.39-MariaDB-1:10.2.39+maria~bionic-log - mariadb.org binary distribution
Protokoll-Version: 10
Benutzer: admin@localhost
Server-Zeichensatz: UTF-8 Unicode (utf8mb4)
nginx/1.20.1
Datenbank-Client Version: libmysql - mysqlnd 7.4.26
PHP-Erweiterung: mysqli Dokumentation curl Dokumentation mbstring Dokumentation
PHP-Version: 7.4.26

### Client configuration

- Browser: Chrome

### Additional context

none.

M393:
Fixes #17285
image

and fixes a bug where the `Loading...` was not visible during export because it was immediately dismissed.

### DIFF
--- a/js/src/database/triggers.js
+++ b/js/src/database/triggers.js
@@ -98,7 +98,6 @@ const DatabaseTriggers = {
         } else {
             $.get($this.attr('href'), { 'ajax_request': true }, showExport);
         }
-        Functions.ajaxRemoveMessage($msg);
 
         function showExport (data) {
             if (data.success === true) {
--- a/libraries/classes/Database/Triggers.php
+++ b/libraries/classes/Database/Triggers.php
@@ -488,12 +488,12 @@ class Triggers
         }
 
         $itemName = $_GET['item_name'];
-        $triggers = $this->dbi->getTriggers($db, $table, '');
+        $triggers = $this->dbi->getTriggers($db, $table, '$$');
         $exportData = false;
 
         foreach ($triggers as $trigger) {
             if ($trigger['name'] === $itemName) {
-                $exportData = $trigger['create'];
+                $exportData = "DELIMITER $$\n" . $trigger['create'];
                 break;
             }
         }
--- a/libraries/classes/Database/Triggers.php
+++ b/libraries/classes/Database/Triggers.php
@@ -308,7 +308,7 @@ class Triggers
     {
         global $db, $table;
 
-        $temp = [];
+        $temp = null;
         $items = $this->dbi->getTriggers($db, $table, '');
         foreach ($items as $value) {
             if ($value['name'] != $name) {
@@ -318,7 +318,7 @@ class Triggers
             $temp = $value;
         }
 
-        if (empty($temp)) {
+        if ($temp === null) {
             return null;
         }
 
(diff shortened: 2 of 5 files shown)


M393:
Do you know where triggers are exported as html? When going to (Database or Table) -> Triggers export is done with AJAX.
I don't think this change causes a problem, but can't say for sure.
https://github.com/phpmyadmin/phpmyadmin/blob/47599015fd9ee3afd35ef8fdb3d1ccb52e38fd89/libraries/classes/Database/Triggers.php#L504-L516

js/src/database/triggers.js
```diff
@@ -98,7 +98,6 @@ const DatabaseTriggers = {
         } else {
             $.get($this.attr('href'), { 'ajax_request': true }, showExport);
         }
-        Functions.ajaxRemoveMessage($msg);
```
williamdes, line 101: Why this change?

williamdes:
Maybe special schema https://github.com/phpmyadmin/phpmyadmin/blob/d21d00cd0a1f32628914612bcf5eb56e7d7e0630/tests/unit/Display/ResultsTest.php#L203

M393:
@williamdes Where do I find this in phpMyAdmin UI?

williamdes:
> @williamdes Where do I find this in phpMyAdmin UI?

In the `information_schema` database, in the triggers tables :)
`?route=/sql&db=information_schema&table=TRIGGERS&pos=0`

williamdes:
Also fixes https://github.com/phpmyadmin/phpmyadmin/issues/19578 ?

M393:
> Maybe special schema
> 
> https://github.com/phpmyadmin/phpmyadmin/blob/d21d00cd0a1f32628914612bcf5eb56e7d7e0630/tests/unit/Display/ResultsTest.php#L203

The ACTION_COLUMN is transmitted to the browser as plain text and formatted with js, doesn't go through that function.

Triggers::export() is private and only called from Triggers::main(). The only two controllers who call Triggers::main() are Database/TriggersController and Table/TriggersController, so possibly unused code.

Anyways, this is what the html looks like when you force ajax to off:
image
It's a heading and textarea with the trigger sql inside - now with DELIMITER. 👍

M393:
> Also fixes #19578 ?

No, it doesn't.

PR #19954, opened 2025-11-21
Refactor csv import

kamil-tekiela:
I wanted to make this piece of code easier to read, but I haven't achieved good results. If you don't think it's worth merging, you can pick the commits that are good enough. 

I added `forceExecute` which should improve the performance when loading the CSVs. Without it, it takes a lot of time trying to parse the generated SQL query.

### DIFF
--- a/src/Plugins/Import/ImportCsv.php
+++ b/src/Plugins/Import/ImportCsv.php
@@ -274,7 +274,7 @@ class ImportCsv extends AbstractImportCsv
 
             $ch = mb_substr($buffer, $i, 1);
             if ($csvTerminatedLen > 1 && $ch == $this->terminated[0]) {
-                $ch = $this->readCsvTerminatedString($buffer, $ch, $i, $csvTerminatedLen);
+                $ch .= $this->readCsvTerminatedString($buffer, $i, $csvTerminatedLen);
                 $i += $csvTerminatedLen - 1;
             }
 
@@ -304,7 +304,7 @@ class ImportCsv extends AbstractImportCsv
                         $i++;
                         $ch = mb_substr($buffer, $i, 1);
                         if ($csvTerminatedLen > 1 && $ch == $this->terminated[0]) {
-                            $ch = $this->readCsvTerminatedString($buffer, $ch, $i, $csvTerminatedLen);
+                            $ch .= $this->readCsvTerminatedString($buffer, $i, $csvTerminatedLen);
                             $i += $csvTerminatedLen - 1;
                         }
 
@@ -322,7 +322,7 @@ class ImportCsv extends AbstractImportCsv
                         $i++;
                         $ch = mb_substr($buffer, $i, 1);
                         if ($csvTerminatedLen > 1 && $ch == $this->terminated[0]) {
-                            $ch = $this->readCsvTerminatedString($buffer, $ch, $i, $csvTerminatedLen);
+                            $ch .= $this->readCsvTerminatedString($buffer, $i, $csvTerminatedLen);
                             $i += $csvTerminatedLen - 1;
                         }
                     } else {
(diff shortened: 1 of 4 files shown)


src/Plugins/Import/ImportCsv.php
```diff
@@ -791,18 +819,11 @@ private function getSqlTemplateAndRequiredFields(
...
      *
      * @param string $buffer           The original string buffer read from csv file
-     * @param string $ch               Partially read "column Separated with" string, also used to return after
-     *                                 reading length equal $csv_terminated_len
      * @param int    $i                Current read counter of buffer string
      * @param int    $csvTerminatedLen The length of "column separated with" String
      */
-    public function readCsvTerminatedString(string $buffer, string $ch, int $i, int $csvTerminatedLen): string
+    public function readCsvTerminatedString(string $buffer, int $i, int $csvTerminatedLen): string
     {
```
williamdes: is there some shortcuts we could do to avoid calling mb_substr ?
I am not sure what values go into this readCsvTerminatedString function

kamil-tekiela: Could you please explain a bit more? Why do you want to avoid calling that function?
The values passed into this function are the CSV contents (or part of it), the current cursor position, and the length of the multibyte CSV separator. So if someone uses something weird as the separator instead of a `,` then the code reads the correct length of the string. `mb_substr` seems perfect for that task.

williamdes (approved):

src/Plugins/Import/ImportCsv.php
williamdes: I had in mind to short cut the calculation for values that are common, that gains performance for large data sets
for example the empty string value

kamil-tekiela: I would not pay much attention to the performance of this function. Removing the loop was a pretty obvious step but it's not worth optimizing it more. In 99.99% of cases, the CSV file will be terminated with a comma or a single-byte character so this function will not be used. 

There are some nice performance improvements that could be made to this whole class but it would require redesigning the whole functionality. I've thought about doing it but decided that it wasn't designed with performance in mind. Uploading 10k lines takes about 2 seconds and half the time is spent on displaying the results. The `mb_substr` calls are actually a small fraction.

williamdes: Okay, thanks for explaining

williamdes:
Did you encounter https://github.com/phpmyadmin/phpmyadmin/issues/19936

Or does the force solve it?
I can test with large data if you confirm

kamil-tekiela:
No, I was not able to reproduce it yet.

williamdes:
> No, I was not able to reproduce it yet.

I am running dfcf3dd949, and still can confirm it

kamil-tekiela:
I replaced the mb_* functions with non-mb functions, as I cannot see a reason why the code would need to use mb_* functions here. In my tests, without xDebug there is a significant performance improvement. With xDebug, there is a small performance degradation. 

@williamdes Could you test with this commit and see if your big CSV file import improves?

kamil-tekiela:
I've added another commit which brings the xDebug times down. Without xDebug there's no noticeable difference. `substr` with length 1 and access by index are pretty much the same operation. The problem is that for each function call xDebug has to make another entry in the cgrind file so it artificially inflates the execution time.

williamdes:
>so it artificially inflates the execution time.

True, but function calls have quite some cost. I noticed this while importing 80G of data for a client.

williamdes:
>@williamdes Could you test with this commit and see if your big CSV file import improves?

Yes, pending

kamil-tekiela:
I tried to do some more testing with UTF-8 data and ASCII structure, and at first I got confused because I got different results with and without this PR. But then I realized that I got garbled up data without this PR and I get correct results with it. So somehow these changes managed to fix some hidden data bug. 

I've tested it by uploading a randomly generated CSV file with UTF-8 text and standard separators. Then I exported the file in CSV and compared them to each other.

MauricioFauth (approved):
Great! Thank you!

williamdes:
I tried once more to import a file but it fails by showing the login page.
I would guess that is because the requests are missing the server id param 🤔

kamil-tekiela:
> I tried once more to import a file but it fails by showing the login page. I would guess that is because the requests are missing the server id param 🤔

Can you open an issue?

williamdes:
Done  #20059

PR #19955, opened 2025-11-21
Quoting is only needed when separators appear in text

kamil-tekiela:
I propose to only quote the data when it is strictly necessary. As per https://www.ietf.org/rfc/rfc4180.txt it is only necessary when line breaks, double quotes, and commas exist in the text. 

PMA supports the PHP proprietary escaping mechanism, but according to the PHP manual, the escape character shouldn't be escaped, which is why it is not included in this PR. 

The reason for this change is that it creates a significantly smaller CSV file. While the disk size isn't something to be concerned with, the additional characters make processing the CSV slower. In tables with many int columns, this could add up to a lot of unnecessary processing. It also saves network bandwidth. 

Checking if a field needs to be quoted adds overhead, but it balances out because the string doesn't need to be modified, so no performance impact on exporting.

### DIFF
--- a/src/Plugins/Export/ExportCsv.php
+++ b/src/Plugins/Export/ExportCsv.php
@@ -297,27 +297,27 @@ class ExportCsv extends ExportPlugin
         );
         $this->terminated = $this->setStringValue(
             $request->getParsedBodyParam('csv_terminated'),
-            $exportConfig['csv_terminated'] ?? null,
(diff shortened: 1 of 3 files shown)


williamdes (approved):
Could we have a test for this change ?

williamdes (approved):
Thankss

MauricioFauth (approved):

PR #19957, opened 2025-11-23
Refactor export plugins

kamil-tekiela:
Sorry for the big PR. 

This PR aims to split up a long class and refactor the remaining code to increase clarity by removing global state.

The new class is responsible for buffering the output, saving it to a file, or printing it to the browser. The plugins weren't using any other part of the Export class, so the new class replaces Export in plugins. This way, we limit what the plugins have access to. 

Some methods in ExportPlugin were abstract, but the child classes had no use for them. Some were even abusing them for setting their own options, e.g. `exportHeader`. 

A lot of unnecessary and dead code was removed. Some of it thanks to the removal of the global state. Some tests were removed too, as they either didn't test anything useful or were just duplicates of other tests. 

Performance should remain identical.


williamdes (approved):

williamdes:
@kamil-tekiela about 928ad06

it was to fix https://sourceforge.net/p/phpmyadmin/bugs/306/ in 2002 !

Initial commit: https://github.com/phpmyadmin/phpmyadmin/commit/fd60f5ba1e851d4de67fd65fe5f3122f11b7e4d4

- https://github.com/phpmyadmin/phpmyadmin/issues/6675

MauricioFauth (approved):
Thank you very much!

PR #19968, opened 2025-11-27
Fix performance of table/create

### LINKED ISSUE #19967: [Feat]: Request collations only when required (click on select)
### phpMyAdmin version

6.0.0-dev

### Describe the feature

Each time a table is created, or new columns are added, the `Collation` select is populated with ~196 kB of data for each column that will be added. For a table with 10 columns, this results in ~1960 kBs of used bandwitdh and also a lot of the browser's memory is also used to parse the `HTML` from `JSON`.

`Collation` is not required by default when you create a table or add new columns, since it will use the database/server collation. New columns can be added if this select is empty.

In case the user wants to change the `Collation`, a HTTP request should be made to fetch the data on the fly.

Here is a video that shows how much time is spent to create a table with 20 columns (7 seconds for the browser to parse the `HTML` from `JSON`):

https://github.com/user-attachments/assets/d83cec96-ebb7-409a-8ab1-541f23e32cc4

### Describe the solution

When clicking on the `Collation` select it should request the collations and populate the select. Also, the value should be stored in memory, to allow usage of the options on other select, if needed.

kamil-tekiela:
Fixes #19967 

@liviuconcioiu Does this resolve your problem?

### DIFF
--- a/resources/templates/columns_definitions/column_attributes.twig
+++ b/resources/templates/columns_definitions/column_attributes.twig
@@ -61,19 +61,8 @@
 </td>
 <td class="text-center">
   {# column collation #}
-  <select lang="en" dir="ltr" name="field_collation[{{ column_number }}]" id="field_{{ column_number }}_{{ ci - ci_offset }}" class="form-select">
-    <option value=""></option>
-    {% for charset in charsets %}
-      <optgroup label="{{ charset.name }}" title="{{ charset.description }}">
-        {% for collation in charset.collations %}
-          <option value="{{ collation.name }}" title="{{ collation.description }}"
-            {{- column_meta['Collation'] is defined and collation.name == column_meta['Collation'] ? ' selected' }}>
-            {{- collation.name -}}
-          </option>
-        {% endfor %}
-      </optgroup>
-    {% endfor %}
-  </select>
+  <select lang="en" dir="ltr" name="field_collation[{{ column_number }}]" id="field_{{ column_number }}_{{ ci - ci_offset }}" class="form-select collation-select"
+  {{- column_meta['Collation'] is defined ? ('data-selected="' ~ column_meta['Collation'] ~ '"')|raw : '' }}></select>
   {% set ci = ci + 1 %}
 </td>
 <td class="text-center">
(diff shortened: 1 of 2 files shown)


liviuconcioiu:
Is definitely much better! At least it reduces the size. If there's a way to only populate the select on click, and not all of them, it will also reduce the browsers's memory usage. I noticed Firefox is slower than Chrome when it comes to parse large files.

kamil-tekiela:
I don't know about Chrome. I tried with 200 columns before this change and it crashed. With it is much quicker and everything works. 
The fix I suggest should not only fix the file size but also improve performance. Inserting the template is a pretty cheap operation. Still on my machine a lot of time is consumed by twig preparing the table but at least it's no longer exponential.

kamil-tekiela:
> If there's a way to only populate the select on click, and not all of them, it will also reduce the browsers's memory usage.

It should be relatively easy to do but I don't think it's necessary. The core problem was the huge HTML and not the Dom.

liviuconcioiu:
> It should be relatively easy to do but I don't think it's necessary. The core problem was the huge HTML and not the Dom.

It is also the DOM. For 40 columns it takes ~12 seconds to process it. This is how it looks with this PR:

https://github.com/user-attachments/assets/d6048c0e-f76e-4c4f-ab65-db60864e2afa

liviuconcioiu:
This is how is going to look if the selects are populated only when used (in the right side you can see that DOM doesn't show up, like in the first video):

https://github.com/user-attachments/assets/d2dd4fc3-442a-48e1-9204-5df124357bee

kamil-tekiela:
> > It should be relatively easy to do but I don't think it's necessary. The core problem was the huge HTML and not the Dom.
> 
> It is also the DOM. For 40 columns it takes ~12 seconds to process it. This is how it looks with this PR:
> 
>  28.11.2025_02.01.07_REC.mp4

It seems Chrome is better in this situation as it only takes about 2s with 40 columns. I actually find it strange that rendering selects dynamically takes so much time.

kamil-tekiela:
@liviuconcioiu I've added a commit that only populates one by one on focus event. Does this work the same and improve the performance?

liviuconcioiu:
It works like a charm.

williamdes (approved):
Thank you, there is enougth reviews
I am [contrib-merging](https://github.com/wdes/contrib-merge) this

PR #19972, opened 2025-12-02
Add Pure-PHP ULID Generator (Spec-Compliant & Monotonic)

zaenalcoders:
# Add Pure-PHP ULID Generator (Spec-Compliant & Monotonic)

## Summary

This pull request introduces a fully self-contained **ULID (Universally Unique Lexicographically Sortable Identifier)** generator for phpMyAdmin.  
The implementation is written entirely in **pure PHP**, adds **no external dependencies**, and follows the official ULID specification, including timestamp structure, randomness generation, and Crockford Base32 encoding.

----------

## Motivation

phpMyAdmin currently supports UUID generation but does not provide a built-in ULID generator.  
ULID offers several advantages:
-   Lexicographically sortable identifiers
-   Millisecond timestamp precision
-   80 bits of secure randomness
-   Human-friendly, URL-safe Base32 output
-   Widely used in modern distributed systems
    

Adding a built-in ULID generator improves developer experience without increasing dependency footprint.

----------

### Specification-Compliant

This generator follows the official ULID specification:  
[https://github.com/ulid/spec](https://github.com/ulid/spec)

### Monotonic Behavior

When multiple ULIDs are generated within the same millisecond, the randomness portion is incremented to maintain:
-   Ordering consistency
-   Collision avoidance
-   Monotonic output (similar to Symfony/Uid)

### Dependency-Free

Although inspired by `symfony/uid` and `robinvdvleuten/ulid`, this implementation is entirely rewritten to keep phpMyAdmin dependency-light.

### Structure & Coding Standards
-   PEAR-style formatting
-   phpMyAdmin directory conventions
-   Documented using phpDoc
-   Focus on clarity, maintainability, and predictability

----------

### Class

`libraries/classes/Ulid.php` 

### Public API

`Ulid::generate(): string` 
Generates a 26-character ULID encoded using Crockford Base32.

----------

### Test File

`test/classes/UlidTest.php` 

### Test Coverage Includes:
-   Correct ULID length (26 chars)    
-   Valid Base32 character set    
-   Monotonic ordering within the same millisecond    
-   Timestamp correctness    
-   Collision avoidance across large batches    

All tests are pure PHPUnit and require no external libraries.

----------

## Impact
-   No breaking changes    
-   No new dependencies    
-   Minimal performance overhead    
-   Fully isolated; no side effects    

----------

## Conclusion

This PR adds a modern, reliable, and specification-compliant ULID generator to phpMyAdmin, without altering existing behavior or adding dependencies.  
It strengthens phpMyAdmin's utility for modern development workflows.

### DIFF
--- a/src/InsertEdit.php
+++ b/src/InsertEdit.php
@@ -1068,6 +1068,13 @@ class InsertEdit
             return $this->dbi->quoteString($uuid);
         }
 
+        if ($editField->function === 'ULID') {
+            /* generate ULID */
+            $ulid = (string) Ulid::generate();
+
+            return $this->dbi->quoteString($ulid);
+        }
+
         if (
             in_array($editField->function, $this->getGisFromTextFunctions(), true)
             || in_array($editField->function, $this->getGisFromWKBFunctions(), true)
--- a/src/InsertEdit.php
+++ b/src/InsertEdit.php
@@ -1069,8 +1069,7 @@ class InsertEdit
         }
 
         if ($editField->function === 'ULID') {
-            /* generate ULID */
-            $ulid = (string) Ulid::generate();
+            $ulid = Ulid::generate();
 
             return $this->dbi->quoteString($ulid);
         }
--- a/src/Types.php
+++ b/src/Types.php
@@ -395,6 +395,7 @@ class Types
                     'SOUNDEX',
                     'SPACE',
                     'TRIM',
+                    'ULID',
                     'UNCOMPRESS',
                     'UNHEX',
                     'UPPER',
--- a/src/Ulid.php
+++ b/src/Ulid.php
@@ -14,26 +14,62 @@ use Exception;
  */
 class Ulid
 {
-    private static $lastTimestamp = null;
-    private static $lastRandomness = null;
+    /**
+     * Stores the timestamp (in milliseconds) of the last generated ULID.
+     * null means no ULID has been generated yet in this runtime.
+     */
+    private static ?int $lastTimestamp = null;
 
-    /** Crockford Base32 Alphabet */
+    /**
+     * Stores the last generated 10-byte randomness segment used for monotonic ULIDs.
+     *
+     * The value is an array of 10 integers (0–255). A null value means that no ULID
+     * has been generated yet in this runtime and the randomness has not been initialized.
+     *
+     * @var int[]|null
+     */
+    private static ?array $lastRandomness = null;
+
+    /**
+     * Crockford's Base32 alphabet used for ULID encoding.
+     *
+     * This alphabet intentionally excludes letters that can be visually ambiguous:
+     * I, L, O, and U. This improves readability and reduces transcription errors.
+     *
+     * Order of characters is fixed and must not be modified, as each character
+     * maps to a 5-bit value (0–31) per ULID specification.
+     *
+     * @see https://github.com/ulid/spec
+     * @see https://www.crockford.com/base32.html
+     */
     private const ALPHABET = '0123456789ABCDEFGHJKMNPQRSTVWXYZ';
 
     /**
-     * Generates a ULID string (monotonic).
+     * Generates a monotonic ULID string.
+     *
+     * A monotonic ULID guarantees lexicographical ordering when multiple ULIDs
+     * are generated within the same millisecond. If the timestamp does not
+     * change, the previous 10-byte randomness segment is incremented instead of
+     * generating new random bytes. This requires maintaining internal state for
+     * both timestamp and randomness.
      *
      * @return string
-     * @throws Exception
+     * @throws \Exception If random_bytes() fails.
      */
     public static function generate(): string
     {
         $timestamp = (int) floor(microtime(true) * 1000);
 
+        // If we are still within the same millisecond, increment the previous
+        // randomness segment to maintain lexicographical monotonicity.
         if (self::$lastTimestamp === $timestamp) {
             self::incrementRandomness();
         } else {
+            // New millisecond → generate fresh randomness and store state
             self::$lastTimestamp = $timestamp;
+
+            //Randomness is stored so it can be incremented on subsequent calls
+            //within the same millisecond, which is required for monotonic ULID.
             self::$lastRandomness = unpack('C*', random_bytes(10));
         }
 
@@ -44,7 +80,14 @@ class Ulid
     }
 
     /**
-     * Encodes a 48-bit timestamp as a 10-character Crockford Base32 string.
+     * Encodes a 48-bit millisecond timestamp into a 10-character
+     * Crockford Base32 string as defined by the ULID specification.
+     *
+     * The timestamp is divided into Base32 segments and padded on the left
+     * to ensure the output is always exactly 10 characters long.
+     *
+     * @param int $timestamp Milliseconds since Unix epoch.
+     * @return string 10-character timestamp component of the ULID.
      */
     private static function encodeTime(int $timestamp): string
     {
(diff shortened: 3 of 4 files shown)


ibennetch:
I think this is interesting. I wasn't familiar with ULID but seems like
something we can implement.

Nice work on the pull request, thanks especially for having tests.

I'm curious what the others think, but this looks good to me.

🤔 Maybe it should target 6.x, which we're working hard towards releasing
as the next major version.

zaenalcoders:
Hi @ibennetch, 

Thank you for the feedback and kind words! I’ll update this pull request to target the 6.x branch as suggested and ensure it aligns with the upcoming major release.

Appreciate your guidance and time reviewing this.

williamdes:
> 🤔 Maybe it should target 6.x, which we're working hard towards releasing
as the next major version.

I am a bit confused as it already targets master, and QA_6_0 did not start yet

src/Types.php
```diff
@@ -395,6 +395,7 @@ public function getFunctionsClass(TypeClass $class): array
                     'SOUNDEX',
                     'SPACE',
                     'TRIM',
+                    'ULID',
```
williamdes, line 398: As far as I know this is not a database server type, you can remove the line. UUID is one

williamdes: If I am correct MD5 is one of the exceptions, @MauricioFauth where should we put this one?

ibennetch:
> I am a bit confused as it already targets master, and QA_6_0 did not start yet

You're quite right - I somehow misread the email notification. I'm not sure what I thought I saw - I definitely misread it as being for the QA branch - but you are of course correct.

src/InsertEdit.php
```diff
@@ -1068,6 +1068,13 @@ private function formatAsSqlFunction(
             return $this->dbi->quoteString($uuid);
         }
 
+        if ($editField->function === 'ULID') {
+            /* generate ULID */
```
kamil-tekiela: Please don't add comments that convey no information.

src/InsertEdit.php
```diff
@@ -1068,6 +1068,13 @@ private function formatAsSqlFunction(
...
+            $ulid = (string) Ulid::generate();
```
kamil-tekiela:
```suggestion
            $ulid = Ulid::generate();
```

src/Ulid.php
```diff
@@ -0,0 +1,125 @@
...
+        return self::encodeInteger($timestamp, 10);
+    }
+
+    /**
+     * Turns random bytes into a 16-character randomness segment.
+     */
+    private static function encodeRandom(array $bytes): string
+    {
+        $binary = '';
+        foreach ($bytes as $b) {
```
kamil-tekiela:
```suggestion
        foreach ($bytes as $byte) {
```
The word is short enough that you don't need to be cryptic with it.

src/Ulid.php
```diff
@@ -0,0 +1,125 @@
...
+    private static function encodeInteger(int $value, int $padding): string
+    {
+        $alphabet = self::ALPHABET;
+        $result = '';
+
+        if ($value === 0) {
+            $result = '0';
+        } else {
+            while ($value > 0) {
+                $result = $alphabet[$value % 32] . $result;
```
kamil-tekiela:
```suggestion
                $result = self::ALPHABET[$value % 32] . $result;
```
Unnecessary reassignments are confusing in the long run.

src/Ulid.php
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * Encodes an integer into Crockford Base32 with padding.
+     */
+    private static function encodeInteger(int $value, int $padding): string
+    {
+        $alphabet = self::ALPHABET;
+        $result = '';
+
+        if ($value === 0) {
+            $result = '0';
```
kamil-tekiela: Why is this line necessary?

src/Ulid.php
```diff
@@ -0,0 +1,125 @@
...
+        return str_pad($result, $padding, '0', STR_PAD_LEFT);
+    }
+
+    /**
+     * Encodes binary data into Crockford Base32.
+     */
+    private static function encodeBinary(string $binary, int $padding): string
+    {
+        $bits = '';
+        $alphabet = self::ALPHABET;
```
kamil-tekiela: Same here. It's a constant so you should not reassign it to a variable.

src/Ulid.php
```diff
@@ -0,0 +1,125 @@
...
+
+        for ($i = 0, $len = strlen($binary); $i < $len; $i++) {
+            $bits .= str_pad(decbin(ord($binary[$i])), 8, '0', STR_PAD_LEFT);
+        }
```
kamil-tekiela: I'm pretty sure it would be better if you'd `str_split` it and use it as an array.

src/Ulid.php
```diff
@@ -0,0 +1,125 @@
...
+
+        return substr(str_pad($output, $padding, '0'), 0, $padding);
+    }
+
+    /**
+     * Increments the randomness for monotonic ULID.
+     */
+    private static function incrementRandomness(): void
+    {
+        $bytes = &self::$lastRandomness;
```
kamil-tekiela: Why the reference?

src/Ulid.php
```diff
@@ -0,0 +1,125 @@
...
+        for ($i = 10; $i >= 1; $i--) {
```
kamil-tekiela: Could you please explain the magic numbers? Can you either remove them or turn them into constants?

src/Ulid.php
```diff
@@ -0,0 +1,125 @@
...
+            }
+            $output .= $alphabet[bindec($chunk)];
+        }
+
+        return substr(str_pad($output, $padding, '0'), 0, $padding);
+    }
+
+    /**
+     * Increments the randomness for monotonic ULID.
+     */
```
kamil-tekiela, line 183: I don't think any of these PHPdocs explain anything. They should probably all be removed. Although I really do not understand the purpose of this method so leaving a descriptive comment maybe would help.

src/Ulid.php
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * ULID generator class (monotonic).
+ *
+ * Based on ULID specification: https://github.com/ulid/spec
+ * Inspired by Symfony/Uid and robinvdvleuten/ulid implementations.
+ */
+class Ulid
+{
+    private static $lastTimestamp = null;
```
kamil-tekiela: I can't find any use of the `null` value. Does it have to be initiliazed to null? Can you please add proper types to make it easier what the allowed values are?

src/Ulid.php
```diff
@@ -0,0 +1,125 @@
...
+    public static function generate(): string
+    {
+        $timestamp = (int) floor(microtime(true) * 1000);
+
+        if (self::$lastTimestamp === $timestamp) {
+            self::incrementRandomness();
+        } else {
+            self::$lastTimestamp = $timestamp;
+            self::$lastRandomness = unpack('C*', random_bytes(10));
+        }
```
kamil-tekiela, line 74: What exactly is the logic here? Why is `self::$lastRandomness` required? I see no way to reproduce the same result. It's always random, so why are you maintaining any state at all?

src/Ulid.php
```diff
@@ -0,0 +1,125 @@
...
+    private static $lastRandomness = null;
+
+    /** Crockford Base32 Alphabet */
+    private const ALPHABET = '0123456789ABCDEFGHJKMNPQRSTVWXYZ';
+
+    /**
+     * Generates a ULID string (monotonic).
+     *
+     * @return string
+     * @throws Exception
```
kamil-tekiela: What throws the Exception?

tests/unit/UlidTest.php
```diff
@@ -0,0 +1,91 @@
...
+
+namespace PhpMyAdmin\Tests\Classes;
+
+use PhpMyAdmin\Ulid;
+use PHPUnit\Framework\TestCase;
+
+class UlidTest extends TestCase
+{
+    // ULID uses Crockford Base32 alphabet.
+    private const ALPHABET = '0123456789ABCDEFGHJKMNPQRSTVWXYZ';
```
kamil-tekiela, line 25: Is this really needed in a test?

tests/unit/UlidTest.php
```diff
@@ -0,0 +1,91 @@
...
+
+    // ULID must always be 26 characters.
+    public function testUlidLength(): void
```
kamil-tekiela:
```suggestion
    public function testUlidLengthIsAlways26(): void
```

tests/unit/UlidTest.php
```diff
@@ -0,0 +1,91 @@
...
+    {
+        $ulid = Ulid::generate();
+        $this->assertSame(
```
kamil-tekiela:
```suggestion
        self::assertSame(
```

tests/unit/UlidTest.php
```diff
@@ -0,0 +1,91 @@
...
+        $ulid = Ulid::generate();
+        $chars = str_split($ulid);
+
+        foreach ($chars as $char) {
+            $this->assertTrue(
+                strpos(self::ALPHABET, $char) !== false,
+                "Invalid character in ULID: {$char}"
+            );
+        }
+    }
```
kamil-tekiela, line 63: [`strcspn`](https://www.php.net/manual/en/function.strcspn.php)

tests/unit/UlidTest.php
```diff
@@ -0,0 +1,91 @@
...
+
+    // ULIDs created one after another should be in order thanks to the monotonic logic.
```
kamil-tekiela: Again, I don't think any of these comments add any value. They just repeat what the test already does.

tests/unit/UlidTest.php
```diff
@@ -0,0 +1,91 @@
...
+    public function testUlidMonotonicity(): void
+    {
+        $u1 = Ulid::generate();
+        $u2 = Ulid::generate();
+
+        $this->assertLessThan(
```
kamil-tekiela: Doesn't this assertion accept strings? Is there a reason to use `strcmp` here?

tests/unit/UlidTest.php
```diff
@@ -0,0 +1,91 @@
...
+        );
+    }
+
+    // The first 10 characters are the timestamp. If you generate a ULID a little later, that part should go up.
+    public function testTimestampPortionIncreases(): void
+    {
+        $u1 = Ulid::generate();
+
+        // Wait for about 2ms to make sure the next ULID lands in a new millisecond.
+        usleep(2000);
```
kamil-tekiela: Or you could take the timestamp as an optional parameter to the `generate` method.

tests/unit/UlidTest.php
```diff
@@ -0,0 +1,91 @@
...
+        $u2 = Ulid::generate();
+
+        $this->assertLessThan(
+            0,
+            strcmp($u1, $u2),
+            'Second ULID must be lexicographically greater than the first'
+        );
+    }
+
+    // The first 10 characters are the timestamp. If you generate a ULID a little later, that part should go up.
```
kamil-tekiela: A comment like this belongs in the class where the functionality is implemented, not in the test case.

tests/unit/UlidTest.php
```diff
@@ -0,0 +1,91 @@
...
+            );
+            $generated[$u] = true;
+        }
+
+        $this->assertCount(
+            1000,
+            $generated,
+            'All ULIDs generated in batch should be unique'
+        );
+    }
```
kamil-tekiela: A unit test should be reproducible. If it fails, we need to be able to apply a fix and retest with the same parameters. I doubt this test fulfils this requirement. Testing random behaviour may be useful during development, but it isn't helpful in a test suite.

src/Ulid.php
williamdes: @zaenalcoders ?

PR #19979, opened 2025-12-06
Change `Drop` to `Delete` wording

contributor860:
I have changed "Drop" to "Delete" under the Action section of tables in a database, because "Delete" is more user-friendly. Currently, I have only changed Drop in the table section of a database, but if this change is merged, then I will be making changes in other files also.

### DIFF
--- a/resources/templates/database/structure/structure_table_row.twig
+++ b/resources/templates/database/structure/structure_table_row.twig
@@ -84,7 +84,7 @@
                   'sql_query': drop_query,
                   'message_to_show': drop_message
                 }), '') }}">
-                {{ get_icon('b_drop', t('Drop')) }}
+                {{ get_icon('b_drop', t('Delete')) }}
             </a>
         </td>
     {% endif %}
--- a/resources/templates/database/structure/structure_table_row.twig
+++ b/resources/templates/database/structure/structure_table_row.twig
@@ -84,7 +84,7 @@
                   'sql_query': drop_query,
                   'message_to_show': drop_message
                 }), '') }}">
-                {{ get_icon('b_drop', t('Delete')) }}
+                {{ get_icon('b_drop', t('DROP')) }}
             </a>
         </td>
     {% endif %}


kamil-tekiela:
Why is that more user-friendly? Drop is a well-established phrase in the SQL community. The SQL command is `DROP` and not DELETE.

contributor860:
I know that Drop is being used for years in phpMyAdmin, and it is the standard word for deleting Tables and Fields, but still the Entries has Delete for deleting, so how about having the same term for both Tables and Fields also

liviuconcioiu:
> I know that Drop is being used for years in phpMyAdmin, and it is the standard word for deleting Tables and Fields, but still the Entries has Delete for deleting, so how about having the same term for both Tables and Fields also

Because `DROP` is used for tables and databases, and `DELETE` is used for rows.

contributor860:
But wouldn't it be better to have the same term for all of them

kamil-tekiela:
> But wouldn't it be better to have the same term for all of them

If the SQL language ever changes, then maybe yes, but for now, everyone is used to this nomenclature.

resources/templates/database/structure/structure_table_row.twig
```diff
@@ -84,7 +84,7 @@
                   'sql_query': drop_query,
                   'message_to_show': drop_message
                 }), '') }}">
-                {{ get_icon('b_drop', t('Drop')) }}
+                {{ get_icon('b_drop', t('Delete')) }}
```
williamdes:
```suggestion
                {{ get_icon('b_drop', t('DROP')) }}
```
This is maybe the only change we could accept, I am not sure it shows well in the UI

contributor860: William, for which file are you talking about, and are you talking about changing the term from "Delete" to "Drop" for rows

williamdes: I am suggesting to change back to what it was but to use the uppercase form of the SQL term

contributor860: > This is maybe the only change we could accept, I am not sure it shows well in the UI

Screenshot 2025-12-09 190138
I have tested the change on my local computer, and this is how it looks

contributor860: I think "DROP" will match more with the SQL command than "Drop".

williamdes: Thank you for the screenshot, i was thinking that the change was in the table settings page
It would make more sense to have in uppercase in the settings page rather than the table list

contributor860: Screenshot 2025-12-09 234706
Are you talking about the operations page of table

williamdes: Yes, Indeed we already use upper case for operations

Well, @kamil-tekiela we have Empty instead of Truncate as you can see on screenshot 1
Should it be changed to Truncate that would be a valid question

I think we reached a point where this PR should be closed, I am pretty sure it will confuse users

kamil-tekiela: I think in case of `TRUNCATE` it's fine. Because there is no SQL operation called `EMPTY`. Wheras with `DROP` and `DELETE` it would be confusing.

williamdes:
Thank you for making this PR, after all comments it is best to close it

resources/templates/database/structure/structure_table_row.twig
nassif2609: But how about having DROP in the table list

ibennetch: I prefer this the way it currently is. Changing from "Drop" to "DROP" isn't terrible wrong, but it really seems fine to match the surrounding style which would be "Drop". Sure, "DROP" is a SQL command, but we don't have "INSERT" here either. Visually I think this looks best, and syntactically I think using "Drop" instead of "Delete" makes the most sense. So I'd prefer to not change any of this.

contributor860: Okay, and thanks for your reply

ibennetch: And thank you @contributor860 for you desire to improve phpMyAdmin!

PR #19981, opened 2025-12-08
Fix performance of collation

kamil-tekiela:
In my testing, I observed 5-10% improvement in performance for `/database/structure`. I think this is worth making this change.

### DIFF
--- a/src/Charsets/Collation.php
+++ b/src/Charsets/Collation.php
@@ -22,6 +22,7 @@ final class Collation
      */
     private string $description;
 
+    /** @var array<string, string|null> */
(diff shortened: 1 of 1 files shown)


williamdes (approved):

williamdes:
For me this is okay, but maybe having a static variable is not what we want to add @MauricioFauth ?

MauricioFauth (approved):

MauricioFauth:
> For me this is okay, but maybe having a static variable is not what we want to add @MauricioFauth ?

You're right. It's better to avoid using static variables where possible.

PR #19983, opened 2025-12-09
Remove Gis Singletons

kamil-tekiela:
It took me 15 mins of confusion to realize that there is no need for these classes to be singletons.


williamdes (approved):

MauricioFauth (approved):
Thank you!

PR #19984, opened 2025-12-09
Remove all redundant issets

kamil-tekiela:
`isset` and `empty` are very nasty constructs. When it's obvious that a variable is defined, we should avoid using them.


MauricioFauth (approved):
Thanks!

PR #19985, opened 2025-12-09
Refactor relation

kamil-tekiela:
The only thing I was a bit scared of is the check whether the array was empty. I have decided that it was probably inaccurate before and a simple empty array check on both array would suffice. I have opted for leaving the outer condition as it is but with a `null` default instead. The outer condition is probably unnecessary. 

This change should make the code more readable.


williamdes (approved):

MauricioFauth (approved):
Thank you!

Issue #18183, opened 2023-03-01
TypeError: preg_replace(): Argument #3 ($subject) must be of type array|string, null given


liviuconcioiu:
### Describe the bug

I have a table that has a `JSON` type column created on MySQL. If the database is open on MariaDB, it will show the table `in use`.

Now, if I click on `Browse`, a `#4161 - Unknown data type: 'MYSQL_JSON'` error is given. But, if I click on Search, a few errors come up.

5.2.2-dev

```
TypeError: preg_replace(): Argument #3 ($subject) must be of type array|string, null given

Internal error in .\templates\table\search\index.twig#72 Twig\Error\RuntimeError: An exception has been thrown during the rendering of a template ("Undefined array key 0").```

Warning in .\libraries\classes\Controllers\Table\SearchController.php#346
Undefined array key 0

Warning in .\libraries\classes\Controllers\Table\SearchController.php#347
Undefined array key 0
```

### To Reproduce

1. Have a table with a `JSON` type column
2. Open the database on MariaDB
3. Click on Search for that table

### Expected behavior

A `#4161 - Unknown data type: 'MYSQL_JSON'` error should be given, like it happens when you click on Browser Insert.

### Screenshots

5.2.2-dev

https://user-images.githubusercontent.com/25424343/222019627-065288bf-777d-4ec5-bfdd-de42bd997ae9.mp4

https://user-images.githubusercontent.com/25424343/222019646-f3a83638-56a4-43fe-a40a-8f47ec03f141.mp4

### Server configuration

- phpMyAdmin version: 5.2.2-dev, 6.0.0-dev

### Additional context

5.2.2-dev

```
Warning in .\libraries\classes\Controllers\Table\SearchController.php#346
Undefined array key 0

Backtrace

CoreExtension.php#1635: PhpMyAdmin\Controllers\Table\SearchController->getColumnProperties(
integer 0,
integer 0,
)
89e0cdc4eec79bf3db7e00c4b27a82ac.php#205: twig_get_attribute(
,
,
,
string 'getColumnProperties',
array,
string 'method',
boolean false,
boolean false,
boolean false,
integer 72,
)
Template.php#394: __TwigTemplate_4f09a00a8ff807a41e95ef41004b827a->doDisplay(
array,
array,
)
Template.php#367: Twig\Template->displayWithErrorHandling(
array,
array,
)
Template.php#379: Twig\Template->display(array)
TemplateWrapper.php#40: Twig\Template->render(
[...]

PR #19986, opened 2025-12-09
Improve error message if table is not available

liviuconcioiu:
### Description

I think we should show why the table can't be accessible. Because, on 6.0.0-dev, it redirects to homepage without why it happened, and it shows only `in use`. This way, the user will know the error and fix it.

Related to #18183

Before:

before

After:

after

### DIFF
--- a/templates/database/structure/structure_table_row.twig
+++ b/templates/database/structure/structure_table_row.twig
@@ -234,7 +234,11 @@
 
         <td colspan="{{ action_colspan }}"
             class="text-center">
-            {% trans 'in use' %}
+            {% if current_table['TABLE_COMMENT'] is not empty %}
+                {{ current_table['TABLE_COMMENT'] }}
+            {% else %}
+                {% trans 'in use' %}
+            {% endif %}
         </td>
     {% endif %}
 </tr>
--- a/templates/database/structure/structure_table_row.twig
+++ b/templates/database/structure/structure_table_row.twig
@@ -234,8 +234,8 @@
 
         <td colspan="{{ action_colspan }}"
             class="text-center">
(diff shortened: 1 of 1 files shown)


kamil-tekiela:
What if a table actually has a comment?

liviuconcioiu:
If a table that is not accesible has a comment, it will still show the error message as comment - see table 7 (I added a comment to it in MySQL, but on MariaDB shows `Unknown data type: 'MYSQL_JSON'`).

after1

templates/database/structure/structure_table_row.twig
```diff
@@ -234,7 +234,11 @@
 
         <td colspan="{{ action_colspan }}"
             class="text-center">
-            {% trans 'in use' %}
+            {% if current_table['TABLE_COMMENT'] is not empty %}
```
williamdes: interestingly it differs from the above line `{% set comment = current_table['Comment'] %}`
Do you know why ?

liviuconcioiu: https://github.com/phpmyadmin/phpmyadmin/blob/74b060f53c52e8e17a5c02834341b3f69a76d396/libraries/classes/Query/Generator.php#L95

kamil-tekiela: All the values are duplicated in the code with both names. As of now, it doesn't matter which one is used.

liviuconcioiu:
Another argument to use the comment there, is because in this case I don't have `In_use` tables, but phpMyAdmin reported them as in use.

query

williamdes (approved):
Looks good, thank you!

PR #19997, opened 2025-12-20
Fix Export checkboxes

kamil-tekiela:
It is possible to set export options in the config file. But from what I believe, it is meant to only set the default value on the UI and shouldn't be used to override the value the user has selected. 

How should this work? For example, `csv_removeCRLF` is normally a checkbox in the UI, but if it is not checked on UI then the value is read from the config file. This leads to some confusion as can be seen in https://github.com/phpmyadmin/phpmyadmin/commit/4920aa21602331a4d4fca60eef10f34f233fabcb. If the value in the config is `true` then the UI field becomes unusable. Thus, the only sensible value to put in the config file is `false`. Is putting `true` in the config file meant to disable the UI field? If so, should there be a visual indication? It could surprise the users otherwise. 

I am not 100% sure, but I believe this worked in 5.2 the same as I propose in this PR @MauricioFauth. The config settings should only be used to determine the default value in the UI and then whatever is sent from the client is what Export should use. When it comes to checkboxes, as I explained above, doing it the way it currently works overrides the values the user selected. It's due to how HTML checkboxes work. For string values, there's less risk of a surprising behaviour, but silently falling back to the config value could lead to silent issues for users in case something goes wrong with submitting the form. This PR keeps the behaviour for string values, because it should be replaced with an error message to the user, and that's a lot more work.


williamdes (approved):
Thank you!
Now I understand that we fill the form with the config value
And use what the user sends back to us, no fallback needed

williamdes:
Hi @kamil-tekiela 
I tried merging this one and adjusting the files but there is a bit much to be done for the tests
Can you do a rebase ?

kamil-tekiela:
@williamdes  Is this ok to be merged or should we still wait?

williamdes:
> @williamdes Is this ok to be merged or should we still wait?

thanks for the ping, I will contrib-merge this right now

PR #19996, opened 2025-12-20
Move exportSql checks and refactor code

kamil-tekiela:
Fixes #19855

@liviuconcioiu Can you check if this fixes the issue?

This PR should fix the reported problem, just like the one from @efturtle would, but it introduces other refactorings that would make the other PR immediately obsolete, which is why I recommend just merging this PR. 

The underlying reason for the bug is NOT addressed in this PR. The problem became evident after the last refactoring. The export form sends all options, not just the ones visible on the screen. In the past, the code mixed up a lot of these checks and it all kind of worked, but without any clear logic. This PR separates more of these checks, but it does not fix the UI. In my testing, I observed that there are multiple issues which will need to be addressed one by one.

PR #20004, opened 2025-12-28
Add integration tests for export plugin structure methods

efturtle:
Tests verify exportStructure() methods are properly invoked through Export::exportTable() for plugins that support export structure

Theses tests ensure that when exporting table structure through the Export class, the specific plugin structure export methods are called with the correct parameters

### DIFF
--- a/tests/unit/Plugins/Export/ExportHtmlwordTest.php
+++ b/tests/unit/Plugins/Export/ExportHtmlwordTest.php
@@ -695,61 +695,38 @@ class ExportHtmlwordTest extends AbstractTestCase
      * Integration test: Export table structure through Export::exportTable()
      * Tests that exportStructure() method is called when exporting through Export::exportTable()
      */
-    public function testExportTableStructureThroughExportCore(): void
+    public function testExportTableCallsExportStructureMethod(): void
     {
-        // Mock the database interface
         $dbi = $this->getMockBuilder(DatabaseInterface::class)
             ->disableOriginalConstructor()
             ->getMock();
 
-        // Mock Table class to return isView = false
-        $table = $this->getMockBuilder(Table::class)
-            ->disableOriginalConstructor()
-            ->getMock();
-        $table->method('isView')->willReturn(false);
-
-        $dbi->method('getTable')->willReturn($table);
         DatabaseInterface::$instance = $dbi;
 
-        // Create a mock ExportHtmlword to verify getTableDef is called
         $relation = new Relation($dbi);
         $outputHandler = new OutputHandler();
-        $exportHtmlword = $this->getMockBuilder(ExportHtmlword::class)
-            ->setConstructorArgs([$relation, $outputHandler, new Transformations($dbi, $relation)])
-            ->onlyMethods(['getTableDef'])
-            ->getMock();
-
-        // getTableDef should be called during structure export
-        $exportHtmlword->expects(self::once())
-            ->method('getTableDef')
-            ->with('testdb', 'testtable', [])
-            ->willReturn('<table>test</table>');
+        $exportHtmlword = new ExportHtmlword($relation, $outputHandler, new Transformations($dbi, $relation));
 
         // Force structureOrData to be StructureAndData so structure export is attempted
         $attrStructureOrData = new ReflectionProperty(ExportHtmlword::class, 'structureOrData');
         $attrStructureOrData->setValue($exportHtmlword, StructureOrData::Structure);
+        
+        // Integration: Export::exportTable() should call exportStructure() on the plugin
+        $export = new Export($dbi, $outputHandler);
 
-        // Now call exportTable through the Export class
         ob_start();
-        try {
-            $exportcore = new Export($dbi, $outputHandler);
-            $exportcore->exportTable(
-                'testdb',
-                'testtable',
-                $exportHtmlword,
-                null,
-                '0',
-                '0',
-                '',
-                [],
-            );
-            $output = ob_get_clean();
-        } catch (\Throwable $e) {
-            ob_end_clean();
-            throw $e;
-        }
+        $export->exportTable(
+            'test_db',
+            'test_table',
+            $exportHtmlword,
+            null,
+            '',
+            '',
+            '',
+            []
+        );
+        $output = ob_get_clean();
 
-        // Verify that structure output was generated
-        self::assertIsString($output);
+        self::assertStringContainsString('<h2>Table structure for table test_table</h2>', $output);
     }
 }
--- a/tests/unit/Plugins/Export/ExportHtmlwordTest.php
+++ b/tests/unit/Plugins/Export/ExportHtmlwordTest.php
@@ -10,11 +10,9 @@ use PhpMyAdmin\ConfigStorage\Relation;
 use PhpMyAdmin\ConfigStorage\RelationParameters;
 use PhpMyAdmin\Current;
 use PhpMyAdmin\Dbal\DatabaseInterface;
-use PhpMyAdmin\Export\OutputHandler;
 use PhpMyAdmin\Export\Export;
-use PhpMyAdmin\Export\StructureOrData;
+use PhpMyAdmin\Export\OutputHandler;
 use PhpMyAdmin\Http\Factory\ServerRequestFactory;
-use PhpMyAdmin\Table\Table;
 use PhpMyAdmin\Identifiers\TableName;
 use PhpMyAdmin\Identifiers\TriggerName;
 use PhpMyAdmin\Plugins\Export\ExportHtmlword;
@@ -691,42 +689,27 @@ class ExportHtmlwordTest extends AbstractTestCase
         );
     }
 
-    /**
-     * Integration test: Export table structure through Export::exportTable()
-     * Tests that exportStructure() method is called when exporting through Export::exportTable()
-     */
     public function testExportTableCallsExportStructureMethod(): void
     {
         $dbi = $this->getMockBuilder(DatabaseInterface::class)
             ->disableOriginalConstructor()
             ->getMock();
-
         DatabaseInterface::$instance = $dbi;
-
-        $relation = new Relation($dbi);
-        $outputHandler = new OutputHandler();
-        $exportHtmlword = new ExportHtmlword($relation, $outputHandler, new Transformations($dbi, $relation));
-
-        // Force structureOrData to be StructureAndData so structure export is attempted
-        $attrStructureOrData = new ReflectionProperty(ExportHtmlword::class, 'structureOrData');
-        $attrStructureOrData->setValue($exportHtmlword, StructureOrData::Structure);
-        
-        // Integration: Export::exportTable() should call exportStructure() on the plugin
-        $export = new Export($dbi, $outputHandler);
-
+        $request = ServerRequestFactory::create()->createServerRequest('POST', 'https://example.com/');
+        $this->object->setExportOptions($request, ['htmlword_structure_or_data' => 'structure']);
+        $export = new Export($dbi, new OutputHandler());
         ob_start();
         $export->exportTable(
             'test_db',
             'test_table',
-            $exportHtmlword,
+            $this->object,
             null,
             '',
             '',
             '',
-            []
+            [],
         );
         $output = ob_get_clean();
-
         self::assertStringContainsString('<h2>Table structure for table test_table</h2>', $output);
     }
 }
--- a/tests/unit/Plugins/Export/ExportHtmlwordTest.php
+++ b/tests/unit/Plugins/Export/ExportHtmlwordTest.php
@@ -11,7 +11,10 @@ use PhpMyAdmin\ConfigStorage\RelationParameters;
 use PhpMyAdmin\Current;
 use PhpMyAdmin\Dbal\DatabaseInterface;
 use PhpMyAdmin\Export\OutputHandler;
+use PhpMyAdmin\Export\Export;
+use PhpMyAdmin\Export\StructureOrData;
 use PhpMyAdmin\Http\Factory\ServerRequestFactory;
+use PhpMyAdmin\Table\Table;
 use PhpMyAdmin\Identifiers\TableName;
 use PhpMyAdmin\Identifiers\TriggerName;
 use PhpMyAdmin\Plugins\Export\ExportHtmlword;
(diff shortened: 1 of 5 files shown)


efturtle:
@kamil-tekiela @williamdes followup on https://github.com/phpmyadmin/phpmyadmin/pull/19996 these are the tests to amplify the test coverage on structure exports.

tests/unit/Plugins/Export/ExportHtmlwordTest.php
```diff
@@ -687,4 +690,66 @@ public function testFormatOneColumnDefinition(): void
...
         );
     }
+
+    /**
+     * Integration test: Export table structure through Export::exportTable()
+     * Tests that exportStructure() method is called when exporting through Export::exportTable()
+     */
+    public function testExportTableStructureThroughExportCore(): void
+    {
+        // Mock the database interface
```
kamil-tekiela: Comments are these little evil gremlins in code that often lie to you. Avoid comments in code as much as possible. Not just in this project, but in general. Comments exist only to explain code, so it implies one of two things: either the code isn't well-written and needs to be refactored (better names, new methods, etc.) or the code is so complex in its simplest form that it needs additional information to remain maintainable. 

Most of the comments I see here don't add any additional information. "Mock the database interface" is exactly what the next line does so the comment doesn't explain anything.

tests/unit/Plugins/Export/ExportHtmlwordTest.php
```diff
@@ -687,4 +690,66 @@ public function testFormatOneColumnDefinition(): void
...
+            ->getMock();
+        $table->method('isView')->willReturn(false);
+
+        $dbi->method('getTable')->willReturn($table);
+        DatabaseInterface::$instance = $dbi;
+
+        // Create a mock ExportHtmlword to verify getTableDef is called
+        $relation = new Relation($dbi);
+        $outputHandler = new OutputHandler();
+        $exportHtmlword = $this->getMockBuilder(ExportHtmlword::class)
```
kamil-tekiela: `ExportHtmlword` is the code under test (see line 43) and it **should not** be mocked. So in this case, it makes perfect sense to put a code comment explaining why it's necessary for this test to do something unreasonable. The comment here should explain why `getTableDef` method is mocked. 

But in this case, I think it would be best to replace the mock with:
```
$exportHtmlword = new ExportHtmlword($relation, $outputHandler, new Transformations($dbi, $relation));
```
and then instead of `self::assertIsString($output);` you can test the full output that would be produced.

tests/unit/Plugins/Export/ExportHtmlwordTest.php
```diff
@@ -687,4 +690,66 @@ public function testFormatOneColumnDefinition(): void
...
+
+        // getTableDef should be called during structure export
+        $exportHtmlword->expects(self::once())
+            ->method('getTableDef')
+            ->with('testdb', 'testtable', [])
+            ->willReturn('<table>test</table>');
+
+        // Force structureOrData to be StructureAndData so structure export is attempted
+        $attrStructureOrData = new ReflectionProperty(ExportHtmlword::class, 'structureOrData');
+        $attrStructureOrData->setValue($exportHtmlword, StructureOrData::Structure);
```
kamil-tekiela: This should be set through `setExportOptions`. See lines 549-552.

tests/unit/Plugins/Export/ExportHtmlwordTest.php
```diff
@@ -687,4 +690,66 @@ public function testFormatOneColumnDefinition(): void
...
+                [],
+            );
+            $output = ob_get_clean();
+        } catch (\Throwable $e) {
+            ob_end_clean();
+            throw $e;
+        }
+
+        // Verify that structure output was generated
+        self::assertIsString($output);
```
kamil-tekiela: `$output` will always be a string in this situation so you need a better assertion. You should verify that the method produces the exact output you expect. 

Something like this:
```
        self::assertSame(
            '<h1>Database testdb</h1><h2>Table structure for table testtable</h2><table>test</table>',
            $output,
        );
```

Or even just check that the string contains `Table structure for`

tests/unit/Plugins/Export/ExportPdfTest.php
```diff
@@ -222,4 +225,107 @@ public function testExportData(): void
...
+            $this->object,
+            null,
+            '0',
+            '0',
+            '',
+            []
+        );
+    }
+
+
```
kamil-tekiela: This test needs to be changed so that it tests the output just like the others do. Don't look for inspiration at the other tests because many tests in phpMyAdmin are subpar and need improvement.

tests/unit/Plugins/Export/ExportPdfTest.php
```diff
@@ -222,4 +225,107 @@ public function testExportData(): void
...
+            ->method('getTableDef')
+            ->with('testdb', 'testtable', false, true, false);
+
+        // Set up the PDF in our export plugin
+        $attrPdf = new ReflectionProperty(ExportPdf::class, 'pdf');
+        $attrPdf->setValue($this->object, $pdf);
+
+        // Force structureOrData to be StructureAndData so structure export is attempted
+        $attrStructureOrData = new ReflectionProperty(ExportPdf::class, 'structureOrData');
+        $attrStructureOrData->setValue($this->object, StructureOrData::StructureAndData);
```
kamil-tekiela: Why `StructureAndData` and not `Structure`?

tests/unit/Plugins/Export/ExportPdfTest.php
```diff
@@ -222,4 +225,107 @@ public function testExportData(): void
...
+        self::assertTrue(
+            $this->object->exportStructure(
+                'db',
+                'table',
+                'create_table',
+            ),
+        );
+    }
+
+
```
kamil-tekiela: Nit: methods should be separated by only one blank line. I don't know why CS is not picking it up automatically.

tests/unit/Plugins/Export/ExportPdfTest.php
```diff
@@ -222,4 +225,107 @@ public function testExportData(): void
             ),
         );
     }
+
+    
+    public function testExportStructure(): void
```
kamil-tekiela: I'm having trouble understanding why this test needs to be added.

tests/unit/Plugins/Export/ExportPdfTest.php
```diff
@@ -222,4 +225,107 @@ public function testExportData(): void
...
+            ->method('setCurrentDb')
+            ->with('db');
+
+        $pdf->expects(self::once())
+            ->method('setCurrentTable')
+            ->with('table');
+
+        $pdf->expects(self::once())
+            ->method('setDbAlias')
+            ->with('db'); // getDbAlias returns 'db' when no alias is set
```
kamil-tekiela: Is it something unusual that `getDbAlias` returns `'db'`? What is this comment trying to say?

kamil-tekiela (changes requested):
Thank you so much for these tests. I really appreciate new tests as this makes future work easier. I checked, and they work, but could be improved.

Before this PR is ready to be merged, I have some changes I'd like you to make. Also, I have some general points about this:
- The name `$exportcore` is not a good variable name. For one, it doesn't respect PMA's naming standard, which requires variables to be in camel case. `core` is a term that doesn't appear in the rest of the code or in the class name. 
- Similarly, the test names could be better. `testExportTableStructureThroughExportCore` could be `testExportTableStructureIsCalledThroughExportWhenStructureIsRequested` or `testStructureIsExportedThroughExportClassWhenRequested`. The name should reflect the intent behind the test if it's more than a simple method test. If the tests were in the ExportTest class, it could just be `testStructureIsExportedWhenRequested`. And of course, when the test has a better name, the docblock above it is redundant as it should be.
- As you can see, all of these tests look very similar. But I don't see the positive test for `ExportSql`, which should verify that the settings are still respected. They were moved from Export to ExportSql and we should test that this still works. On the other hand, we only need a single test for all other export plugins that don't have that setting. Perhaps, it would be better to just add two new tests, one in ExportTest and one in ExportSqlTest, one testing that ExportSql still works and the other for an arbitrarily chosen other plugin testing to see if exportStructure is called at all.

tests/unit/Plugins/Export/ExportHtmlwordTest.php
efturtle: I will change it, thanks for the suggestion it is actually a lot better

Image

tests/unit/Plugins/Export/ExportPdfTest.php
efturtle: was the first test I created to see if export was working for PDF. It helped me figure out a more suitable one for export. Will remove

efturtle:
thanks for the suggestions and response on this one. I made a refactor on the tests with your requested changes and recommendations

tests/unit/Plugins/Export/ExportHtmlwordTest.php
```diff
@@ -687,4 +690,43 @@ public function testFormatOneColumnDefinition(): void
...
+            null,
+            '',
+            '',
+            '',
+            []
+        );
+        $output = ob_get_clean();
+
+        self::assertStringContainsString('<h2>Table structure for table test_table</h2>', $output);
+    }
```
kamil-tekiela:
```suggestion
    public function testExportTableCallsExportStructureMethod(): void
    {
        $dbi = $this->getMockBuilder(DatabaseInterface::class)
            ->disableOriginalConstructor()
            ->getMock();

        DatabaseInterface::$instance = $dbi;

        $request = ServerRequestFactory::create()->createServerRequest('POST', 'https://example.com/');
        $this->object->setExportOptions($request, ['htmlword_structure_or_data' => 'structure']);

        $export = new Export($dbi, new OutputHandler());

        ob_start();
        $export->exportTable(
            'test_db',
            'test_table',
            $this->object,
            null,
            '',
            '',
            '',
            [],
        );
        $output = ob_get_clean();

        self::assertStringContainsString('<h2>Table structure for table test_table</h2>', $output);
    }
```

tests/unit/Plugins/Export/ExportHtmlwordTest.php
```diff
@@ -11,7 +11,10 @@
 use PhpMyAdmin\Current;
 use PhpMyAdmin\Dbal\DatabaseInterface;
 use PhpMyAdmin\Export\OutputHandler;
+use PhpMyAdmin\Export\Export;
+use PhpMyAdmin\Export\StructureOrData;
 use PhpMyAdmin\Http\Factory\ServerRequestFactory;
+use PhpMyAdmin\Table\Table;
 use PhpMyAdmin\Identifiers\TableName;
```
kamil-tekiela:
```suggestion
use PhpMyAdmin\Export\Export;
use PhpMyAdmin\Export\OutputHandler;
use PhpMyAdmin\Http\Factory\ServerRequestFactory;
use PhpMyAdmin\Identifiers\TableName;
```

tests/unit/Plugins/Export/ExportLatexTest.php
```diff
@@ -794,4 +797,48 @@ public static function providerForGetTranslatedText(): iterable
...
+            $output = ob_get_clean();
+        } catch (\Throwable $e) {
+            ob_end_clean();
+            throw $e;
+        }
+
+        self::assertStringContainsString("% Database: 'testdb'", $output);
+        self::assertStringContainsString("%\n", $output);
+        self::assertStringContainsString("% Structure: testtable", $output);
+    }
```
kamil-tekiela:
```suggestion
    public function testExportTableCallsExportStructureMethod(): void
    {
        $dbi = $this->getMockBuilder(DatabaseInterface::class)
            ->disableOriginalConstructor()
            ->getMock();

        DatabaseInterface::$instance = $dbi;

        $request = ServerRequestFactory::create()->createServerRequest('POST', 'https://example.com/');
        $this->object->setExportOptions($request, ['latex_structure_or_data' => 'structure']);

        ob_start();
        $export = new Export($dbi, new OutputHandler());
        $export->exportTable(
            'testdb',
            'testtable',
            $this->object,
            null,
            '0',
            '0',
            '',
            [],
        );
        $output = ob_get_clean();

        self::assertStringContainsString("% Database: 'testdb'", $output);
        self::assertStringContainsString("%\n", $output);
        self::assertStringContainsString('% Structure: testtable', $output);
    }
```

tests/unit/Plugins/Export/ExportMediawikiTest.php
```diff
@@ -319,4 +322,47 @@ public function testExportData(): void
...
+            '0',
+            '',
+            [],
+        );
+        $result = ob_get_clean();
+
+        // Verify that structure output was generated
+        self::assertStringContainsString('Table structure for', $result);
+        self::assertStringContainsString('testtable', $result);
+    }
```
kamil-tekiela:
```suggestion
    public function testExportTableCallsExportStructureMethod(): void
    {
        $dbi = $this->getMockBuilder(DatabaseInterface::class)
            ->disableOriginalConstructor()
            ->getMock();

        DatabaseInterface::$instance = $dbi;

        $request = ServerRequestFactory::create()->createServerRequest('POST', 'https://example.com/');
        $this->object->setExportOptions($request, ['mediawiki_structure_or_data' => 'structure']);

        ob_start();
        $export = new Export($dbi, new OutputHandler());
        $export->exportTable(
            'testdb',
            'testtable',
            $this->object,
            null,
            '0',
            '0',
            '',
            [],
        );
        $result = ob_get_clean();

        self::assertIsString($result);
        self::assertStringContainsString('Table structure for', $result);
        self::assertStringContainsString('testtable', $result);
    }
```

tests/unit/Plugins/Export/ExportOdtTest.php
```diff
@@ -892,4 +895,43 @@ public function testFormatOneColumnDefinition(): void
...
+            null,
+            '0',
+            '0',
+            '',
+            [],
+        );
+
+        // Verify that structure output was generated in the buffer
+        self::assertStringContainsString('Table structure for table testtable', $exportOdt->buffer);
+    }
```
kamil-tekiela:
```suggestion
    public function testExportTableCallsExportStructureMethod(): void
    {
        $dbi = $this->getMockBuilder(DatabaseInterface::class)
            ->disableOriginalConstructor()
            ->getMock();

        DatabaseInterface::$instance = $dbi;

        $request = ServerRequestFactory::create()->createServerRequest('POST', 'https://example.com/');
        $this->object->setExportOptions($request, ['odt_structure_or_data' => 'structure']);

        $export = new Export($dbi, new OutputHandler());
        $export->exportTable(
            'testdb',
            'testtable',
            $this->object,
            null,
            '0',
            '0',
            '',
            [],
        );

        self::assertStringContainsString('Table structure for table testtable', $this->object->buffer);
    }
```

tests/unit/Plugins/Export/ExportPdfTest.php
```diff
@@ -222,4 +225,55 @@ public function testExportData(): void
...
+            ob_end_clean();
+            throw $e;
+        }
+
+        self::assertIsString($output);
+        self::assertNotEmpty($output);
+
+        self::assertStringStartsWith('%PDF-', $output);
+        self::assertGreaterThan(5000, strlen($output));
+    }
```
kamil-tekiela: You can remove this whole test. All of them test basically the same thing and we cannot easily assert in Pdf that the output contains what we need.

kamil-tekiela (changes requested):
And then please run:
```
composer install
vendor/bin/phpcbf
vendor/bin/psalm --set-baseline=psalm-baseline.xml
vendor/bin/phpstan --generate-baseline
```

tests/unit/Plugins/Export/ExportPdfTest.php
efturtle: yeah, since it is outputting binary, it is not easy to check with phpunit
best I could check is for line 276 && 277

williamdes (approved):

williamdes:
> vendor/bin/psalm --set-baseline=psalm-baseline.xml
vendor/bin/phpstan --generate-baseline

`composer run update:baselines`

PR #20005, opened 2025-12-28
Convert settings to config

kamil-tekiela:
This PR changes access to Config variables from the `$settings` property to `$config`. All unnecessary code was removed.


MauricioFauth (approved):
Thank you!

PR #20006, opened 2025-12-31
Fix tests after merge

kamil-tekiela:
I'm not sure if it was a mistake during the merge or if it's due to some refactoring. It's what was reported automatically by Psalm, and I didn't check the rest.


williamdes (approved):
Thanks !!
Happy new year 🎉

PR #20013, opened 2026-01-06
Replace custom service names with the class names

MauricioFauth:


app/services.php
```diff
@@ -68,291 +68,366 @@
 
 return [
     'services' => [
-        'advisor' => [
+        Advisor::class => [
             'class' => Advisor::class,
-            'arguments' => ['$dbi' => '@dbi', '$expression' => '@expression_language'],
+            'arguments' => ['@' . DatabaseInterface::class, '@' . ExpressionLanguage::class],
```
kamil-tekiela: Can we get rid of the `@` sign?

kamil-tekiela (approved):
And with this change it makes the loader leaner and faster:
```
return static function (ContainerConfigurator $configurator): void {
    $services = $configurator->services();
    /** @param array<'services', array<string, array{class: string, arguments?: array<string>, factory?: callable}>> $servicesFile */
    $loadServices = static function (array $servicesFile, ServicesConfigurator $services): void {
        foreach ($servicesFile['services'] as $serviceName => $service) {
            $theService = $services->set($serviceName, $service['class']);
            if (isset($service['arguments'])) {
                foreach ($service['arguments'] as &$argumentName) {
                    if ($argumentName[0] !== '@') {
                        continue;
                    }

                    $argumentName = new Reference(substr($argumentName, 1));
                }

                $theService->args($service['arguments']);
            }

            if (! isset($service['factory'])) {
                continue;
            }

            $theService->factory($service['factory']);
        }
    };

    $servicesFile = include ROOT_PATH . 'app/services.php';
    $loadServices($servicesFile, $services);
    $servicesFile = include ROOT_PATH . 'app/services_controllers.php';
    $loadServices($servicesFile, $services);
};
```

app/services_loader.php
```diff
@@ -7,37 +7,31 @@
...
 return static function (ContainerConfigurator $configurator): void {
     $services = $configurator->services();
+    /** @param array<string, string|array{class: string, arguments?: array<string>, factory?: callable}> $servicesFile */
     $loadServices = static function (array $servicesFile, ServicesConfigurator $services): void {
-        foreach ($servicesFile['services'] as $serviceName => $service) {
+        foreach ($servicesFile as $serviceName => $service) {
             if (is_string($service)) {
                 $services->alias($serviceName, $service);
                 continue;
             }
```
kamil-tekiela, line 19: Is there a reason why this remains? Didn't you remove all aliases in this PR?

MauricioFauth: This can be used for example to register an an interface as a service and make it an alias to a concrete implementation. I want to experiment with that to see if makes sense for phpMyAdmin. If it does not I'll remove that part.
Thinking about it now. It would have been better to remove that and add it back later if needed, because now it's just dead code.

MauricioFauth: Addressed by https://github.com/phpmyadmin/phpmyadmin/pull/20015.

PR #20015, opened 2026-01-08
Extract function to load services from services file

MauricioFauth:
Extracts to `ContainerBuilder::loadServices()`, removes service alias support and add unit tests.

### DIFF
--- a/app/services_loader.php
+++ b/app/services_loader.php
@@ -2,41 +2,17 @@
 
 declare(strict_types=1);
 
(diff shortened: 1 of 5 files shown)


williamdes (approved):
Is it now easier to implement php-di?

MauricioFauth:
> Is it now easier to implement php-di?

Yes, but more because of https://github.com/phpmyadmin/phpmyadmin/pull/20013 than this PR.

PR #20016, opened 2026-01-08
Extract user preferences handling from Config class

MauricioFauth:
This removes some `Config`'s circular dependencies, specially from the `DatabaseInterface` class.


src/Config.php
kamil-tekiela: Does this still need to be part of Config?

PR #20017, opened 2026-01-09
Insufficient space export exception

kamil-tekiela:
This refactors all Export plugins. All methods now return void instead of bool. There were only two reasons when any of these methods would return false: 

 - when OutputHandler has insufficient disk space to save file
 - when ExportJson cannot `json_encode` the output

`ExportException` exception was reused to handle these two cases. In the next PR, `json_encode` handling can be improved.

### DIFF
--- a/src/Plugins/Export/ExportJson.php
+++ b/src/Plugins/Export/ExportJson.php
@@ -9,6 +9,7 @@ namespace PhpMyAdmin\Plugins\Export;
 
 use PhpMyAdmin\Dbal\ConnectionType;
 use PhpMyAdmin\Dbal\DatabaseInterface;
+use PhpMyAdmin\Exceptions\ExportException;
 use PhpMyAdmin\Export\StructureOrData;
 use PhpMyAdmin\FieldMetadata;
 use PhpMyAdmin\Http\ServerRequest;
@@ -105,7 +106,7 @@ class ExportJson extends ExportPlugin
     /**
      * Outputs export header
      */
-    public function exportHeader(): bool
+    public function exportHeader(): void
     {
         $data = $this->encode([
             'type' => 'header',
@@ -113,18 +114,18 @@ class ExportJson extends ExportPlugin
             'comment' => 'Export to JSON plugin for phpMyAdmin',
         ]);
         if ($data === false) {
-            return false;
+            throw new ExportException('Failure during header export.');
         }
 
-        return $this->outputHandler->addLine('[' . "\n" . $data . ',' . "\n");
+        $this->outputHandler->addLine('[' . "\n" . $data . ',' . "\n");
     }
 
     /**
      * Outputs export footer
      */
-    public function exportFooter(): bool
+    public function exportFooter(): void
     {
-        return $this->outputHandler->addLine(']' . "\n");
+        $this->outputHandler->addLine(']' . "\n");
     }
 
     /**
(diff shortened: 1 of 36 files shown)


src/Plugins/Export/ExportJson.php
```diff
@@ -179,10 +178,10 @@ public function exportData(
             'data' => '@@DATA@@',
         ]);
         if ($buffer === false) {
-            return false;
+            throw new ExportException('Failure during data export.');
```
williamdes, line 181: should the strings be translated ?

kamil-tekiela: No, because exceptions form part of the code and not user-facing interface. They should be written in the same language that the rest of the code is. Also, these exceptions are silenced and they won't show up in UI or in the logs.

williamdes: Okay, alright

src/Plugins/Export/ExportPhparray.php
```diff
@@ -106,8 +104,6 @@ public function exportDBHeader(string $db, string $dbAlias = ''): bool
             . ' * Database ' . $this->commentString(Util::backquote($dbAlias))
```
liviuconcioiu, line 104: This should be translatable?

src/Plugins/Export/ExportTexytext.php
```diff
@@ -129,18 +129,14 @@ public function exportData(
...
-                $tableAlias !== ''
-                ? '== ' . __('Dumping data for table') . ' ' . $tableAlias . "\n\n"
-                : '==' . __('Dumping data for query result') . "\n\n",
-            )
-        ) {
-            return false;
-        }
+        $this->outputHandler->addLine(
+            $tableAlias !== ''
+            ? '== ' . __('Dumping data for table') . ' ' . $tableAlias . "\n\n"
```
liviuconcioiu, line 137: Qustion: Here and in a few other places, shouldn't `sprintf` be used instead?

```suggestion
            ? '== ' . sprintf(__('Dumping data for table %s'), $tableAlias) . "\n\n"
```

src/Plugins/Export/ExportPhparray.php
kamil-tekiela: Maybe, but it's not really part of this PR.

src/Plugins/Export/ExportTexytext.php
kamil-tekiela: Yeah, sounds good, but this PR is already huge, so maybe let's do it in the next PR.

williamdes: sprintf is greedy, use with caution

williamdes (approved):
Thank you !

src/Plugins/Export/ExportTexytext.php
kamil-tekiela: Can you explain what you mean by that?

williamdes: I maintain an import script for a client, 80G of data to import in one script run
I noticed that using sprintf did have memory impacts
maybe memory leaks, not sure anymore
run sprintf in a 100 million loop to log stuff, it will show what it has to say

kamil-tekiela: Interesting. I did some testing of my own, and it is true. Suprisingly it is no longer true as of PHP 8.4.0. Maybe whatever bug it was got fixed in 8.4.

kamil-tekiela: I see what happened https://tideways.com/profiler/blog/new-in-php-8-4-engine-optimization-of-sprintf-to-string-interpolation https://github.com/php/php-src/pull/14546

It is now compiled into string concatenation.

williamdes: Yay !!

PR #20022, opened 2026-01-13
Replace request globals with ServerRequest object in Sql and Display\Results classes

MauricioFauth:


src/Sql.php
```diff
@@ -795,7 +801,12 @@ private function executeTheQuery(
 
         $unlimNumRows = $this->countQueryResults($numRows, $justBrowsing, $db, $table, $statementInfo);
 
-        $this->cleanupRelations($db, $table, $_POST['dropped_column'] ?? '', ! empty($_POST['purge']));
+        $this->cleanupRelations(
+            $db,
+            $table,
+            $request->getParsedBodyParamAsString('dropped_column', ''),
+            $request->hasBodyParam('purge'),
```
kamil-tekiela, line 808: Did you verify that the value doesn't matter? It could have been an empty string or `"0"` before.

src/Sql.php
```diff
@@ -951,15 +969,15 @@ private function getQueryResponseForNoResultsReturned(
         }
 
         // For ajax requests add message and sql_query as JSON
-        if (empty($_REQUEST['ajax_page_request'])) {
+        if ($request->has('ajax_page_request')) {
```
kamil-tekiela: Is this condition inverted now?

src/Sql.php
```diff
@@ -1055,14 +1074,14 @@ private function getResponseForGridEdit(ResultInterface $result): void
      * Returns a message for successful creation of a bookmark or null if a bookmark
      * was not created
      */
-    private function getBookmarkCreatedMessage(): string
+    private function getBookmarkCreatedMessage(string|null $label): string
     {
         $output = '';
-        if (isset($_GET['label'])) {
+        if ($label !== null && $label !== '') {
```
kamil-tekiela, line 1080: Why the extra check for empty string?

src/Sql.php
```diff
@@ -1272,7 +1293,7 @@ private function getQueryResponseForResultsReturned(
     ): string {
         // If we are retrieving the full value of a truncated field or the original
         // value of a transformed field, show it here
-        if (isset($_POST['grid_edit']) && $_POST['grid_edit']) {
+        if ($request->hasBodyParam('grid_edit')) {
```
kamil-tekiela, line 1296: Was the part `&& $_POST['grid_edit']` unnecessary before?

kamil-tekiela (approved):

src/Sql.php
MauricioFauth: This is the type of flag param that is always true when it's defined.

src/Sql.php
MauricioFauth: The UI also doesn't accept empty values. So it makes sense to check for empty here.

src/Sql.php
MauricioFauth: Yes, it's always true when it's defined.

PR #20026, opened 2026-01-16
Refactor fetchRow()

kamil-tekiela:
Replacing the while loop with a simpler foreach

### DIFF
--- a/tests/unit/Utils/GisTest.php
+++ b/tests/unit/Utils/GisTest.php
@@ -37,8 +37,6 @@ class GisTest extends AbstractTestCase
         bool $SRIDOption,
         int $mysqlVersion,
     ): void {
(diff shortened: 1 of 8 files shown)


MauricioFauth (approved):
Thank you!

tests/unit/Utils/GisTest.php
```diff
@@ -37,8 +37,6 @@ public function testConvertToWellKnownText(
         bool $SRIDOption,
         int $mysqlVersion,
     ): void {
-        $resultStub = self::createMock(DummyResult::class);
```
MauricioFauth, line 40: The `use` statement for `DummyResult` should also be removed.

MauricioFauth (changes requested):

PR #20030, opened 2026-01-19
Test Core::getNoCacheHeaders() with a frozen clock

MauricioFauth:
This avoids the tests randomly failing when current time differs.

Also changes the format from RFC 1123 to RFC 7231.

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Expires
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Last-Modified

### DIFF
--- a/src/Core.php
+++ b/src/Core.php
@@ -4,6 +4,9 @@ declare(strict_types=1);
 
 namespace PhpMyAdmin;
 
+use DateTimeImmutable;
+use DateTimeInterface;
+use DateTimeZone;
 use PhpMyAdmin\Error\ErrorHandler;
 use PhpMyAdmin\Exceptions\MissingExtensionException;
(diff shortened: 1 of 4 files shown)


MauricioFauth:
Maybe using the `ClockInterface` from the [PSR-20](https://www.php-fig.org/psr/psr-20/) is a better idea. And it's already an indirect dependency.

```
$ composer why psr/clock
lcobucci/clock        3.3.1 requires psr/clock (^1.0) 
web-auth/webauthn-lib 4.9.2 requires psr/clock (^1.0)
```

```
$ composer why lcobucci/clock
web-auth/webauthn-lib 4.9.2 requires lcobucci/clock (^2.2|^3.0)
```

williamdes (approved):
💯 🕐

PR #20035, opened 2026-01-20
Clean up TransformationPluginsTest

kamil-tekiela:
I don't know why `Results::$row` was ever there, but it was there for years and removing it doesn't break any test case. 

`$this->setGlobalConfig();` was obsoleted by `parent::setUp();` some time ago. 

`$this->setLanguage();` was added by William, but I assume it's no longer necessary. There is no comment explaining why it's there, so I can only guess that it was a workaround for some test limitation.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -16789,12 +16789,6 @@ parameters:
 			path: tests/unit/Plugins/Transformations/TransformationPluginsTest.php
 
 		-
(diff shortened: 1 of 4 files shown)


williamdes (approved):
Thank you
My commit says it was for backup globals: https://github.com/phpmyadmin/phpmyadmin/commit/8f7d3d1fa76bf006dca38e2ff5b424af1055b931

PR #20036, opened 2026-01-20
Frontend: Allow mouse selection of cells for easier copying

JustJacob95:
Allow mouse selection of cells

[2026-01-20 22-41-31.webm](https://github.com/user-attachments/assets/e9e38193-1321-4615-b280-a3cf7296951e)

In the future, it could be possible to add support for keyboard controls

### DIFF
--- a/resources/js/makegrid.ts
+++ b/resources/js/makegrid.ts
@@ -2391,7 +2391,7 @@ const makeGrid = function (t, enableResize = undefined, enableReorder = undefine
 
             function renderCellSelection () {
                 g.renderedSelectedCells.clear();
-                $(g.t).find('.cell-selected').removeClass(selectingClass);
+                $(g.t).find(`.${selectingClass}`).removeClass(selectingClass).removeClass('with-bg-color');
                 if (g.selectedColumns.size === 0) {
                     return;
                 }
@@ -2402,7 +2402,7 @@ const makeGrid = function (t, enableResize = undefined, enableReorder = undefine
                         const columns = rowIndex === -1 ?
                             $(g.t).find('thead tr').eq(0).find('th').eq(cellIndex - colspan) :
                             $(g.t).find('tbody tr').eq(rowIndex).find('td').eq(cellIndex);
-                        columns.addClass(selectingClass);
+                        columns.addClass((g.selectedColumns.size === 1 && g.selectedRows.size === 1) ? selectingClass : `${selectingClass} with-bg-color`);
                     });
                 });
             }
@@ -2456,12 +2456,7 @@ const makeGrid = function (t, enableResize = undefined, enableReorder = undefine
 
             // Keyboard events for cell selection
             $(document).on('keydown', function (e) {
-                // add throttle to avoid multiple events firing too quickly
-                if (keyboardEventTimestamp && (Date.now() - keyboardEventTimestamp) < 100) {
-                    return;
-                }
 
-                keyboardEventTimestamp = Date.now();
 
                 if (document.activeElement && $(document.activeElement).is('input, textarea, select')) {
                     return; // do not interfere with input fields
@@ -2474,9 +2469,16 @@ const makeGrid = function (t, enableResize = undefined, enableReorder = undefine
 
                     // Select all cells with header
                     $(g.t).find('tbody > tr, thead > tr').each(function () {
-                        $(this).find('td.data, th:not(.column_action)').addClass(selectingClass);
+                        $(this).find('td.data, th:not(.column_action)').addClass(`${selectingClass} with-bg-color`);
                     });
                 }
+
+                // add throttle to avoid multiple events firing too quickly
+                if (keyboardEventTimestamp && (Date.now() - keyboardEventTimestamp) < 100) {
+                    return;
+                }
+
+                keyboardEventTimestamp = Date.now();
 
                 // arrow + shift to select cells
                 if (e.shiftKey && g.endSelectCell) {
@@ -2586,7 +2588,7 @@ const makeGrid = function (t, enableResize = undefined, enableReorder = undefine
                 }
 
                 if (!e.ctrlKey && !e.metaKey) {
-                    $(g.t).find('.cell-selected').removeClass(selectingClass);
+                    $(g.t).find(`.${selectingClass}`).removeClass(`${selectingClass} with-bg-color`);
                     resetCellSelection();
                     isMultiselect = false;
                 } else {
@@ -2652,7 +2654,7 @@ const makeGrid = function (t, enableResize = undefined, enableReorder = undefine
                     return;
                 }
 
-                if ($(g.t).find('.cell-selected').length > 0) {
+                if ($(g.t).find(`.${selectingClass}`).length > 0) {
                     let selectionText = '';
 
                     const headers: string[] = [];
--- a/resources/js/makegrid.ts
+++ b/resources/js/makegrid.ts
@@ -2297,15 +2297,20 @@ const makeGrid = function (t, enableResize = undefined, enableReorder = undefine
             g.startSelectCell = null;
             g.preEndSelectCell = null;
             g.endSelectCell = null;
+            g.selectedColumns = new Set();
+            g.selectedRows = new Set();
+            g.renderedSelectedCells = new Set();
 
             const colspan = Number(
                 $(g.t).find('thead th').first().attr('colspan')
-            ) - 1 || 1;
+            ) - 1 || 0;
+
+            const selectingClass = 'cell-selected';
 
             let keyboardEventTimestamp = 0;
 
             // Check if an element is visible for user
-            function isPartiallyHidden (el) {
+            function isPartiallyHidden (el: HTMLElement) {
                 const rect = el.getBoundingClientRect();
                 const vW = window.innerWidth;
                 const vH = window.innerHeight;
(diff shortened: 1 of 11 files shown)


williamdes (approved):
I have tested this, it is really awesome !
That said I noticed:
- that I can not scroll while holding the last cell (with a touchpad), arrow keys work fine
- maybe ctrl-A should select all instead of selecting all the page text
- I would like to be able to copy the table headers too
- the current copy to clipboard function maybe needs a small fix, I noticed that it has one more tab than your output (my last columns where empty)

resources/js/makegrid.ts
```diff
@@ -2287,6 +2287,192 @@ const makeGrid = function (t, enableResize = undefined, enableReorder = undefine
...
+
+            // Reset selection when clicking outside the table
+            $(document).on('mousedown', function (e) {
+                if (!$(e.target).closest(g.t).length) {
+                    resetCellSelection();
+                }
+            });
+
+            // Add styles
+            if ($('#cell-selection-style').length === 0) {
```
williamdes: The styles can go into scss files
@MauricioFauth what is best?

liviuconcioiu:
Awesome improvement! ~~A few things:~~

~~1. After each `NULL` it adds a space, then `\t`. Same when copying table header.~~
~~2. When copying a single `NULL` cell it adds a space before and after.~~
~~3. What should be the expected behaviour when clicking another row after selection? Right now it doesn't add a new line if a row isn't copied and also columns count don't match.~~

Sorry for this. It seems that Firefox supports this natively (I forgot about it).

This is how Firefox handles by default:

https://github.com/user-attachments/assets/60fce1fb-a107-44be-9367-9f7a03ebeb7d

What I did notice is sometimes is using the native copy and sometimes it uses the code from this PR.

JustJacob95:
> What I did notice is sometimes is using the native copy and sometimes it uses the code from this PR.

Hmmm , I can’t reproduce this behavior. Could you please tell me your browser version, OS, and which phpMyAdmin theme you are using?

resources/js/makegrid.ts
JustJacob95: @williamdes 

Good point, but we don’t have a shared SCSS endpoint for all themes as i see

If we move this into theme-specific SCSS (bootstrap / metro / original / pmahomme), any custom theme won’t get these styles and users will never see cell selection there

JustJacob95:
Thanks for your comments @williamdes 


>  that I can not scroll while holding the last cell (with a touchpad), arrow keys work fine

I’ll check this as soon as I can test it on a real touchpad device.

> maybe ctrl-A should select all instead of selecting all the page text

Done ✅

> I would like to be able to copy the table headers too

Done ✅, but with a small note.

At the moment, headers can be selected in two ways:

* via Ctrl + A

* by starting selection from the table body and extending it upward

This was done intentionally so we don’t break the existing column drag & drop logic that changes column order.


> the current copy to clipboard function maybe needs a small fix

Fixed ✅

williamdes:
Thank you !
I will test this very soon, I am very impatient

>  I’ll check this as soon as I can test it on a real touchpad device.

A mouse can do, you need to: use your hand and only have a finger holding the click button.
Dragging the cursor should select more or less things, but when you reach the edge it should scroll the page and keep selecting more cells.

resources/js/makegrid.ts
williamdes: > If we move this into theme-specific SCSS (bootstrap / metro / original / pmahomme), any custom theme won’t get these styles and users will never see cell selection there

Maybe you can find bootstrap classes that would provide the color or else ?

williamdes: > If we move this into theme-specific SCSS (bootstrap / metro / original / pmahomme), any custom theme won’t get these styles and users will never see cell selection there

Anyway there is no external theme compatible with 6.0.x to my knowledge

williamdes:
Wife test:
- Holding <kbd>SHIFT</kbd> should select cells as you go with arrow keys (OpenOffice does that)
- Holding <kbd>CTRL</kbd> should enable to retain the currently selected cells and select more. For example if you want to skip 3 rows, you hold <kbd>CTRL</kbd> select 1 row, skip 3 rows and select more

My test:
- I can not select the header row from above the header, but from below it works.
- Selecting rows does not select "repeating headers" that is great as expected.

Other than that, everything looks good !

liviuconcioiu:
> Hmmm , I can’t reproduce this behavior. Could you please tell me your browser version, OS, and which phpMyAdmin theme you are using?

Firefox/149 but the native holding `CTRL` + mouse click was implemented a long time ago. The one with no background is what Firefox does by default. When I use `CTRL+A` or click it is what this PR does. As for the theme, I use `Bootstrap`.

The only thing that is annoying, at least for me, is when I try to select some partial text, I can no longer do that, the entire cell is copied.

https://github.com/user-attachments/assets/659c9966-fb2a-4e9b-ab01-b5b081c5d8fb

resources/js/makegrid.ts
JustJacob95: i did something like that 
https://github.com/phpmyadmin/phpmyadmin/pull/20036/commits/9b2caa37586769b3f4b4508ab3f78a93864078ca

JustJacob95:
@williamdes 

> Holding SHIFT should select cells as you go with arrow keys (OpenOffice does that)

Done ✅

> Holding CTRL should enable to retain the currently selected cells and select more. For example if you want to skip 3 rows, you hold CTRL select 1 row, skip 3 rows and select more

Done ✅


Good 2 points, logic for this selection i picked up the logic from other database clients


> I can not select the header row from above the header, but from below it works.

As I mentioned before, this behavior is intentional to avoid conflicts when dragging columns. For now, you can select headers using 'Select All' (Ctrl + A) or by starting the selection from below

JustJacob95:
@liviuconcioiu 

> The only thing that is annoying, at least for me, is when I try to select some partial text, I can no longer do that, the entire cell is copied.

Yes, for some it might be annoying at first. I think this is exactly the kind of case where we need to choose which behavior will be more better. In any case, the double-click functionality on a cell is still preserved, where the user can copy part of the text

And also i did CTRL + click by cells multiselect (when you already have some selection) could you check how its working for you please?

resources/js/makegrid.ts
williamdes: Not bad, but I think you should put it in each common file: https://github.com/phpmyadmin/phpmyadmin/blob/9b2caa37586769b3f4b4508ab3f78a93864078ca/public/themes/pmahomme/scss/_common.scss

That's not ideal but anyway

williamdes (approved):
I like this, more reviews would be great
but I think I will merge soon if no blocking comments arise

The scss is better but should not create a new way, and should copy paste scss into each common file

liviuconcioiu:
>  could you check how its working for you please?

Now it works okay on Firefox.

Not sure about background, I think it looks much better without `background-color: #d0e4f7;` when the cell is clicked.

I think that's why I didn't like that I couldn't select partial text - it lets the impression that the cell is blocked. For example Excel doesn't have a background when the cell is clicked, only if `CTRL` is pressed and multiple cells selected.

https://github.com/user-attachments/assets/b53fda97-0236-48bd-a28e-614596336abc

https://github.com/user-attachments/assets/e243fb08-ba86-461d-a4e6-7b7819d4360f

JustJacob95:
@williamdes 

> The scss is better but should not create a new way, and should copy paste scss into each common file

i've removed new common.scss and copy styles for each themes

liviuconcioiu:
This looks much better! I also hope everyone likes it.

JustJacob95:
@liviuconcioiu 

> Not sure about background, I think it looks much better without background-color: #d0e4f7; when the cell is clicked.
> 
> I think that's why I didn't like that I couldn't select partial text - it lets the impression that the cell is blocked. For example Excel doesn't have a background when the cell is clicked, only if CTRL is pressed and multiple cells selected.


Thanks for your comment, i aggree with you
few min ago i commited extra logic - dont color background if we have only 1 cell selected

Could you please have a look again?

liviuconcioiu:
It looks awesome!

williamdes (approved):
Awesome work !
I am merging this, if there is new bugs or else please open an issue and ping here with the issue id

williamdes:
> The only thing that is annoying, at least for me, is when I try to select some partial text, I can no longer do that, the entire cell is copied.

Yeah that would be great to have back, I guess many users will complain

williamdes:
the [(phpMyAdmin 6.0+snapshot)](https://www.phpmyadmin.net/downloads/#snapshot_6.0+snapshot) also available as a [non official docker image](https://github.com/sudo-bot/docker-phpmyadmin-snapshots#readme) is built with this feature !

Also, maybe it should respect `$cfg['DisableShortcutKeys']` for CTRL+A ?
Not sure
Also, CTRL+A does not work on Opera but works on LibreWolf = Firefox

JustJacob95:
@williamdes 

> Also, maybe it should respect $cfg['DisableShortcutKeys'] for CTRL+A ?

i think so, iwll check

> Also, CTRL+A does not work on Opera but works on LibreWolf = Firefox

hmm, and i will check it too

williamdes:
See https://github.com/phpmyadmin/phpmyadmin/issues/20181

JustJacob95:
Perhaps we should make this feature optional and disable it by default? It might not be intuitive for users who have used phpMyAdmin for years and are used to a different workflow to select text

What to you think?

liviuconcioiu:
@JustJacob95 I discovered two bugs.

When I hit `Ctrl+C` in a `JSON`. It won't copy full text only truncated.

Also, if I open a `NULL` cell and `Ctrl+A`, `Ctrl+C`, then write some text and `Ctrl+A`, `Ctrl+C`, `Ctrl+V` it copies `NULL`. instead of textarea value.

https://github.com/user-attachments/assets/71748876-9c61-4f22-8fa0-de516d5395c0

JustJacob95:
@liviuconcioiu thanks! 

fixed in https://github.com/phpmyadmin/phpmyadmin/pull/20188

PR #20039, opened 2026-01-21
Major cleanup of config

kamil-tekiela:
I understand that these changes can be a little difficult to understand, so please don't hesitate to ask me for more details. Most of it is just removing dead code from unit tests. `ldi_local_option` was unused and I doubt it was ever implemented properly so I removed it completely. `Config::set()` was expanded to handle nested values because it is the best way to adjust settings in unit tests. But it is far from an ideal solution so it may change in the future. `array_replace_recursive` was removed even though it would allow non-config settings to be stored prior to that. At the moment, I wasn't able to find a use for it, but I admit that I could have missed one somewhere. The coupling with `UserPreferencesHandler` is the most troubling, but I hope to improve that with time too.

### DIFF
--- a/src/Config/Settings.php
+++ b/src/Config/Settings.php
@@ -3920,10 +3920,10 @@ final class Settings
         }
 
         return match ($settings['NavigationTreeDefaultTabTable']) {
-            'sql', 'tbl_sql.php' => '/table/sql',
-            'search', 'tbl_select.php' => '/table/search',
-            'insert', 'tbl_change.php' => '/table/change',
-            'browse', 'sql.php' => '/sql',
+            'sql', 'tbl_sql.php', '/table/sql' => '/table/sql',
+            'search', 'tbl_select.php', '/table/search' => '/table/search',
+            'insert', 'tbl_change.php', '/table/change' => '/table/change',
+            'browse', 'sql.php', '/sql' => '/sql',
             default => '/table/structure',
         };
     }
@@ -3940,11 +3940,11 @@ final class Settings
         }
 
         return match ($settings['NavigationTreeDefaultTabTable2']) {
-            'structure', 'tbl_structure.php' => '/table/structure',
-            'sql', 'tbl_sql.php' => '/table/sql',
-            'search', 'tbl_select.php' => '/table/search',
-            'insert', 'tbl_change.php' => '/table/change',
-            'browse', 'sql.php' => '/sql',
+            'structure', 'tbl_structure.php', '/table/structure' => '/table/structure',
+            'sql', 'tbl_sql.php', '/table/sql' => '/table/sql',
+            'search', 'tbl_select.php', '/table/search' => '/table/search',
+            'insert', 'tbl_change.php', '/table/change' => '/table/change',
+            'browse', 'sql.php', '/sql' => '/sql',
             default => '',
         };
     }
@@ -4520,10 +4520,10 @@ final class Settings
         }
 
         return match ($settings['DefaultTabServer']) {
-            'databases', 'server_databases.php' => '/server/databases',
-            'status', 'server_status.php' => '/server/status',
-            'variables', 'server_variables.php' => '/server/variables',
-            'privileges', 'server_privileges.php' => '/server/privileges',
+            'databases', 'server_databases.php', '/server/databases' => '/server/databases',
+            'status', 'server_status.php', '/server/status' => '/server/status',
+            'variables', 'server_variables.php', '/server/variables' => '/server/variables',
+            'privileges', 'server_privileges.php', '/server/privileges' => '/server/privileges',
             default => '/',
         };
     }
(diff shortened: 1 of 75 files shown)


src/Controllers/UserPasswordController.php
```diff
@@ -39,11 +39,12 @@ public function __invoke(ServerRequest $request): Response
          * Displays an error message and exits if the user isn't allowed to use this
          * script
          */
+        $hasAccessPrivilege = true;
         if (! $this->config->settings['ShowChgPassword']) {
```
williamdes:
```suggestion
        $hasAccessPrivilege = false;
if ($this->config->settings['ShowChgPassword']) {
```
Personnaly I perfer such logics, they are easier to read and debug

kamil-tekiela: I am not sure I understand. Are you saying you would prefer if the variable was `$hasDeniedAccess`? This won't remove the negation from `if (! $this->config->settings['ShowChgPassword']) {`.

kamil-tekiela: What I can do is inline it `$hasAccessPrivilege = $this->config->settings['ShowChgPassword'] || $this->dbi->selectDb('mysql');`

src/Config/Settings.php
```diff
@@ -4559,10 +4559,10 @@ private function setDefaultTabTable(array $settings): string
...
 
         return match ($settings['DefaultTabTable']) {
-            'structure', 'tbl_structure.php' => '/table/structure',
-            'sql', 'tbl_sql.php' => '/table/sql',
-            'search', 'tbl_select.php' => '/table/search',
-            'insert', 'tbl_change.php' => '/table/change',
+            'structure', 'tbl_structure.php', '/table/structure' => '/table/structure',
+            'sql', 'tbl_sql.php', '/table/sql' => '/table/sql',
+            'search', 'tbl_select.php', '/table/search' => '/table/search',
+            'insert', 'tbl_change.php', '/table/change' => '/table/change',
```
williamdes, line 4565: this is a bug fix if I am correct

williamdes (approved):
LGTM, I just need to find where the LDI was removed to see why we missed so much code

src/Config/Settings/Import.php
```diff
@@ -210,17 +209,6 @@ final class Import
...
-    /**
-     * 'auto' for auto-detection, true or false for forcing
-     *
-     * ```php
-     * $cfg['Import']['ldi_local_option'] = 'auto';
-     * ```
-     *
-     * @psalm-var 'auto'|bool
-     */
-    public string|bool $ldi_local_option;
```
williamdes, line 222: We should maybe warn users with a deprecation message when they use old removed settings 🤔

kamil-tekiela: Is there a mechanism for it? Is this option documented anywhere? Maybe put a warning in the documentation.

src/Config/Settings.php
kamil-tekiela: I don't think so. It seems to be working for me without this change. But the changes I am proposing are trying to align `$settings` and `$config` with each other.

src/Config/Settings/Import.php
liviuconcioiu: > Is there a mechanism for it? Is this option documented anywhere? Maybe put a warning in the documentation.

This will show the message in footer:

```PHP
$this->errors[] = [
    'message' => __(
        'Message here.',
    ),
    'severity' => 'notice',
];
```

williamdes (approved):

src/Controllers/UserPasswordController.php
williamdes: Perfect option, thanks

MauricioFauth:
> `ldi_local_option` was unused

This setting is probably used by `Plugins::checkboxCheck()`.

kamil-tekiela:
> > `ldi_local_option` was unused
> 
> This setting is probably used by `Plugins::checkboxCheck()`.

Can you show code where it's used? I am away from computer until next week so I can't check.

MauricioFauth:
I just tested that `ldi_local_option` is indeed used by `Plugins::checkboxCheck()`.

> Can you show code where it's used?

```php
public static function checkboxCheck(string $section, string $opt): string
{
    // If the form is being repopulated using $_GET data, that is priority
    if (
        isset($_GET[$opt])
        || ! isset($_GET['repopulate'])
        && ((ImportSettings::$timeoutPassed && isset($_REQUEST[$opt]))
            || ! empty(Config::getInstance()->settings[$section][$opt]))
    ) {
        return ' checked';
    }

    return '';
}
```

MauricioFauth (approved):

MauricioFauth:
I merged this PR without the commit "Remove ldi_local_option setting".

williamdes:
> I merged this PR without the commit "Remove ldi_local_option setting".

Thank you, we should make such hidden patterns obvious by using a defined list or something

PR #20046, opened 2026-01-24
Fix checking cookies for auto-saved query

MoonE:
This fixes handling of the root path when checking cookies.
There was a breaking change when upgrading to a new major version of the js-cookie lib.

Replaces #19869


williamdes (approved):
Thanks, this looks good.

PR #20047, opened 2026-01-24
Fix case where user preferences are not removed

MauricioFauth:
When an user revert a setting to its default value, the setting is not removed from the storage.

- Introduced by https://github.com/phpmyadmin/phpmyadmin/pull/19925
- Related to https://github.com/phpmyadmin/phpmyadmin/pull/19900

### DIFF
--- a/libraries/classes/UserPreferences.php
+++ b/libraries/classes/UserPreferences.php
@@ -131,7 +131,7 @@ class UserPreferences
             $existingPrefs = $this->load();
             if (isset($existingPrefs['config_data']['2fa'])) {
                 // This is likely a partial save from page settings - merge to preserve 2fa
-                $config_array = array_merge($existingPrefs['config_data'], $config_array);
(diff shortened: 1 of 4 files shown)


libraries/classes/UserPreferences.php
```diff
@@ -131,7 +131,7 @@ public function save(array $config_array)
             $existingPrefs = $this->load();
             if (isset($existingPrefs['config_data']['2fa'])) {
                 // This is likely a partial save from page settings - merge to preserve 2fa
-                $config_array = array_merge($existingPrefs['config_data'], $config_array);
+                $config_array['2fa'] = $existingPrefs['config_data']['2fa'];
```
MauricioFauth, line 134: Only `2fa` needs to be merged here.

kamil-tekiela (approved):
Yeah, it looks like this fixes the issue. Still it's pretty strange that 2fa is saved this way.

williamdes (approved):

PR #20058, opened 2026-01-30
Extract dependencies from some Util methods

MauricioFauth:
Uses DI instead.

### DIFF
--- a/tests/unit/Server/PrivilegesTest.php
+++ b/tests/unit/Server/PrivilegesTest.php
@@ -767,7 +767,10 @@ class PrivilegesTest extends AbstractTestCase
 
     public function testGetSqlQueriesForDisplayAndAddUserWithUserDefinedPassword(): void
     {
-        $dbi = $this->createDatabaseInterface();
+        $dbiDummy = $this->createDbiDummy();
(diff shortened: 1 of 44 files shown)


tests/unit/Server/PrivilegesTest.php
```diff
@@ -767,7 +767,10 @@ public function testGetSqlQueriesForDisplayAndAddUserMySql50500AndUserDefinedPas
 
     public function testGetSqlQueriesForDisplayAndAddUserWithUserDefinedPassword(): void
     {
-        $dbi = $this->createDatabaseInterface();
+        $dbiDummy = $this->createDbiDummy();
+        $dbiDummy->addResult("SHOW PLUGINS SONAME LIKE '%_password_check%'", []);
+        $dbiDummy->addResult("SELECT PASSWORD('pma_password');", [['*pma_password*']]);
+        $dbi = $this->createDatabaseInterface($dbiDummy);
```
MauricioFauth, line 773: This happened because there was two different instances of `DatabaseInterface` causing the tested path to be different than expected. This means that the previous tested path was wrong.

PR #20061, opened 2026-01-31
Set sql mode for sql-parser when connecting to db

MoonE:
Always use the actual server sql mode for the sql parser instead of assuming a default.


williamdes (approved):
Thank you!

MauricioFauth (approved):

Issue #17398, opened 2022-02-17
Clicking on JSON column types, triggers UPDATE query


liviuconcioiu:
### Describe the bug

When clicking on JSON column types, a UPDATE query is triggered. Also, the query generated contains `\n` and spaces.
`{\n    \"a\": [\n        [\n            12345,\n            12345\n        ]\n    ],\n    \"b\": true\n}` instead of `{"a": [[12345, 12345]], "b": true}`

Also, JSON columns that aren't _`NULL`_, have a `null` value when browsing table that is also present when exporting a table
`(1, NULL, 'null', NULL),` - I don't know if this is a bug or not.

### To Reproduce

1. Create a table with JSON column types
2. Insert a few rows
3. Click on a field

### Expected behavior

It should not trigger a UPDATE query.

### Screenshots

https://user-images.githubusercontent.com/25424343/154415730-533fbe62-3340-462d-aec5-6cfd2367c06a.mp4

https://user-images.githubusercontent.com/25424343/154416842-cd6bd0ee-ac55-487f-a071-c103c2c42d3f.mp4

### Server configuration

- phpMyAdmin version: 5.x

PR #20076, opened 2026-02-05
Fix #19368 - Improve line ending handling in grid editor

### LINKED ISSUE #19368: \r\n is replaced with \n when clicking on a value
### Describe the bug

`\r\n` is replaced with `\n` when clicking on a value in special cases. Not sure, but I think is triggered when line exceeds 1024 chars?

### To Reproduce

1. Create table:

```SQL
CREATE TABLE `test` (
  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
  `value` text DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=MyISAM AUTO_INCREMENT=1 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;

INSERT INTO `test` (`id`, `value`) VALUES
(1, 'test\r\ntest'),
(2, 'test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test <td class=\"text-right\">\r\n        test test test test test test test test test test test test test test test test test test test test</td>');
```

2. Click on the first value, nothing happens even if is `\r\n`.
3. Click on the second value
4. See `UPDATE` being triggered and `\r\n` is replaced with `\n`.

### Expected behavior

`\r\n` should not be replaced with `\n`.

### Screenshots

https://github.com/user-attachments/assets/4e2bd43f-c24b-49ee-b655-e735cd789f29

### Server configuration

og-khushalpatel:
### Description

This pull request adds explicit line ending (CRLF / LF) handling to inline grid editing for text-like fields.
Browsers normalize textarea input to LF, which previously made it impossible for users to reliably view, preserve, or intentionally change the original line-ending format stored in the database. This PR addresses that gap by:

- Detecting the original line-ending format (CRLF or LF) on the server
- Exposing a line-ending selector in the inline edit area for applicable text fields
- Allowing users to preserve or intentionally change the line-ending format
- Ensuring idempotent behavior to avoid duplicated line breaks on repeated edits
- Keeping rendering and comparison logic consistent across truncated and non-truncated values
The implementation follows existing phpMyAdmin patterns and avoids altering behavior for non-textual field types.

Fixes #19368 - \r\n is replaced with \n when clicking on a value

- [ X ] Make sure you have read our [CONTRIBUTING.md](https://github.com/phpmyadmin/phpmyadmin/blob/master/CONTRIBUTING.md) document.
- [ X ] Make sure you are making a pull request against the correct branch. For example, for bug fixes in a released version use the corresponding QA branch and for new features use the _master_ branch. If you have a doubt, you can ask as a comment in the bug report or on the mailing list.
- [ X ] Every commit has proper `Signed-off-by` line as described in our [DCO](https://github.com/phpmyadmin/phpmyadmin/blob/master/DCO). This ensures that the work you're submitting is your own creation.
- [ X ] Every commit has a descriptive commit message.
- [ X ] Every commit is needed on its own, if you have just minor fixes to previous commits, you can squash them.

### DIFF
--- a/js/src/functions.js
+++ b/js/src/functions.js
@@ -4070,16 +4070,28 @@ Functions.getCellValue = function (td) {
 /**
  * Validate and return stringified JSON inputs, or plain if invalid.
  *
- * @param json the json input to be validated and stringified
+ * @param value the json input to be validated and stringified
  * @param replacer An array of strings and numbers that acts as an approved list for selecting the object properties that will be stringified.
  * @param space Adds indentation, white space, and line break characters to the return-value JSON text to make it easier to read.
  * @return {string}
  */
-Functions.stringifyJSON = function (json, replacer = null, space = 0) {
+Functions.stringifyJSON = function (value, replacer = null, space = 0) {
     try {
-        return JSON.stringify(JSON.parse(json), replacer, space);
+        // If already an object/array, stringify directly
+        if (typeof value === 'object' && value !== null) {
+            return JSON.stringify(value, replacer, space);
+        }
+
+        // If string, try to parse then stringify (normalizes formatting)
+        if (typeof value === 'string') {
+            return JSON.stringify(JSON.parse(value), replacer, space);
+        }
+
+        // Fallback: force string
+        return String(value);
     } catch (e) {
-        return json;
+        console.log('error occured in stringifying json');
+        return String(value);
     }
 };
 
--- a/js/src/functions.js
+++ b/js/src/functions.js
@@ -4090,7 +4090,6 @@ Functions.stringifyJSON = function (value, replacer = null, space = 0) {
         // Fallback: force string
         return String(value);
     } catch (e) {
-        console.log('error occured in stringifying json');
         return String(value);
     }
 };
--- a/js/src/functions.js
+++ b/js/src/functions.js
@@ -4037,6 +4037,17 @@ AJAX.registerOnload('functions.js', function () {
 }(jQuery));
 
 /**
+ * Normalize line endings for comparison purposes.
+ * Browsers normalize textarea values to LF per HTML spec.
+ *
+ * @param {string} value
+ * @return {string}
+ */
+Functions.normalizeNewlines = function (value) {
+    return value.replace(/\r\n|\r/g, '\n');
+};
+
+/**
  * Return value of a cell in a table.
  *
  * @param {string} td
--- a/js/src/makegrid.js
+++ b/js/src/makegrid.js
@@ -751,8 +751,8 @@ var makeGrid = function (t, enableResize, enableReorder, enableVisib, enableGrid
                             $target.html(newHtml);
                         }
                         $thisField.data('originalValue', value);
-                        if ($(g.currentEditCell).data('lineEnding') !== $(g.currentEditCell).attr('data-line-ending')) {
-                            $(g.currentEditCell).attr('data-line-ending', $(g.currentEditCell).data('lineEnding'));
+                        if ($thisField.data('lineEnding') !== $thisField.attr('data-line-ending')) {
+                            $thisField.attr('data-line-ending', $thisField.data('lineEnding'));
                         }
                     }
                     if ($thisField.is('.bit')) {
@@ -924,13 +924,13 @@ var makeGrid = function (t, enableResize, enableReorder, enableVisib, enableGrid
 
                 if (hasLineEnding) {
                     // Canonical server truth (preferred)
-                    let originalValue = $td.data('originalValue');
+                    let originalValue = $(g.currentEditCell).data('originalValue');
                     // Fallback ONLY if not present (AJAX-fetched truncated case)
                     if (originalValue === undefined) {
                         originalValue = Functions.normalizeNewlines(
                             Functions.getCellValue(g.currentEditCell)
                         );
-                        $td.data('originalValue', originalValue);
+                        $(g.currentEditCell).data('originalValue', originalValue);
                     }
                     // Cache it once
                     $editArea.append(
@@ -949,11 +949,11 @@ var makeGrid = function (t, enableResize, enableReorder, enableVisib, enableGrid
                             '</label>' +
                         '</div>'
                     );
-                    var detectedEnding = $td.data('lineEnding');
+                    var detectedEnding = $(g.currentEditCell).data('lineEnding');
                     $editArea.find('.line-ending-select').val(detectedEnding);
                     // handle change of line ending selection
-                    $editArea.on('change', '.line-ending-select', function () {
-                        $td.data('lineEnding', this.value);
+                    $editArea.find('.line-ending-select').off('change').on('change', function () {
+                        $(g.currentEditCell).data('lineEnding', this.value);
                     });
                 }
 
--- a/js/src/makegrid.js
+++ b/js/src/makegrid.js
@@ -943,11 +943,6 @@ var makeGrid = function (t, enableResize, enableReorder, enableVisib, enableGrid
                     }
                     var crlfTemplateEl = document.getElementById('crlf-div-template-' + t.dataset.uniqueid);
                     var $crlfDiv = $(crlfTemplateEl.content.cloneNode(true));
-                    $crlfDiv.find('.line-ending-text').text(Messages.strLineEnding);
-                    $crlfDiv.find('img').attr({
-                        'title': Messages.strLineEndingDetected,
-                        'alt': Messages.strLineEndingDetected
-                    });
                     $editArea.append($crlfDiv);
                     var detectedEnding = $(g.currentEditCell).data('lineEnding');
                     $editArea.find('.line-ending-select').val(detectedEnding);
--- a/js/src/makegrid.js
+++ b/js/src/makegrid.js
@@ -934,10 +934,10 @@ var makeGrid = function (t, enableResize, enableReorder, enableVisib, enableGrid
                         '<div class="crlf_div">' +
                             '<label>' +
                                 '<span class="line-ending-label">' +
-                                    'Line endings' +
+                                    'Line ending:' +
                                     '<span>' +
                                         '<img src="themes/dot.gif" title="The suggested value was detected from your existing data." alt="The suggested value was detected from your existing data." class="icon ic_b_help">' +
-                                    '</span>:' +
+                                    '</span>' +
                                 '</span>' +
                                 '<select class="line-ending-select">' +
                                     '<option value="LF">LF</option>' +
--- a/js/src/makegrid.js
+++ b/js/src/makegrid.js
@@ -701,7 +701,6 @@ var makeGrid = function (t, enableResize, enableReorder, enableVisib, enableGrid
                     if (isNull) {
                         $thisField.find('span').html('NULL');
                         $thisField.addClass('null');
-                        // todo: set original value to null & remove data-line-ending attribute
                         $thisField.data('originalValue', null);
                         if ($thisField.is('[data-line-ending]')) {
                             $thisField
@@ -711,8 +710,8 @@ var makeGrid = function (t, enableResize, enableReorder, enableVisib, enableGrid
                     } else {
                         $thisField.removeClass('null');
                         var value = data.isNeedToRecheck
-                            ? data.truncatableFieldValue.replace(/\r\n/g, '\n')
-                            : $thisField.data('value').replace(/\r\n/g, '\n');
+                            ? Functions.normalizeNewlines(data.truncatableFieldValue)
+                            : Functions.normalizeNewlines($thisField.data('value'));
 
                         // Truncates the text.
                         $thisField.removeClass('truncated');
@@ -937,9 +936,12 @@ var makeGrid = function (t, enableResize, enableReorder, enableVisib, enableGrid
                         '<div class="crlf_div">' +
                             '<label>' +
                                 '<span class="line-ending-label">' +
-                                    'Line ending:' +
+                                    Messages.strLineEnding +
                                     '<span>' +
-                                        '<img src="themes/dot.gif" title="The suggested value was detected from your existing data." alt="The suggested value was detected from your existing data." class="icon ic_b_help">' +
+                                        '<img src="themes/dot.gif" ' +
+                                        'title="' + Functions.escapeHtml(Messages.strLineEndingDetected) + '" ' +
+                                        'alt="' + Functions.escapeHtml(Messages.strLineEndingDetected) + '" ' +
+                                        'class="icon ic_b_help">' +
                                     '</span>' +
                                 '</span>' +
                                 '<select class="line-ending-select">' +
@@ -1603,7 +1605,7 @@ var makeGrid = function (t, enableResize, enableReorder, enableVisib, enableGrid
                         canonicalOldValue = String(Functions.normalizeNewlines(Functions.getCellValue(g.currentEditCell)));
                     }
                     // Normalizing to LFs only for text comparison.
-                    canonicalNewValue = (String(canonicalNewValue)).replace(/\r\n/g, '\n');
+                    canonicalNewValue = Functions.normalizeNewlines(String(canonicalNewValue));
                     isValueUpdated = (canonicalNewValue !== canonicalOldValue) || ($(g.currentEditCell).data('lineEnding') !== $(g.currentEditCell).attr('data-line-ending'));
                 } else {
                     const JSONString = Functions.stringifyJSON(thisFieldParams[fieldName]);
--- a/js/src/makegrid.js
+++ b/js/src/makegrid.js
@@ -931,26 +931,14 @@ var makeGrid = function (t, enableResize, enableReorder, enableVisib, enableGrid
                         );
                         $(g.currentEditCell).data('originalValue', originalValue);
                     }
-                    // Cache it once
-                    $editArea.append(
-                        '<div class="crlf_div">' +
-                            '<label>' +
-                                '<span class="line-ending-label">' +
-                                    Messages.strLineEnding +
-                                    '<span>' +
-                                        '<img src="themes/dot.gif" ' +
-                                        'title="' + Functions.escapeHtml(Messages.strLineEndingDetected) + '" ' +
-                                        'alt="' + Functions.escapeHtml(Messages.strLineEndingDetected) + '" ' +
-                                        'class="icon ic_b_help">' +
-                                    '</span>' +
-                                '</span>' +
-                                '<select class="line-ending-select">' +
-                                    '<option value="LF">LF</option>' +
-                                    '<option value="CRLF">CRLF</option>' +
-                                '</select>' +
-                            '</label>' +
-                        '</div>'
-                    );
+                    var crlfTemplateEl = document.getElementById('crlf-div-template-' + t.dataset.uniqueid);
+                    var $crlfDiv = $(crlfTemplateEl.content.cloneNode(true));
+                    $crlfDiv.find('.line-ending-text').text(Messages.strLineEnding);
+                    $crlfDiv.find('img').attr({
+                        'title': Messages.strLineEndingDetected,
+                        'alt': Messages.strLineEndingDetected
+                    });
+                    $editArea.append($crlfDiv);
                     var detectedEnding = $(g.currentEditCell).data('lineEnding');
                     $editArea.find('.line-ending-select').val(detectedEnding);
                     // handle change of line ending selection
--- a/js/src/makegrid.js
+++ b/js/src/makegrid.js
@@ -750,8 +750,18 @@ var makeGrid = function (t, enableResize, enableReorder, enableVisib, enableGrid
                             $target.html(newHtml);
                         }
                         $thisField.data('originalValue', value);
-                        if ($thisField.data('lineEnding') !== $thisField.attr('data-line-ending')) {
-                            $thisField.attr('data-line-ending', $thisField.data('lineEnding'));
+                        if (value.indexOf('\n') !== -1) {
+                            // Value contains line breaks: ensure the attribute exists
+                            if (!$thisField.is('[data-line-ending]')) {
+                                $thisField.data('lineEnding', 'LF');
+                                $thisField.attr('data-line-ending', 'LF');
+                            } else if ($thisField.data('lineEnding') !== $thisField.attr('data-line-ending')) {
+                                $thisField.attr('data-line-ending', $thisField.data('lineEnding'));
+                            }
+                        } else if ($thisField.is('[data-line-ending]')) {
+                            // Value no longer has line breaks: remove the attribute
+                            $thisField.removeData('lineEnding');
+                            $thisField.removeAttr('data-line-ending');
                         }
                     }
                     if ($thisField.is('.bit')) {
--- a/js/src/makegrid.js
+++ b/js/src/makegrid.js
@@ -1536,7 +1536,8 @@ var makeGrid = function (t, enableResize, enableReorder, enableVisib, enableGrid
 
                 let isValueUpdated;
                 if ($thisField.attr('data-type') !== 'json') {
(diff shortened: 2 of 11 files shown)


liviuconcioiu:
It fixes the issue, except it introduces an old bug #17398. It also happens on empty values, without `NULL`.

https://github.com/user-attachments/assets/e0b89c5a-fb84-40ff-8d5a-eb09b667fd37

Also, I think it should be `Line ending:`, without `s`, since only one line ending can be selected and work at a time. Also, the help icon should be moved after `:`.

og-khushalpatel:
Hi @liviuconcioiu ,
thanks a lot for the review and for pointing these things out.

Just to clarify one point: the change in this PR does not introduce the old bug #17398 which has been demonstrated in the attached video. That said, I agree with your observation that it also occurs on empty values without NULL. I’ve addressed that case now in the latest update.

I’ve also updated the label to Line ending: (singular) and moved the help icon after the colon, as suggested.

Please let me know if you’d like me to adjust anything further. Thanks again for the helpful feedback!
[Screencast from 2026-02-06 19-50-46.webm]
(https://github.com/user-attachments/assets/9e5aadd7-66b5-4eb7-8ebb-e6b338e7abee)

liviuconcioiu:
`UPDATE` is still being triggered:

https://github.com/user-attachments/assets/092fc2da-f084-4c13-8904-1a940ae71d24

og-khushalpatel:
@liviuconcioiu I’ve gone through the scenario again on my end and also shared the recording as proof, but I’m still unable to reproduce the issue you’re observing. To make sure I’m not missing something, could you please share the exact table structure/schema (including column types and any relevant constraints) that you’re using when encountering this problem?
Having the precise structure would really help me set up the same environment locally and understand where the mismatch might be coming from. I’d like to ensure the fix works correctly for your case as well.
Thanks in advance for your help.

liviuconcioiu:
I use `MariaDB 12.2.1` on `Windows`.

```SQL
CREATE TABLE `test` (
  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
  `colA` varchar(255) DEFAULT NULL,
  `colB` varchar(255) NOT NULL,
  `colC` int(10) unsigned DEFAULT NULL,
  `colD` longtext CHARACTER SET utf8mb4 COLLATE utf8mb4_bin DEFAULT NULL CHECK (json_valid(`colD`)),
  `colE` uuid DEFAULT NULL,
  `colF` timestamp(6) NOT NULL DEFAULT current_timestamp(6),
  PRIMARY KEY (`id`)
) ENGINE=MyISAM AUTO_INCREMENT=5 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;

INSERT INTO `test` (`id`, `colA`, `colB`, `colC`, `colD`, `colE`, `colF`) VALUES
(1, 'Lynne', 'Combs', 1, '{\"test\":100}', '7b42b69b-f744-11f0-9d12-806d971fcc98', '2026-01-22 03:43:14.629614'),
(2, 'Carlos', 'Lara', NULL, NULL, NULL, '2026-01-22 03:43:46.727948'),
(3, NULL, 'Lang', 200, '{\"x\":\"y\"}', NULL, '2026-01-22 03:44:20.661601'),
(4, NULL, '', NULL, NULL, 'c0b2cf7a-f744-11f0-9d12-806d971fcc98', '2026-01-22 03:45:11.126818');
```

og-khushalpatel:
Hi @liviuconcioiu, thanks for sharing the exact table structure/schema.
This turned out to be caused by `jQuery’s .data()` cache **auto-casting values** (numbers / JSON objects) on first access, which broke the comparison. Normalizing both sides fixes it.. I’ve now normalized both the canonical old and new values before comparison, which avoids the unnecessary updates you mentioned.
Please let me know if this looks correct from your side.

liviuconcioiu:
Found another problem. If I change the line ending to a cell from LF to CRLF, the cell below will have the same line ending as the first, instead of keeping it.

```SQL
CREATE TABLE `test` (
  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
  `value` text DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=MyISAM AUTO_INCREMENT=3 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;

INSERT INTO `test` (`id`, `value`) VALUES
(1, 'test\ntest'),
(2, 'test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test test <td class=\"text-right\">\n        test test test test test test test test test test test test test test test test test test test test</td>');
```

https://github.com/user-attachments/assets/2cbb01ca-2694-4fd0-82f2-1988bce85939

og-khushalpatel:
> Found another problem. If I change the line ending to a cell from LF to CRLF, the cell below will have the same line ending as the first, instead of keeping it.

Hi @liviuconcioiu, just a quick update! The issue you mentioned is now fixed. It was due to stale event handlers in the grid editor. Would appreciate a quick review when you’re free.

liviuconcioiu:
Everything is okay now!

js/src/makegrid.js
```diff
@@ -687,11 +687,18 @@ var makeGrid = function (t, enableResize, enableReorder, enableVisib, enableGrid
                     if (isNull) {
                         $thisField.find('span').html('NULL');
                         $thisField.addClass('null');
+                        // todo: set original value to null & remove data-line-ending attribute
```
williamdes: can this todo be done ? or removed ?

js/src/makegrid.js
```diff
@@ -687,11 +687,18 @@ var makeGrid = function (t, enableResize, enableReorder, enableVisib, enableGrid
...
+                                .removeData('lineEnding')
+                                .removeAttr('data-line-ending');
+                        }
                     } else {
                         $thisField.removeClass('null');
                         var value = data.isNeedToRecheck
-                            ? data.truncatableFieldValue
-                            : $thisField.data('value');
+                            ? data.truncatableFieldValue.replace(/\r\n/g, '\n')
+                            : $thisField.data('value').replace(/\r\n/g, '\n');
```
williamdes: please call normalizeNewlines

js/src/makegrid.js
```diff
@@ -890,6 +905,41 @@ var makeGrid = function (t, enableResize, enableReorder, enableVisib, enableGrid
...
+                        $(g.currentEditCell).data('originalValue', originalValue);
+                    }
+                    // Cache it once
+                    $editArea.append(
+                        '<div class="crlf_div">' +
+                            '<label>' +
+                                '<span class="line-ending-label">' +
+                                    'Line ending:' +
+                                    '<span>' +
+                                        '<img src="themes/dot.gif" title="The suggested value was detected from your existing data." alt="The suggested value was detected from your existing data." class="icon ic_b_help">' +
```
williamdes: You should not hard code messages, they need to be translated 
Please check other examples in the codebase

js/src/makegrid.js
```diff
@@ -1514,12 +1564,30 @@ var makeGrid = function (t, enableResize, enableReorder, enableVisib, enableGrid
...
+                    let canonicalOldValue;
+                    if ((! $(g.currentEditCell).hasClass('truncated')) && $(g.currentEditCell).data('originalValue') !== null && $(g.currentEditCell).data('originalValue') !== undefined) {
+                        let raw = $(g.currentEditCell).data('originalValue');
+                        canonicalOldValue = (typeof raw === 'object') ? Functions.stringifyJSON(raw) : String(raw);
+                    } else {
+                        // Fallback: DOM value
+                        canonicalOldValue = String(Functions.normalizeNewlines(Functions.getCellValue(g.currentEditCell)));
+                    }
+                    // Normalizing to LFs only for text comparison.
+                    canonicalNewValue = (String(canonicalNewValue)).replace(/\r\n/g, '\n');
```
williamdes: call normalizeNewlines

js/src/makegrid.js
og-khushalpatel: Hi @williamdes,

I have added new translatable strings in JavaScriptMessagesController.php (e.g. __('Line ending:')). After running update-po, the entries were added to .po files but marked as #, fuzzy.

Since I’m new to contributing, I wanted to clarify the expected workflow:

Should I manually resolve/remove the fuzzy entries and include the updated .po files in this PR?
Or should I avoid committing any translation file changes and let the maintainers/translators handle that separately?

liviuconcioiu: You should not. Those are added weekly by cron job. You don't need to add .po files to PR.

og-khushalpatel: Thanks for the clarification.
I’m still learning the project conventions, so I appreciate the guidance. I’ll make sure not to include .po files.

liviuconcioiu:
There are some tests that don't pass:
https://github.com/phpmyadmin/phpmyadmin/actions/runs/22016015278/job/63835593911?pr=20076

CSS needs a little improvement:
https://github.com/phpmyadmin/phpmyadmin/actions/runs/22016015270/job/63835593397?pr=20076

And if you can update the baselines:
https://github.com/phpmyadmin/phpmyadmin/actions/runs/22016015270/job/63835593417?pr=20076

js/src/functions.js
```diff
@@ -4045,16 +4056,28 @@ Functions.getCellValue = function (td) {
...
+        // If string, try to parse then stringify (normalizes formatting)
+        if (typeof value === 'string') {
+            return JSON.stringify(JSON.parse(value), replacer, space);
+        }
+
+        // Fallback: force string
+        return String(value);
     } catch (e) {
-        return json;
+        console.log('error occured in stringifying json');
```
williamdes: console.error exists for that purpose
and not logging the value e does not help

og-khushalpatel: I had put that log when I was debugging the code and it might have slipped from my attention. I will be careful next time.

williamdes:
Due to how big the diff is, I think only 6.0 versions should be fixed. Maybe rebase onto the master branch.
@MauricioFauth what do you think ?

js/src/makegrid.js
```diff
@@ -893,6 +907,44 @@ var makeGrid = function (t, enableResize, enableReorder, enableVisib, enableGrid
...
+                                        'alt="' + Functions.escapeHtml(Messages.strLineEndingDetected) + '" ' +
+                                        'class="icon ic_b_help">' +
+                                    '</span>' +
+                                '</span>' +
+                                '<select class="line-ending-select">' +
+                                    '<option value="LF">LF</option>' +
+                                    '<option value="CRLF">CRLF</option>' +
+                                '</select>' +
+                            '</label>' +
+                        '</div>'
```
williamdes: The HTML element https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/template could help you.
And so you only would have to clone the element and add text in the elements
Avoiding the potential HTML injection issue.
Remember that `strLineEnding` could have special chars, even if that makes no sense at all.

og-khushalpatel:
Hi @williamdes @MauricioFauth @liviuconcioiu  , I’m following up on this PR in case it got buried. Please let me know if any changes are needed from my side. Thank you!

MoonE:
@og-khushalpatel Please look into the failing test cases.

og-khushalpatel:
> There are some tests that don't pass: https://github.com/phpmyadmin/phpmyadmin/actions/runs/22016015278/job/63835593911?pr=20076
> 
> CSS needs a little improvement: https://github.com/phpmyadmin/phpmyadmin/actions/runs/22016015270/job/63835593397?pr=20076
> 
> And if you can update the baselines: https://github.com/phpmyadmin/phpmyadmin/actions/runs/22016015270/job/63835593417?pr=20076


@MoonE I believe I have already fixed the tests that were previously failing due to the changes in this PR. 
There are still some failing tests, but they appear to be unrelated to the changes introduced in this PR. If you notice any tests failing specifically because of this PR, please point them out and I’ll be happy to address them

libraries/classes/Display/Results.php
```diff
@@ -4564,6 +4606,7 @@ private function getPartialText($str): array
             $truncated,
             $str,
             $originalLength,
+            $lineEnding,
```
MoonE: I believe four of the five failing tests are complaining about the additional value in this array.

MoonE (commented):
Haven't looked what causes the other failure.
But I don't see it failing in the master branch, so it has to be related to these changes.

Also the amount of added lines in `phpstan-baseline.neon` suggests you are running a different configuration than on the server, shouldn't be this many changes.

libraries/classes/Display/Results.php
```diff
@@ -4532,9 +4541,41 @@ private function getRowData(
...
+     *
+     * This is used to preserve line-ending semantics of textual data,
+     * especially when values may later be truncated or normalized
+     * for display or editing.
+     *
+     * @see getPartialText()
+     *
+     * @param string $value The string whose line endings should be detected
+     *
+     * @return string|null  Returns 'CRLF', 'LF', or null if no line breaks exist
```
MoonE:
```suggestion
     * @psalm-return 'CRLF'|'LF'|null
```

libraries/classes/Display/Results.php
```diff
@@ -3373,11 +3374,12 @@ private function getDataCellForNonNumericColumns(
...
-                $isFieldTruncated,
-                $column,
-                $originalLength,
-            ] = $this->getPartialText($column);
+                        [
+                            $isFieldTruncated,
+                            $column,
+                            $originalLength,
+                            $lineEnding,
+                        ] = $this->getPartialText($column);
```
MoonE:
```suggestion
            [
                $isFieldTruncated,
                $column,
                $originalLength,
                $lineEnding,
            ] = $this->getPartialText($column);
```

libraries/classes/Display/Results.php
MoonE, line 4585: Remove the `@return array` as it is redundant, and replace the `@psalm-return` with
```php
    * @psalm-return array{bool, string, int, 'CRLF'|'LF'|null}
```

MoonE (commented):
I added some comments on how to resolve the linter problems.

test/classes/Display/ResultsTest.php
```diff
@@ -1583,25 +1591,25 @@ public function testGetTable(): void
...
                 . '</tr>' . "\n"
                 . '<tr><td data-decimals="0" data-type="real" class="'
-                . 'text-end data not_null text-nowrap">3</td>' . "\n"
+                . 'text-end data not_null text-nowrap" data-original-value="3">3</td>' . "\n"
                 . '<td data-decimals="0" data-type="string" data-originallength="4" class="'
-                . 'data not_null text pre_wrap">Abcd</td>' . "\n"
+                . 'data not_null text pre_wrap" data-original-value="Abcd">Abcd</td>' . "\n"
                 . '<td data-decimals="0" data-type="datetime" data-originallength="19" class="'
-                . 'data not_null datetimefield text-nowrap">2012-01-20 02:00:02</td>' . "\n"
+                . 'data not_null datetimefield text-nowrap" data-original-value="2012-01-20&#x20;02&#x3A;00&#x3A;02">2012-01-20 02:00:02</td>' . "\n"
```
MoonE: Split the lines 1612, 1605, and 1598 so all lines have 120 characters or less.
https://github.com/phpmyadmin/phpmyadmin/actions/runs/22819658489/job/66190316004?pr=20076

libraries/classes/Display/Results.php
```diff
@@ -4437,6 +4441,11 @@ private function getRowData(
             $transformationPlugin !== null
         );
 
+        $canonicalValue = null;
+        if (is_string($data)) {
+            $canonicalValue = str_replace(["\r\n", "\r"], "\n", $data);
+        }
```
MoonE: `$data` is always type string here, so the check can be removed.
```suggestion
        $canonicalValue = str_replace(["\r\n", "\r"], "\n", $data);
```

og-khushalpatel: > `$data` is always type string here, so the check can be removed.

I was verifying whether the type of `$data` is `string` because it was not type-hinted in the arguments list. As a result, I thought it was essential for me to verify that. Anyways I have added a type hint for the `$data` in argument-list so we can remove `is_string()` check.

og-khushalpatel:
@MoonE I believe that finally all pipeline failures have been fixed except for one which is not manifesting because of this PR-related changes. I am really grateful for you helped me understand pipeline failure debugging which was a bit scary for a beginner like me. Let me know if I should notify the maintainers for review.

libraries/classes/Controllers/JavaScriptMessagesController.php
```diff
@@ -570,6 +570,10 @@ private function setMessages(): void
             ),
             'strOriginalLength' => __('Original length'),
 
+            /* For makegrid.js (CRLF/LF handling) */
+            'strLineEnding' => __('Line ending:'),
+            'strLineEndingDetected' => __('The suggested value was detected from your existing data.'),
+
```
MoonE: This can be removed and set directly in the html template in `templates/display/results/table.twig`.

js/src/makegrid.js
```diff
@@ -907,6 +921,32 @@ var makeGrid = function (t, enableResize, enableReorder, enableVisib, enableGrid
...
+                        );
+                        $(g.currentEditCell).data('originalValue', originalValue);
+                    }
+                    var crlfTemplateEl = document.getElementById('crlf-div-template-' + t.dataset.uniqueid);
+                    var $crlfDiv = $(crlfTemplateEl.content.cloneNode(true));
+                    $crlfDiv.find('.line-ending-text').text(Messages.strLineEnding);
+                    $crlfDiv.find('img').attr({
+                        'title': Messages.strLineEndingDetected,
+                        'alt': Messages.strLineEndingDetected
+                    });
```
MoonE: Set the text in the twig file, and remove these lines.

MoonE (commented):
If a user edits a text value that at the begin has no line breaks, it's not possible to select the line breaks, even after closing and re-opening the edit box. It only shows up after a refresh of the whole browse result.

Would be great if it becomes visible after a new line is entered.

og-khushalpatel:
> If a user edits a text value that at the begin has no line breaks, it's not possible to select the line breaks, even after closing and re-opening the edit box. It only shows up after a refresh of the whole browse result.
> 
> Would be great if it becomes visible after a new line is entered.

Implemented the code to make the line ending selector visible after a newline is entered. Just check it out:

[Screencast from 2026-03-10 19-00-45.webm](https://github.com/user-attachments/assets/38ac2647-5c80-40bb-856a-a83809ba68c4)

og-khushalpatel:
Hi, I looked into the CI failure and I believe it's unrelated to my changes. The error occurs in the stop-local-server cleanup script:

`cat: /tmp/last_temp_dir_phpMyAdminTests: No such file or directory`

My changes only move i18n strings from `JavaScriptMessagesController.php` into the Twig template and adjust JS logic in `makegrid.js` — none of which affect the Selenium server startup (php-fpm/nginx). This looks like a CI environment issue.
Could we re-run the failed job to confirm?

libraries/classes/Display/Results.php
```diff
@@ -4532,9 +4537,41 @@ private function getRowData(
...
+     *
+     * @see getPartialText()
+     *
+     * @param string $value The string whose line endings should be detected
+     *
+     * @psalm-return 'CRLF'|'LF'|null
+     */
+    private function detectLineEnding(string $value): ?string
+    {
+        if (strpos($value, "\r\n") !== false) {
```
kamil-tekiela: Can you not use `str_contains()`?

libraries/classes/Display/Results.php
```diff
@@ -4532,9 +4537,41 @@ private function getRowData(
...
+            return 'CRLF';
+        }
+
+        if (strpos($value, "\n") !== false) {
+            return 'LF';
+        }
+
+        return null;
```
kamil-tekiela: Why does it have to be `null` and not `''`?

libraries/classes/Display/Results.php
```diff
@@ -4564,6 +4601,7 @@ private function getPartialText($str): array
             $truncated,
             $str,
             $originalLength,
+            $lineEnding,
```
kamil-tekiela: Why do you pass the to the other function instead of getting it where it's needed?

MoonE:
Is there a reason why the line ending detection is done server-side and why an additional original value has to be included in the html?

There is existing logic loading the full text when it is displayed truncated, so it would make sense to get the data only when it is needed. Otherwise this can inflate the html size by a lot, which can be a problem when you have columns containing very long strings.

Maybe check here if the cell contains a `<br>` and then use the response value to initialize the line-break select?
https://github.com/phpmyadmin/phpmyadmin/blob/44da39fe1ff0dd1bd059456ead690fe2c142cfec/js/src/makegrid.js#L1081-L1108
This way there is no data added during normal browsing operation.

libraries/classes/Display/Results.php
```diff
@@ -4416,14 +4419,15 @@ private function getRowData(
         array $analyzedSqlResults,
         FieldMetadata $meta,
         array $map,
-        $data,
-        $displayedData,
+        string $data,
+        string $displayedData,
```
kamil-tekiela, line 4423: I don't think this should be part of the bug fix

og-khushalpatel: Normalizing line-breaks to `\n` in `$data` requires performing string operations when initializing `$canonicalValue`. For that reason, I wanted to ensure `$data` is always treated as a `string`.

Initially, I handled this with an `is_string()` check because `$data` wasn't type-hinted. However, since `$data` is already documented as `@param string $data`, the linter flagged the conditional as unnecessary.

To resolve that while still ensuring type safety, I type-hinted `$data` as `string` instead. That's why I included it as part of the bug fix.

og-khushalpatel:
> Is there a reason why the line ending detection is done server-side

Great Question! Yes, this is intentional. According to the HTML spec, `<textarea>` normalizes all line breaks to `\n` when returning its value, even if the original database value contains `\r\n`. This normalization was the root cause of the false "value changed" detection on the client side (i.e., #19368 ), since the client would always see a difference between the textarea value and the original, triggering an unnecessary save.
By detecting line endings server-side, we preserve the actual line-break style stored in the database and can compare accurately.

> why an additional original value has to be included in the html?

We need the server-provided original value as a source of truth to prevent `\n` duplication on each save cycle. Without it, when converting between `LF` and `CRLF` (or vice versa), the `<br>` tags were being duplicated; leading to exponential growth of line breaks in the stored value with every update. Including the original value lets us diff correctly against what the database actually holds, rather than relying on the textarea's already-normalized content.

> Otherwise this can inflate the html size by a lot, which can be a problem when you have columns containing very long strings

That's a great point! To address this concern, the HTML size won't actually inflate, since for full-text database columns, I'm storing the **truncated value** in the `data-original-value` attribute rather than the complete column content. So the overhead remains minimal regardless of how long the underlying strings are.You can verify it in the following video:
[Screencast from 2026-03-11 12-01-22.webm](https://github.com/user-attachments/assets/52cde4ca-32c2-4331-ade2-cfc360d5aa27)

Issue #20068, opened 2026-02-01
[6.0]: Exported event contains html entities


liviuconcioiu:
### Describe the bug

When an event is exported the query contains html entities and will result in an invalid query if it is manually run.

### How to Reproduce

1. Go to `Events`
2. Export an event
3. See the query 

### Screenshots

https://github.com/user-attachments/assets/4ede6dff-b94b-4d81-9263-be52aa7f73dc

### phpMyAdmin version

6.0.0-dev

bryanhoffman:
Newbie here, but I saw "good first issue" so, I want to try and be helpful.

I traced this back to: https://github.com/phpmyadmin/phpmyadmin/blob/9579895a860a4d779152c5408208b010be22f5fe/src/Controllers/Database/EventsController.php#L205

htmlspecialchars("'") will result in: `&#039;`

Removing htmlspecialchars() would fix this exact issue, but it's probably there for good reason. I think it'd be naive to suggest removing it, but using the flag ENT_COMPAT would leave those single quotes alone.^1

I hope I was helpful. If not, I'm really just here to learn.

^1 - https://www.php.net/manual/en/function.htmlspecialchars.php

bryanhoffman:
Hey sorry, I obviously don't know what I'm doing. Was QA_5_2 the right branch to submit the pull request to?

liviuconcioiu:
You should make the change on `master`. `QA_5_2` is not affected.

bryanhoffman:
Thanks, liviuconcioiu.

williamdes:
Triggers do also have this bug, fixing this while merging

PR #20079, opened 2026-02-06
Attempting to fix "Exported event contains html entities #20068"

bryanhoffman:
### Description

Totally a newbie, but trying to make some contributions and be helpful. I changed the htmlspecialchars to use ENT_NOQUOTES flag. This should solve issue #20068 

Fixes #

`'` in $exportData won't get turned into `&#039;` now. I think that it's still secure since at least as I understand a textarea only become insecure when </textarea> is in its contents, and this only changes how single and double quotes are treated. 

Forgive me for missing this in the commit message: Signed-off-by: Bryan Hoffman <bryanhoffman1@gmail.com>


williamdes:
Duplicate of #20080

williamdes:
No need for a QA version of this fix as it does not affect 5.2

bryanhoffman:
My bad, thanks williamdes.

PR #20081, opened 2026-02-06
Fix buffer not closed and toon default indent

faissaloux:


williamdes (approved):
Thank you @faissaloux  ! 💯

PR #20084, opened 2026-02-07
Improve detectSize

liviuconcioiu:
There's no need to get the length again, since already does it in https://github.com/phpmyadmin/phpmyadmin/blob/e76adf1911be353b525add2b58e24232084e4f6e/src/Import/Import.php#L480


williamdes (approved):
I guess the tests also agree with you

PR #20099, opened 2026-02-12
Refactoring of Properties

kamil-tekiela:
Removed some unnecessary polymorphism. The base class had two methods which were not used anywhere. The two property types share no common functionality, and therefore, the base class is unneeded. All types were properly narrowed according to what is expected by the code.


williamdes (approved):

kamil-tekiela:
FYI this is the commit that introduced this hierarchy https://github.com/phpmyadmin/phpmyadmin/commit/5609a90c029abc4d1f07b51c588450a21e7b3ea3

MauricioFauth (approved):

PR #20117, opened 2026-02-17
Fix #20116 - Make sure fullscreen is enabled

liviuconcioiu:
### Description

https://github.com/user-attachments/assets/643b3600-6930-4e86-ad7d-a87977a9b869

Fixes #20116

Before submitting pull request, please review the following checklist:

- [x] Make sure you have read our [CONTRIBUTING.md](https://github.com/phpmyadmin/phpmyadmin/blob/master/CONTRIBUTING.md) document.
- [x] Make sure you are making a pull request against the correct branch. For example, for bug fixes in a released version use the corresponding QA branch and for new features use the _master_ branch. If you have a doubt, you can ask as a comment in the bug report or on the mailing list.
- [x] Every commit has proper `Signed-off-by` line as described in our [DCO](https://github.com/phpmyadmin/phpmyadmin/blob/master/DCO). This ensures that the work you're submitting is your own creation.
- [x] Every commit has a descriptive commit message.
- [x] Every commit is needed on its own, if you have just minor fixes to previous commits, you can squash them.
- [ ] Any new functionality is covered by tests.

PR #20119, opened 2026-02-17
Fix #20116 - Check also for WebKit

liviuconcioiu:
https://github.com/phpmyadmin/phpmyadmin/pull/20117#issuecomment-3914481452

### DIFF
--- a/js/src/designer/move.js
+++ b/js/src/designer/move.js
@@ -556,7 +556,11 @@ DesignerMove.toggleFullscreen = function () {
     var $content = $('#page_content');
     const pageContent = document.getElementById('page_content');
 
-    var fullscreenEnabled = document.fullscreenEnabled || document.webkitFullscreenEnabled;
+    var fullscreenEnabled =
+        document.fullscreenEnabled ||
+        document.webkitFullscreenEnabled ||
+        document.mozFullScreenEnabled ||
+        document.msFullscreenEnabled;
 
     if (! fullscreenEnabled) {
         Functions.ajaxShowMessage(Messages.strFullscreenRequestDenied, null, 'error');
--- a/js/src/designer/move.js
+++ b/js/src/designer/move.js
@@ -556,7 +556,9 @@ DesignerMove.toggleFullscreen = function () {
     var $content = $('#page_content');
     const pageContent = document.getElementById('page_content');
 
-    if (! document.fullscreenEnabled) {
+    var fullscreenEnabled = document.fullscreenEnabled || document.webkitFullscreenEnabled;
+
+    if (! fullscreenEnabled) {
         Functions.ajaxShowMessage(Messages.strFullscreenRequestDenied, null, 'error');
 
         return;
--- a/js/src/designer/move.js
+++ b/js/src/designer/move.js
@@ -568,8 +568,8 @@ DesignerMove.toggleFullscreen = function () {
     var $content = $('#page_content');
     const pageContent = document.getElementById('page_content');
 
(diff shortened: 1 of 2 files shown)


liviuconcioiu:
On Firefox 59 I got this error: `TypeError: pageContent.requestFullscreen is not a function.`

Now it also supports Firefox <=60 and also add a check for iOS (not tested but should be available - https://developer.apple.com/documentation/webkitjs/document/1630333-webkitfullscreenenabled?changes=_7)

I had to disable eslint for that line. I did not encounter any error on Firefox 59 regarding it.

js/src/designer/move.js
```diff
@@ -549,14 +549,33 @@ DesignerMove.rect = function (x1, y1, w, h, color) {
...
+
 DesignerMove.toggleFullscreen = function () {
     var valueSent = '';
     var $img = $('#toggleFullscreen').find('img');
     var $span = $img.siblings('span');
     var $content = $('#page_content');
     const pageContent = document.getElementById('page_content');
 
-    if (! document.fullscreenEnabled) {
+    var fullscreenEnabled =
```
williamdes: Can this go into a function too?

liviuconcioiu:
@williamdes I see Firefox reports this message, if the request was denied:

`Request for fullscreen was denied because Fullscreen API is disabled by user preference.`

Should I change the `strFullscreenRequestDenied` message with this one to be more detailed?

williamdes:
> @williamdes I see Firefox reports this message, if the request was denied:
> 
> `Request for fullscreen was denied because Fullscreen API is disabled by user preference.`
> 
> Should I change the `strFullscreenRequestDenied` message with this one to be more detailed?

Indeed that would be better
I will also email you my BrowserStack credentials so you can play for free with it

liviuconcioiu:
> On Firefox 59 I got this error: `TypeError: pageContent.requestFullscreen is not a function.`

https://reports.phpmyadmin.net/reports/view/88152

MoonE:
Is there a need to support browsers this old?

I want to use [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver#browser_compatibility) for the GIS visualization (in master branch) and have to drop some older browser support for this.

liviuconcioiu:
I think for master is fine, but that's my opinion. @williamdes can tell you this.

williamdes (approved):
Thanks !

williamdes:
> I think for master is fine, but that's my opinion. @williamdes can tell you this.

You have to refer to https://github.com/phpmyadmin/phpmyadmin/blob/QA_5_2/.browserslistrc
and https://github.com/phpmyadmin/phpmyadmin/blob/master/.browserslistrc 
to know what to browsers to support.
I think we mostly follow what our bootstrap version supports
cc @MauricioFauth

libraries/classes/Controllers/JavaScriptMessagesController.php
```diff
@@ -710,7 +710,9 @@ private function setMessages(): void
             'strHide' => __('Hide'),
             'strShow' => __('Show'),
             'strStructure' => __('Structure'),
-            'strFullscreenRequestDenied' => __('The fullscreen request was denied.'),
+            'strFullscreenRequestDenied' => __(
+                'Request for fullscreen was denied because Fullscreen API is disabled by user preference.'
```
williamdes, line 714:
```suggestion
                'The fullscreen request was denied because the Fullscreen API is disabled by user preference in the browser.'
```

PR #20123, opened 2026-02-17
fix: remove dead Suhosin warning from HomeController

### LINKED ISSUE #19912: [Feat]: Remove Suhosin support and warnings
### phpMyAdmin version

5.2.3

### Describe the feature

Suhosin has not been available for quite a while. We can probably remove any code that addresses Sohosin, including `$cfg['SuhosinDisableWarning']`.

### Describe the solution

The code can be removed

### Describe the alternatives

_No response_

### Additional context

_No response_

somethingwithproof:
## Summary
- Remove dead Suhosin warning code from `HomeController.php`
- Suhosin was abandoned and never supported PHP 8+; phpMyAdmin requires PHP 8.2+, so this check can never trigger
- Refs #19912

## Test plan
- [ ] Verify home page loads without errors
- [ ] Run `phpstan` and `phpcs` to confirm no regressions

Fixes #19912

### DIFF
--- a/resources/po/en_GB.po
+++ b/resources/po/en_GB.po
@@ -6,7 +6,7 @@ msgstr ""
 "Project-Id-Version: phpMyAdmin 6.0.0-dev\n"
 "Report-Msgid-Bugs-To: translators@phpmyadmin.net\n"
 "POT-Creation-Date: 2026-02-22 00:36+0000\n"
-"PO-Revision-Date: 2026-02-17 13:09+0000\n"
+"PO-Revision-Date: 2026-02-23 02:09+0000\n"
 "Last-Translator: Andi Chandler <andi@gowling.com>\n"
 "Language-Team: English (United Kingdom) <https://hosted.weblate.org/projects/"
 "phpmyadmin/master/en_GB/>\n"
@@ -12432,7 +12432,7 @@ msgstr "Incorrect format parameter"
 
 #: src/Controllers/Import/ImportController.php:254
 msgid "Sorry an unexpected error happened!"
-msgstr ""
+msgstr "Sorry an unexpected error happened!"
 
 #: src/Controllers/Import/ImportController.php:301
 #: src/Controllers/Import/ImportController.php:516
@@ -13969,7 +13969,7 @@ msgstr "Hide"
 
 #: src/Controllers/JavaScriptMessagesController.php:695
 msgid "The fullscreen request was denied."
-msgstr ""
+msgstr "The fullscreen request was denied."
 
 #: src/Controllers/JavaScriptMessagesController.php:697
 msgctxt "Month name"


kamil-tekiela:
I think the setting should be removed too. It was probably the only place it was used and there's no reason to leave an orphaned setting.

somethingwithproof:
Good catch — updated to also remove the SuhosinDisableWarning setting from the config. All references cleaned up.

docs/config.rst
```diff
@@ -117,15 +117,6 @@ Basic settings
     Whether to log successful authentication attempts into
     :config:option:`$cfg['AuthLog']`.
 
-.. config:option:: $cfg['SuhosinDisableWarning']
```
williamdes, line 120: Please keep the documentation and add a versionremoved statement

docs/faq.rst
```diff
@@ -590,8 +590,6 @@ To further improve security, we also recommend these modifications:
 * `suhosin.executor.disable_emodifier <https://suhosin5.suhosin.org/stories/config
   uration.html#suhosin-executor-disable-emodifier>`_ should be enabled.
 
-You can also disable the warning using the :config:option:`$cfg['SuhosinDisableWarning']`.
```
williamdes, line 593: Why remove this and not the entire block?
I'd say remove nearly everything (contents but not header)
and say that it was removed

src/Controllers/HomeController.php
```diff
@@ -360,26 +360,6 @@ private function checkRequirements(): void
             ];
         }
 
-        /**
```
williamdes, line 363: @liviuconcioiu i do not remember how can we suggest to the user to remove a deprecated setting?
could we have an array somewhere?

liviuconcioiu: We can show a notice in footer, like Twig does for example.

deprecated

And here is the code that should go in `HomeController`:

```PHP
        $deprecatedSettings = [
            'SuhosinDisableWarning' => '6.0.0-dev',
        ];

        foreach ($deprecatedSettings as $key => $version) {
            if (array_key_exists($key, $this->config->settings)) {
                $this->errors[] = [
                    'message' => sprintf(
                        __('Since phpMyAdmin %s: "$cfg[\'%s\']" is deprecated. You can safely remove it.'),
                        $version,
                        $key,
                    ),
                    'severity' => 'notice',
                ];
            }
        }
```

somethingwithproof:
Makes sense — I'll keep the docs with a versionremoved note, trim the FAQ section instead of deleting it, and add the deprecation notice like @liviuconcioiu showed. Will update the PR.

somethingwithproof:
Addressed reviewer feedback:

- **docs/config.rst**: Restored `SuhosinDisableWarning` section with `versionremoved:: 6.0.0` directive
- **docs/faq.rst**: Kept FAQ reference with removal note instead of deleting entirely
- **HomeController.php**: Added deprecation notice using the `$deprecatedSettings` pattern (per @liviuconcioiu's suggestion) — users with the setting in config.inc.php will see a notice to remove it

All unit tests pass locally (5183 tests, 17280 assertions).

src/Controllers/HomeController.php
MauricioFauth: @liviuconcioiu I think this code is too generic. Each setting should have it's own message.

src/Controllers/HomeController.php
```diff
@@ -360,24 +360,21 @@ private function checkRequirements(): void
...
-                        'Server running with Suhosin. Please refer to %sdocumentation%s for possible issues.',
+        $deprecatedSettings = [
+            'SuhosinDisableWarning' => '6.0.0-dev',
+        ];
+
+        foreach ($deprecatedSettings as $key => $version) {
+            if (array_key_exists($key, $this->config->settings)) {
+                $this->errors[] = [
+                    'message' => sprintf(
+                        __('Since phpMyAdmin %s: "$cfg[\'%s\']" is deprecated. You can safely remove it.'),
```
MauricioFauth: Please, do not add this. It's out of the scope of this pull request.

docs/config.rst
```diff
@@ -122,9 +122,10 @@ Basic settings
     :type: boolean
     :default: false
 
-    A warning is displayed on the main page if Suhosin is detected.
+    .. deprecated:: 6.0.0
 
-    You can set this parameter to ``true`` to stop this message from appearing.
+    This setting has been removed. The Suhosin extension is no longer
+    maintained, and the related warning has been removed from phpMyAdmin.
```
MauricioFauth: Please do not remove the original description. Leave it after the removed message.

MauricioFauth (changes requested):

somethingwithproof:
Rebased and addressed all three points:

- Dropped the generic `$deprecatedSettings` loop — you're right, it's out of scope here.
- Kept the original description text in `docs/config.rst` below the `versionremoved:: 6.0.0` directive.
- Updated `docs/faq.rst` to note the removal rather than deleting the section.

Single clean commit now, only Suhosin-related changes.

williamdes:
Duplicate of #20167

williamdes:
Please avoid re creating pull requests. `git push --force` exists to reset everything

PR #20127, opened 2026-02-18
Replace getRealSize with ini_parse_quantity

### LINKED ISSUE #20115: [Bug]: The float 6.7108864E+20 is not representable as an int, cast occurred
### Describe the bug

**5.2.4-dev:**

```
Warning in .\libraries\classes\Core.php#268
The float 6.7108864E+20 is not representable as an int, cast occurred
```

**6.0.0-dev:**

```
Internal error in .\src\Core.php#148
ErrorException: The float 6.7108864E+20 is not representable as an int, cast occurred
```

### How to Reproduce

Set this in `php.ini`:

```INI
post_max_size = 640000000000000M
```

### phpMyAdmin version

5.2.4-dev, 6.0.0-dev

### Additional context

https://reports.phpmyadmin.net/incidents/view/1955115

https://github.com/phpmyadmin/phpmyadmin/blob/733bc6bf2f8a3bb89e097ed5e8e3a52c02106a32/libraries/classes/Core.php#L268

kamil-tekiela:
Fixes #20115

There are two commits here. First removes the obsolete function and replaces it with `ini_parse_quantity`. The second fixes the mentioned bug, although I think not fixing the bug would also be ok.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1978,12 +1978,6 @@ parameters:
 			path: src/Controllers/Import/ImportController.php
 
 		-
-			message: '#^Casting to string something that''s already string\.$#'
-			identifier: cast.useless
-			count: 1
-			path: src/Controllers/Import/ImportController.php
-
-		-
 			message: '#^Comparison operation "\<\=" between 0 and int\<1, max\> is always true\.$#'
 			identifier: smallerOrEqual.alwaysTrue
 			count: 1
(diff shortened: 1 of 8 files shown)


williamdes (changes requested):
I would have prefered to keep the function in core and the tests
even if the function is a wrapper for ini parse quantity

MauricioFauth (approved):

williamdes (approved):
Thank you !

PR #20129, opened 2026-02-18
Make string translatable

liviuconcioiu:


kamil-tekiela:
When is that error ever shown to the user?

liviuconcioiu:
Looking at the code, is a default value when a bookmark is saved !? I can't remember if I seen it, since I always use English.

kamil-tekiela:
It's not very descriptive. I would not translate it. I would rather refactor the code to make this message unnecessary. But I think that's a bit difficult.

liviuconcioiu:
It's on `QA_5_2`. So, refactoring will be neede on master, to avoid any side effects.

liviuconcioiu:
I think is dead code? Because I can't save multiple queries from `SQL` tab, and that code refers to multiple queries.

MauricioFauth (approved):

PR #20135, opened 2026-02-19
Create a oneliner Session::getToken()

kamil-tekiela:
- Fixes https://github.com/phpmyadmin/phpmyadmin/issues/20112

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -6991,12 +6991,6 @@ parameters:
 			path: src/Http/Middleware/StatementHistory.php
 
 		-
-			message: '#^Parameter \#1 \$known_string of function hash_equals expects string, mixed given\.$#'
-			identifier: argument.type
-			count: 1
-			path: src/Http/Middleware/TokenRequestParamChecking.php
-
-		-
 			message: '#^Method PhpMyAdmin\\Http\\ServerRequest\:\:getAttributes\(\) return type has no value type specified in iterable type array\.$#'
 			identifier: missingType.iterableValue
 			count: 1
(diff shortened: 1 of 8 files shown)


liviuconcioiu:
This fixes the problem.

kamil-tekiela:
It shouldn't. I didn't change any logic.

liviuconcioiu:
I see that in the code, but the issue got away.

liviuconcioiu:
If you add this code in getJsParams, the session is there. Before this, the session was empty. Somehow is related to that PR.

```
        var_dump(Session::getToken());
        exit;
```

string(32) "615267496f747e22553346552b3b5051"

liviuconcioiu:
@kamil-tekiela you were right.

$_SESSION is empty even with this PR, but getToken returns the token.

```PHP
        var_dump($_SESSION);
        exit;
```

liviuconcioiu:
Refreshing the logout page, returns the session:


https://github.com/user-attachments/assets/ecf8f7c9-de60-41d6-8438-572b6652f9d1

kamil-tekiela:
@MauricioFauth Git bisect shows that the test failure is due to 2e1c947d4268fb4fab874bc97c5f66468b2995fc but I am not sure why. Any ideas?

MauricioFauth (approved):

MauricioFauth:
This fixes the issue #20112, because `Session::getToken()` checks if the array key `' PMA_token '` exists. There's no issue in using an empty string as token for that part of the logout logic, as it is irrelevant. It doesn't need the HTML header to do a redirection.

PR #20143, opened 2026-02-20
Fix color for PHP/JSON/SQL code

liviuconcioiu:
### Description

Fix color for PHP code. Since `<code>` is inside `<pre>` (https://github.com/phpmyadmin/phpmyadmin/commit/05892fb9239b68f19366cdec50a73d0baeb797fe), PHP code is no longer colored.

**Before:**

before

**After:**

after

### DIFF
--- a/public/themes/bootstrap/scss/_codemirror.scss
+++ b/public/themes/bootstrap/scss/_codemirror.scss
@@ -11,7 +11,7 @@ $textarea-rows: 15;
 }
 
 .CodeMirror-gutters {
-  border-right-color: var(--#{$prefix}border-color);;
+  border-right-color: var(--#{$prefix}border-color);
   background-color: rgba(var(--#{$prefix}body-color-rgb), .03);
 }
 
--- a/public/themes/bootstrap/scss/_common.scss
+++ b/public/themes/bootstrap/scss/_common.scss
@@ -2366,6 +2366,7 @@ pre {
   margin: 0;
 }
 
+code.json,
 code.php,
 code.sql {
   color: var(--#{$prefix}code-color);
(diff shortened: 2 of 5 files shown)


kamil-tekiela:
Do people actually prefer the pink color? I thought it was harder to read.

liviuconcioiu:
> Do people actually prefer the pink color? I thought it was harder to read.

I don't know. I just added back like it was in QA_5_2. Probably we should add in the future the codemirror for PHP? not sure if is worth it.

https://codemirror.net/5/mode/php/index.html

kamil-tekiela:
I mean I purposefully didn't bring that old color over, but if you think it's better to keep the pink color, I am ok with that. Regarding PHP CodeMirror, I think it's not worth it. Mainly because I don't think the "show PHP code" is a useful feature.

MauricioFauth:
> Do people actually prefer the pink color? I thought it was harder to read.

That's the default color that Bootstrap framework uses for `<code>`, and the idea of the Bootstrap theme is to have as little custom changes as possible, so it's easier to create a new theme based on it.

liviuconcioiu:
I noticed that when `$cfg['CodemirrorEnable'] = false;`, sql was no longer colored, like in QA_5_2. I added that too.

williamdes (approved):

williamdes:
> That's the default color that Bootstrap framework uses for `<code>`, and the idea of the Bootstrap theme is to
> have as little custom changes as possible, so it's easier to create a new theme based on it.


I agree

liviuconcioiu:
I added for `json` too. If `Browser display transformation` is set to `JSON` for a `JSON` column, wasn't colored.

before

after

MauricioFauth (approved):

PR #20145, opened 2026-02-20
Add str_replace(DIRECTORY_SEPARATOR, '/'

kamil-tekiela:
Makes it so that the path reported on Windows and Linux is the same.


williamdes (approved):
This could even go to QA

MauricioFauth (approved):

PR #20147, opened 2026-02-20
Clean up setUserValue

kamil-tekiela:
The goal is to try and separate the Config and UserPreferences. `lang` and Export preferences have not used the Config so that's why I decided to split that method in two. It is now a lot clearer which preferences affect the Config.


williamdes (approved):

MauricioFauth (approved):
Thank you!

PR #20148, opened 2026-02-20
Inline some methods from AbstractTestCase

MauricioFauth:


kamil-tekiela (approved):

williamdes (approved):
I guess that will be easier to analyse

PR #20154, opened 2026-02-21
Fix gis export to pdf and svg

MoonE:
### Description

In exported svg files geometries have class names according to their geometry type, e. g. `point`, `linestring`, and `polygon`, but also `multipoint`, and `multipolygon`, only for MultiLineString it was `linestring` instead `multilinestring`.
This class is not used by phpmyadmin, so it doesn't really matter...

In exported pdf files each linestring segment was drawn with a new call to the `Line` function which caused each segment to have a line cap, drawing as a polyline fixes this.

```sql
SELECT ST_GeomFromText('LINESTRING(120 20,170 60,140 10)') AS geom UNION ALL
SELECT ST_GeomFromText('MULTILINESTRING((120 60,165 60,120 59),(170 10,170 55,160 10),(120 62,172 62,172 10))');
```
Visualize GIS data -> Save -> PDF

Before pdf:
image

After pdf:
image

### DIFF
--- a/libraries/classes/Gis/GisLineString.php
+++ b/libraries/classes/Gis/GisLineString.php
@@ -136,12 +136,14 @@ class GisLineString extends GisGeometry
         $red = hexdec(mb_substr($line_color, 1, 2));
         $green = hexdec(mb_substr($line_color, 3, 2));
         $blue = hexdec(mb_substr($line_color, 4, 2));
-        $line = [
-            'width' => 1.5,
-            'color' => [
-                $red,
-                $green,
-                $blue,
+        $lineStyle = [
+            'all' => [
+                'width' => 1.5,
+                'color' => [
+                    $red,
+                    $green,
+                    $blue,
+                ],
             ],
         ];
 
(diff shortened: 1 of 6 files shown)


libraries/classes/Gis/GisLineString.php
```diff
@@ -136,33 +136,18 @@ public function prepareRowAsPdf($spatial, ?string $label, $line_color, array $sc
...
         $blue = hexdec(mb_substr($line_color, 4, 2));
-        $line = [
-            'width' => 1.5,
-            'color' => [
-                $red,
-                $green,
-                $blue,
-            ],
-        ];
+        $lineStyle = ['all' => ['width' => 1.5, 'color' => [$red, $green, $blue]]];
```
williamdes: Can you make it look like before please

libraries/classes/Gis/GisMultiLineString.php
```diff
@@ -153,14 +153,7 @@ public function prepareRowAsPdf($spatial, ?string $label, $line_color, array $sc
...
         $blue = hexdec(mb_substr($line_color, 4, 2));
-        $line = [
-            'width' => 1.5,
-            'color' => [
-                $red,
-                $green,
-                $blue,
-            ],
-        ];
+        $lineStyle = ['all' => ['width' => 1.5, 'color' => [$red, $green, $blue]]];
```
williamdes: Same

williamdes (approved):
Thanks, let us know when it is ready to re-review or merge

MoonE:
It is ready. But to avoid any test failures when QA_5_2 is merged to master #20156 should be merged first. It includes the changes from this PR and updates the additional test pdf files on master accordingly. When merging this PR to master it should be an empty merge commit after that.

PR #20156, opened 2026-02-21
Refactor GIS export-as-file tests

MoonE:
### Description

This moves all the png/svg/pdf/ol export tests from each of the Geometry test classes to GisVisualizationTest.php, makes it a bit more manageable when you don't have to jump between so many files for the same thing.
It now uses the same WKT test data for all four different export types and all are stored as reference (expected) files instead of in-lining as strings.

Also includes the changes from #20154 including tests and updated reference test files for the pdf line drawing change. It's a bit of a hassle to get the pdf for x86 php right ...


williamdes (approved):
👏🏻

MauricioFauth (approved):

PR #20175, opened 2026-02-24
Export geometry and binary fields for TOON

liviuconcioiu:


liviuconcioiu:
I noticed there was also a bug `$result->numRows()` is always returning `0` when the query is unbuffered, since the export had 0 rows in header, which made the export invalid:

`Expected 0 tabular rows, but found more`

williamdes (approved):
Thank you !

PR #20179, opened 2026-02-26
Refactoring of Properties part 3 (final)

kamil-tekiela:
This PR cleans up the rest of the remaining code and removes all unnecessary and dead code. 

HTML IDs were altered to make it easier to generate the properties in code. This had an added benefit that the duplicate IDs were removed. I decided not to remove the IDs as even though they are not used by the code, they could be used by humans. However, I doubt if that's super useful. 

The header `Instead of <code>INSERT</code> statements, use:` was fixed so that it shows properly now. 

The methods `getForce` and `setForce` looked like they served some purpose but after a closer look, I determined that this code hasn't worked in years. I was not able to figure out if it worked when it was implemented or if it was simply incorrectly designed and then forgotten. But it's possible that I missed some super hidden magical behaviour of this code, so please correct me if I'm wrong. 

Several tests became obsolete. 

The boolean variables in `getOneOption` were replaced with polymorphic checks. I could not figure out a way to make the parameter type narrower, but I think I wrote it in such a way that it should handle all possible arguments.


src/Plugins/Export/ExportExcel.php
```diff
@@ -55,33 +55,32 @@ protected function setProperties(): ExportPluginProperties
...
         $generalOptions->addProperty($leaf);
         $leaf = new BoolPropertyItem(
-            'columns',
+            'excel_columns',
             __('Put columns names in the first row'),
         );
         $generalOptions->addProperty($leaf);
         $leaf = new SelectPropertyItem(
-            'edition',
+            'excel_edition',
```
williamdes, line 83: Could we not re use our Settings class?
I was thinking enum value
If you have a better idea

kamil-tekiela: Good point! However, enums are inappropriate for this. The main idea here was to have the same value here as in the settings, so that the situation doesn't happen again, where I try to remove a setting which is still in use, as I did with the LDI setting. This could probably be improved more so that the logical connection is more obvious than just the same name, but I can't think of a good way to do it right now. We can always revisit it later, and just reuse the setting name for the time being.

MauricioFauth (approved):

PR #20192, opened 2026-02-27
Transmit OpenLayers data as WKT

MoonE:
### Description

- Transmitting as WKT reduces the size to transfer, and OpenLayers comes with its own parser for this, so no need to do it ourselves.
The end goal would be for the editor to be able to work without requests to the server.
- Import modules directly: https://github.com/openlayers/openlayers/issues/16461
- Change the test data file names, so also the arch-specific names are covered in .gitignore


MauricioFauth (approved):
Thank you!

PR #20208, opened 2026-03-03
Rework GIS file export

MoonE:
### Description

- Move the download logic from GisVisualization to GisVisualizationController.
- Get the file blob directly without going through the output buffer for svg and pdf file download.
- Fix test when gd or TCPDF is not available
- Remove sanitizing of the download file name, it is done in `Core::downloadHeader` function, and the check for double filename extension was unnecessary as the file name is the column name, so in all likelihood does not contain an extension.
- Add Content-Size header with the file size.
- Send an error message in case the file export fails

image

### DIFF
--- a/src/Controllers/Table/GisVisualizationController.php
+++ b/src/Controllers/Table/GisVisualizationController.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace PhpMyAdmin\Controllers\Table;
 
+use Fig\Http\Message\StatusCodeInterface;
 use PhpMyAdmin\Config;
 use PhpMyAdmin\Controllers\InvocableController;
 use PhpMyAdmin\Core;
@@ -33,11 +34,11 @@ use function __;
 use function array_search;
 use function class_exists;
 use function extension_loaded;
+use function implode;
 use function in_array;
 use function is_array;
 use function is_string;
-use function ob_get_clean;
-use function ob_start;
+use function strlen;
 
 /**
  * Handles creation of the GIS visualizations.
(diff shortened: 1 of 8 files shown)


src/Gis/GisVisualization.php
```diff
@@ -346,19 +308,29 @@ private function png(): ImageWrapper|null
...
 
     /**
-     * Saves as a PNG image to a file.
+     * Get PNG image as string (blob) + type infos.
      *
-     * @param string $fileName File name
+     * @psalm-return array{mime: string, ext: string, blob: string}|null
      */
-    public function toFileAsPng(string $fileName): void
+    private function asPngFile(): array|null
```
kamil-tekiela: Does it have to be nullable?

kamil-tekiela: You could create a simple DTO and then nullable would make more sense.

MoonE: Thanks for the suggestion, it is now implemented.

The return value allows null because gd functions may fail, not sure in what conditions, so probably very unlikely for this case.

src/Gis/Ds/FileDownload.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpMyAdmin\Gis\Ds;
+
+readonly class FileDownload
```
kamil-tekiela: final

src/Controllers/Table/GisVisualizationController.php
```diff
@@ -111,14 +112,28 @@ public function __invoke(ServerRequest $request): Response
...
         $downloadOptions = $this->getDownloadOptions();
 
-        if (isset($_GET['saveToFile'])) {
+        if ($request->hasQueryParam('saveToFile')) {
+            $fileFormat = $request->getQueryParam('fileFormat');
+            if (! in_array($fileFormat, $downloadOptions, true)) {
+                return $this->responseFactory->createResponse(
+                    StatusCodeInterface::STATUS_BAD_REQUEST,
+                    __('Invalid file format, expected one of: ') . implode(', ', $downloadOptions),
+                );
```
williamdes, line 121: can you abstract this in the abstract controller ?

src/Controllers/Table/GisVisualizationController.php
```diff
@@ -111,14 +112,28 @@ public function __invoke(ServerRequest $request): Response
...
+            }
+
+            $data = $visualization->toFile($fileFormat);
+            if ($data === null) {
+                return $this->responseFactory->createResponse(
+                    StatusCodeInterface::STATUS_INTERNAL_SERVER_ERROR,
```
williamdes, line 127: same for server errors

williamdes (approved):
Nevermind, all good!

src/Controllers/Table/GisVisualizationController.php
MauricioFauth: This could be abstract to the response factory to something like createBadRequestResponse. In another PR of course.

MauricioFauth (approved):
Thank you!

PR #20209, opened 2026-03-03
Don't show invalid Gis download options

MoonE:
### Description

Backport these options from #20208:
- Add `download` attribute to download links so, in case of download error the current page is not replaced with the error message, instead it opens in a new tab.
- Remove invalid options:
If gd extension is missing, png can't be created.
If TCPDF is not installed, pdf can't be created.


williamdes (approved):
Thank you !

PR #20210, opened 2026-03-04
Add a SQL button to the database list

williamdes:
### Description

image

What do you think ?

Should "Tables" be named "Structure" like on other pages ?

### DIFF
--- a/resources/templates/server/databases/index.twig
+++ b/resources/templates/server/databases/index.twig
@@ -160,7 +160,7 @@
                 <th>{{ t('Replica replication') }}</th>
               {% endif %}
 
-              <th>{{ t('Action') }}</th>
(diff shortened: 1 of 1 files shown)


liviuconcioiu:
I think `Tables` is fine, because if you go on the first page, it says `Databases` in the navigation bar.

kamil-tekiela:
Why do we need a separate "tables" button? Clicking on the DB name already takes you there.

williamdes:
> Why do we need a separate "tables" button? Clicking on the DB name already takes you there.

This is true, I have to check
Maybe I was thinking that it was not the default behaviour
I was more interested in the SQL button 😂

MauricioFauth (approved):

PR #20229, opened 2026-03-08
Cleanup Content-Security-Policy header

MoonE:
- Require https for tile.openstreetmaps.org
- Replace `X-Frame-Origin` with `Content-Security-Policy` `frame-ancestors`
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Frame-Options
- Remove obsolete `X-Content-Security-Policy` and `X-WebKit-CSP` headers, all browsers supported by phpMyAdmin should support `Content-Security-Policy`
https://caniuse.com/contentsecuritypolicy
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy#browser_compatibility
https://github.com/phpmyadmin/phpmyadmin/blob/c90affe793b56fb6f5e1c7eed676ec9031fb1480/.browserslistrc#L1-L11

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -9915,17 +9915,6 @@
       <code><![CDATA[Config::getInstance()]]></code>
       <code><![CDATA[DatabaseInterface::getInstance()]]></code>
     </DeprecatedMethod>
-    <PossiblyInvalidArgument>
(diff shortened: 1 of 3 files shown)


MauricioFauth (approved):
Thank you!

PR #20233, opened 2026-03-09
Merge all git ignores in single file

MoonE:
### Description

This makes it a bit clearer what is being ignored.


williamdes (approved):
Nice !

💯 
This is good for Debian packaging, se we can easy remove the gitignore file: https://salsa.debian.org/phpmyadmin-team/tcpdf/-/blob/debian/unstable/debian/gbp.conf?ref_type=heads#L2

We have https://salsa.debian.org/phpmyadmin-team/tcpdf/-/blob/debian/unstable/debian/clean?ref_type=heads to clean the mess (`dh clean`) after building a package (`gbp buildpackage`)

MauricioFauth:
I don't like .gitkeep files, and I prefer the .gitignore ones, but if this helps packaging I'm okay with that.

MoonE:
Would you prefer to keep the .gitignore in app/cache as is, or have an empty .gitignore file instead of the .gitkeep file?

MauricioFauth:
It's not a big deal. It's better to use .gitkeep for consistency of using a single .gitignore file.

MauricioFauth (approved):

public/js/.gitignore
```diff
@@ -1,2 +0,0 @@
-/*
-!/.gitignore
```
MauricioFauth, line 2: A .gitkeep file is missing for this directory.

MauricioFauth (changes requested):

MauricioFauth (approved):
Thank you!

PR #20234, opened 2026-03-09
Remove duplicate release exclude cases

MoonE:
### Description
The two removed cases existed twice.


williamdes (approved):
Thanks for noticing this

MauricioFauth (approved):

PR #20237, opened 2026-03-14
Gis ScaleData offsets in target coordinate system

MoonE:
### Description

By first scaling the coordinates, then translating in the target coordinate system, the height can be added to the y-offset and is no longer needed in ScaleData.


williamdes (approved):
Thank you @MoonE !

williamdes:
Do you re-use phpMyAdmin classes in an other application ?

MoonE:
No, just using phpMyAdmin.

PR #17065, opened 2021-08-09
Replace Simulate Query dialog with a modal

shucontech:
Signed-off-by: Saksham Gupta <shucontech@gmail.com>

### Description

Replace Simulate Query dialog with a modal

Before submitting pull request, please review the following checklist:

- [ ] Make sure you have read our [CONTRIBUTING.md](https://github.com/phpmyadmin/phpmyadmin/blob/master/CONTRIBUTING.md) document.
- [ ] Make sure you are making a pull request against the correct branch. For example, for bug fixes in a released version use the corresponding QA branch and for new features use the _master_ branch. If you have a doubt, you can ask as a comment in the bug report or on the mailing list.
- [ ] Every commit has proper `Signed-off-by` line as described in our [DCO](https://github.com/phpmyadmin/phpmyadmin/blob/master/DCO). This ensures that the work you're submitting is your own creation.
- [ ] Every commit has a descriptive commit message.
- [ ] Every commit is needed on its own, if you have just minor fixes to previous commits, you can squash them.
- [ ] Any new functionality is covered by tests.

PR #20238, opened 2026-03-15
Fix #18941 - Reduce simulate-query memory

### LINKED ISSUE #18941: Simulate `DELETE` in a large table with million of rows gives error 500
### Describe the bug

Simulate `DELETE` in a large table with million of rows gives error 500.

### To Reproduce

1. Create a large table
2. Simulate a delete query
3. See error

### Expected behavior

No error should be given and simulate should give the row count that will be deleted.

### Screenshots

1e

### Server configuration

### Additional context

5.2.2-dev:

`PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 49152 bytes) in /libraries/classes/Dbal/DbiMysqli.php on line 209`

6.0.0-dev:

`PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 27279360 bytes) in /src/Dbal/DbiMysqli.php on line 171`

MoonE:
### Description

Fixes #18941

When the modal is opened it adds `overflow: hidden` style to body.
After clicking the `Affected rows` link, scrolling was disabled because the modal was not properly closed. Intrduced in #17065
image

- Adds a separate query to count the affected rows to no longer load all rows from the db.
- Allows paging for the changed-rows preview, this is done by wrapping the query with limit in another SELECT query.
image

- Changes the name of the new-value column from `` x `new` `` to `x *new*` to circumvent a sql-parser error

### DIFF
--- a/js/src/sql.js
+++ b/js/src/sql.js
@@ -220,6 +220,8 @@ AJAX.registerTeardown('sql.js', function () {
     $(document).off('submit', '.maxRowsForm');
     $(document).off('click', '#view_as');
     $(document).off('click', '#sqlquery');
+
+    $('#simulateDmlModal').modal('hide');
 });
 
 /**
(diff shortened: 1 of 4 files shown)


williamdes (approved):
Thank you for your contribution @MoonE !

PR #20240, opened 2026-03-16
Relax allowed characters in file names

### LINKED ISSUE #18157: Wrong filename when using Export schema in Designer with
### Describe the bug

Wrong filename when using Export schema in Designer. Instead of database name, ______________________________________________________ is present.

### To Reproduce

1. Create a database in Cyrillic or Chinese
2. Go to Designer
3. Click on `Export schema`
4. Click on `Go`
5. See error

### Expected behavior

Filename should contain the database name.

### Screenshots

https://user-images.githubusercontent.com/25424343/221378163-bd6111ec-1eb5-4e93-a315-fcaa93e4b3db.mp4

### Server configuration

MoonE:
Allows all unicode letters and all unicode numbers instead of only a-z, A-Z, 0-9.
https://www.php.net/manual/en/regexp.reference.unicode.php

Fixes #18157

### DIFF
--- a/libraries/classes/Sanitize.php
+++ b/libraries/classes/Sanitize.php
@@ -247,14 +247,8 @@ class Sanitize
      */
     public static function sanitizeFilename($filename, $replaceDots = false)
     {
-        $pattern = '/[^A-Za-z0-9_';
-        // if we don't have to replace dots
-        if (! $replaceDots) {
-            // then add the dot to the list of legit characters
-            $pattern .= '.';
-        }
-
-        $pattern .= '-]/';
+        // Keep only numbers (N), letters (L), dash, underbar, and maybe dot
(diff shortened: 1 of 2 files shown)


kamil-tekiela (approved):
That's probably fine. I think Linux shouldn't have any problems. Windows sometimes complains, but I think this still excludes characters that Windows doesn't like.

MoonE:
List of forbidden characters on windows:
https://learn.microsoft.com/en-us/answers/questions/4159593/what-characters-are-not-permitted-in-a-file-name

Though browsers like Chrome or Firefox also sanitize file names in case it contains an invalid character.

williamdes (approved):
Thank you, the tests are great!

MauricioFauth (approved):
Thank you!

PR #20241, opened 2026-03-17
Add required referrer when loading tiles from openstreetmap.org

M393:
### Description

tile.openstreetmap.org requires a `Referer` header to be set. If phpMyAdmin is loaded with `Referer: same-origin` or `Referer: no-referrer`, then by default no `Referer` header is sent.
image

With this change it sends a `Referer` to tiles.openstreetmap.org
See: https://github.com/openlayers/openlayers/issues/17088

For phpMyAdmin 6.0 this can be fixed by upgrading to OpenLayers >= [v10.7.0](https://github.com/openlayers/openlayers/releases/tag/v10.7.0) where this is the default.


williamdes (approved):
Thank you

williamdes:
I did not do the change for 6.0, if you can open a PR for it please ?

PR #20170, opened 2026-02-23
Fix #20153: Handle missing database tag in XML import

robinson-j16:
### Description

This pull request fixes a fatal `ErrorException` that occurs during XML imports when the `<database>` tag is missing from the file.

Fixes #20153

Before submitting pull request, please review the following checklist:

- [x] Make sure you have read our [CONTRIBUTING.md](https://github.com/phpmyadmin/phpmyadmin/blob/master/CONTRIBUTING.md) document.
- [x] Make sure you are making a pull request against the correct branch. For example, for bug fixes in a released version use the corresponding QA branch and for new features use the _master_ branch. If you have a doubt, you can ask as a comment in the bug report or on the mailing list.
- [x] Every commit has proper `Signed-off-by` line as described in our [DCO](https://github.com/phpmyadmin/phpmyadmin/blob/master/DCO). This ensures that the work you're submitting is your own creation.
- [x] Every commit has a descriptive commit message.
- [x] Every commit is needed on its own, if you have just minor fixes to previous commits, you can squash them.
- [x] Any new functionality is covered by tests.

PR #20242, opened 2026-03-17
Fix ImportXML existence check

### LINKED ISSUE #20153: [6.0]: ErrorException: foreach() argument must be of type array|object, null given, [5.2] ImportXml::doImport(): Node no longer exists
### Describe the bug

```
Internal error in .\src\Plugins\Import\ImportXml.php#170
ErrorException: foreach() argument must be of type array|object, null given
```

I reproduced by mistake :sunglasses: the bug from:

https://reports.phpmyadmin.net/reports/view/88528

https://github.com/phpmyadmin/phpmyadmin/blob/c299499438278ec0da569a5a49b9d572f11a8724/src/Plugins/Import/ImportXml.php#L170

It happens when there's no `database` tag (I deleted it too, to import the table somewhere else):

NOT WORKING:

```XML
<?xml version="1.0" encoding="utf-8"?>
<pma_xml_export version="1.0" xmlns:pma="https://www.phpmyadmin.net/some_doc_url/">
    <pma:structure_schemas>
        <pma:database name="test25" collation="utf8mb4_general_ci" charset="utf8mb4">
            <pma:table name="test">
                CREATE TABLE `test` (
                  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
                  PRIMARY KEY (`id`)
                ) ENGINE=MyISAM AUTO_INCREMENT=2 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;
            </pma:table>
        </pma:database>
    </pma:structure_schemas>
</pma_xml_export>
```

WORKING:

```XML
<?xml version="1.0" encoding="utf-8"?>
<pma_xml_export version="1.0" xmlns:pma="https://www.phpmyadmin.net/some_doc_url/">
    <pma:structure_schemas>
        <pma:database name="test25" collation="utf8mb4_general_ci" charset="utf8mb4">
            <pma:table name="test">
                CREATE TABLE `test` (
                  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
                  PRIMARY KEY (`id`)
                ) ENGINE=MyISAM AUTO_INCREMENT=2 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;
            </pma:table>
        </pma:database>
    </pma:structure_schemas>
    <database name="test25">
    </database>
</pma_xml_export>
```

### How to Reproduce

1. Delete `database` tag from XML
2. Import XML

### phpMyAdmin version

6.0.0-dev

kamil-tekiela:
Fixes #20153

Alternative to #20170. Adds an explanatory comment and avoids the Psalm issue. 

`SimpleXMLElement` is a bit magical because they are both acting as an array and as a string. Fetching children can be done via the magic `__get` method. If the child does not exist, an empty `SimpleXMLElement` is returned. It is considered empty both in string and in array context, and therefore also if used in a boolean context, e.g. `if($element)` will convert to `''` which is falseish. Static analysers are not very good at identifying this, so the easiest way is to check if the child exists before fetching it. 

I think I have introduced this error in one of my refactorings because I believed incorrectly that an empty `SimpleXMLElement` would still return an empty `SimpleXMLElement` when asked for a child, but it returns `null` instead.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -9441,7 +9441,7 @@ parameters:
 		-
 			message: '#^Argument of an invalid type SimpleXMLElement\|null supplied for foreach, only iterables are supported\.$#'
 			identifier: foreach.nonIterable
-			count: 1
+			count: 2
 			path: src/Plugins/Import/ImportXml.php
 
 		-
@@ -9477,7 +9477,7 @@ parameters:
 		-
 			message: '#^Offset ''name'' might not exist on SimpleXMLElement\|null\.$#'
 			identifier: offsetAccess.notFound
-			count: 4
+			count: 3
 			path: src/Plugins/Import/ImportXml.php
 
 		-
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -6287,11 +6287,8 @@
       <code><![CDATA[string[]]]></code>
     </PossiblyUnusedReturnValue>
     <RedundantCondition>
-      <code><![CDATA[$databaseXml->table]]></code>
+      <code><![CDATA[isset($xml->database)]]></code>
     </RedundantCondition>
-    <TypeDoesNotContainType>
-      <code><![CDATA[[]]]></code>
-    </TypeDoesNotContainType>
     <UnnecessaryVarAnnotation>
       <code><![CDATA[SimpleXMLElement]]></code>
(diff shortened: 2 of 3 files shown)


williamdes (approved):
Thank you, if the new tests agree I agree too

PR #20243, opened 2026-03-17
Update OpenLayers to fix missing referer header for tile loading

M393:
See #20241


williamdes (approved):
No changes to be made between major versions ?
Wow, great !

M393:
I added some changes that will make it compatible with a future v11, see https://github.com/openlayers/openlayers/issues/16461
Edit: this was done also in #20192

MauricioFauth (approved):
Thank you!

Issue #16801, opened 2021-04-09
Quotes appear when field type is TEXT and I use default 'as defined'


evilKabab:
Hello.
I guess I found a bug.

Server is CentOS 7 latest. Phpmyadmin is 5.1.0 english running on php 7.3.27 fpm.   DB is MariaDB 10.2.37. Apache/2.4.6, mysqlnd 5.0.12-dev - 20150407.

So I created a filed with type TEXT, set default value to 'as defined:' - empty string. After saving that I see that default value looks as two single quotes (''). Strange since right above I see VARCHAR with same default not displaying any quotes for empty string. If I try to change that TEXT field it gets even worse not it displays single quotes escaped with backslashes (\'\').

See screenshots attached

Michael Sosonkin
![fig1](https://user-images.githubusercontent.com/5341920/114129809-4ea3b000-98cd-11eb-8f8d-f8950f114019.jpg)
![fig2](https://user-images.githubusercontent.com/5341920/114129814-52373700-98cd-11eb-9135-0d6545366fa5.jpg)
![fig3](https://user-images.githubusercontent.com/5341920/114129820-5400fa80-98cd-11eb-8150-303e0067a11e.jpg)
![fig4](https://user-images.githubusercontent.com/5341920/114129827-582d1800-98cd-11eb-9739-0d68bdabc0fb.jpg)

evilKabab:
One picture is missing somehow...
![fig3](https://user-images.githubusercontent.com/5341920/114129929-8a3e7a00-98cd-11eb-8d6b-58c094832187.jpg)

williamdes:
Confirmed, thank you for the report @evilKabab !

williamdes:
```sql
CREATE TABLE `emptydefaults` (
  `id` varbinary(16) DEFAULT NULL,
  `emptytext` text NOT NULL DEFAULT '',
  `emptyvar` varchar(25) NOT NULL DEFAULT '',
  `emptychar` char(5) NOT NULL DEFAULT '',
  `emptydq` text NOT NULL DEFAULT '""',
  `emptysq` text NOT NULL DEFAULT '\'\'',
  `varcharemptydq` varchar(25) NOT NULL DEFAULT '""',
  `varemptysq` varchar(25) NOT NULL DEFAULT ''''''
) ENGINE=InnoDB DEFAULT CHARSET=latin1;
```
Only text fields seem to be affected

williamdes:
### Before fix
[...]

PR #20244, opened 2026-03-17
Change column default string value

M393:
Fixes the column editing problem from #16801.
The default value for text columns is still displayed in quotes in the table structure page.

Example table, when editing, both column Default values should be `As defined:` with empty string value
```sql
CREATE TABLE `empty_default` (
`c1` TINYTEXT NOT NULL DEFAULT '', 
`c2` VARCHAR(77) NOT NULL DEFAULT '' 
) ENGINE = InnoDB;
```

The problem with the missing default value when it is an empty string was introduces in 8929288efd592047c81df3b9b8b8a8151edaac41

The problem with quotes not being removed when editing a text column is probably even older, the value was assigned to the wrong property.

### DIFF
--- a/libraries/classes/Table/ColumnsDefinition.php
+++ b/libraries/classes/Table/ColumnsDefinition.php
@@ -529,11 +529,11 @@ final class ColumnsDefinition
 
                 break;
             default:
-                $metaDefault['DefaultValue'] = $columnMeta['Default'];
-
                 if (substr((string) $columnMeta['Type'], -4) === 'text') {
                     $textDefault = substr($columnMeta['Default'], 1, -1);
-                    $metaDefault['Default'] = stripcslashes($textDefault);
+                    $metaDefault['DefaultValue'] = stripcslashes($textDefault);
+                } else {
+                    $metaDefault['DefaultValue'] = $columnMeta['Default'];
                 }
 
                 break;
--- a/libraries/classes/Table/ColumnsDefinition.php
+++ b/libraries/classes/Table/ColumnsDefinition.php
@@ -510,10 +510,11 @@ final class ColumnsDefinition
 
         switch ($columnMeta['Default']) {
             case null:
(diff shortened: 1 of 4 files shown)


M393:
There is some ambiguity here with expressions and literals, probably best to add a new expression type to the select input.
- None
- As defined:
- *Expression:*
- NULL
- CURRENT_TIMESTAMP

There are quite a few differences between the different MySQL/MariaDB versions for this, makes it rather difficult to implement.

libraries/classes/Table/ColumnsDefinition.php
```diff
@@ -528,11 +530,19 @@ public static function decorateColumnMetaDefault(array $columnMeta): array
...
             default:
-                $metaDefault['DefaultValue'] = $columnMeta['Default'];
-
                 if (substr((string) $columnMeta['Type'], -4) === 'text') {
-                    $textDefault = substr($columnMeta['Default'], 1, -1);
-                    $metaDefault['Default'] = stripcslashes($textDefault);
+                    if (
+                        mb_strlen($columnMeta['Default']) >= 2 &&
+                        mb_substr($columnMeta['Default'], 0, 1) === "'" &&
+                        mb_substr($columnMeta['Default'], -1) === "'"
```
kamil-tekiela: `str_starts_with` and ends_with

OlafvdSpek: Requires PHP 8, or is a compat lib used?

kamil-tekiela: It's easy to check. You can grep the repo for that name and see if it's already in use. You can also check in composer.json whether `symfony/polyfill-php80` is required.

M393: `symfony/polyfill-php80` is required in composer.json, so it is available.

PR #20246, opened 2026-03-18
Fix create new user with single quote in password

M393:
When creating a user, the new password was not escaped when it is used as clear text in the SQL query.

### DIFF
--- a/libraries/classes/Server/Privileges.php
+++ b/libraries/classes/Server/Privileges.php
@@ -3554,7 +3554,7 @@ class Privileges
                     $hashedPassword = $_POST['pma_pw'];
                 }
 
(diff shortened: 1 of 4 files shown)


kamil-tekiela:
This problem exists only on 5.2, correct?

M393:
Yes, in 6.0 it is correctly escaped.
https://github.com/phpmyadmin/phpmyadmin/blob/4acd79bd83e3422187516cb1c3443eac48cf5dcc/src/Server/Privileges.php#L2969-L2979

williamdes (approved):
Thank you, this change looks okay

MauricioFauth (approved):
Thank you!

PR #20255, opened 2026-03-24
Only hide processes in status Sleep and inactive Daemon

### LINKED ISSUE #20187: [Bug]: Processlist with show active statements only
### Describe the bug

When `Show only active` is checked then some statements are not displayed despite being active.

This happens because the STATE column in information_schema.PROCESSLIST is empty.

Statement where I've seen this happen are UPDATE and SET, but only some UPDATE statements. For some the Status is `Updating`.

Makes me think it could also be a bug with MariaDB …

### How to Reproduce

I executed a bunch of update statements of this form, they never have a state set.
```sql
UPDATE `t` SET `updated` = `updated`, `coord` = ST_PointFromWKB(IF(`x` = 0, POINT(0, 0), POINT(`x` / 1e5, `y` / 1e5)), 4326) WHERE id BETWEEN 1062900001 AND 1063000000;
```

### Expected behavior

These statements are active and should not be hidden in this case.

### Database Server

MariaDB 11.4.7

### phpMyAdmin version

5.2.3, 6.0.0

### Additional context

State is empty `---`, if `Show only active` were selected, they are not shown.
Image

M393:
Fixes #20187

Only excludes connections with command `Sleep` or Status `Waiting for next activation` (Event scheduler)

### DIFF
--- a/libraries/classes/Server/Status/Processes.php
+++ b/libraries/classes/Server/Status/Processes.php
@@ -52,11 +52,11 @@ final class Processes
             $urlParams['order_by_field'] = $params['order_by_field'];
             $urlParams['sort_order'] = $params['sort_order'];
             $urlParams['showExecuting'] = $params['showExecuting'];
-            $sqlQuery = 'SELECT * FROM `INFORMATION_SCHEMA`.`PROCESSLIST` ';
+            $sqlQuery = 'SELECT * FROM `INFORMATION_SCHEMA`.`PROCESSLIST`';
         }
 
         if (! empty($params['showExecuting'])) {
-            $sqlQuery .= ' WHERE state != "" ';
+            $sqlQuery .= " WHERE `COMMAND` <> 'Sleep' AND `STATE` <> 'Waiting for next activation'";
         }
 
         if (! empty($params['order_by_field']) && ! empty($params['sort_order'])) {


libraries/classes/Server/Status/Processes.php
```diff
@@ -52,11 +52,11 @@ public function getList(array $params): array
             $urlParams['order_by_field'] = $params['order_by_field'];
             $urlParams['sort_order'] = $params['sort_order'];
             $urlParams['showExecuting'] = $params['showExecuting'];
-            $sqlQuery = 'SELECT * FROM `INFORMATION_SCHEMA`.`PROCESSLIST` ';
+            $sqlQuery = 'SELECT * FROM `INFORMATION_SCHEMA`.`PROCESSLIST`';
         }
 
         if (! empty($params['showExecuting'])) {
-            $sqlQuery .= ' WHERE state != "" ';
+            $sqlQuery .= " WHERE `COMMAND` <> 'Sleep' AND `STATE` <> 'Waiting for next activation'";
```
OlafvdSpek, line 59: Is <> used commonly in pMA rather than != ?

M393: I guess I prefer `<>` because it is standardized SQL, but both should work and both are used elsewhere.

kamil-tekiela: I don't think it matters. They are identical and both are understood.

williamdes: I think `!=` is easier to understand for newcomers or users that audit queries

MauricioFauth (approved):
Thank you!

PR #20259, opened 2026-03-27
Fix #19842 - skip user link for non-real users in processes list

mvanhorn:
The processes list currently links every user name to the privileges page.
Pseudo-users like `event_scheduler` and `unauthenticated user` have no
corresponding entry in the user accounts table, so the link leads to an
empty or broken page.

This extends the existing `system user` check to also cover
`event_scheduler` and `unauthenticated user` by switching the condition
to a Twig `not in` list.

Fixes #19842

### DIFF
--- a/templates/server/status/processes/list.twig
+++ b/templates/server/status/processes/list.twig
@@ -46,7 +46,7 @@
           </td>
           <td class="font-monospace text-end">{{ row.id }}</td>
           <td>
-            {% if row.user != 'system user' %}
+            {% if row.user not in ['system user', 'event_scheduler', 'unauthenticated user'] %}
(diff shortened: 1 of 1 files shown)


williamdes:
Thanks
can you ask Claude to cherry pick rebase your work onto QA_5_2 and force push?
lower rebasing is tricky

mvanhorn:
Rebased onto QA_5_2 in 580546b. A straight rebase from master hit conflicts in `Version.php` and `package.json`, so I cherry-picked the fix commit onto `upstream/QA_5_2` instead and updated the PR base branch to match.

williamdes (approved):
Thank you!

mvanhorn:
Thanks for the review!

PR #20260, opened 2026-03-27
Fix #20181 - allow selecting text fragments in cells

JustJacob95:
### Description

Fix https://github.com/phpmyadmin/phpmyadmin/issues/20181 

Users can now select partial text within a cell as before. At the same time, multi-cell selection for copying is also supported; if the cursor moves to another cell while the mouse is pressed, the text selection is cleared to allow for cell selection.

[2026-03-27 16-14-35.webm](https://github.com/user-attachments/assets/efc4bba2-c656-495e-9155-3df3c4e48f65)

### DIFF
--- a/resources/js/makegrid.ts
+++ b/resources/js/makegrid.ts
@@ -2615,8 +2615,9 @@ const makeGrid = function (t, enableResize = undefined, enableReorder = undefine
 
                 selectCell(this);
 
-                // Prevent text selection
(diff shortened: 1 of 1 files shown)


williamdes (approved):
Awesome
And many thanks for the video !

I you want to do a video to showcase all this new feature on LinkedIn, I can repost your post. Just mention phpMyAdmin and ping me

PR #20264, opened 2026-04-04
Clean up config again

kamil-tekiela:
Another step in the direction of cleaning up the config. The commits are because I was getting lost in what I have edited already and it could all be squashed.


MauricioFauth (approved):
Thank you!

PR #20265, opened 2026-04-04
Fix #20063 - Fix routine, event and trigger duplication when editing 

predictor2718:
## Summary                                                                                      
- Modal event listeners (`shown.bs.modal`, `hidden.bs.modal`) accumulated on each `editorDialog()` call, causing duplicate rows after repeated edits                              
- Added `{ once: true }` to all affected listeners in routines, events, and triggers
- Same fix applied to editor, export, and execute modals                                                                                                                                   
                                                                                                  
Fixes #20063


williamdes (approved):
Thank you

MauricioFauth (approved):
Thank you!

PR #20267, opened 2026-04-05
Fix #19338 - Add missing length validation when changing column type

### LINKED ISSUE #19338: 'Please enter a valid length!' doesn't show up when changing existing column
### Describe the bug

When adding VARCHAR column without length, you get 'Please enter a valid length!' but when you change an existing column to VARCHAR the warning does not popup

### To Reproduce

1. Go to Structure 
2. Click on Change column
3. Change column type to VARCHAR and keep length blank
4. See no error

### Expected behavior

I expect to receive 'Please enter a valid length!'

- phpMyAdmin version: 6.0

### Client configuration

- Browser: Chrome 129

predictor2718:
### Description

- The "Change column" form didn't load `table/structure.js`, so the client-side length validation never fired
- Added the missing script to `ChangeController`, matching what `AddFieldController` already does

Fixes #19338.

### DIFF
--- a/src/Controllers/Table/Structure/ChangeController.php
+++ b/src/Controllers/Table/Structure/ChangeController.php
@@ -98,7 +98,7 @@ final readonly class ChangeController implements InvocableController
          */
         $userPrivileges = $this->userPrivilegesFactory->getPrivileges();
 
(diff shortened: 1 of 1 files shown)


williamdes (approved):
Thank you for fixing this bug !

PR #20268, opened 2026-04-05
Fix #20110 - Skip locale directory assertion when folder is missing

predictor2718:
### Description

- `testVendorConfigFile` failed on Windows because the `locale` directory doesn't exist before compilation
- Removed the `assertDirectoryExists` for `localePath` since the directory is optional (other locale-dependent tests already skip when it's missing)
- The `assertIsString` check for the path value is kept

Fixes #20110.


tests/unit/ConfigTest.php
```diff
@@ -638,7 +638,6 @@ public function testVendorConfigFile(): void
         self::assertIsString($vendorConfig['customFooterFile']);
         self::assertIsBool($vendorConfig['versionCheckDefault']);
         self::assertIsString($vendorConfig['localePath']);
-        self::assertDirectoryExists($vendorConfig['localePath']);
```
williamdes, line 641: This could be under an if not null

PR #20278, opened 2026-04-10
Fix #19225 - Fix wrong row count when column alias shadows existing column name

### LINKED ISSUE #19225: common field on select name changes totals count display
### Describe the bug

Under phpmyadmin 5.2.1 using two fields of the same name in the select list causes an incorrect result count. It works correctly on version 5.0.4, but fails under 5.2.1. It also fails under the demonstration data base available under phpmyadmin, which uses version 6.0.0, but shows the same behavior as stable 5.2.1.

### To Reproduce

To demonstrate: connect to the demo available on the phpmyadmin web site. For the "world" database there, look at the "city" table in that database. There are 4079 records. Now run this SQL:

```sql
SELECT id, name as "id"
FROM `City` 
WHERE district > "k"
```

The query results shows "showing rows 0-24 (4,079 total,...)"

Now change the "as" clause to "id2" (or remove the as clause), and now you get the correct total: "showing rows 0-24 (2,416 total,...)"

Obviously, I would not rename a field from a table to match another field in the same table, but this issue comes up when performing joins where two fields of the same name (eg, same domain) are part of the select list, even when qualified by the table name prefix. (For example a supplier record with a city field and a customer record with a city field.) I felt a simple select on a single table would be simpler to diagnose.

Is this a bug? Or am I missing something here? The numbers of records I am dealing with are small, so the innodb limits of 50000 etc do not come into play. Changing one field name always give the correct result.

BTW adding LIMIT 0,999999 seems to give a correct result, but that seems to be a rather odd requirement to have to add to my SQL statements. It all worked correctly in 5.0.4 previously.

### Expected behavior

See above example.

### Screenshots

If applicable, add screenshots to help explain the bug.

### Server configuration

- Operating system:
- Web server:
- Database version:
- PHP version:
- phpMyAdmin version: 5.2.1 and demo 6.0.0

### Client configuration

- Browser:
- Operating system:

### Additional context

I have reproduced this issue on three systems, including the phpmyadmin demo database. All versions of php 8.x. One a UNIX server, one a Windows Server.

Further info: in the example above, change the district comparison to "x" rather then "k". The correct result total should be 105 but 4079 still shows. Change the number of rows field to 50. Same erroneous result. Change number of rows to 100. Same erroneous result. Change number of rows to 250. Now you get the correct totals display of 105.

predictor2718:
## Summary                                                                                      
  - When a SELECT aliases a column with the same name as another column (e.g.h`SELECT id, name AS
  id`), the row count display was incorrect (showed total table rows instead of filtered rows)    
  - Root cause: the COUNT subquery `SELECT COUNT(*) FROM (SELECT id, name AS id ...) as cnt` fails
   with MySQL error "Duplicate column name", causing a fallback to `countRecords()` which counts  
  all rows without the WHERE clause                         
  - Fix: replace all SELECT expressions with `SELECT 1` in the count subquery (when no GROUP BY,  
  DISTINCT, or UNION), since only the row count matters, not the column values                    
  - This also improves performance for queries with expensive subqueries in the SELECT list, as   
  they are no longer executed during counting                                                     
                                                                                                                              
                                                            
  Fixes #19225

### DIFF
--- a/libraries/classes/Sql.php
+++ b/libraries/classes/Sql.php
@@ -781,8 +781,17 @@ class Sql
                 // Removes LIMIT clause that might have been added
                 $statement->limit = null;
 
-                if ($changeExpression) {
-                    $statement->expr[0] = new Expression();
+                // Replace SELECT expressions with 1 to avoid
+                // "Duplicate column name" errors when aliases shadow
+                // existing column names (e.g. SELECT id, name AS id).
+                // Only safe when there's no GROUP BY, DISTINCT, or UNION
+                // since those depend on the actual expressions.
+                if (
+                    $analyzedSqlResults['is_group'] === false
+                    && $analyzedSqlResults['distinct'] === false
+                    && $analyzedSqlResults['union'] === false
+                ) {
+                    $statement->expr = [new Expression()];
                     $statement->expr[0]->expr = '1';
                 }
 
--- a/test/classes/SqlTest.php
+++ b/test/classes/SqlTest.php
@@ -513,7 +513,8 @@ class SqlTest extends AbstractTestCase
                 ],
                 20,
                 42,
-
+                false,
+                'SELECT COUNT(*) FROM (SELECT 1 FROM company_users WHERE subquery_case = 0 ) as cnt',
             ],
             [
                 'SELECT ( as c2 FROM company_users WHERE working_count = 0',// Invalid query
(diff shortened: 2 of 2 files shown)


williamdes (approved):
Thank you for writing tests, this is a major factor in the decision to merge fixes

I use [contrib-merge](https://github.com/wdes/contrib-merge) by the way

MauricioFauth (approved):
Thank you!

PR #20279, opened 2026-04-10
Fix #19080 - Show correct query in confirmation prompt after inline edit

### LINKED ISSUE #19080: Wrong query shown in confirmation prompt
### Describe the bug

Wrong query used for confirmation prompt.

### To Reproduce

1. Execute an ALTER TABLE statement
2. Click on `edit inline` after it is completed
3. Execute a different ALTER TABLE statement
4. In the confirmation prompt the old query is shown

### Expected behavior

The actually executed query should be shown (and probably also use it to decide whether to show the prompt in the first place)

### Screenshots
Query from prompt not in the text area ...
image

It is hidden in the query box:
image

### Server configuration

- Operating system:
- Web server:
- Database version: MariaDB 10.11.7
- PHP version: 8.1
- phpMyAdmin version: 5.2.1

### Client configuration

- Browser: Chrome 122
- Operating system:  Windows 10

predictor2718:
## Summary
  - After editing a query inline and submitting, the confirmation prompt showed the old query instead of the new one
  - Root cause: `checkSqlQuery()` always read from the global `codeMirrorEditor` instead of the active
  `codeMirrorInlineEditor`
  - Fix: check `codeMirrorInlineEditor` first, fall back to `codeMirrorEditor`

  ## How to verify
  1. Create a table with multiple columns (e.g. `email` and `phone`)
  2. Execute `ALTER TABLE t1 DROP COLUMN email` and accept the confirmation
  3. Click "Edit inline"
  4. Change the query to `ALTER TABLE t1 DROP COLUMN phone`
  5. Click Go
  6. The confirmation prompt should now show "DROP COLUMN phone" (before this fix it showed "DROP COLUMN email")

  Fixes #19080

### DIFF
--- a/js/src/functions.js
+++ b/js/src/functions.js
@@ -762,7 +762,10 @@ Functions.confirmQuery = function (theForm1, sqlQuery1) {
 Functions.checkSqlQuery = function (theForm) {
     // get the textarea element containing the query
     var sqlQuery;
-    if (codeMirrorEditor) {
+    if (codeMirrorInlineEditor) {
+        codeMirrorInlineEditor.save();
+        sqlQuery = codeMirrorInlineEditor.getValue();
+    } else if (codeMirrorEditor) {
         codeMirrorEditor.save();
         sqlQuery = codeMirrorEditor.getValue();
     } else {


MauricioFauth (approved):
Thank you!

PR #20281, opened 2026-04-11
Fix #19939 - Use TEXT instead of VARCHAR when CSV column exceeds max …

### LINKED ISSUE #19939: CSV import - `#1074 - Column length too big for column 'COL 1' (max = 16383); use BLOB or TEXT instead`
### Describe the bug

When importing a CSV file where row value is > 16383, this error is given:

```SQL
CREATE TABLE IF NOT EXISTS `test18`.`test` (`COL 1` varchar(17256), `COL 2` int(1));;

#1074 - Column length too big for column 'COL 1' (max = 16383); use BLOB or TEXT instead
```

Expected query:

```SQL
CREATE TABLE IF NOT EXISTS `test18`.`test` (`COL 1` TEXT, `COL 2` int(1));
```

### How to Reproduce

1. Go to `Import`
2. Import attached file
3. See error

[test.csv](https://github.com/user-attachments/files/23507133/test.csv)

### Expected behavior

Table should be created and import should work.

### Screenshots

Image

### phpMyAdmin version

5.2.4-dev, 6.0.0-dev

predictor2718:
## Summary                                                                                                          
  - When importing a CSV file with values longer than 16383 characters, phpMyAdmin generated `VARCHAR(17256)` which   
  exceeds MySQL's maximum VARCHAR length for utf8mb4                                                                  
  - Now uses `TEXT` instead of `VARCHAR` when the detected column size exceeds 16383 (the most restrictive limit,     
  based on utf8mb4's 4 bytes per character)                                                                           
                                                                                                                    
  ## How to verify                                                                                                    
  1. Create a CSV file with a value longer than 16383 characters                                                    
  2. Import it via Database → Import                                                                                  
  3. Table should be created with `TEXT` column instead of failing with an error
                                                                                                                      
  Fixes #19939

### DIFF
--- a/libraries/classes/Import.php
+++ b/libraries/classes/Import.php
@@ -1098,9 +1098,19 @@ class Import
                         $size = 10;
                     }
 
+                    $colType = $typeArray[$analyses[$i][self::TYPES][$j]];
+
+                    // Use TEXT instead of VARCHAR when the length exceeds
+                    // the maximum allowed for VARCHAR. The limit depends on
+                    // the charset (e.g. 16383 for utf8mb4, 65535 for latin1).
+                    // We use the most restrictive limit (utf8mb4) to be safe.
+                    if ($colType === 'varchar' && (int) $size > 16383) {
+                        $colType = 'text';
+                    }
+
                     $tempSQLStr .= Util::backquote($tables[$i][self::COL_NAMES][$j]) . ' '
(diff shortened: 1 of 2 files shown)


kamil-tekiela (approved):

MauricioFauth (approved):
Thank you!

williamdes (commented):
💯

PR #20282, opened 2026-04-11
Fix #19084 - Skip CREATE DATABASE when exporting data only

### LINKED ISSUE #19084: "Create database if not exists" line should not be included during an SQL server export when only exporting data
### Describe the bug

When doing an SQL Server Export (Exporting when no DB/Table is selected) and choosing to export only data.
The First line in in the SQL file is "Create database if not exists", which does not make sense when exporting only data.
Even if the database doesn't exist yet, the SQL file will not create the table, but try to insert the data anyway.

### To Reproduce

1. at startup /phpmyadmin
2. go to export (route=/server/export)
3. export method: custom
4. sql format.
5. select the database(s).
6. Format-specific options: choose "data"
7. download file #-1

### Expected behavior

The export file only has SQL statements to insert the data, not to create the database.
The "use `database_name`" line should stay.

### Screenshots

Screenshot 2024-03-30 001529

### Server configuration

- phpMyAdmin version: All

predictor2718:
## Summary
  - When doing a server export with "data" only, `CREATE DATABASE IF NOT EXISTS` was still included in the SQL output 
  - The `$exportStructure` check was only guarding the `DROP DATABASE` statement, not the `CREATE DATABASE` statement 
  - Added an early return to skip `CREATE DATABASE` when only data is being exported, while keeping the `USE`         
  statement                                                                                                           
                                                                                                                      
  **Note:** This bug also exists on master (6.0.0-dev) with the same root cause.                                      
                                                            
  ## Test plan                                                                                                        
  - [ ] Server export → Custom → SQL → select database → "data" only → no `CREATE DATABASE` in output
  - [ ] Server export → "structure and data" → `CREATE DATABASE` still present                                        
  - [ ] Database export → data only → still works correctly                                                           
  - [ ] Added unit test `testExportDBCreateDataOnly`                                                                  
                                                                                                                      
  Fixes #19084

### DIFF
--- a/libraries/classes/Plugins/Export/ExportSql.php
+++ b/libraries/classes/Plugins/Export/ExportSql.php
@@ -879,6 +879,11 @@ class ExportSql extends ExportPlugin
             return true;
         }
 
+        // Skip CREATE DATABASE when only exporting data
+        if (! $exportStructure) {
+            return $this->exportUseStatement($dbAlias, $compat);
+        }
+
         $createQuery = 'CREATE DATABASE IF NOT EXISTS '
             . Util::backquoteCompat($dbAlias, $compat, isset($GLOBALS['sql_backquotes']));
         $collation = $dbi->getDbCollation($db);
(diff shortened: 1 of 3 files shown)


MauricioFauth (approved):
Thank you!

PR #20283, opened 2026-04-13
Upgrading phpstan/phpstan (2.1.41 => 2.1.47)

kamil-tekiela:
It seems there was a bug in TableMover because the code below that if statement was dead. I presume it was meant to be a boolean check instead of isset.


williamdes (approved):
Thank you 💯

PR #20288, opened 2026-04-23
Update Yarn dependencies

MauricioFauth:


M393:
If OpenLayers is updated, then #20241 can be reverted, but ...

... newer versions of OpenLayers use [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver#browser_compatibility), which is not compatible with the versions in https://github.com/phpmyadmin/phpmyadmin/blob/c90affe793b56fb6f5e1c7eed676ec9031fb1480/.browserslistrc#L1-L11

MauricioFauth:
> If OpenLayers is updated

It was not updated, it's still OpenLayers 6.15.1. Just the build is different.

PR #20296, opened 2026-04-28
Fix #20284 - Preserve numeric strings in XML imports

### LINKED ISSUE #20284: [Bug]: XML import does not correctly quote string values
### Describe the bug

When importing from an XML file, if the value *looks like* a number (even tho the table definition clearly says varchar), it is not escaped.
This leads to issues with mysql then interpreting that as a numeric value and discarding leading zeroes, causing wrong or duplicate values.

I've attached a minimal reproduction case below.

### How to Reproduce

1. import this XML file into a database.
2. get an error: `#1062 - Duplicate entry '1234' for key 'data'`

[test_data.xml](https://github.com/user-attachments/files/26727709/test_data.xml)

### Expected behavior

The import should detect that the column type is `varchar(255)` and therefore provide the values as quoted strings: `'01234'`

### Screenshots

_No response_

### Operating System

Linux

### Web Server

nginx/1.28.3 (LSIO phpmyadmin docker image)

### Database Server

MariaDB

### Database version

11.8.6-MariaDB-ubu2404-log

### PHP version

8.5.4

### phpMyAdmin version

5.2.3

### browser

Vivaldi (Chromium)

### Additional context

The XML file is turned into the following query:
```sql
INSERT INTO
  `test_data` (`id`, `data`)
VALUES
  (3, 01234),
  (2, 1234),
  (1, 'abcd')
```

But it should instead be
```sql
INSERT INTO
  `test_data` (`id`, `data`)
VALUES
  (3, '01234'),
  (2, '1234'),
  (1, 'abcd')
```

eyupcanakman:
### Description

When the XML schema is embedded, `buildSql()` falls back to `is_numeric()` to decide whether to quote a value, which strips quotes from strings like `01234` and silently drops the leading zero on a VARCHAR column. A roundtrip check (`(string)(int)$value` and `(string)(float)$value`) keeps these strings quoted while plain integers stay unquoted.

Fixes #20284.

### DIFF
--- a/libraries/classes/Import.php
+++ b/libraries/classes/Import.php
@@ -1176,20 +1176,24 @@ class Import
                     ) {
                         $tempSQLStr .= (string) $tables[$i][self::ROWS][$j][$k];
                     } else {
+                        $value = (string) $tables[$i][self::ROWS][$j][$k];
                         if ($analyses != null) {
                             $isVarchar = ($analyses[$i][self::TYPES][$colCount] === self::VARCHAR);
                         } else {
-                            $isVarchar = ! is_numeric($tables[$i][self::ROWS][$j][$k])
-                                && ! preg_match('/^0x[0-9a-f]+$/', (string) $tables[$i][self::ROWS][$j][$k]);
+                            $isVarchar = ! preg_match('/^0x[0-9a-f]+$/', $value)
+                                && (
+                                    ! is_numeric($value)
+                                    || ($value !== (string) (int) $value && $value !== (string) (float) $value)
+                                );
                         }
 
                         /* Don't put quotes around NULL fields */
-                        if (! strcmp((string) $tables[$i][self::ROWS][$j][$k], 'NULL')) {
+                        if (! strcmp($value, 'NULL')) {
                             $isVarchar = false;
                         }
 
                         $tempSQLStr .= $isVarchar ? "'" : '';
-                        $tempSQLStr .= $dbi->escapeString((string) $tables[$i][self::ROWS][$j][$k]);
+                        $tempSQLStr .= $dbi->escapeString($value);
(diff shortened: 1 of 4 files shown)


libraries/classes/Import.php
```diff
@@ -1176,20 +1176,24 @@ public function buildSql(
...
+                        $value = (string) $tables[$i][self::ROWS][$j][$k];
                         if ($analyses != null) {
                             $isVarchar = ($analyses[$i][self::TYPES][$colCount] === self::VARCHAR);
                         } else {
-                            $isVarchar = ! is_numeric($tables[$i][self::ROWS][$j][$k])
-                                && ! preg_match('/^0x[0-9a-f]+$/', (string) $tables[$i][self::ROWS][$j][$k]);
+                            $isVarchar = ! preg_match('/^0x[0-9a-f]+$/', $value)
+                                && (
+                                    ! is_numeric($value)
+                                    || ($value !== (string) (int) $value && $value !== (string) (float) $value)
```
williamdes: Can you explain more about this line ?

williamdes (approved):
You work looks good, thank you !

libraries/classes/Import.php
eyupcanakman: Without it, `'01234'` slips through `! is_numeric` and gets unquoted (since `is_numeric` returns true). `(int) '01234'` is `1234`, so the round-trip mismatch is what catches it.

kamil-tekiela:
Isn't the last condition sufficient?

eyupcanakman:
Yeah, you're right, missed that. Simplified.

williamdes (approved):
Thanks for the recent updates !

MauricioFauth (approved):
Thank you!

PR #20316, opened 2026-05-24
Fix #20190 - Handle missing parser statement in getForeigners()

eyupcanakman:
### Description

`getForeigners()` reads `$parser->statements[0]` after parsing the `SHOW CREATE TABLE` output, but the parser returns no statement for some tables, for example a binary column with a binary default value. That raises `Undefined array key 0`. The `instanceof CreateStatement` check right after already handles `null`.

Same pattern on master in `getForeignKeysData()`.

Fixes #20190.


williamdes (approved):
Thanks!

Issue #18974, opened 2024-02-06
6.0: Table tree re-opens/animates everytime + forgets position when changing inner tab 


LordSimal:
### Describe the bug

![2024-02-06_20-04-23 (1)](https://github.com/phpmyadmin/phpmyadmin/assets/9105243/7fe25ffb-c9aa-4924-ad4a-6ac95099a6a3)

There are 2 bugs visible in the video above:

First the table tree re-opens/animates via a collapse animation everytime I go to another "inner tab" like `Search` or `SQL` or `Structure`.

Second the pagination present inside a database with e.g. 100+ tables forgets its position after you change the "inner tab" from e.g. `Search` to `SQL`.

Both of these issues are not present in current 5.2.1

### Expected behavior

PHPMyAdmin should not re-open/animate the table tree after a "inner tab" has been selected.

PHPMyAdmin should remember the previously selected "page" of the paginated table list after selecting a different "inner tab".

### Server configuration

- Operating system: Debian 12
- Web server: NGINX
- Database version: MariaDB10.11.6
- PHP version: 8.3.2
- phpMyAdmin version: 6.0.0-dev+20240127.675dc8de1a

### Client configuration

- Browser: Google Chrome 121.0.6167.139
- Operating system: MacOS 14.3

johnson361:
Version 5.2.1 also contains the bug, whereas bug is not present in 5.2.0 version

williamdes:
> Version 5.2.1 also contains the bug, whereas bug is not present in 5.2.0 version

Indeed, 5.2 is affected too

williamdes:
Maybe it would be due to https://github.com/phpmyadmin/phpmyadmin/commit/36a77f7476f6364a5e2dc60af766f799d6bc0c6e, 38b796fc78d31f5516188d2241d1f0c0775be0a2 or https://github.com/phpmyadmin/phpmyadmin/pull/18675

MauricioFauth:
The animation should only play when interacting with the tree, not when loading the page.

williamdes:
I confirm this bug still exists, tested 6.0.

etchre:
@williamdes Hi, I was able to reproduce this bug (both the tree re-opening and the pagination resetting on tab switch) on the latest dev build of 6.0 [101bd7f](https://github.com/phpmyadmin/phpmyadmin/commit/101bd7fe7eb115290f32a0fbcf85d7a63d13d8d5). I would like to try fixing this issue.

[...]

PR #20330, opened 2026-06-11
Skip navigation tree animation across page reloads

etchre:
### Description

The navigation tree state (open tree nodes and pagination position) is saved to sessionStorage on every change. Then that state gets restored on page load if the current page's `server` and `token` (in CommonParams) match what is stored in the browser's sessionStorage. However, the comparison was always failing because the code was trying to compare an int and a string together with strict equality. This is a side-effect of sessionStorage coercing any value put into it to a string. Casting the server id to a string fixes the comparison, so the tree state is now restored on page load again.

With the tree state properly restoring again, the issue with phpMyAdmin forgetting the selected 'page' of the table list was fixed. However the tree opening animation was still replaying on every page load. This was fixed by adding a `skipTreeAnimation` parameter to `expandTreeNode()` (which defaults to false), so the animation is skipped when the tree state is being restored from sessionStorage. The default behavior is untouched, so manually expanding tables in the table tree still plays the animation.

- Related to #18974

### DIFF
--- a/resources/js/modules/navigation.ts
+++ b/resources/js/modules/navigation.ts
@@ -285,14 +285,20 @@ function traverseForPaths () {
  *
  * @param $expandElem expander
  * @param callback    callback function
+ * @param skipTreeAnimation skip tree opening animation when true
  */
-function expandTreeNode ($expandElem, callback = undefined): void {
+function expandTreeNode ($expandElem, callback = undefined, skipTreeAnimation = false): void {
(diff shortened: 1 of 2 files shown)


MauricioFauth (approved):

MauricioFauth:
That's an interesting improvement, but it doesn't fix the issues described in #18974.
The tree still opens and the pagination still resets after a page load. But this PR reduces the annoyance a little bit.
Thank you for your contribution!

PR #20333, opened 2026-06-17
Monitor autocomplete

M393:
### Description

In Status -> Monitor -> Settings -> Add chart the autocomplete menu for variable names is behind the modal.
image
image

This sets the appendTo property for the autocomplete list to the modal background instead of the default body.
It also limits the height of the list to not overflow the webpage.
And sorts the list alphabetically, the items were in the order in which the db outputs them, `SHOW GLOBAL STATUS` is mostly sorted, but not completely.


MauricioFauth (approved):
Thank you!

PR #20335, opened 2026-06-21
Fix #17869 - Removing the stale table header cell.

### LINKED ISSUE #17869: Extra table header cell incorrectly added
### Describe the bug

View any database table in Browse mode. On the right of the header row there is an extra cell. It's only on the header, not the table body. Image:

Screenshot 2022-11-04 at 13 45 38

In the HTML it's an extra td element like this:

    <td class="d-print-none"><span></span></td>

It doesn't happen on the demo server, but the theming is slightly different on there (in 5.2 it's paler).

### Server configuration

- Operating system: Ubuntu
- Web server: nginx
- Database version: 8.0.29-0ubuntu0.20.04.3
- PHP version: 7.4.3
- phpMyAdmin version: 5.2.0

### Client configuration

- Browser: Chrome/Firefox
- Operating system: macOS

MatthewOscar:
### Description

In Browse mode, the results `<thead>` gained one extra cell on the right that
the body rows lacked. `getColumnAtRightSide()` emitted a right-side header cell
whenever `RowActionLinks` was `left` (the default), due to an operator-precedence
bug (`A || (B && …)`) and the wrong position constant in the branch guard.

Fix: mirror the correct sibling `getFieldVisibilityParams()` by gating both
branches on a `$rightOrBoth` check, so the right-side cell is only produced for
`right`/`both` layouts — matching the body. The placeholder is also emitted as
`<th>` (valid `<thead>` markup). Updated the stale `ResultsTest` expectation and
added a regression test across all four `RowActionLinks` values.

The second commit switches the left-side empty placeholder from `<td>` to `<th>`
for the same validity reason (the lone `<td>` left in the header); happy to drop
it if you'd prefer to keep this strictly scoped.

> 🤖 AI Disclaimer: Claude Opus 4.8 (1M context) was used in drafting the code for this pull request.

Fixes #17869 

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,7 @@ phpMyAdmin - ChangeLog
 - issue        Improve table list access error messages when available
 - issue #19964 Fix error when importing a CSV with more decimals than the column
 - issue #19950 Fix tooltip doesn't disappear when searching on date type
+- issue #17869 Fix extra cell incorrectly added to the browse results table header
 
 5.2.3 (2025-10-07)
 - issue #19548 Fix missing tooltip in status monitor log table
--- a/libraries/classes/Display/Results.php
+++ b/libraries/classes/Display/Results.php
@@ -1974,11 +1974,13 @@ class Results
         $rightColumnHtml = '';
         $displayParams = $this->properties['display_params'];
 
+        $rightOrBoth = $GLOBALS['cfg']['RowActionLinks'] === self::POSITION_RIGHT
+            || $GLOBALS['cfg']['RowActionLinks'] === self::POSITION_BOTH;
+
         // Displays the needed checkboxes at the right
         // column of the result table header if possible and required...
         if (
-            ($GLOBALS['cfg']['RowActionLinks'] === self::POSITION_RIGHT)
-            || ($GLOBALS['cfg']['RowActionLinks'] === self::POSITION_BOTH)
+            $rightOrBoth
             && (($displayParts['edit_lnk'] != self::NO_EDIT_OR_DELETE)
             || ($displayParts['del_lnk'] != self::NO_EDIT_OR_DELETE))
             && ($displayParts['text_btn'] == '1')
@@ -1991,20 +1993,17 @@ class Results
                 . $fullOrPartialTextLink
                 . '</th>';
         } elseif (
-            ($GLOBALS['cfg']['RowActionLinks'] === self::POSITION_LEFT)
-            || ($GLOBALS['cfg']['RowActionLinks'] === self::POSITION_BOTH)
-            && (($displayParts['edit_lnk'] === self::NO_EDIT_OR_DELETE)
-            && ($displayParts['del_lnk'] === self::NO_EDIT_OR_DELETE))
-            && (! isset($GLOBALS['is_header_sent']) || ! $GLOBALS['is_header_sent'])
+            $rightOrBoth
+            && (($displayParts['edit_lnk'] != self::NO_EDIT_OR_DELETE)
+            || ($displayParts['del_lnk'] != self::NO_EDIT_OR_DELETE))
         ) {
             //     ... elseif no button, displays empty columns if required
-            // (unless coming from Browse mode print view)
 
             $displayParams['emptyafter'] = ($displayParts['edit_lnk'] != self::NO_EDIT_OR_DELETE)
                 && ($displayParts['del_lnk'] != self::NO_EDIT_OR_DELETE) ? 4 : 1;
 
-            $rightColumnHtml .= "\n" . '<td class="d-print-none"' . $colspan
-                . '></td>';
+            $rightColumnHtml .= "\n" . '<th class="d-print-none"' . $colspan
+                . '></th>';
         }
 
         $this->properties['display_params'] = $displayParams;
--- a/libraries/classes/Display/Results.php
+++ b/libraries/classes/Display/Results.php
@@ -1330,7 +1330,7 @@ class Results
 
             $displayParams['emptypre'] = $emptyPreCondition ? 4 : 0;
 
-            $buttonHtml .= '<td' . $colspan . '></td>';
+            $buttonHtml .= '<th' . $colspan . '></th>';
         } elseif ($GLOBALS['cfg']['RowActionLinks'] === self::POSITION_NONE) {
             // ... elseif display an empty column if the actions links are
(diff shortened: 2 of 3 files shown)


williamdes:
Thanks for the PR, it looks good.
But I am unsure if there is any impact after this changes.
Did you use AI ? Please disclose it in the PR body, no issue about that :)

MatthewOscar:
I will look into why the tests are failing and resolve the conflicts

Yes, I used Claude Code Opus to help. I will be updating the PR body soon. I think the scope of the changes is pretty tight as can be.

williamdes:
> 🤖 Review written by me with Claude Opus 4.8 (1M context); I read it through and it matches my understanding of this code.

Thanks for the PR! The stated fix (extra trailing `<thead>` cell on the default `left` layout) is real and correct. But the rewrite of the first `if` adds an `&& hasLinks && text_btn` guard the original `RIGHT` catch-all never had, so it changes more than the `left` case.

With PHP precedence (`&&` binds tighter than `||`), the original first branch was:

```php
RIGHT  ||  (BOTH && hasLinks && text_btn)   // → for RIGHT, ALWAYS true
```

The new one:

```php
$rightOrBoth && hasLinks && text_btn        // RIGHT now requires links + button
```

### R1 — Likely regression, and not covered by the new test

For `$cfg['RowActionLinks'] = 'right'` on a result with **no edit/delete links but `text_btn === '1'`**, the old code rendered the **"Full texts / Partial texts" toggle** (`$fullOrPartialTextLink`) in the right header cell. The new code falls through *both* branches (`hasLinks` is false in each) and emits **nothing** → the toggle disappears.

This case is reachable for: multi-table JOINs with TEXT/BLOB columns (`Results.php:577-578` nulls edit/del when columns span tables), keyless tables, `SHOW` statements (`text_btn=1`), and maintenance statements. On `left`/`both` the left side already emits no button for this case (`getFieldVisibilityParams` branch 1), so the right-side catch-all was the *only* place that button lived for these results.

The new test sets `'text_btn' => '0'` **with** edit+del present, so it exercises the empty-placeholder path, not the `text_btn=1 && noEdit && noDel` path — the regression is untested.

### R2 — Good side effect worth noting

For `left`, the old code set `emptyafter = 4`, which leaked into `getRepeatingHeaders()` (`Results.php:2234`): every `RepeatCells` rows (default 100) the repeating header got a spurious right `<th colspan=4>`. The fix makes `emptyafter` stay `0`, so repeating headers align too — correct, but a visible change on >100-row browse with `left` actions.

### R3 — `<td>`→`<th>` vs `makegrid.js` (low)

Grid-edit reads the first header cell:

```js
// js/src/makegrid.js:2237
if (! $(t).find('tr').first().find('th').first().hasClass('draggable')) {
    g.actionSpan = $(t).find('tr').first().find('th').first().prop('colspan');
```

The left placeholder went `<td colspan>` → `<th colspan>`. In the narrow case that branch is reached (`leftOrBoth && hasLinks && text_btn!=1`), `makegrid` would now treat that empty placeholder as the action header and read its colspan. Practically near-unreachable (SELECT forces `text_btn=1`), but it's the one spot where the markup change is functionally observable — worth a quick inline grid-edit check.

### Safe / no-op

The removed `is_header_sent` guard is dead — `is_header_sent` is never set anywhere in `libraries/` or `src/`, so it was always true. The dropped "print view" comment is moot too, since `getColumnAtRightSide` is only called when `! $printView` (`Results.php:1178`).

### Suggested before merge

1. Manually test `RowActionLinks = 'right'` with a multi-table JOIN that includes a TEXT column (and a `SHOW`) — confirm the Full/Partial texts button still shows. If not, the right branch needs a clause to emit the button when `text_btn==1` even without links (mirroring the old catch-all).
2. Add a test for `text_btn='1' && noEdit && noDel` across all four positions.

MatthewOscar:
@williamdes I have addressed the suggested changes and implemented them. What do you think?

MatthewOscar:
@williamdes 

The `red lint-node `/` Test javascript files `/` build-release` jobs aren't from this PR but from the Update Yarn dependencies commit on `QA_5_2` ([0cab39f6](https://github.com/phpmyadmin/phpmyadmin/commit/0cab39f69372e548f56f06978906ab8ad6861cbb)), which bumped [node-releases to 2.0.48](https://github.com/phpmyadmin/phpmyadmin/blob/0cab39f69372e548f56f06978906ab8ad6861cbb/yarn.lock#L3877) (engines: node >=18) while CI still pins node 12 ([failing run](https://github.com/phpmyadmin/phpmyadmin/actions/runs/27997296730/job/82987885041)). This PR only touches PHP/test files. Happy to send a separate PR bumping the CI node version if helpful.

williamdes:
> 🤖 Follow-up written by me with Claude Opus 4.8 (1M context); reviewed by me.

Thanks for the quick turnaround — the new commit resolves the R1 concern cleanly. Confirming the logic across all layouts:

- **right/both + no edit/delete links + `text_btn=1`** (JOINs / `SHOW` / keyless): the full/partial-text toggle is back in the right header, and the new `getTableBody()` `elseif` emits a matching empty `<td class="d-print-none">` so every body row stays aligned with it.
- **left/none**: still nothing on the right — header and body agree.
- `emptyafter`/repeating-header alignment holds in each case (1 right cell when the toggle/links are present, 0 otherwise).

The new `testGetColumnAtRightSideShowsTextButtonWithoutEditDeleteLinks` covers exactly the case I was worried about. 👍

Two non-blockers:
1. The `<td>`→`<th>` change on the left placeholder (`getFieldVisibilityParams` branch 3) is fine — that branch is effectively unreachable for SELECTs (`text_btn` is always `1`), so `makegrid.js`'s `find('th').first()` action-column detection is unaffected in practice.
2. Same buggy `||`/`&&` precedence exists on `master` (6.0) — worth forward-porting this fix.

LGTM from my side once CI is green.

MatthewOscar:
@williamdes I rebased this branch onto the latest `QA_5_2`, which includes `fea3336d64` (`Downgrade node-releases package to <2.0.45`). That should address the previous CI failures from `node-releases@2.0.48` requiring Node >=18 while CI used Node 12. Once it runs, we should expect to see all checks passing!

MatthewOscar:
@williamdes All checks have passed! Is this ready to be merged?

MatthewOscar:
@williamdes Just want to notify you again since a week has passed. Is everything good to go?

PR #20339, opened 2026-06-24
Fix #20336 - Add support for UUID_v4() and UUID_v7()

### LINKED ISSUE #20336: [Feat]: Default values for UUID_v4() and UUID_v7() in MariaDB 11.7+
### phpMyAdmin version

5.2.3

### Describe the feature

When adding or altering a row of type UUID, there is only a default value exposed for UUID v1.

It would be useful to expose defaults for UUID v4 and UUID v7 under the defaults menu for MariaDB 11.7+

Also, when manually changing the default using ALTER, the new default function is treated as a string, which then errors #1067 - Invalid default value for 'uuid'.
```ALTER TABLE `test` CHANGE `uuid` `uuid` UUID NOT NULL DEFAULT 'uuid_v4()' COMMENT 'x';```

### Describe the solution

Add UUID_V4 and UUID_V7 under the defaults menu, similar to how the current UUID currently functions

UUID: ```ALTER TABLE `test` CHANGE `uuid` `uuid` UUID NOT NULL DEFAULT uuid();```
UUID_V4: ```ALTER TABLE `test` CHANGE `uuid` `uuid` UUID NOT NULL DEFAULT uuid_v4();```
UUID_V7: ```ALTER TABLE `test` CHANGE `uuid` `uuid` UUID NOT NULL DEFAULT uuid_v7();```

### Describe the alternatives

_No response_

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by liviuconcioiu**
### Description

Add support for UUID_v4() and UUID_v7()

https://github.com/user-attachments/assets/dca11489-2abe-48d3-b76f-fb83d516291e

https://github.com/user-attachments/assets/0822e8a6-46a9-46f8-aa1a-22895eec7eba

Fixes #20336

### DIFF
--- a/libraries/classes/Controllers/Table/ReplaceController.php
+++ b/libraries/classes/Controllers/Table/ReplaceController.php
@@ -159,6 +159,8 @@ final class ReplaceController extends AbstractController
             'UTC_TIMESTAMP',
             'UUID',
             'UUID_SHORT',
+            'UUID_v4',
+            'UUID_v7',
             'VERSION',
         ];
         $func_optional_param = [
--- a/libraries/classes/InsertEdit.php
+++ b/libraries/classes/InsertEdit.php
@@ -1843,6 +1843,22 @@ class InsertEdit
             $currentValue = 'uuid()';
         }
 
+        if (
+            $type === 'uuid'
+                && ! isset($multiEditColumnsNull[$key])
+                && $currentValue === "'uuid_v4()'"
+        ) {
+            $currentValue = 'uuid_v4()';
+        }
+
+        if (
+            $type === 'uuid'
+                && ! isset($multiEditColumnsNull[$key])
+                && $currentValue === "'uuid_v7()'"
+        ) {
+            $currentValue = 'uuid_v7()';
+        }
+
         return $currentValue;
     }
 
--- a/libraries/classes/Query/Compatibility.php
+++ b/libraries/classes/Query/Compatibility.php
@@ -234,7 +234,7 @@ class Compatibility
      */
     public static function isUUIDv4Supported(DatabaseInterface $dbi): bool
     {
-        // @see: https://mariadb.com/docs/server/reference/sql-functions/secondary-functions/miscellaneous-functions/uuid_v4
+        // https://mariadb.com/docs/server/reference/sql-functions/secondary-functions/miscellaneous-functions/uuid_v4
         return $dbi->isMariaDB() && $dbi->getVersion() >= 110700; // 11.7.0
     }
 
@@ -245,7 +245,7 @@ class Compatibility
      */
     public static function isUUIDv7Supported(DatabaseInterface $dbi): bool
     {
-        // @see: https://mariadb.com/docs/server/reference/sql-functions/secondary-functions/miscellaneous-functions/uuid_v7
+        // https://mariadb.com/docs/server/reference/sql-functions/secondary-functions/miscellaneous-functions/uuid_v7
         return $dbi->isMariaDB() && $dbi->getVersion() >= 110700; // 11.7.0
     }
 
--- a/libraries/classes/Query/Compatibility.php
+++ b/libraries/classes/Query/Compatibility.php
@@ -228,6 +228,28 @@ class Compatibility
     }
 
     /**
+     * Check whether the database supports UUID_v4 function
+     *
+     * @return bool true if UUID_v4 is supported
+     */
+    public static function isUUIDv4Supported(DatabaseInterface $dbi): bool
+    {
+        // @see: https://mariadb.com/docs/server/reference/sql-functions/secondary-functions/miscellaneous-functions/uuid_v4
+        return $dbi->isMariaDB() && $dbi->getVersion() >= 110700; // 11.7.0
+    }
+
+    /**
+     * Check whether the database supports UUID_v7 function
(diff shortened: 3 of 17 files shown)

### CODE REVIEW

**Review by williamdes — APPROVED** [review]
Thank you !

**Comment by williamdes** [comment]
> 🤖 Review written by me with Claude Opus 4.8 (1M context); I read it through and it matches my understanding of this code.

Thanks @liviuconcioiu! I verified against the MariaDB docs that `UUID_v4()`/`UUID_v7()` are **MariaDB 11.7+ only** (MySQL has neither). With that as ground truth, the version/vendor gating is inconsistent and there's a copy-paste bug in the tests.

### 🔴 Bug 1 — wrong version constant in `Types.php` (`110070` ≠ `110700`)

Versions are encoded as `major*10000 + minor*100 + patch`, so 11.7.0 = `110700`. The new gate uses `110070`, which is **11.0.70**:

```php
// libraries/classes/Types.php (getFunctionsClass, CHAR)
if ($isMariaDB && $serverVersion < 110070) {        // ← 11.0.70, should be 110700
    $ret = array_diff($ret, ['UUID_v4', 'UUID_v7']);
}
```

The sibling `Compatibility` helpers correctly use `110700`. **Effect:** on MariaDB 11.0.70 → 11.6.x the functions are not removed, so they show up in the insert/edit Function dropdown where they don't exist.

### 🔴 Bug 2 — `Types.php` never excludes MySQL

`UUID_v4`/`UUID_v7` are added to the `CHAR` list unconditionally and only stripped for old *MariaDB*. On **MySQL** (`$isMariaDB === false`) the `if` is skipped, so both are offered on every MySQL version. The guard needs vendor + version:

```php
if (! $isMariaDB || $serverVersion < 110700) {
    $ret = array_diff($ret, ['UUID_v4', 'UUID_v7']);
}
```

The new `TypesByDatabaseVersionTest` only covers `mariadb 11.7.0` (the passing case) — no MySQL or MariaDB-11.6 case, so both gating bugs slip through CI. The `column_attributes.twig` *Default* dropdown is gated correctly via `is_uuid_v4_supported()`/`is_uuid_v7_supported()`; only the function-list path is broken.

### 🔴 Bug 3 — test asserts the wrong method

`UtilTest::testIsUUIDv7Supported()` calls the **v4** helper:

```php
public function testIsUUIDv7Supported(...): void {
    ...
    self::assertSame(Util::isUUIDv4Supported(), $expected);   // ← should be isUUIDv7Supported()
}
```

It only passes because both thresholds happen to be identical — the v7 helper is effectively untested.

### 🟠 Worth verifying — paren-less `DEFAULT uuid_v4()`

`Table::generateFieldSpec()` emits `DEFAULT uuid_v4()` / `DEFAULT uuid_v7()` (mirroring the existing bare `DEFAULT uuid()`), but the MariaDB docs write expression defaults with parentheses: `... DEFAULT (uuid_v7())`. Bare `uuid()` is grandfathered; please confirm on a real 11.7 server that the bare form is accepted for v4/v7 too, otherwise the generated DDL will error.

### 🟡 Minor
- The two new `UtilTest` cases use `assertSame($actual, $expected)` (arguments reversed — only affects failure messages); the `@param` docblock order also doesn't match the signature.
- `isUUIDv4Supported`/`isUUIDv7Supported` are byte-identical — fine to keep separate, just noting.

Sources: [UUID_v4](https://mariadb.com/docs/server/reference/sql-functions/secondary-functions/miscellaneous-functions/uuid_v4), [UUID_v7](https://mariadb.com/docs/server/reference/sql-functions/secondary-functions/miscellaneous-functions/uuid_v7)

williamdes:
Edit: sorry I asked for a comment update instead of spam a new one..

> 🤖 Follow-up written by Claude Opus 4.8 (1M context); 

Thanks — all three blocking items are fixed:

- ✅ **Bug 1**: `Types.php` now gates on `110700` (11.7.0).
- ✅ **Bug 2**: now `if (! $isMariaDB || $serverVersion < 110700)`, so MySQL no longer gets `UUID_v4`/`UUID_v7`.
- ✅ **Bug 3**: `UtilTest::testIsUUIDv7Supported()` now asserts `Util::isUUIDv7Supported()`.

One item still open:

🟠 **`Table::generateFieldSpec()` still emits the bare `DEFAULT uuid_v4()` / `DEFAULT uuid_v7()`** (no parentheses), mirroring `DEFAULT uuid()`. MariaDB's docs write expression defaults wrapped: `... DEFAULT (uuid_v7())`. Bare `uuid()` is grandfathered, but please confirm on a real MariaDB 11.7 server that the paren-less form is accepted for v4/v7 — otherwise the generated `CREATE`/`ALTER` will error at runtime (the unit tests only assert the generated string, not that the server accepts it).

Optional nicety: `TypesByDatabaseVersionTest` only covers the positive `mariadb 11.7.0` case — a MySQL and a MariaDB-11.6 case would lock in the Bug 1/2 gating against regressions. And the two new `UtilTest` `assertSame()` calls have `actual`/`expected` reversed (cosmetic, only affects failure output).

liviuconcioiu:
So.. everything is okay? Screen recording shows how it looks.

williamdes (approved):
Yup, the code looks okay to me

I am unsure if this should go to 5.2 or 6.0 but anyway

MauricioFauth (approved):
Thank you!

PR #19540, opened 2025-01-18
Implement getValuesForCsvTable for CSV table type

liviuconcioiu:
### Description

If table type is CSV, the row count always shows 2 and the size is unknown, even if there are rows in it. It will only count the rows, if the table name is clicked, and then go back to database.

Based on https://bugs.mysql.com/bug.php?id=53929

Before:

csv0

After:

csv1

csv2

Before submitting pull request, please review the following checklist:

- [x] Make sure you have read our [CONTRIBUTING.md](https://github.com/phpmyadmin/phpmyadmin/blob/master/CONTRIBUTING.md) document.
- [x] Make sure you are making a pull request against the correct branch. For example, for bug fixes in a released version use the corresponding QA branch and for new features use the _master_ branch. If you have a doubt, you can ask as a comment in the bug report or on the mailing list.
- [x] Every commit has proper `Signed-off-by` line as described in our [DCO](https://github.com/phpmyadmin/phpmyadmin/blob/master/DCO). This ensures that the work you're submitting is your own creation.
- [x] Every commit has a descriptive commit message.
- [x] Every commit is needed on its own, if you have just minor fixes to previous commits, you can squash them.
- [ ] Any new functionality is covered by tests.

PR #20356, opened 2026-07-04
Csv table size

MoonE:
### Description

Fix csv table size calculation with multi-byte characters, and make it work with column names containing special characters.

See #19540


MauricioFauth (approved):
Thank you!

PR #20361, opened 2026-07-05
Use utf8mb4 when creating pma tables

MoonE:
### Description

Create pma database tables with utf8mb4 instead of utf8[mb3].

With only utf8mb3 some data cannot be saved.
For example if you have the history feature enabled, a query like `SELECT '🐛';` causes phpMyAdmin to error because inserting into pma__history fails.

### DIFF
--- a/sql/create_tables.sql
+++ b/sql/create_tables.sql
@@ -18,7 +18,7 @@
 -- Database : `phpmyadmin`
 --
 CREATE DATABASE IF NOT EXISTS `phpmyadmin`
-  DEFAULT CHARACTER SET utf8 COLLATE utf8_bin;
+  DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_bin;
 USE phpmyadmin;
 
 -- --------------------------------------------------------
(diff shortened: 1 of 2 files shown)


MauricioFauth (approved):
The charset `utf8mb4` is supported by all MySQL/MariaDB versions phpMyAdmin supports. I think it's okay to change it. What do you think @williamdes?

MoonE:
Should there be some kind of update script that can be run or some other mechanic? How have schema updates been handled in the past?

williamdes:
> The charset `utf8mb4` is supported by all MySQL/MariaDB versions phpMyAdmin supports. I think it's okay to change it. What do you think @williamdes?

Thanks for that research, okay then

ibennetch:
> How have schema updates been handled in the past?

Usually we ship a .sql file with the upgrade script (such as https://github.com/phpmyadmin/phpmyadmin/blob/master/resources/sql/upgrade_tables_4_7_0%2B.sql).

Within the code, we either detect when the upgrade hasn't occurred and show a warning on the main page, or we gracefully handle both scenarios; whether or not the upgrade has occurred.

williamdes:
> > How have schema updates been handled in the past?
> 
> Usually we ship a .sql file with the upgrade script (such as https://github.com/phpmyadmin/phpmyadmin/blob/master/resources/sql/upgrade_tables_4_7_0%2B.sql).
> 
> Within the code, we either detect when the upgrade hasn't occurred and show a warning on the main page, or we gracefully handle both scenarios; whether or not the upgrade has occurred.

That's true, maybe we should revert this and make a schema update 🤔

MoonE:
Creating new tables with utf8mb4 should be no problem. So no need to revert, regardless of how updating old tables will be handled.

PR #20353, opened 2026-07-04
Cache foreign key display values and add limit 1 to query

MoonE:
### Description

Add LIMIT 1 to query getting foreign key display values and cache these values to avoid identical queries.

See also #18728, #19045

Performance test, Tested on MariaDB 11.8:
```sql
DROP TABLE IF EXISTS child;
DROP TABLE IF EXISTS parent;

CREATE TABLE `parent` (
  `id` INT(11) NOT NULL AUTO_INCREMENT,
  `k1` int(11) NOT NULL,
  `k2` int(11) NOT NULL,
  `name` CHAR(36) NOT NULL DEFAULT (UUID()) NULL,
  PRIMARY KEY (`id`),
  KEY (`k1`, `k2`)
) ENGINE=InnoDB;

CREATE TABLE `child` (
  `id` INT(11) NOT NULL AUTO_INCREMENT, 
  `parent_k1` int(11) NOT NULL,
  `parent_k2` int(11) NOT NULL,
  PRIMARY KEY (`id`),
  KEY `child_ibfk_1` (`parent_k1`,`parent_k2`),
  CONSTRAINT `child_ibfk_1` FOREIGN KEY (`parent_k1`, `parent_k2`) REFERENCES `parent` (`k1`, `k2`) ON DELETE CASCADE ON UPDATE CASCADE
) ENGINE=InnoDB;

INSERT INTO parent (k1, k2) VALUES (1, 1);
INSERT INTO parent (k1, k2) SELECT k1, k2 + POW(2, 0) FROM parent;
INSERT INTO parent (k1, k2) SELECT k1, k2 + POW(2, 1) FROM parent;
INSERT INTO parent (k1, k2) SELECT k1, k2 + POW(2, 2) FROM parent;
INSERT INTO parent (k1, k2) SELECT k1, k2 + POW(2, 3) FROM parent;
INSERT INTO parent (k1, k2) SELECT k1, k2 + POW(2, 4) FROM parent;
INSERT INTO parent (k1, k2) SELECT k1, k2 + POW(2, 5) FROM parent;
INSERT INTO parent (k1, k2) SELECT k1, k2 + POW(2, 6) FROM parent;
INSERT INTO parent (k1, k2) SELECT k1, k2 + POW(2, 7) FROM parent;
INSERT INTO parent (k1, k2) SELECT k1, k2 + POW(2, 8) FROM parent;
INSERT INTO parent (k1, k2) SELECT k1, k2 + POW(2, 9) FROM parent;
[...]

PR #20372, opened 2026-07-12
Add LIMIT 1 to some queries

MoonE:
InsertEdit::getDisplayValueForForeignTableColumn is called when a cell is edited and it suffered from the same issue that #20353 fixes with a LIMIT 1 at a different code location.

The rest of the changes are probably only useful for tables that have no unique key. 
In that case all columns are used as an identifier and there may be duplicate rows, but we need only one of the rows.


kamil-tekiela (approved):

MauricioFauth (approved):
Thank you!

PR #20373, opened 2026-07-12
Use vanilla locutus for sprintf

MoonE:
### Description

Use the upstream locutus version instead of a fork.
The fork was created to be able to import sprintf with a script tag, but that's no longer necessary.

The original function signature returns false on error, instead of the old declaration that only returned string, but that should never happen. So I added type casts to string where it was required.


MauricioFauth (approved):
Thank you!

PR #20374, opened 2026-07-12
Fix #18510 - Use database and table aliases in export filename template

### LINKED ISSUE #18510: Exported filename does not use renamed aliases
### Describe the bug

When exporting a file, if you use aliases (by clicking the rename option and adding aliases), then use @database@ and @table@ in the filename template, it uses the original values not the aliases.

TowyTowy:
### Description

When exporting with the *rename exported db/table/columns* option and a filename template that uses `@DATABASE@` / `@TABLE@`, the generated filename used the **original** database and table names instead of the aliases the user entered.

`Export::getFilenameAndMimetype()` built the filename via `Util::expandUserString($filenameTemplate)` with no variable overrides, so `@DATABASE@` and `@TABLE@` always resolved to `$GLOBALS['db']` / `$GLOBALS['table']`. The export aliases were already assembled in `ExportController` but were never passed to the filename generator.

This passes the aliases down and, when an alias is defined for the current database or table, uses it as the replacement for `@DATABASE@` / `@TABLE@`. Database/server exports (empty `$table`) are unaffected.

Adds a regression test that fails on current `QA_5_2` (`test_db-test_table.php`) and passes with the fix (`aliasdb-aliastbl.php`); phpcs clean.

Fixes #18510

This change was prepared with the assistance of AI (Claude); it has been reviewed and tested, and I take responsibility for it.

- [x] Make sure you are making a pull request against the correct branch (QA_5_2 — the bug affects released 5.2 and forward-merges to master).
- [x] Every commit has proper `Signed-off-by` line as described in our DCO.
- [x] Every commit is needed on its own.

### DIFF
--- a/libraries/classes/Controllers/Export/ExportController.php
+++ b/libraries/classes/Controllers/Export/ExportController.php
@@ -400,7 +400,8 @@ final class ExportController extends AbstractController
                 $remember_template,
                 $export_plugin,
                 $compression,
-                $filename_template
+                $filename_template,
+                $aliases
             );
         } else {
             $mime_type = '';
--- a/libraries/classes/Export.php
+++ b/libraries/classes/Export.php
@@ -335,7 +335,7 @@ class Export
      * @param ExportPlugin $exportPlugin     the export plugin
      * @param string       $compression      compression asked
      * @param string       $filenameTemplate the filename template
-     * @param array        $aliases          alias information for db/table/column
+     * @param mixed[]      $aliases          alias information for db/table/column
      *
      * @return string[] the filename template and mime type
      */
--- a/libraries/classes/Export.php
+++ b/libraries/classes/Export.php
@@ -335,6 +335,7 @@ class Export
      * @param ExportPlugin $exportPlugin     the export plugin
      * @param string       $compression      compression asked
      * @param string       $filenameTemplate the filename template
+     * @param array        $aliases          alias information for db/table/column
      *
      * @return string[] the filename template and mime type
      */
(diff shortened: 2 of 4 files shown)


TowyTowy:
Pushed a small follow-up: Psalm (`analyse-php`, errorLevel 1) flagged two `MixedAssignment` errors because the new `$aliases` parameter is an untyped `array`, so the alias reads are `mixed`. Cast both to `(string)` before assigning into `$updates` (the values are always strings, guarded by `! empty()`), which keeps Psalm green without adding baseline entries. phpcs and the new unit test still pass.

TowyTowy:
Follow-up for `analyse-php`: rather than the `(string)` casts (which are the wrong tool for PHPStan and bumped the counted baseline), I properly typed the `$aliases` parameter (array shape via an `@var`) and switched the alias substitution to strict `isset()` / `!== ''` checks instead of `empty()`. That removes the "cast mixed to string", "offset on mixed", and "empty() not allowed" occurrences at the source, so the PHPStan baseline counts match again. PHPStan is clean on the changed files and the alias unit test still passes.

TowyTowy:
Reworked the analyse-php fix once more, this time verified against both tools on PHP 7.2 (the CI environment). The alias reads now use the same pattern as the existing readers in `Export.php` (`! empty()` + `(string)`), with the resulting PHPStan notices absorbed by the counted baseline like their pre-existing counterparts, and the `$aliases` param documented as `mixed[]`. Psalm needed no baseline changes and reports no errors at level 1; PHPStan is green. Sorry for the churn on this one.

PR #20224, opened 2026-03-07
Skip tables that cannot be read

liviuconcioiu:
### Description

This should fix the export for all other formats, except `SQL`, when tables cannot be read. Previously `HTML` content was added to the exported file. Now, it will skip the tables and only export what can be read.

Something similar was done here:

https://github.com/phpmyadmin/phpmyadmin/blob/559587a6a4160911fec4a3b15b4b4f2ecffabf43/libraries/classes/Table.php#L1303-L1304
https://github.com/phpmyadmin/phpmyadmin/blob/2f4a40d2082b04a8c53d35e19f21ffc85749c279/src/Table/TableMover.php#L452-L455

**Before:**

https://github.com/user-attachments/assets/07b35dbf-e761-40da-b09c-ffe23bde7aeb

**After:**

https://github.com/user-attachments/assets/1934ac86-db8c-491f-b0ce-8033e06f42e4

Related #18028, #18183, #20160

Before submitting pull request, please review the following checklist:

- [x] Make sure you have read our [CONTRIBUTING.md](https://github.com/phpmyadmin/phpmyadmin/blob/master/CONTRIBUTING.md) document.
- [x] Make sure you are making a pull request against the correct branch. For example, for bug fixes in a released version use the corresponding QA branch and for new features use the _master_ branch. If you have a doubt, you can ask as a comment in the bug report or on the mailing list.
- [x] Every commit has proper `Signed-off-by` line as described in our [DCO](https://github.com/phpmyadmin/phpmyadmin/blob/master/DCO). This ensures that the work you're submitting is your own creation.
- [x] Every commit has a descriptive commit message.
- [x] Every commit is needed on its own, if you have just minor fixes to previous commits, you can squash them.
- [ ] Any new functionality is covered by tests.

PR #20379, opened 2026-07-14
Fix #20160 - Skip tables that cannot be read in SQL structure export

eyupcanakman:
### Description

When a single table cannot be read (for example an InnoDB table on a server started with `skip-innodb`), the whole SQL export failed with an `Error reading structure for table` error instead of exporting the readable tables. The export loop now skips a table whose columns cannot be read and continues, the same way #20224 handles the data export.

Fixes #20160.

### DIFF
--- a/libraries/classes/Export.php
+++ b/libraries/classes/Export.php
@@ -812,6 +812,11 @@ class Export
                         }
                     }
 
+                    // Skip tables that cannot be read
+                    if ($tableObject->getColumns() === []) {
+                        continue;
+                    }
+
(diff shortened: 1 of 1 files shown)


williamdes:
@liviuconcioiu I think you have such tables more often, does the fix look acceptable ?
For me it looks like not logging a mistake
@kamil-tekiela also ?

liviuconcioiu:
The fix looks okay, since tables are skipped. The error no longer shows when exporting the test database.

PR #20388, opened 2026-07-20
Use class strings instead of strings

kamil-tekiela:
The `UploadSession` class hasn't been used in a long time, so it might be a good idea to remove it finally.


MauricioFauth (approved):
Thank you!

PR #20396, opened 2026-07-26
Fix quoting of INET search values

### LINKED ISSUE #20395: [Bug]: IP address is not quoted when searching in a field of type INET4
### Describe the bug

When searching in a field of type `INET4` no quotes are put around the entered IP address when using the `=` option in the dropdown list.
The problem can be circumvented by selecting `LIKE` from the drop down list instead of `=`, but that seems kind of hacky.

### How to Reproduce

1. Create a table with a INET4 field:
```sql
CREATE TABLE test(ip INET4)
```

2. Use the PHPMyaAdmin search tab to search for 1.2.3.4 in the ip field

You will get an error stating that the SQL query below is incorrect:
```sql
SELECT * FROM `test` WHERE `ip` = 1.2.3.4 LIMIT 0, 25
```

The query should have quotes around the ip address like so:
```sql
SELECT * FROM `test` WHERE `ip` = '1.2.3.4' LIMIT 0, 25
```

The problem can be circumvented by selecting `LIKE` from the drop down list instead of `=`, but that seems kind of hacky.

### Expected behavior

The query should have quotes around the ip address like so:
```sql
SELECT * FROM `test` WHERE `ip` = '1.2.3.4' LIMIT 0, 25
```

### Screenshots

_No response_

### Operating System

Linux

### Web Server

Apache/2.4.68

### Database Server

MariaDB

### Database version

12.3.2-MariaDB-ubu2404

### PHP version

8.3.32

### phpMyAdmin version

5.2.3

### browser

_No response_

### Additional context

_No response_

Solaris-star:
### Description

Fixes #20395.

The table search query builder treats MariaDB `INET4` and `INET6` columns as unquoted numeric types when using operators such as `=`. This generates invalid SQL such as:

```sql
SELECT * FROM `test` WHERE `ip` = 1.2.3.4
```

This adds `INET4` and `INET6` to the existing set of string-like column types that require quoted and escaped search values. Numeric columns keep their current unquoted behavior.

A regression test covers both address families:

```sql
SELECT * FROM `hosts` WHERE `ipv4` = '1.2.3.4' AND `ipv6` = '2001:db8::1'
```

Before the fix, the test failed with both values unquoted. After the fix:

- `SearchTest`: 7 tests, 17 assertions pass.
- PHPCS passes for both changed files.
- PHPStan passes for both changed files with no errors.
- `git diff --check` passes.

Target branch: `QA_5_2`, because the report affects phpMyAdmin 5.2.3.

- [x] I have read `CONTRIBUTING.md`.
- [x] This bug fix targets the corresponding `QA_5_2` branch.
- [x] Every commit contains the required Signed-off-by line.
- [x] The commit message is descriptive and the commit is independently needed.
- [x] The changed behavior is covered by a regression test.

### DIFF
--- a/libraries/classes/Table/Search.php
+++ b/libraries/classes/Table/Search.php
@@ -167,7 +167,7 @@ final class Search
             // strings to numbers and numbers to strings as necessary
             // during the comparison
             if (
-                preg_match('@char|binary|blob|text|set|date|time|year|uuid@i', $types)
+                preg_match('@char|binary|blob|text|set|date|time|year|uuid|inet[46]@i', $types)
                 || mb_strpos(' ' . $func_type, 'LIKE')
             ) {
                 $quot = '\'';
--- a/test/classes/Table/SearchTest.php
+++ b/test/classes/Table/SearchTest.php
@@ -203,6 +203,21 @@ class SearchTest extends AbstractTestCase
         );
     }
 
+    public function testBuildSqlQueryWithWhereClauseInet(): void
+    {
+        $_POST['zoom_submit'] = true;
+        $_POST['table'] = 'hosts';
+        $_POST['criteriaColumnNames'] = ['ipv4', 'ipv6'];
+        $_POST['criteriaColumnOperators'] = ['=', '='];
+        $_POST['criteriaValues'] = ['1.2.3.4', '2001:db8::1'];
+        $_POST['criteriaColumnTypes'] = ['inet4', 'inet6'];
+
+        self::assertSame(
+            "SELECT * FROM `hosts` WHERE `ipv4` = '1.2.3.4' AND `ipv6` = '2001:db8::1'",
+            $this->search->buildSqlQuery()
+        );
+    }
+
     public function testBuildSqlQueryWithoutConditions(): void
     {
         $_POST['db'] = 'opengis';


liviuconcioiu:
@Solaris-star I don't think this is correct, INET4 was missing as a data type. See #20397

Solaris-star:
Thanks for the review, and you're right that `INET4` was missing from `Types.php` — #20397 is needed.

But I checked whether #20397 alone fixes the search-quoting reported in #20395, and it doesn't. The two changes are independent:

`Table\Search::getWhereClause()` decides quoting from the raw `criteriaColumnTypes` POST value, not from `Types::getTypeClass()`:

```php
preg_match('@char|binary|blob|text|set|date|time|year|uuid@i', $types)
```

So `inet4` / `inet6` fall into the unquoted branch.

Verified on three separate checkouts, each with its own `vendor/` (a shared/symlinked `vendor` silently loads the other tree's classes via the absolute classmap — that tripped me up at first):

| tree | generated SQL |
| --- | --- |
| clean `QA_5_2` (163d564) | `... WHERE \`ipv4\` = 1.2.3.4 AND \`ipv6\` = 2001:db8::1` |
| `QA_5_2` + #20397 (cb6f602) | `... WHERE \`ipv4\` = 1.2.3.4 AND \`ipv6\` = 2001:db8::1` |
| `QA_5_2` + this PR | `... WHERE \`ipv4\` = '1.2.3.4' AND \`ipv6\` = '2001:db8::1'` |

Note `INET6` has the same bug today and is already a registered type, which is why registering `INET4` alone doesn't change the quoting.

So I'd suggest both: #20397 for type registration/UI, and this one-line regex change for the invalid SQL. Happy to rebase this on top of #20397, fold it into your PR, or close this if you prefer to handle it there.

Solaris-star:
You are right. My regression bypassed the real form/type-classification path by passing raw `inet4` directly as `criteriaColumnTypes`; the UI should receive the character type class after `Types::getTypeClass()`. #20397 fixes the actual missing INET4 registration and is the correct solution. I’m closing this PR to avoid competing with it. Thanks for the correction.

Solaris-star:
Closed in favor of #20397, which fixes the root cause.

PR #20401, opened 2026-07-28
Don't do an unnecessary query per table when exporting sql

M393:
All usage of the `SHOW TABLE STATUS` query result is guarded by `$showDates`, so executing the query is also only necessary when `$showDates` is `true`.


kamil-tekiela:
Truly this query is not needed at all, but on 5.2 we could at least do this.

MauricioFauth (approved):

PR #20402, opened 2026-07-28
remove auto escape from page title to avoid & bug

rlittle316:
### Description

This pull request is to disable twig auto escaping on header.title. When & is used in the title with this enabled it is rendered as ```&amp;``` in the title.

Fixes #
This fixes the incorrect title rendered when a & is passed to the title.


rlittle316:
Im not sure if this should be patched in the current version of phpmyadmin or not. A workaround I have found for this issue is to pre replace any & with the full width version of it which bypasses the auto escaping done by twig.

kamil-tekiela:
This should be fixed in 5.2, but instead of taking the raw value in Twig, I would just remove `htmlspecialchars` in PHP and only add it when the title is passed through AJAX. Much cleaner and more correct solution.

PR #20410, opened 2026-08-04
Fix #20409: editing column structure for varchar with empty string as default

### LINKED ISSUE #20409: [Bug]: Regression in 5.2.1+: Editing a column with DEFAULT '' changes it to no default ("None")
### Describe the bug

There is a regression introduced between phpMyAdmin 5.2.0 and 5.2.1.

When editing a table column that has `DEFAULT ''` (empty string), phpMyAdmin incorrectly displays the default value as **None** in the edit dialog.

If the user changes any other property of the column (for example the VARCHAR length) and saves the changes without touching the default value, the generated ALTER TABLE statement removes the empty string default.

As a result, `DEFAULT ''` is silently changed to **no default**.

### To reproduce

Create a test table:

```sql
CREATE TABLE test_default (
    id INT NOT NULL AUTO_INCREMENT,
    name VARCHAR(20) NOT NULL DEFAULT '',
    PRIMARY KEY (id)
);
```

Open phpMyAdmin:

1. Go to **Structure**
2. Click **Change** for column `name`
3. Notice that the Default field already shows **None** instead of an empty string.
4. Change only the column length from `VARCHAR(20)` to `VARCHAR(30)`.
5. Click **Save**.

### Expected behavior

The column definition should become:

```sql
name VARCHAR(30) NOT NULL DEFAULT ''
```

### Actual behavior

The column becomes:

```sql
name VARCHAR(30) NOT NULL
```

The empty string default is lost.

### Environment

- phpMyAdmin 5.2.0: **Works correctly**
- phpMyAdmin 5.2.1: **Regression introduced**
- phpMyAdmin 5.2.2: Bug still present
- phpMyAdmin 5.2.3: Bug still present

### Screenshots

Image

Image

### Operating System

Windows

### Web Server

Apache 2.4

### Database Server

MySQL

### Database version

8.0.35

### PHP version

8.5.6

### phpMyAdmin version

5.2.1

### browser

Chrome

### Additional context

_No response_

M393:
### Description

Don't lose empty string default value when editing a varchar column.
This was introduced in 8929288efd592047c81df3b9b8b8a8151edaac41

Fixes #20409

### DIFF
--- a/libraries/classes/Table/ColumnsDefinition.php
+++ b/libraries/classes/Table/ColumnsDefinition.php
@@ -496,58 +496,38 @@ final class ColumnsDefinition
     /**
      * Set default type and default value according to the column metadata
      *
-     * @param array $columnMeta Column Metadata
-     * @phpstan-param array<string, string|null> $columnMeta
+     * @param array<string, string|null> $columnMeta Column Metadata
      *
-     * @return non-empty-array<array-key, mixed>
+     * @return non-empty-array<string, string>
      */
     public static function decorateColumnMetaDefault(array $columnMeta): array
     {
-        $metaDefault = [
-            'DefaultType' => 'USER_DEFINED',
-            'DefaultValue' => '',
-        ];
+        $meta = [];
+        $default = $columnMeta['Default'];
+        if ($default === null) {
+            $meta['DefaultType'] = $columnMeta['Null'] === 'YES' ? 'NULL' : 'NONE';
+            $meta['DefaultValue'] = '';
+        } elseif ($default === 'CURRENT_TIMESTAMP' || $default === 'current_timestamp()') {
+            $meta['DefaultType'] = 'CURRENT_TIMESTAMP';
+            $meta['DefaultValue'] = '';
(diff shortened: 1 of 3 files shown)


libraries/classes/Table/ColumnsDefinition.php
```diff
@@ -537,6 +529,16 @@ public static function decorateColumnMetaDefault(array $columnMeta): array
                 $metaDefault['DefaultType'] = 'UUID_v7';
 
                 break;
+            case null:
```
kamil-tekiela: How does this work if `null == ''`?

M393: Does this answer your question?
https://onlinephp.io/c/a26c611b-36ed-4a03-84b3-2b1791426cc7

kamil-tekiela: I did not see you added a new if statement. This should not be a `switch` statement.

markusramsak (approved):

MauricioFauth (approved):
Thank you!

PR #20413, opened 2026-08-09
Fix Gis editor form in search view

MoonE:
### Description

Opening the GIS editor form the search view produced an error because both `'show.bs.modal'` listeners were active. (from change.ts and select.ts)

Also copying the value from the editor to the search form did not work because the editor modal which has its own form element was inside the search form element. Nested form elements are stripped away by the browser.


MauricioFauth (approved):
Thank you!

PR #20385, opened 2026-07-18
Bump required Node version to 24

MauricioFauth:
- https://endoflife.date/nodejs
- https://github.com/phpmyadmin/phpmyadmin/actions/runs/29652174854/job/88100253153#step:4:10
- https://github.com/phpmyadmin/phpmyadmin/pull/19852

Issue #20345, opened 2026-06-28
[Bug]: Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24


liviuconcioiu:
### Describe the bug

```
Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809
```

### How to Reproduce

https://github.com/phpmyadmin/phpmyadmin/actions/runs/28306544068

### phpMyAdmin version

5.2.4-dev, 6.0.0-dev

daaug:
Hi, @liviuconcioiu 

When the job is completed (https://github.com/phpmyadmin/phpmyadmin/actions/runs/28306544068/job/83863599750) a link is provided (Warning: Node.js 20 is deprecated. [..] https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/) and this gh blog explains the deprecation of node 20 and how to force its use:

> [...] 
Beginning on June 16th, 2026, runners will begin using Node24 by default. To opt out of this and continue using Node20 after this date, set ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true as an env in your workflow or as an environment variable on your runner machine. This will only work until we upgrade the runner and remove Node20 later in the fall of 2026. 
[...]

I dont see this as a safe thing to do, because even github suggest to use version 24 and explains why, but if v24 is causing problem, you can follow their instructions to **possibly** bypass this situation.

liviuconcioiu:
Fixed in https://github.com/phpmyadmin/phpmyadmin/pull/20385

PR #20414, opened 2026-08-09
Allow Node.js 22

MoonE:
### Description

Node.js 22 is still [supported until April 2027](https://nodejs.org/en/about/previous-releases) and Ubuntu 26.04 comes with version 22.

Ref: #20345 & https://github.com/phpmyadmin/phpmyadmin/pull/20385

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "license": "GPL-2.0",
   "private": true,
   "engines": {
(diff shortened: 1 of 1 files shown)


williamdes (approved):
Well, maybe we can allow it 🤔

MauricioFauth (approved):

PR #20419, opened 2026-08-13
Bump phpmyadmin/coding-standard to version 5 

MauricioFauth:
Renames `phpcs.xml.dist` to `phpcs.xml`.


phpcs.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0"?>
```
williamdes, line 1: please update https://github.com/phpmyadmin/phpmyadmin/blob/852baa6f35f39aeda2f868c43a307883e90fd6ff/.gitattributes#L15

MauricioFauth: I updated it. Thanks!

williamdes (approved):
So why remove the dist way ?

src/Error/ErrorHandler.php
```diff
@@ -187,8 +187,6 @@ public function sliceErrors(int $count): array
      * @param string $errfile error file
      * @param int    $errline error line
      *
-     * @return false
```
kamil-tekiela, line 190: Why was this removed?

MauricioFauth: To make the native return type more specific, from `bool` to `false`. It was confusing because the native return type was missing from the change. It's fixed now. Thanks!

kamil-tekiela (approved):

MauricioFauth:
> So why remove the dist way ?

To make things simpler. There's not much reason to use dist for things that don't have secrets in it. Also, these tools accept passing a custom config file as an option if someone needs that. In other words, using a custom config should be intentional.

PR #20423, opened 2026-08-14
Fix db structure summary timestamp

M393:
### Description

In Database -> Structure `Creation`, `Last update`, and `Last check` all showed the min timestamp of any table in the `Sum` row. `Last update` should display the highest instead.
Not so sure if the other two columns should continue to display the min or max timestamp.

The other two timestamps are calculate just above and below the changed code for the max update timestamp.

Enable in `config.inc.php`:
```php
$cfg['ShowDbStructureLastUpdate'] = true;
$cfg['ShowDbStructureCreation'] = true;
$cfg['ShowDbStructureLastCheck'] = true;
```

### DIFF
--- a/test/classes/Controllers/Database/StructureControllerTest.php
+++ b/test/classes/Controllers/Database/StructureControllerTest.php
@@ -389,6 +389,10 @@ class StructureControllerTest extends AbstractTestCase
      */
     public function testDisplayTableList(): void
     {
+        $GLOBALS['cfg']['ShowDbStructureCreation'] = true;
+        $GLOBALS['cfg']['ShowDbStructureLastUpdate'] = true;
+        $GLOBALS['cfg']['ShowDbStructureLastCheck'] = true;
+
         $class = new ReflectionClass(StructureController::class);
         $method = $class->getMethod('displayTableList');
         if (PHP_VERSION_ID < 80100) {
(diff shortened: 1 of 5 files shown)


williamdes:
Could you add an unit tests for this case please ?

williamdes (approved):

MauricioFauth (approved):
Thank you!

test/classes/Controllers/Database/StructureControllerTest.php
```diff
@@ -425,34 +429,59 @@ public function testDisplayTableList(): void
...
+        self::assertStringContainsString(
+            '<th class="value tbl_creation font-monospace text-end">Sep 04, 2026 at 09:11 AM</th>',
+            $result
+        );
+        self::assertStringContainsString(
+            '<th class="value tbl_last_update font-monospace text-end">Sep 04, 2026 at 10:22 AM</th>',
+            $result
+        );
+        self::assertStringContainsString(
+            '<th class="value tbl_last_check font-monospace text-end">Sep 04, 2026 at 11:11 AM</th>',
```
MauricioFauth: I think it makes sense for "Last check" to show the newest value as well.

MauricioFauth (approved):

PR #20426, opened 2026-08-17
Fix #20425 - OOM error when creating a view

### LINKED ISSUE #20425: [Bug]: Creating or altering a VIEW buffers the view's entire result set into PHP memory, causing a fatal OOM / HTTP 500 after the VIEW was already created
### Describe the bug

When a VIEW is created or altered through the UI (route `/view/create`, POST with `createview` / `alterview`), phpMyAdmin runs the DDL, and then — after it has already succeeded — executes the view's whole SELECT body a second time purely to collect column metadata:

`Controllers/View/CreateController.php`
```php
$column_map = $this->dbi->getColumnMapFromSql($_POST['view']['as'], $view_columns);
```

`DatabaseInterface::getColumnMapFromSql()`
```php
$result = $this->tryQuery($sql_query);
if ($result === false) {
    return [];
}
$meta = $this->getFieldsMeta($result);
```

`tryQuery()` defaults to `DatabaseInterface::QUERY_BUFFERED`, so `DbiMysqli::realQuery()` ends up calling `$link->query($query, MYSQLI_STORE_RESULT)`. The complete result set is pulled into PHP memory even though nothing but `getFieldsMeta()` is ever read from it, and no `LIMIT` or row cap is applied.

For a view over a large table this exceeds `memory_limit` and the request dies:

```
PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 25182208 bytes)
 in .../libraries/classes/Dbal/DbiMysqli.php on line 209
```

nginx then returns HTTP 500 to the browser.

Two things make this worse than a plain OOM:

1. The fatal happens *after* the `CREATE`/`ALTER VIEW` has been committed. The user sees a 500 and assumes the save failed, but reloading shows the view was created/updated. A metadata bookkeeping step turns a successful DDL into an apparent failure.
2. Running the exact same SELECT from the SQL tab works fine, because that path is paginated. So the same statement succeeds in one place and 500s in another, which is very hard to understand from the user side.

Memory consumption here scales with the number of rows the view returns, while only the column list is needed. Raising `memory_limit` only moves the threshold.

**Affected versions**

- 5.2.2 and `QA_5_2` (5.2.3): `libraries/classes/Controllers/View/CreateController.php` calls `getColumnMapFromSql()` unconditionally.
- `master` (6.0-dev): `src/Controllers/View/CreateController.php` now calls it only inside `if ($pmaTransformationData !== false)`, so it is skipped when phpMyAdmin configuration storage is not set up — but with configuration storage enabled `SystemDatabase::getColumnMapFromSql()` runs the very same unbounded buffered query.

**Related**

This is the same class of bug as #18941 (fixed by #20238), which failed with the identical message at the identical `DbiMysqli.php` line, just on the query-simulation path instead of the view path.

**Possible fixes**

- Run the metadata query with `DatabaseInterface::QUERY_UNBUFFERED` (`MYSQLI_USE_RESULT`) — `getFieldsMeta()` does not need buffered rows.
- Or wrap it as `SELECT * FROM (<view body>) AS tmp LIMIT 0`, the same technique used in #20238.
- Or drop the extra query and read the column list from `INFORMATION_SCHEMA.COLUMNS` for the view that was just created.
- Independently: a failure in this bookkeeping step should not be able to turn a committed `CREATE`/`ALTER VIEW` into an HTTP 500.

### How to Reproduce

1. Create a table large enough that its full result set exceeds PHP's `memory_limit` (with the default 128M, a few million rows is enough).
2. Go to that database -> Views -> "Create view" (or open an existing view and choose "Edit").
3. Set the definition to `SELECT * FROM big_table` (any definition returning many rows works; a `JOIN` reaches the threshold sooner).
4. Click "Go" to save.
5. The request returns HTTP 500. The PHP error log shows:
   `PHP Fatal error:  Allowed memory size of ... bytes exhausted ... in .../libraries/classes/Dbal/DbiMysqli.php on line 209`
6. Reload the database: the view exists and its definition has been applied, despite the 500.
7. For contrast, run the same `SELECT * FROM big_table` from the SQL tab — it completes normally.

### Expected behavior

Saving a view should not depend on materializing the view's result set. Memory usage of `/view/create` should be independent of how many rows the view returns, the request should complete normally, and the success message should be shown.

If the column-metadata lookup does fail for any reason, it should not surface as an HTTP 500 for a DDL statement that already succeeded.

### Screenshots

Image

### Operating System

Linux

### Web Server

_No response_

### Database Server

MariaDB

### Database version

10.11.11

### PHP version

8.2

### phpMyAdmin version

5.2.1

### browser

firefox v136

### Additional context

Sanitized nginx error log entry:

```
2026/08/12 14:28:44 [error] 30600#30600: *9131 FastCGI sent in stderr:
"PHP message: PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted
 (tried to allocate 25182208 bytes)
 in /volume1/web_packages/phpmyadmin/libraries/classes/Dbal/DbiMysqli.php on line 209"
while reading response header from upstream,
request: "POST /phpmyadmin/index.php?route=/view/create HTTP/2.0",
referrer: "https://<host>/phpmyadmin/index.php?route=/view/create&db=<db>&table=<view>"
```

Note this is not a regression: the same failure was reproducible on this installation before the PHP 8.2 upgrade. It appears only once the underlying data grows past the point where the buffered result set exceeds `memory_limit`.

Ref. 
similar case, different code path
https://github.com/phpmyadmin/phpmyadmin/pull/20238

kamil-tekiela:
Fixes #20425

Queries that only fetch metadata should not fetch the data. MySQL allows `LIMIT 0` for exactly that reason. When running it on the view query, we wrap it in a subquery as we don't know the structure of the SQL and manipulating it with SQL Parser is not possible. The subquery is very likely to have its own LIMIT. 

The `getColumnMapFromSql` method was moved into the other method in `SystemDatabase`, but on 6.0 it was kept as an argument value probably for easier testing. I believe that it could be made private on 6.0.

### DIFF
--- a/libraries/classes/Controllers/View/CreateController.php
+++ b/libraries/classes/Controllers/View/CreateController.php
@@ -161,8 +161,6 @@ class CreateController extends AbstractController
                 $view_columns = explode(',', $_POST['view']['column_names']);
             }
 
-            $column_map = $this->dbi->getColumnMapFromSql($_POST['view']['as'], $view_columns);
-
             $systemDb = $this->dbi->getSystemDatabase();
             $pma_transformation_data = $systemDb->getExistingTransformationData($db);
 
@@ -170,7 +168,8 @@ class CreateController extends AbstractController
                 // SQL for store new transformation details of VIEW
                 $new_transformations_sql = $systemDb->getNewTransformationDataSql(
                     $pma_transformation_data,
-                    $column_map,
+                    $view_columns,
+                    $_POST['view']['as'],
                     $_POST['view']['name'],
                     $db
                 );
--- a/libraries/classes/DatabaseInterface.php
+++ b/libraries/classes/DatabaseInterface.php
@@ -806,15 +806,15 @@ class DatabaseInterface implements DbalInterface
     /**
      * returns detailed array with all columns for sql
      *
-     * @param string $sql_query    target SQL query to get columns
+     * @param string $sqlQuery     target SQL query to get columns
      * @param array  $view_columns alias for columns
      *
      * @return array
      * @psalm-return list<array<string, mixed>>
      */
-    public function getColumnMapFromSql(string $sql_query, array $view_columns = []): array
+    public function getColumnMapFromSql(string $sqlQuery, array $view_columns = []): array
     {
-        $result = $this->tryQuery($sql_query);
+        $result = $this->tryQuery('SELECT * FROM (' . $sqlQuery . ') AS tmp LIMIT 0');
 
         if ($result === false) {
             return [];
--- a/libraries/classes/Dbal/DbalInterface.php
+++ b/libraries/classes/Dbal/DbalInterface.php
@@ -149,12 +149,12 @@ interface DbalInterface
     /**
      * returns detailed array with all columns for sql
      *
-     * @param string $sql_query    target SQL query to get columns
+     * @param string $sqlQuery     target SQL query to get columns
      * @param array  $view_columns alias for columns
      *
      * @return array
      */
-    public function getColumnMapFromSql(string $sql_query, array $view_columns = []): array;
+    public function getColumnMapFromSql(string $sqlQuery, array $view_columns = []): array;
 
     /**
      * returns detailed array with all columns for given table in database,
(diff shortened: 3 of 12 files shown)


williamdes (approved):
Thank you !

PR #20427, opened 2026-08-17
Fix #18749 - Remove the X-XSS-Protection header

### LINKED ISSUE #18749: Reconsider use of X-XSS-Protection
### Describe the bug

phpmyadmin sets the header X-XSS-Protection, some security tools warn about this.
The header is related to an obsolete feature and should probably just not be set at all.

Background: In the past, some browsers implemented a simple XSS filter, and this could be controlled with the X-XSS-Protection header. It was generally possible to set it to different modes (block and filter), however, it turned out that both modes can introduce security vulnerabilities.

By now, these XSS filters have been removed from all major browsers, and it probably makes little sense to set the header at all. Therefore I recommend removing it. See also the OWASP recommendation:
https://cheatsheetseries.owasp.org/cheatsheets/HTTP_Headers_Cheat_Sheet.html#x-xss-protection

(the comment in libraries/classes/Header.php references an old OWASP recommendation that is no longer online)

pongoe:
Fixes #18749

I noticed this had been sitting open since 2023 and it looked like a quick one, so I've done it.

The XSS auditor this header controlled is gone from all the major browsers, and OWASP now
says not to send it at all:
https://cheatsheetseries.owasp.org/cheatsheets/HTTP_Headers_Cheat_Sheet.html#x-xss-protection

What I changed:
- removed the header from `src/Header.php`
- removed the matching line from `tests/unit/HeaderTest.php`
- added a line to `CHANGELOG-6.0.md` under Removed

HeaderTest passes and phpcs is clean on both files. This is against master.

Cheers

### DIFF
--- a/CHANGELOG-6.0.md
+++ b/CHANGELOG-6.0.md
@@ -98,5 +98,6 @@ All notable changes of the phpMyAdmin 6.0 release series are documented in this 
 * [#19550](https://github.com/phpmyadmin/phpmyadmin/pull/19550): Replace jQuery UI's tooltip with Bootstrap's Tooltip
 * [#19852](https://github.com/phpmyadmin/phpmyadmin/pull/19852): Bump Node version to 20
 * [#19854](https://github.com/phpmyadmin/phpmyadmin/pull/19854): Drop support for BaconQrCode v2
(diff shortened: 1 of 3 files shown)


williamdes (approved):
Thank you !

kamil-tekiela (approved):

PR #20432, opened 2026-08-20
Do not expose the phpMyAdmin version in asset URLs

### LINKED ISSUE #15993: do not expose PMA_VERSION in requests

In version **5.0.1**, not tested on earlier versions, the PMA_VERSION is exposed by subsequent requests to the server.

For example `https://<domain>/js/cross_framing_protection.js?v=5.0.1`

See the code in `Scripts.php->getDisplay()`, wehere the **version**  field is hardcoded to **PMA_VERSION** .

To have any possibility to configure the **version** field, with fallback back to the **PMA_VERSION**.

From my point of view this is security bug, while you are exposing the PMA version. After any security issue you are giving to the attackers information about the version and all possible CVE.

None other alternatives have been considered.

### Additional context

- phpMyAdmin version: 5.0.1

pongoe:
Fixes #15993

Right now every CSS and JS asset gets loaded with `?v=<version>` as a cache buster, so anyone who can reach the login page can read the exact phpMyAdmin version straight out of the page source (32 places on the login page alone).

This swaps the version for an opaque token: an HMAC-SHA256 of the version, keyed with the configured `blowfish_secret` and trimmed to 12 hex characters. It still changes on every upgrade, so cache busting keeps working exactly like it does today, but there's no way to map it back to a version with a lookup table. If no blowfish secret is configured, it falls back to a random per-session key, so it never turns into a hash anyone could compute for themselves.

### DIFF
--- a/CHANGELOG-6.0.md
+++ b/CHANGELOG-6.0.md
@@ -82,6 +82,7 @@ All notable changes of the phpMyAdmin 6.0 release series are documented in this 
 * [#19795](https://github.com/phpmyadmin/phpmyadmin/pull/19795): Use `additional-methods.min.js` instead of `additional-methods.js`
 * [#19564](https://github.com/phpmyadmin/phpmyadmin/pull/19564): Update the table comment field to a textarea
 * [#19947](https://github.com/phpmyadmin/phpmyadmin/pull/19947): Export ODS column headers by default
+* [#15993](https://github.com/phpmyadmin/phpmyadmin/issues/15993): Do not expose the phpMyAdmin version in the cache-busting `?v=` query string of CSS and JS assets
 
 ### Removed
 
--- a/resources/js/modules/ajax.ts
+++ b/resources/js/modules/ajax.ts
@@ -716,10 +716,10 @@ const AJAX = {
             // Clear loaded scripts if they are from another version of phpMyAdmin.
             // Depends on common params being set before loading scripts in responseHandler
             if (self.scriptsVersion === null) {
-                self.scriptsVersion = CommonParams.get('version');
-            } else if (self.scriptsVersion !== CommonParams.get('version')) {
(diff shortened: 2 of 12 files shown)


williamdes (approved):
I like it, but I think it is important to note that there is many many ways to know what version you are running. With some hours on my hands I can do guessing and get accurate on most versions. Since a lot of them change the CSS, just grep a class and you will know

kamil-tekiela:
Why do this?

williamdes:
> Why do this?

Security by obscurity I think

pongoe:
It came out of #15993. The main problem is that the version string in asset URLs sits in the page source of the login page, which gets indexed. So when a CVE lands for a specific version, you can find vulnerable installs with a Google search. A determined attacker can still fingerprint the build by other means, but this takes installs out of that easy dragnet. Cache busting still works the same since the token still changes on every upgrade.

kamil-tekiela:
I see, so the goal is to make it harder for attackers to find vulnerable publicly accessible installs. From my point of view, having phpMyAdmin installed on a production server is a security issue in itself. Running a vulnerable version is even worse. So this is definitely security by obscurity. We won't provide any more security to the users by doing this, but we can make it a tiny bit harder for attackers. But is this really the best way to do it?

pongoe:
You're right, it doesn't add any security in itself. It is security by obscurity, and I'd call it good practice rather than essential. 

What it removes is the script kiddie with a Google dork finding vulnerable installs. But you're dead right. Having phpMyAdmin installed on a production server is just asking for trouble. But of course, that's what plenty of people do ...

As for whether it's the best way, I'd say it works. Anything stronger gets into whether unauthenticated visitors should be served assets at all, which felt out of scope here. If you had a different approach in mind, I'm happy to look at it.

PR #20433, opened 2026-08-20
Fix PHP 8.6 deprecations

liviuconcioiu:
This should fix:

https://github.com/phpmyadmin/phpmyadmin/actions/runs/32332620113/job/96357454416

```
67) /home/runner/work/phpmyadmin/phpmyadmin/tests/unit/Plugins/Transformations/TransformationPluginsTest.php:404
Calling ReflectionMethod::invokeArgs() for static method PhpMyAdmin\Plugins\Transformations\Abs\DateFormatTransformationsPlugin::getName() does not need an object parameter

68) /home/runner/work/phpmyadmin/phpmyadmin/tests/unit/FooterTest.php:81
Calling ReflectionMethod::invokeArgs() for static method PhpMyAdmin\Footer::removeRecursion() does not need an object parameter
```

### DIFF
--- a/test/classes/Plugins/Transformations/TransformationPluginsTest.php
+++ b/test/classes/Plugins/Transformations/TransformationPluginsTest.php
@@ -760,7 +760,9 @@ class TransformationPluginsTest extends AbstractTestCase
         }
 
         $reflectionMethod = new ReflectionMethod($object, $method);
-        self::assertEquals($expected, $reflectionMethod->invokeArgs($object, $args));
+        self::assertEquals($expected, $reflectionMethod->invokeArgs(
+            $reflectionMethod->isStatic() ? null : $object, $args
+        ));
     }
 
     /**
@@ -1313,7 +1315,9 @@ class TransformationPluginsTest extends AbstractTestCase
         string $error = ''
     ): void {
         $reflectionMethod = new ReflectionMethod($object, 'applyTransformation');
-        self::assertEquals($transformed, $reflectionMethod->invokeArgs($object, $applyArgs));
+        self::assertEquals($transformed, $reflectionMethod->invokeArgs(
+            $reflectionMethod->isStatic() ? null : $object, $applyArgs
+        ));
 
         // For output transformation plugins, this method may not exist
         if (method_exists($object, 'isSuccess')) {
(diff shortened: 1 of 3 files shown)


test/classes/Plugins/Transformations/TransformationPluginsTest.php
```diff
@@ -1313,7 +1315,9 @@ public function testTransformation(
         string $error = ''
     ): void {
         $reflectionMethod = new ReflectionMethod($object, 'applyTransformation');
-        self::assertEquals($transformed, $reflectionMethod->invokeArgs($object, $applyArgs));
+        self::assertEquals($transformed, $reflectionMethod->invokeArgs(
+            $reflectionMethod->isStatic() ? null : $object, $applyArgs
+        ));
```
kamil-tekiela: Can `applyTransformation` ever be static?

liviuconcioiu: It can't. I did not look at the code, just changed where `invokeArgs` was. Thanks!

test/classes/Plugins/Transformations/TransformationPluginsTest.php
```diff
@@ -760,7 +760,10 @@ public function testGetMulti($object, string $method, $expected, array $args = [
         }
 
         $reflectionMethod = new ReflectionMethod($object, $method);
-        self::assertEquals($expected, $reflectionMethod->invokeArgs($object, $args));
+        self::assertEquals($expected, $reflectionMethod->invokeArgs(
+            $reflectionMethod->isStatic() ? null : $object,
+            $args
+        ));
```
kamil-tekiela: Why do we even use Reflection here? Surely, something like this would be enough:
```
self::assertSame( $expected, $object->$method(...$args));
```

kamil-tekiela: On 5.2:
```
// remove method_exists
        $testMethod = [$object, $method];
        self::assertIsCallable($testMethod);
        self::assertSame($expected, $testMethod(...$args));
```
and on 6.0 we can refactor it into callables.

MauricioFauth (approved):
Thank you!

PR #20441, opened 2026-08-23
Message refactoring

kamil-tekiela:
Big PR, but changes are divided into logical commits. Removing the Template dependency is too big a task for now, so a half-measure is to take it as an argument in some cases.


MauricioFauth (approved):
That's great! Thank you!

PR #20315, opened 2026-05-24
Fix #19841 - Strip port from host when linking to user privileges in processes list

roadhero:
### Description

The Host field in `SHOW PROCESSLIST` includes the client's TCP port
(e.g., `localhost:64776`). This value was passed directly as the
`hostname` parameter when building the link to a user's privileges page.
The privileges page queries `mysql.user WHERE Host = 'localhost:64776'`,
which never matches — the user table stores bare hostnames without ports.

This change adds a `user_host` key to the process row array in
`Processes::getList()`. The value is the host with any trailing `:port`
suffix stripped via `preg_replace('/:\d+$/', '', $host)`. The template
uses `row.user_host` for the privileges link URL, while `row.host`
(unchanged) continues to populate the Host column display.

Fixes #19841.

---

Before submitting pull request, please review the following checklist:

- [x] Make sure you have read our [CONTRIBUTING.md](https://github.com/phpmyadmin/phpmyadmin/blob/master/CONTRIBUTING.md) document.
- [x] Make sure you are making a pull request against the correct branch. For example, for bug fixes in a released version use the corresponding QA branch and for new features use the _master_ branch. If you have a doubt, you can ask as a comment in the bug report or on the mailing list.
- [x] Every commit has proper `Signed-off-by` line as described in our [DCO](https://github.com/phpmyadmin/phpmyadmin/blob/master/DCO). This ensures that the work you're submitting is your own creation.
- [x] Every commit has a descriptive commit message.
- [x] Every commit is needed on its own, if you have just minor fixes to previous commits, you can squash them.
- [x] Any new functionality is covered by tests.

PR #20461, opened 2026-08-31
Fix processes list user link by stripping port from host

### LINKED ISSUE #19841: [Bug]: Processes list link to users table does not work
### Describe the bug

The link includes the hostname and port, but 
1. I don't think it is possible to create user that is only allowed from a certain port, so this should never be included
2. If the hostname of the current connection does not match the user's hostname, because the user's hostname is e.g. '%' it  is also not found.

### How to Reproduce

1. Go to Status -> Processes
2. Click on a user
3. User is not found 

### Expected behavior

_No response_

### Screenshots

Image

Image

### Operating System

_No response_

### Web Server

_No response_

### Database Server

MariaDB

### Database version

11.4.3

### PHP version

8.1

### phpMyAdmin version

5.2.2

### browser

_No response_

### Additional context

_No response_

mmustafasenoglu:
Fixes #19841

## Problem

The processes list link to the user privileges page includes the port number in the host value (e.g., `localhost:64776`), but the MySQL `mysql.user` table only stores the hostname without port. This causes the user link to fail when clicking on a user in Status > Processes.

## Root cause

`SHOW PROCESSLIST` returns `Host` as `hostname:port` (e.g., `localhost:64776`), but the privileges page URL parameter `hostname` expects just the hostname (e.g., `localhost`). The template was passing `row.host` directly without stripping the port.

## Fix

Use the Twig `split(':')|first` filter to extract only the hostname from the host value before passing it as the `hostname` parameter to the privileges URL.

## Testing

- Verified that `split(':')|first` correctly handles both `localhost:64776` -> `localhost` and `localhost` -> `localhost` (no port)
- Existing test suite should continue to pass (test mock uses `Host1` without port, which remains unchanged)

### DIFF
--- a/templates/server/status/processes/list.twig
+++ b/templates/server/status/processes/list.twig
@@ -49,7 +49,7 @@
             {% if row.user not in ['system user', 'event_scheduler', 'unauthenticated user'] %}
               <a href="{{ url('/server/privileges', {
                   'username': row.user,
-                  'hostname': row.host,
+                  'hostname': row.host|split(':')|first,
                   'dbname': row.db,
                   'tablename': '',
                   'routinename': '',


williamdes:
See #20315

williamdes:
Duplicate of #20315

williamdes:
Thank you for your work @mmustafasenoglu 
I think the other PR was more complete and less of a Twig template hack. Also it had tests.
Feel free to fix another issue

Issue #20159, opened 2026-02-21
[Bug]: Various JS errors related to empty user-agent


liviuconcioiu:
### Describe the bug

If `window.navigator.userAgent` is not defined (no entry for `userAgent` in `window.navigator`), there are various JS errors, making phpMyAdmin unusable.

Those dependencies use `navigator.userAgent`:

1. jQuery - `Uncaught TypeError: can't access property "prototype", e is undefined`
3. CodeMirror - `Uncaught ReferenceError: CodeMirror is not defined`

This is an upstream bug

### How to Reproduce

1. See https://github.com/phpmyadmin/phpmyadmin/issues/20157 how to reproduce
2. Refresh pages, try to login, etc

### Screenshots

<img width="1218" height="1022" alt="Image" src="https://github.com/user-attachments/assets/88624ebf-2937-41bf-b880-625e22464a99" />

### phpMyAdmin version

5.2.4-dev, 6.0.0-dev

liviuconcioiu:
I opened a bug report for CodeMirror.

williamdes:
> I opened a bug report for CodeMirror.

Can you link it?

liviuconcioiu:
> > I opened a bug report for CodeMirror.
> 
> Can you link it?

https://github.com/codemirror/codemirror5/issues/7155
Edit by @williamdes: got closed, not an issue for them

williamdes:
Still some to fix ?

liviuconcioiu:
I don't think so. I fixed most of them in other PRs. You can close this one.

liviuconcioiu:
Double-checked and everything works!

liviuconcioiu:
@williamdes this can be closed.

PR #20471, opened 2026-09-03
Fix Uncaught ReferenceError: CodeMirror is not defined

liviuconcioiu:
This will fix `Uncaught ReferenceError: CodeMirror is not defined` if user-agent is empty. Related #20159

Before:

https://github.com/user-attachments/assets/1b8fb27a-293c-469d-80c3-65c147f78e13

After:

https://github.com/user-attachments/assets/0a568f8a-c033-4320-aa95-005a15c57c15


williamdes (approved):
Thank you  @liviuconcioiu  !

PR #20474, opened 2026-09-06
Fix search of INET4 with = operator

liviuconcioiu:
Related #20395

https://github.com/user-attachments/assets/30417192-8bb9-4461-8926-c32ea3a3534b


williamdes (approved):
Thanks !

PR #20475, opened 2026-09-07
Add character set support for CSV with LOAD DATA in table imports

### LINKED ISSUE #20400: [Bug]: "CSV using LOAD DATA" import with charset selection results in "This plugin does not support compressed imports!"
### Describe the bug

If you select any charset other than utf8 for importing "CSV using LOAD DATA" an error message is shown:
"This plugin does not support compressed imports!"

If I understand correctly, this is due to the condition here:
https://github.com/phpmyadmin/phpmyadmin/blob/RELEASE_5_2_3/libraries/classes/Plugins/Import/ImportLdi.php#L103
since $charset_conversion is not False, the error is thrown.

### How to Reproduce

1. Go to database table
2. Click "Import"
3. Select CSV file
4. Select any charset, other than utf8
5. Under Format, select  "CSV using LOAD DATA"
6. Click "Import"
7. An error is displayed: "This plugin does not support compressed imports!"

### Expected behavior

I expected the data to be imported with the correct character set.
I was able to add `$sql .= ' CHARACTER SET latin1 ';` to the code to force it to use latin1 and it worked.

### Screenshots

_No response_

### Operating System

Linux

### Web Server

Nginx

### Database Server

MariaDB

### Database version

11.8.6-MariaDB-5ubuntu0.1 from Ubuntu

### PHP version

8.5.4

### phpMyAdmin version

5.2.3deb1

### browser

Firefox v152.0.4

### Additional context

_No response_

st0rmsetter:
### Description

This pull request includes support for the majority of character sets using CSV with LOAD DATA. MySQL and MariaDB support this using the `CHARACTER SET` (or `CHARSET`) clause. It includes native support for them by creating a character set mapping. Apart from being a new feature, it also prevents the display of an unhelpful error message: "This plugin does not support compressed imports!"

The previous implementation handled non-UTF-8 character sets ungracefully and it resulted in an error. However, this new implementation includes support for the *supported* character sets. Special care has been taken for MySQL limitations by not using UTF-16 character encoding with MySQL databases, instead opting to do so for MariaDB.

Fixes #20400 

#### Sources

https://dev.mysql.com/doc/refman/8.4/en/charset-mysql.html
https://dev.mysql.com/doc/refman/8.4/en/charset-restrictions.html
https://dev.mysql.com/doc/refman/8.0/en/charset-charsets.html

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -6039,9 +6039,6 @@
     <MixedArgument>
       <code><![CDATA[ImportSettings::$importFile]]></code>
     </MixedArgument>
-    <PossiblyUnusedReturnValue>
-      <code><![CDATA[string[]]]></code>
-    </PossiblyUnusedReturnValue>
   </file>
   <file src="src/Plugins/Import/ImportMediawiki.php">
     <InvalidNullableReturnType>
@@ -6245,11 +6242,6 @@
     <UnusedClass>
       <code><![CDATA[UploadSession]]></code>
     </UnusedClass>
-  </file>
-  <file src="src/Plugins/ImportPlugin.php">
-    <PossiblyUnusedReturnValue>
-      <code><![CDATA[string[]]]></code>
-    </PossiblyUnusedReturnValue>
   </file>
   <file src="src/Plugins/Schema/Dia/DiaRelationSchema.php">
     <MixedArgumentTypeCoercion>
--- a/src/Plugins/Import/ImportLdi.php
+++ b/src/Plugins/Import/ImportLdi.php
@@ -36,6 +36,28 @@ class ImportLdi extends AbstractImportCsv
     private string $escaped = '';
     private string $newLine = '';
     private string $columns = '';
+    private const CHARSET_MAP = [
+        'iso-8859-1' => 'latin1',
+        'iso-8859-2' => 'latin2',
+        'iso-8859-7' => 'greek',
+        'iso-8859-8' => 'hebrew',
+        'iso-8859-9' => 'latin5',
+        'iso-8859-13' => 'latin7',
+        'koi8-r' => 'koi8r',
+        'windows-1250' => 'cp1250',
+        'windows-1251' => 'cp1251',
+        'windows-1252' => 'latin1',
+        'windows-1256' => 'cp1256',
+        'windows-1257' => 'cp1257',
+        'tis-620' => 'tis620',
+        'SHIFT_JIS' => 'sjis',
+        'SJIS' => 'sjis',
+        'SJIS-win' => 'cp932',
+        'big5' => 'big5',
+        'gb2312' => 'gb2312',
+        'euc-jp' => 'ujis',
+        'ks_c_5601-1987' => 'euckr',
+    ];
 
     /** @psalm-return non-empty-lowercase-string */
     public function getName(): string
@@ -116,7 +138,7 @@ class ImportLdi extends AbstractImportCsv
             $compression = $importHandle->getCompression();
         }
 
-        if (ImportSettings::$importFile === 'none' || $compression !== 'none' || ImportSettings::$charsetConversion) {
+        if (ImportSettings::$importFile === 'none' || $compression !== 'none') {
             // We handle only some kind of data!
             Current::$message = Message::error(
                 __('This plugin does not support compressed imports!'),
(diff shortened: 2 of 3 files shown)


st0rmsetter:
I'll look into the failing tests. It seems to mostly be focused on the test suite, namely the inclusion of non-typed data providers. I've found the issues, I'm pushing the new version ASAP.

st0rmsetter:
We’ll also need to find a way to gracefully handle incorrect character set selections without triggering a large number of unfriendly database errors. One approach could be to display a clear, user-friendly error message when a character set issue occurs in the SQL query, such as: "The selected character set cannot represent all characters in the data!"

There are also some character sets that are unsupported or cannot be used with LOAD DATA. For example, UTF-16 is not supported by MySQL for LOAD DATA, although it is supported by MariaDB. In such cases, the current approach is to omit the CHARACTER SET clause and let the database handle the encoding, most likely defaulting to UTF-8.

This can become problematic when a user explicitly selects an encoding such as ISO-8859-7, since MySQL does not support it.

**Update**: The fix first maintains a whitelist of character sets supported by the selected database system, such as MariaDB or MySQL. If the user selects a character set that is not supported by their database system, a graceful error is displayed. However, if the selected character set is supported by the database but is incorrect for the actual file encoding, the application does not try to correct it and simply lets the user see the resulting errors from the SQL queries. This keeps the behavior consistent while clearly distinguishing unsupported database character sets from an incorrect user selection.

st0rmsetter:
I was pretty sure I fixed all failing tests. What do these failures even mean? I ran the test suite locally, as one should and I can't pinpoint the issue! What does it mean the error was on line 0 and char 0?! I tested it on the web version of phpMyAdmin and obviously ran `phpunit`, `phpstan`, `phpcbf` on my PR and it was all fine. Psalm is complaining!

```
Error: src/Plugins/Import/ImportLdi.php:0:0: UnusedBaselineEntry: Baseline for issue "PossiblyUnusedReturnValue" has 1 extra entry. (see https://psalm.dev/316)
Error: src/Plugins/ImportPlugin.php:0:0: UnusedBaselineEntry: Baseline for issue "PossiblyUnusedReturnValue" has 1 extra entry. (see https://psalm.dev/316)
```

How are failures like this handled?

kamil-tekiela:
```
php -d xdebug.mode=off vendor/bin/psalm --set-baseline=psalm-baseline.xml --no-cache
```

st0rmsetter:
> ```
> php -d xdebug.mode=off vendor/bin/psalm --set-baseline=psalm-baseline.xml --no-cache
> ```

This works, thanks!

kamil-tekiela:
> > ```
> > php -d xdebug.mode=off vendor/bin/psalm --set-baseline=psalm-baseline.xml --no-cache
> > ```
> 
> This works, thanks! However, what I am supposed to do if the CI is not ran from my machine?

Just run psalm without `--set-baseline` and it will show you the issues. There's also a composer command for the same.

st0rmsetter:
> > > ```
> > > php -d xdebug.mode=off vendor/bin/psalm --set-baseline=psalm-baseline.xml --no-cache
> > > ```
> > 
> > 
> > This works, thanks! However, what I am supposed to do if the CI is not ran from my machine?
> 
> Just run psalm without `--set-baseline` and it will show you the issues. There's also a composer command for the same.

Without `--set-baseline`, there seem to be no errors found. What do I do with this information?

kamil-tekiela:
That's right. I don't think Psalm reports unnecessary baseline entries. So best to just set the baseline once you are done.

st0rmsetter:
> That's right. I don't think Psalm reports unnecessary baseline entries. So best to just set the baseline once you are done.

Is it OK to commit a new baseline? I heard changes should be focused on their respective areas.

kamil-tekiela:
Yes, you even should. Baseline should always be kept up-to-date. If your code causes changes to the baseline then you should include it in the PR.

st0rmsetter:
> Yes, you even should. Baseline should always be kept up-to-date. If your code causes changes to the baseline then you should include it in the PR.

I'd like to thank you for helping me out. I'm a PHP developer and I've recently trying dipping my toes with open-source projects. I knew PHPStan and PHPUnit but certainly not Psalm and its baselines. I just wanted to fix an issue however I ended up wasting the time of others who had no business with me. I committed the brand new baseline and all signs are green. 

I could have never imagined I would have received help and guidance in a pull request, but well here I am. I'd like to apologize to everybody whose time I wasted and also for the CI minutes I've wasted. 

This PR is now complete and anybody is free to check it out. I've *fixed* the failing tests.

kamil-tekiela:
No problem. Regardless of the outcome of this PR, I am very happy that you would like to contribute to open source. Every contribution is very much appreciated. Don't worry about small mistakes; we all make them, especially at the start.

Issue #20477, opened 2026-09-08
Dead-code audit with a code-graph tool (plouf.rs) - Things to improve so tools can better understand the codebase


williamdes:
I'm **Claude** (Anthropic's Claude Code assistant). @williamdes asked me to open this issue.

phpMyAdmin was used as a real-world test case while improving a static code-graph / dead-code tool, [plouf.rs](https://github.com/wdes/plouf.rs) (Rust; PHP via [Mago](https://github.com/carthage-software/mago) + JS via oxc; MPL-2.0). A first pass flagged 57 PHP symbols and 24 TS symbols as "unreferenced". Every one was then verified against the source: **they are all live** — each flag is a false positive from a dispatch/registration mechanism a static pass cannot follow. No dead code was found; posting the catalogue for the record. A checked box = verified live.

## PHP — all verified live

- [x] **Schema drawing methods** — `RelationStats*/TableStats*::relationDraw`/`tableDraw`, and `Eps`/`Svg`/`Pdf` `line`/`rect`/`showXY`/`printElement`/`cellScale`/`lineScale`/`setXScale`/`setXyScale` — called on an **abstract-base-typed** receiver (`$relation->relationDraw(...)` where `$relation` is a `RelationStats`), with the method defined only in subclasses, so no static call resolves to the concrete class. `src/Plugins/Schema/Pdf/PdfRelationSchema.php:423`, `src/Plugins/Schema/Svg/SvgRelationSchema.php:257`.
- [x] **`Pdf` overrides** — `src/Pdf.php` `Error`/`Footer`/`_putpages`, `src/Plugins/Schema/Pdf/Pdf.php` `Header`/`Footer`/`cellScale`/… — override a **vendored** base (`class Pdf extends TCPDF`, `src/Pdf.php:24`); TCPDF dispatches them, so there is no in-repo caller.
- [x] **`Database/Designer/Common`** (`addNewRelation`, `createNewPage`, `getAllKeys`, `getColumnsInfo`, `getTablesInfo`, `saveTablePositions`, …) and **`DesignerTable::getTableEngine`** — called through typed-property / return-typed receivers from the Designer controllers.
- [x] **`Partitioning/Partition`** (`addSubPartition`, `getSubPartitions`, `hasSubPartitions`, `getDescription`) — called on `Partition` instances obtained from a return-typed factory.
[...]

PR #20481, opened 2026-09-08
Refactor getPartitions()

kamil-tekiela:
I don't know why some tools are confused by this loop, but the temporary variable was unnecessary, so hopefully this makes it more clear. 

Related to #20477


src/Partitioning/Partition.php
M393:
```suggestion
        if (! self::havePartitioning()) {
            return [];
        }
```
While you're at it change it to an early return here?

MauricioFauth (approved):
Thank you!

PR #20483, opened 2026-09-08
Remove @see declarations

kamil-tekiela:
I never understood their purpose, but many of them are stale and it seems that LLMs read them, judging by #20477. I don't think we need to keep them around, especially in their outdated form.


williamdes (approved):
I agree

PR #20487, opened 2026-09-16
fix: support SJIS-win exports with iconv

### LINKED ISSUE #20128: [Bug]: ErrorException: iconv(): Wrong encoding, conversion from "utf-8" to "SJIS-win//TRANSLIT" is not allowed
### Describe the bug

```
Internal error in .\libraries\classes\Encoding.php#208
ErrorException: iconv(): Wrong encoding, conversion from "utf-8" to "SJIS-win//TRANSLIT" is not allowed
```

### How to Reproduce

```SQL
CREATE TABLE `test_table` (
  `id` int(11) NOT NULL,
  `col` varchar(50) DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=MyISAM DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;

INSERT INTO `test_table` (`id`, `col`) VALUES (1, 'テスト');
```

2. Export table as SQL with custom options
3. Select `SJIS-win` from `Character set of the file`.
4. Click on `Export`
5. See error

### Screenshots

https://github.com/user-attachments/assets/5903b8f8-ce90-4545-85e5-ff0303b114d3

### PHP version

8.5.3

### phpMyAdmin version

5.2.4-dev, 6.0.0-dev

mvanhorn:
### Description

First confirm the alias diagnosis on an affected iconv backend by comparing conversion to `SJIS-win`, `SJIS-win//TRANSLIT`, `CP932`, and `CP932//TRANSLIT`, recording PHP and iconv implementation versions; the expected distinction is support for the charset name, not a blanket incompatibility with transliteration. In `src/Encoding.php`, normalize the exact `SJIS-win` name case-insensitively to `CP932` for both source and destination only when the selected engine is iconv, immediately before the existing conversion dispatch. Preserve the equality fast path, configured iconv suffix, mbstring behavior, and engine-none behavior, and do not substitute plain `SJIS` or suppress conversion errors.

Custom SQL export with `SJIS-win` selected fails with an iconv wrong-encoding error on the reporter's PHP 8.5.3 installation, affecting both 5.2.4-dev and 6.0.0-dev. The issue supplies a MyISAM table and a Japanese `テスト` row as a concrete reproduction. On the current master checkout, `Encoding::convertString()` passes the configured charset straight to iconv and appends the default `//TRANSLIT` suffix; `SJIS-win` is offered by the default charset configuration. The thread suggests removing the suffix, but does not establish that transliteration is the cause; acceptance of charset aliases varies between iconv implementations.

Fixes #20128


mvanhorn:
The PHP 8.5 job wasn't failing an assertion — it was hitting `max_execution_time`. `EncodingTest::testIconvSjisWin` hangs: the SJIS-win path normalises to CP932 and hands it to `iconv()` with the caller's `//TRANSLIT` or `//IGNORE` suffix, and on glibc that combination can fail to consume any input for an unconvertible sequence, so the conversion loop spins until PHP kills it at 300s.

Fixed in 1e73b67:

- prefer mbstring when the build lists `SJIS-win` natively, since mbstring implements Windows CP932 directly rather than via an iconv alias
- when falling back to iconv for CP932, drop the `//TRANSLIT` / `//IGNORE` suffixes, so an unconvertible sequence returns false instead of spinning
- a false result from either backend now yields the original string rather than propagating `false`
- the mbstring capability lookup is cached, since `convertString()` runs once per exported line

Two things worth a reviewer's eye. Dropping `//TRANSLIT` on the iconv CP932 fallback is a real behaviour change — unconvertible characters are no longer transliterated on that path. I think that's the right trade against a hang, but say the word if you'd rather keep transliteration and bound the loop a different way. And I could not reproduce the hang on my machine (macOS, libiconv 1.11) — it's glibc-specific, so the fix is reasoned from the mechanism and confirmed by the suite now finishing in ~35s rather than by watching the original spin stop.

